PR #1193, opened 2019-10-22
[WIP] LG activity copy behavior: copies all possible lg templates

yeze322:
## Description

**Issues**: 

- Not all LG fields will be copied when copying an action. We should consider covering several kinds of LG activities:
  1. activity (used in SendActivity)
  2. prompt (BotAsks - XXXInput)
  3. invalidPrompt
  4. unrecognizedPrompt
  5. defaultValueResponse

- Resolves missed comments in #1096 

**Changes**:
1. Include prompt fields when copying LG template
2. Move the `copyLgTemplate` handler's implementation from 'copyUtils.ts' in 'shared' lib to Shell (lgHandlers.ts)
3. Pass `copyLgTemplate` through ExtensionContainer to visual editor
4. Add UT for lgHandlers, update UT of copyUtils

**Remaining issue**:
- `updateLgTemplate` is not concurrency-safe
- `removeLgTemplate` is not concurrency-safe
(addressed in #1210 )

## Task Item

fixes #1191 
fixes #1213 

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Code refactor (non-breaking change which improve code quality, clean up, add tests, etc)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Doc update (document update)

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have functionally tested my change

## Screenshots

Please include screenshots or gifs if your PR include UX changes.

Issue #1191, opened 2019-10-22
TextInput prompts not copied correctly


yeze322:
## Version
What version of the Composer are you using? Paste the build SHA found on the about page.

## Describe the bug
1. In 'BotAsks', the copied activity doesn't have an emulator
![image](https://user-images.githubusercontent.com/8528761/67280298-721f4300-f4ff-11e9-8414-fc4b53c68a6d.png)
2. In 'Exceptions', 3 other kinds of prompts not copied
![image](https://user-images.githubusercontent.com/8528761/67280375-98dd7980-f4ff-11e9-9943-d3543550e0ec.png)

## To Reproduce
1. Create a prompt text action
2. Fill those prompts
3, copy & paste

## Expected behavior
All prompts should be copied correctly.

PR #1501, opened 2019-11-05
chore: copy Constructor in 'shared' lib

yeze322:
## Description

Implement the copy constructor in dialogFactory to enable action deep clone.

**Obi types need sepcial treatment**
1. Nested
    - IfCondition
    - SwitchCondition
    - Foreach
    - Foreachpage
    - EditActions
2. With LG templates
    - SendActivity
    - All kinds of *Input (TextInput, ChoiceInput, .etc.)

## Task Item

refs #1193, #1191 

## Type of change

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have commented my code, particularly in hard-to-understand areas

## Screenshots


yeze322:
@a-b-r-o-w-n @boydc2014 please review the current pattern we use.

Composer/packages/lib/shared/src/types/sdk.ts
```diff
@@ -3,15 +3,21 @@
...
 /* eslint-disable @typescript-eslint/no-empty-interface */
 
-interface BaseSchema {
+export interface DesignerData {
+  name?: string;
+  description?: string;
+  id: string;
+}
+
+export interface BaseSchema {
```
a-b-r-o-w-n: This was intentionally a private interface. Why do we need to export?

Composer/packages/lib/shared/src/types/sdk.ts
```diff
@@ -3,15 +3,21 @@
...
   /** Defines the valid properties for the component you are configuring (from a dialog .schema file) */
   $type: string;
   /** Inline id for reuse of an inline definition */
   $id?: string;
   /** Copy the definition by id from a .dialog file. */
   $copy?: string;
   /** Extra information for the Bot Framework Composer. */
-  $designer?: OpenObject;
+  $designer?: DesignerData;
```
a-b-r-o-w-n: 👍

Composer/packages/lib/shared/src/types/sdk.ts
```diff
@@ -229,6 +240,32 @@ export interface SwitchCondition extends BaseSchema {
...
+  itemsProperty?: string;
+  pageSize?: number;
+  actions?: MicrosoftIDialog[];
+}
+
+export interface EditActions extends BaseSchema {
+  changeType: string;
+  actions?: MicrosoftIDialog[];
+}
+
```
a-b-r-o-w-n: Thanks for improving these!

They should be added to the `MicrosoftIDiloag` union type as well.

Composer/packages/lib/shared/src/copyUtils/shallowCopyAdaptiveAction.ts
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { BaseSchema } from '../types';
+
+import { ExternalApi } from './ExternalApi';
+
+export function shallowCopyAdaptiveAction(input: BaseSchema, externalApi: ExternalApi): BaseSchema {
```
a-b-r-o-w-n:
```suggestion
export function shallowCopyAdaptiveAction<T = MicrosoftIDialog>(input: T, externalApi: ExternalApi): T {
```

Composer/packages/lib/shared/src/copyUtils/copySwitchCondition.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import { ExternalApi } from './ExternalApi';
+import { copyAdaptiveActionList } from './copyAdaptiveActionList';
+import { shallowCopyAdaptiveAction } from './shallowCopyAdaptiveAction';
+
+export const copySwitchCondition = async (
+  input: SwitchCondition,
+  externalApi: ExternalApi
+): Promise<SwitchCondition> => {
+  const copy: SwitchCondition = shallowCopyAdaptiveAction(input, externalApi);
```
a-b-r-o-w-n:
```suggestion
  const copy = shallowCopyAdaptiveAction(input, externalApi);
```

Composer/packages/lib/shared/src/copyUtils/copySendActivity.ts
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { SendActivity } from '../types';
+
+import { ExternalApi } from './ExternalApi';
+import { shallowCopyAdaptiveAction } from './shallowCopyAdaptiveAction';
+
+export const copySendActivity = async (input: SendActivity, externalApi: ExternalApi): Promise<SendActivity> => {
+  const copy: SendActivity = shallowCopyAdaptiveAction(input, externalApi);
```
a-b-r-o-w-n:
```suggestion
  const copy = shallowCopyAdaptiveAction(input, externalApi);
```

Composer/packages/lib/shared/src/copyUtils/copyInputDialog.ts
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { InputDialog } from '../types';
+
+import { ExternalApi } from './ExternalApi';
+import { shallowCopyAdaptiveAction } from './shallowCopyAdaptiveAction';
+
+export const copyInputDialog = async (input: InputDialog, externalApi: ExternalApi): Promise<InputDialog> => {
+  const copy: InputDialog = shallowCopyAdaptiveAction(input, externalApi) as InputDialog;
```
a-b-r-o-w-n:
```suggestion
  const copy = shallowCopyAdaptiveAction(input, externalApi);
```

Composer/packages/lib/shared/src/copyUtils/copyIfCondition.ts
```diff
@@ -0,0 +1,22 @@
...
+// Licensed under the MIT License.
+
+import { IfCondition } from '../types';
+
+import { ExternalApi } from './ExternalApi';
+import { copyAdaptiveActionList } from './copyAdaptiveActionList';
+import { shallowCopyAdaptiveAction } from './shallowCopyAdaptiveAction';
+
+export const copyIfCondition = async (input: IfCondition, externalApi: ExternalApi): Promise<IfCondition> => {
+  const copy: IfCondition = shallowCopyAdaptiveAction(input, externalApi);
```
a-b-r-o-w-n:
```suggestion
  const copy = shallowCopyAdaptiveAction(input, externalApi);
```

Composer/packages/lib/shared/src/copyUtils/copyForeach.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import { Foreach, ForeachPage } from '../types';
+
+import { ExternalApi } from './ExternalApi';
+import { shallowCopyAdaptiveAction } from './shallowCopyAdaptiveAction';
+import { copyAdaptiveActionList } from './copyAdaptiveActionList';
+
+type ForeachAction = Foreach | ForeachPage;
+export const copyForeach = async (input: ForeachAction, externalApi: ExternalApi): Promise<ForeachAction> => {
+  const copy: ForeachAction = shallowCopyAdaptiveAction(input, externalApi);
```
a-b-r-o-w-n:
```suggestion
  const copy = shallowCopyAdaptiveAction(input, externalApi);
```

Composer/packages/lib/shared/src/copyUtils/copyEditActions.ts
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under the MIT License.
+
+import { EditActions } from '../types';
+
+import { ExternalApi } from './ExternalApi';
+import { shallowCopyAdaptiveAction } from './shallowCopyAdaptiveAction';
+import { copyAdaptiveActionList } from './copyAdaptiveActionList';
+
+export const copyEditActions = async (input: EditActions, externalApi: ExternalApi): Promise<EditActions> => {
+  const copy: EditActions = shallowCopyAdaptiveAction(input, externalApi) as EditActions;
```
a-b-r-o-w-n:
```suggestion
  const copy = shallowCopyAdaptiveAction(input, externalApi);
```

Composer/packages/lib/shared/src/copyUtils/copyAdaptiveActionList.ts
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { copyAdaptiveAction } from './copyAdaptiveAction';
+import { ExternalApi } from './ExternalApi';
+
+export async function copyAdaptiveActionList(actions, externalApi: ExternalApi): Promise<any[]> {
```
a-b-r-o-w-n:
```suggestion
export async function copyAdaptiveActionList(actions: MicrosoftIDialog[], externalApi: ExternalApi): Promise<MicrosoftIDialog[]> {
```

Composer/packages/lib/shared/src/copyUtils/copyAdaptiveActionList.ts
```diff
@@ -0,0 +1,16 @@
...
+  if (!Array.isArray(actions)) return [];
+
+  const results: any[] = [];
```
a-b-r-o-w-n:
```suggestion
  const results: MicrosoftIDialog[] = [];
```

Composer/packages/lib/shared/src/copyUtils/shallowCopyAdaptiveAction.ts
yeze322: Template here is great! I will merge all following change requests to one commit instead of accept each of them

Composer/packages/lib/shared/src/types/sdk.ts
yeze322: Not necessariIy. It's a leak, removed.

Composer/packages/lib/shared/src/copyUtils/shallowCopyAdaptiveAction.ts
yeze322: We can either write a generic function here
```ts
function shallowCopy<T extends BaseSchema>(input: T): T {}
```
or use the `MicrosoftIDialog` type directly
```ts
function shallowCopy(input: MicrosoftIDialog): MicrosoftIDialog {}
```
`T = MicrosoftIDialog` seems not working since `MicrosoftIDialog` is already a set of types.
Now I use the second solution to avoid exporting `BaseSchema`.

yeze322: `MicrosoftIDialog` is a union of types, we cannot use it in a generic function.
image

yeze322: There is another thing confused me. Semantically, schema ought have a `MicrosoftIAction` interface. In current botbuilder-dotnet schema, `actions` is not differentiated and typed with `MicrosoftIDialog` which means that it could be a recognizer / dialog / trigger. How do you understand that @a-b-r-o-w-n ?

Composer/packages/lib/shared/src/copyUtils/copyAdaptiveActionList.ts
yeze322: This is resolved. Typed in commit 3bbc01c4.

Composer/packages/lib/shared/src/copyUtils/copyAdaptiveActionList.ts
yeze322: resolved by 5562a51

Composer/packages/lib/shared/src/copyUtils/shallowCopyAdaptiveAction.ts
a-b-r-o-w-n: I like `function shallowCopy<T extends BaseSchema = BaseSchema>(input: T): T {}`

yeze322: Okay great!

yeze322: resolved

Composer/packages/lib/shared/src/types/sdk.ts
yeze322: Mark as resolve after we use template in `shallowCopyAdaptiveAction<T as extends>`

Composer/packages/lib/shared/src/types/sdk.ts
yeze322: they are added, mark as resolved

Composer/packages/lib/shared/__tests__/copyUtils/copyAdaptiveAction.test.ts
```diff
@@ -0,0 +1,325 @@
...
+    copyLgTemplate: (id, x) => Promise.resolve(x + '(copy)'),
+  };
+
+  it('should return {} when input is invalid', async () => {
+    expect(await copyAdaptiveAction(null, externalApi)).toEqual({});
+    expect(await copyAdaptiveAction({}, externalApi)).toEqual({});
+    expect(await copyAdaptiveAction({ name: 'hi' }, externalApi)).toEqual({});
+  });
+
+  it('can copy BeginDialog', async () => {
```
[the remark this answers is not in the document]
yeze322: yes! this is a really good advide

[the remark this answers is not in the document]
yeze322: resolved by 69c376

yeze322:
resolved all Andy's comments. now working on adapting new lg api

Composer/packages/lib/shared/__tests__/copyUtils/copyAdaptiveAction.test.ts
```diff
@@ -0,0 +1,65 @@
...
+
+      for (const $type of registeredUniqueTypes) {
+        await copyAdaptiveAction({ $type }, externalApi);
+        expect(CopyConstructorMap[$type]).toHaveBeenCalledTimes(1);
+      }
+
+      for (const $type of registeredUniqueTypes) {
+        await copyAdaptiveAction({ $type }, externalApi);
+        expect(CopyConstructorMap[$type]).toHaveBeenCalledTimes(2);
+      }
```
a-b-r-o-w-n: I don't get this. What is it testing?

Composer/packages/lib/shared/__tests__/copyUtils/copyAdaptiveAction.test.ts
```diff
@@ -0,0 +1,65 @@
...
+describe('copyAdaptiveAction', () => {
+  it('should return {} when input is invalid', async () => {
+    expect(await copyAdaptiveAction('hello', externalApi)).toEqual('hello');
+
+    expect(await copyAdaptiveAction(null as any, externalApi)).toEqual({});
+    expect(await copyAdaptiveAction({} as any, externalApi)).toEqual({});
+    expect(await copyAdaptiveAction({ name: 'hi' } as any, externalApi)).toEqual({});
+  });
+
+  describe('should call certain handler in CopyConstructorMap', () => {
```
a-b-r-o-w-n: From a test organization point of view, this is not very descriptive.

here's my suggestion for this file:

```
describe('copyAdaptiveAction')
  it('should return {} when input is invalid')
 
  for type in registeredUniqueTypes
    it(`should invoke correct handler for ${type}')

  it('should invoke default handler for other types')
```

Composer/packages/lib/shared/__tests__/copyUtils/copyAdaptiveAction.test.ts
```diff
@@ -0,0 +1,65 @@
...
+    let originCopyConstructorMap = {};
+    beforeAll(() => {
+      originCopyConstructorMap = { ...CopyConstructorMap };
+    });
+
+    beforeEach(() => {
+      Object.keys(CopyConstructorMap).forEach(key => {
+        CopyConstructorMap[key] = jest.fn();
+      });
+    });
```
a-b-r-o-w-n: This is better done with jest module mocking.

```ts
jest.mock('../../src/copyUtils/copyAdaptiveAction', () => ({
  'Microsoft.TextInput': jest.fn(),
  ...otherTypes
});

// in a test
expect(CopyConstructorMap['Microsoft.TextInput']).toHaveBeenCalledTimes(1);
```

The path may need to be the exact same one used in the app code. I forget.

Composer/packages/server/src/models/bot/botProject.ts
```diff
@@ -182,7 +182,7 @@ export class BotProject {
         ...mainDialog.content.$designer,
         name,
         description,
-      };
+      } as any;
```
a-b-r-o-w-n: Why is this needed?

Composer/packages/lib/shared/__tests__/copyUtils/copyAdaptiveAction.test.ts
yeze322: great! I will accept this one

Composer/packages/lib/shared/__tests__/copyUtils/copyAdaptiveAction.test.ts
yeze322: just to make sure copiers are called every time. I will remove it because of it's not necessary

Composer/packages/server/src/models/bot/botProject.ts
yeze322: the `id` field is missing in this object, triggered a ts error after we adding the `DesingerData` in sdk type. I just realized that it could be rewritten by leveraging the utility in `dialogFactory`.

yeze322: after understanding the logic here, I rewrite it like this:
image

Composer/packages/lib/shared/src/copyUtils/copyInputDialog.ts
```diff
@@ -0,0 +1,30 @@
...
+import { InputDialog } from '../types';
+
+import { ExternalApi } from './ExternalApi';
+import { shallowCopyAdaptiveAction } from './shallowCopyAdaptiveAction';
+
+export const copyInputDialog = async (input: InputDialog, externalApi: ExternalApi): Promise<InputDialog> => {
+  const copy = shallowCopyAdaptiveAction(input, externalApi);
+  const nodeId = copy.$designer ? copy.$designer.id : '';
+
+  if (input.prompt) {
```
boydc2014: input.prompt looks like a requried field. so perhaps we can save some checking there?

Composer/packages/lib/shared/src/copyUtils/copyInputDialog.ts
```diff
@@ -0,0 +1,30 @@
...
+    copy.prompt = await externalApi.copyLgTemplate(input.prompt, nodeId);
+  }
+
+  if (input.unrecognizedPrompt) {
```
boydc2014: Similar things like this "unrecoginzedPrompt". Not sure this is required or not, just want to make sure we leverage the type checking as much as possible. 

Because we control the "constructor", we could be able to enforce more strict type (less optional field). Not for just reducing 1 or 2 line codes, more importantly, have the type enforced at first place, will reduce the possible `state` in our system, means we will reduce the overall complexity of the system.

Composer/packages/lib/shared/src/copyUtils/copyForeach.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import { ExternalApi } from './ExternalApi';
+import { shallowCopyAdaptiveAction } from './shallowCopyAdaptiveAction';
+import { copyAdaptiveActionList } from './copyAdaptiveActionList';
+
+type ForeachAction = Foreach | ForeachPage;
+export const copyForeach = async (input: ForeachAction, externalApi: ExternalApi): Promise<ForeachAction> => {
+  const copy = shallowCopyAdaptiveAction(input, externalApi);
+
+  if (Array.isArray(input.actions)) {
```
boydc2014: is it possible to remove this checking, since you already checked that inside copyList function

boydc2014:
Looks good, i leave a few comments about if we can remove some checking, by enforcing the structure more strictly, which would help reduce the overall complexity.  It might not be the major focus of this PR, but it's a good direction to follow.

Composer/packages/lib/shared/src/copyUtils/copyInputDialog.ts
yeze322: it's declared as required in sdk but might be empty in real dialog json. The checking here is to avoid the case that we generate a json like
```
{
  ...
  prompt: null | undefined
}
```

yeze322: So the check should not be `if (input.prompt)`, it will miss copying an empty string. Maybe `if (input.prompt !== undefined)`?

Composer/packages/lib/shared/src/copyUtils/copyForeach.ts
yeze322: it's similar to previous comment, `actions` are declared as optional in SDK. If we don't check it here, the copy result will be different from input value (`[]` against `undefined`)

yeze322:
I got Dong's point, we have declared some fields as required in sdk but during construction we missed the default values for some required fields. Though the 'copy constructor' should respect the original input to copy the data as it was, it's still a 'constructor' which means that it should generate default values for those required fields.

yeze322:
However, I know a strange behavior which makes me cannot assgin a default value for those propmpt fields:
empty-string-in-activity
@boydc2014

cwhitten:
> I got Dong's point, we have declared some fields as required in sdk but during construction we missed the default values for some required fields. Though the 'copy constructor' should respect the original input to copy the data as it was, it's still a 'constructor' which means that it should generate default values for those required fields.

`assignDefaults` exported from `dialogFactory.ts` should handle assigning any/all default values as defined off of the SDK.

yeze322:
@cwhitten I've added default fields for `SendActivity` and `*Input` type. (`initialDialogShape` in `dialogFactory`)

PR #1502, opened 2019-11-05
fix: yarn start:dev can't work in windows

lei9444:
## Description

1. Add [cross-env](https://www.npmjs.com/package/cross-env) to set environment variables for different platforms.

image

2. In development "import settings from '../settings/settings';" will import setting.json. This will make some path can't be found. so I renamed the setting.ts to index.ts.

## Type of change

## Checklist

## Screenshots

### DIFF
--- a/Composer/yarn.lock
+++ b/Composer/yarn.lock
@@ -5158,8 +5158,8 @@ create-react-context@^0.2.1:
 
 cross-env@^6.0.3:
   version "6.0.3"
-  resolved "https://registry.yarnpkg.com/cross-env/-/cross-env-6.0.3.tgz#4256b71e49b3a40637a0ce70768a6ef5c72ae941"
-  integrity sha512-+KqxF6LCvfhWvADcDPqo64yVIB31gv/jQulX2NGzKS/g3GEVz6/pt4wjHFtFWsHMddebWD/sDthJemzM4MaAag==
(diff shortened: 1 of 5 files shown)


Composer/yarn.lock
```diff
@@ -5158,8 +5158,8 @@ create-react-context@^0.2.1:
 
 cross-env@^6.0.3:
   version "6.0.3"
-  resolved "https://registry.yarnpkg.com/cross-env/-/cross-env-6.0.3.tgz#4256b71e49b3a40637a0ce70768a6ef5c72ae941"
-  integrity sha512-+KqxF6LCvfhWvADcDPqo64yVIB31gv/jQulX2NGzKS/g3GEVz6/pt4wjHFtFWsHMddebWD/sDthJemzM4MaAag==
+  resolved "https://botbuilder.myget.org/F/botbuilder-declarative/npm/cross-env/-/cross-env-6.0.3.tgz#4256b71e49b3a40637a0ce70768a6ef5c72ae941"
```
boydc2014: we should not change this

lei9444:
> Sorry about introducing this! cross-env is already a workspace dependency at the root so it shouldn't need to be added to dependencies. If it does, it should be listed as a devDependency instead.

Hi, @a-b-r-o-w-n  I have removed cross-env from the dependencies and it works well.

Issue #1500, opened 2019-11-05
Only support westus deployment in Deployment UX


luhan2017:
**Is your feature request related to a problem? Please describe.**
#1459 

**Describe the solution you'd like**
By default select region: westus in the deployment UX

**Describe alternatives you've considered**

**Additional context** 
remove this and by default select west us
![image](https://user-images.githubusercontent.com/32191031/68176723-f1614c00-ffc0-11e9-8137-96cc88fa9bed.png)

VanyLaw:
Hi @hibrenda, I think it's a little bit similar with the concept @mareekuh suggested in [here](https://github.com/microsoft/BotFramework-Composer/issues/1295).

If we support WestUs as default and remove other options, Maybe we should use TextField instead and make it not editable. Or just simply don't show this value to user, hardcode it in code.

If we can support 3 regions below, I think it's better to keep those options, and make WestUs as default.
- (US)West US
- (Asia Pacific) Australia East
- (Europe) West Europe

hibrenda:
@mareekuh @cwhitten  what is your thoughts ? I prefer the simple version which align with the new design.

Issue #1459, opened 2019-11-01
Error assigning LUIS production keys to newly created LUIS app


yochay:
## Version
cf7eea30 (master 11/1)

## Describe the bug
Deployment script fails on assigning luis app with luisAppIds with the endpoint key

## To Reproduce
1. created a new Interruptions bot
2. run deploy script (results here-under). Successfully created all resources 
```
cd C:/Data/repos/BotFramework-Composer/MyBots/Interuption2
pwsh ./Scripts/create.ps1 -name Interuption2 -environment NewInterutpion -location WestUS2 -appPassword appkeyhere -luisAuthoringKey keygoeshere

C:\Data\repos\BotFramework-Composer\MyBots\Interuption2>pwsh ./Scripts/create.ps1 -name Interuption2 -environment NewInterutpion -location WestUS2 -appPassword '1qazxsw2!QAZXSW@' -luisAuthoringKey d016585ce5c9446bbdbac5ea95660e6b
C:\Program Files (x86)\Microsoft SDKs\Azure\CLI2\lib\site-packages\requests\__init__.py:91: RequestsDependencyWarning: urllib3 (1.25) or chardet (3.0.4) doesn't match a supported version!
> Creating resource group ...
> Validating Azure deployment ...
C:\Program Files (x86)\Microsoft SDKs\Azure\CLI2\lib\site-packages\requests\__init__.py:91: RequestsDependencyWarning: urllib3 (1.25) or chardet (3.0.4) doesn't match a supported version!
> Deploying Azure services (this could take a while)...
C:\Program Files (x86)\Microsoft SDKs\Azure\CLI2\lib\site-packages\requests\__init__.py:91: RequestsDependencyWarning: urllib3 (1.25) or chardet (3.0.4) doesn't match a supported version!
> Updating appsettings.json ...
Specify --help for a list of available options and commands.
Unrecognized command or argument 'init'
Could not find the global property 'UserSecretsId' in MSBuild project 'C:\Data\repos\BotFramework-Composer\MyBots\Interuption2\BotProject.csproj'. Ensure this property is set in the project or use the '--id' command line option.
> Done.
- App Id: 47743043-87b3-4d95-b1df-qqqqqqqq
- App Password: qqqqqqqqq
- Resource Group: Interuption2-NewInterutpion
- ServicePlan: Interuption2-NewInterutpion
- Bot Name: Interuption2
- Web App Name : Interuption2
```

[...]

PR #1503, opened 2019-11-05
remove unworkable luis region

VanyLaw:
## Description
Fix this [Issue](https://github.com/microsoft/BotFramework-Composer/issues/1500)
Remove unavailable luis region and set "WestUs" as default.

## Task Item
https://github.com/microsoft/BotFramework-Composer/issues/1459
https://github.com/microsoft/BotFramework-Composer/issues/1500

## Type of change

## Checklist

## Screenshots

region


Composer/packages/client/src/pages/setting/deployment/deployWizardStep-createDeploy.js
```diff
@@ -23,6 +24,7 @@ import { styles } from './styles';
 import { regionOptions } from './luisRegions.js';
 
 export const DeployWizardStepCreate = props => {
+  const defaultRegion = 'WestUS';
```
boydc2014: I don't think we should hard code here, we should assume the options[0] is default.

VanyLaw: Yeah, but I am not sure the other 2 regions is it OK for everyone. because currently our luis authoring accounts are mostly for westus region. Should we give a hint to user?

cwhitten:
We want to keep WestUS as the only option in the dropdown. We keep the control and will add other regions as they become available. _bonus_: inspect office-ui-fabric-react to see if we can set this value by default.

Composer/packages/client/src/pages/setting/deployment/deployWizardStep-createDeploy.js
VanyLaw: removed other 2 regions, use options[[0] as default.

VanyLaw:
> We want to keep WestUS as the only option in the dropdown. We keep the control and will add other regions as they become available. _bonus_: inspect office-ui-fabric-react to see if we can set this value by default.

Ok, I only remove the other options and keep original control. I use defaultSelectedKey to set this value by default.

PR #1522, opened 2019-11-06
update main workflow to run for pull requests created from forks

a-b-r-o-w-n:
Using `push` only runs when the base branch is in the upstream repository so does not work with forks.


.github/workflows/main.yml
```diff
@@ -1,9 +1,16 @@
 name: Composer CI
 
-on: [push]
+on:
+  pull_request:
+    types: [opened, reopend, synchronize]
```
cwhitten: is reopend really a type?

a-b-r-o-w-n: nope!
nice catch

a-b-r-o-w-n:
```suggestion
    types: [opened, reopened, synchronize]
```

PR #1536, opened 2019-11-08
update $schema to point to composer sdk.schema

### LINKED ISSUE #875: $schema is incorrect for examples
The current samples point to ../../app.schema which does not exist.  We should make sure it does in order to support tooling and visual studio code.  It should point to the SDK schema, i.e. 
"$schema": "https://raw.githubusercontent.com/microsoft/botbuilder-dotnet/4.Future/schemas/sdk.schema"
Since composer is a private repo, we cannot point to the composer checked-in schema but we could point to the relative schema, i.e.
  "$schema": "../../../Composer/packages/server/schemas/sdk.schema"

luhan2017:
## Description

update $schema to point to composer sdk.schema

## Task Item

closes #875 

## Type of change

## Checklist


cwhitten:
`stable` makes the most sense for now

PR #1548, opened 2019-11-11
feat: update default bots folder

### LINKED ISSUE #1363: Update default folder when Creating a new bot
## Version
cd1cc71f (from Master 10/30/2019)

## Describe the bug
When creating a new bot, use a dedicated folder that is outside of Composer repo. The default project folder for new bots should not be MyBots under Composer. 

image

For windows use %userprofile%\Documents\Composer\
- create the folder if doesn't exists 

For Mac it should be in /user/documents/composer 
- create the folder if doesn't exists

### LINKED ISSUE #1506: need to take care location in bot creation
Complete the necessary portions of this template and delete the rest.

## Describe the bug
User could specify the bot location during the bot creation and next time the location will be loaded from recent usage. Default folder location defined in #1363. 

If user create the bot by specifying the location, and then delete such folder, user will see error shows as following. and click next, bot will be created into default folder "MyBot".
image

So we need to check if recent usage is still valid ,  if folder has been moved, we should load default folder. If something wrong with default folder, at least we should let user to update the bot create location.  

Steps 
- create a new bot with valid folder
- delete folder 
- go to homepage and create a new bot again
- hit error 

## Version
master

## Browser
What browser are you using?

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS 

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

## Expected behavior

## Screenshots
If applicable, add screenshots/gif/video to help explain your problem.

liweitian:
## Description
1. The default folder for creating a new bot is path.join(os.homedir(), 'Documents', 'Composer') now.
2. If the last accessed path is invalid, load default folder.

## Task Item

Closes #1506 
Closes #1363 

## Type of change
Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/Composer/packages/server/src/services/storage.ts
+++ b/Composer/packages/server/src/services/storage.ts
@@ -86,7 +86,8 @@ class StorageService {
   };
 
   public updateCurrentPath = (path: string) => {
-    Store.set(this.STORE_KEY, path);
+    this.storageConnections[0].path = path;
+    Store.set(this.STORE_KEY, this.storageConnections);
   };
 
   private isBotFolder = (path: string) => {
--- a/Composer/packages/server/src/services/storage.ts
+++ b/Composer/packages/server/src/services/storage.ts
@@ -18,6 +18,11 @@ class StorageService {
 
   constructor() {
     this.storageConnections = Store.get(this.STORE_KEY);
+    this.storageConnections.forEach(s => {
+      if (!fs.existsSync(s.defaultPath)) {
+        fs.mkdirSync(s.defaultPath);
+      }
+    });
   }
 
   public getStorageClient = (storageId: string): IFileStorage => {
@@ -41,7 +46,12 @@ class StorageService {
   public getStorageConnections = (): StorageConnection[] => {
     return this.storageConnections.map(s => {
       const temp = Object.assign({}, s);
-      temp.path = Path.resolve(s.path); // resolve path if path is relative, and change it to unix pattern
+      if (fs.existsSync(s.path)) {
+        temp.path = Path.resolve(s.path); // resolve path if path is relative, and change it to unix pattern
+      } else {
+        temp.path = Path.resolve(s.defaultPath);
+      }
+      temp.defaultPath = Path.resolve(s.defaultPath);
       return temp;
     });
   };
--- a/Composer/packages/server/src/services/storage.ts
+++ b/Composer/packages/server/src/services/storage.ts
@@ -85,6 +85,10 @@ class StorageService {
     }
   };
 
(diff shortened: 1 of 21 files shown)


Composer/packages/server/src/services/storage.ts
```diff
@@ -18,6 +18,9 @@ class StorageService {
 
   constructor() {
     this.storageConnections = Store.get(this.STORE_KEY);
+    this.storageConnections.forEach(s => {
+      this.createFolderRecurively(s.defaultPath);
+    });
```
a-b-r-o-w-n: What if the user has changed their default folder and then deleted the one we create? Won't this create /Documents/Composer every time?

Composer/packages/server/src/services/storage.ts
```diff
@@ -41,7 +44,12 @@ class StorageService {
   public getStorageConnections = (): StorageConnection[] => {
     return this.storageConnections.map(s => {
       const temp = Object.assign({}, s);
-      temp.path = Path.resolve(s.path); // resolve path if path is relative, and change it to unix pattern
+      // if the last accessed path exist
+      if (fs.existsSync(s.path)) {
+        temp.path = Path.resolve(s.path); // resolve path if path is relative, and change it to unix pattern
+      } else {
+        temp.path = Path.resolve(s.defaultPath);
+      }
```
a-b-r-o-w-n: What happens if neither path exists? We might want to ensure that the default path exists here as well.

Composer/packages/server/src/services/storage.ts
liweitian: the default path is set in file data.template.json. It does not allow users to change it. These lines of code it to make sure the default path is always valid.

Composer/packages/server/src/services/storage.ts
```diff
@@ -41,7 +42,13 @@ class StorageService {
   public getStorageConnections = (): StorageConnection[] => {
     return this.storageConnections.map(s => {
       const temp = Object.assign({}, s);
-      temp.path = Path.resolve(s.path); // resolve path if path is relative, and change it to unix pattern
+      // if the last accessed path exist
+      if (fs.existsSync(s.path)) {
+        temp.path = Path.resolve(s.path); // resolve path if path is relative, and change it to unix pattern
+      } else {
+        this.createDefaultBotFolders();
```
boydc2014: you are creating all folders inside a for loop. It's a O(n2) operation. 

my suggestion is call this method "ensureFolderExisit", and put this after this map.

cwhitten:
@liweitian there is a lint failure

PR #1555, opened 2019-11-12
build: improve app bundle size and speed up client builds

### LINKED ISSUE #1289: Investigate the memory usage of webpack
We will hit the OOF memory very often, especially when building with docker. 
```
$ node scripts/build.js
/bin/sh: git: not found
Creating an optimized production build...
Browserslist: caniuse-lite is outdated. Please run next command `yarn upgrade`

<--- Last few GCs --->

[471:0x55563aaaaae0]   288341 ms: Mark-sweep 1330.7 (1422.8) -> 1330.1 (1422.8) MB, 954.8 / 0.0 ms  (average mu = 0.150, current mu = 0.013) allocation failure scavenge might not succeed
[471:0x55563aaaaae0]   289519 ms: Mark-sweep 1330.9 (1423.3) -> 1330.3 (1423.8) MB, 1174.5 / 0.0 ms  (average mu = 0.072, current mu = 0.003) allocation failure scavenge might not succeed


<--- JS stacktrace --->

==== JS stack trace =========================================
```

It's because in the linux version we choose for docker, node.js have default 1.76 GB memory limit. 

But we can't simply add the memory as we do in CI, because the up-limit of docker-compose build container is 2GB. Really not too much space there. 

This issue is to investigate, do we really need so much memory? If we can optimize some webpack config, probably the memory usage can drop down significantly. (typical webpack pattern)

a-b-r-o-w-n:
## Description

Introduces multiple concepts to address bundle size as well as build performance.
#### Dynamic imports + React Suspense 
This enables webpack to create smaller chunks which reduces the code needed to be sent over the wire for first load. This also improves build speed.

#### Smarter imports
Eslint rules help us import packages from ES6 modules (like lodash and office ui). Webpack can tree shake these imports resulting in smaller bundles.

#### Improved typescript support in client
Uses `ts-loader` instead of babel to parse ts files in the client. It is faster than babel, but did expose some issues (`@jsx` pragma required for any file using `css` prop -- eslint && fixable).

#### Updates to webpack dependencies
This is suggested by the webpack docs to get the best performance. Some packages jumped major versions. I did not see any issues with the build result though.

## Task Item

closes #1289


boydc2014:
Fantastic ! @lei9444 can you help check this?

PR #1601, opened 2019-11-19
fix: misused 'useEffect' hook in Visual Editor (#1496)

### LINKED ISSUE #1496: Visual Editor redraw problem (performance)
## Describe the bug
Visual Editor shouldn't redraw the whole element tree when operating on it. We should optimize the rendering performance to avoid frequent redraw.

We can focus on optimizing two scenarios for now:
1. click on a node will render twice 
2. change focus state will redraw other nodes

## Version
What version of the Composer are you using? Paste the build SHA found on the about page (`/about`).

## Browser
What browser are you using?

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS 

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

## Expected behavior

## Screenshots
If applicable, add screenshots/gif/video to help explain your problem.

yeze322:
## Description

**How does the redraw promblem happen**
My observation was based on two scenarios: 1. change focus state by clicking a node 2. change node data by adding/deleting a node.
According to the profiler's result, every time we do those two operations, Visual Editor will be redrawn twice: 
- the first redraw is triggered by Shell state update (expected);
- the second redraw, which is redundant, is triggered by the `useEffect` hook which syncs Shell props to local state(unexpected).

So the redraw problem is basically caused by the misused `useEffect` hook and unnecessary local states in Visual Editor.

**Why there are 'useEffect' hooks**
At the very beginning, Visual Editor didn't sync states except 'obi json' with Shell, so all its states were local. As the evolution of Visual Editor, we are syncing more states between Visual and Shell, as a result, we hoisted many local states to shell layer. Thus `useEffect` hook was used an observer to sync Shell props to Visual Editor's local states which will always trigger another Visual Editor rendering.

**Solution**
As describe before, It is obviously an anti-pattern usage of the `useEffect` according to the official doc, indicates our state management is not good enough. According to our discussion, the long term plan is, we migrate to the `useReducer` hook to follow a redux-like state management pattern.

In the short run, for fixing the redraw issue, the solution is to let Visual Editor be rendered only by props passed from Shell and remove some local states such as `focusedId`.

Previously, the state chain is:
```
Shell props => (render) => local state => (render)
```
now, the state chain is:
```
Shell props => (render)
```

It's a quite straight forward fix after we hoisted visual editor states to Shell, but I think it would be much clearer by providing more context about it.

## Task Item

Fixes #1496 

## Type of change

## Checklist

## Screenshots

### DIFF
--- a/Composer/packages/extensions/visual-designer/src/index.tsx
+++ b/Composer/packages/extensions/visual-designer/src/index.tsx
@@ -4,7 +4,7 @@
 /** @jsx jsx */
 import { jsx, CacheProvider } from '@emotion/core';
 import createCache from '@emotion/cache';
-import React, { useRef, useState, useEffect } from 'react';
+import React, { useRef } from 'react';
 import isEqual from 'lodash/isEqual';
 import formatMessage from 'format-message';
 
@@ -63,7 +63,7 @@ const VisualDesigner: React.FC<VisualDesignerProps> = ({
   const focusedId = Array.isArray(focusedActions) && focusedActions[0] ? focusedActions[0] : '';
 
   // NOTE: avoid re-render. https://reactjs.org/docs/context.html#caveats
-  const [context, setContext] = useState({
+  const nodeContext = {
     focusedId,
     focusedEvent,
     focusedTab,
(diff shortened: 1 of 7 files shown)


Composer/packages/extensions/visual-designer/src/index.tsx
```diff
@@ -62,8 +62,7 @@ const VisualDesigner: React.FC<VisualDesignerProps> = ({
 
   const focusedId = Array.isArray(focusedActions) && focusedActions[0] ? focusedActions[0] : '';
 
-  // NOTE: avoid re-render. https://reactjs.org/docs/context.html#caveats
-  const [context, setContext] = useState({
```
yeze322: Notes here: the usage of `useState` here was correct before we hoisting the `focusedId` to Shell but not necessary any more after our hoisting PRs.

Issue #1559, opened 2019-11-12
investigate running e2e tests using the cypress runner


a-b-r-o-w-n:
Cypress hosts a service which will execute our e2e tests in parallel. It's free for open source projects so I think it's worth looking into in hopes of 1. speeding the builds up. 2. gaining stability and trust in our e2e suite.

cwhitten:
addressed in https://github.com/microsoft/BotFramework-Composer/pull/1602

PR #1602, opened 2019-11-19
ci: run cypress tests in parallel

a-b-r-o-w-n:
## Description

Uses parallel builds to split up e2e tests on azure.

## Task Item

refs #1559


Composer/cypress.json
```diff
@@ -1,4 +1,5 @@
 {
+  "projectId": "27ikaq",
```
cwhitten: what's this?

a-b-r-o-w-n: this is for connection with the cypress dashboard.

Issue #1291, opened 2019-10-28
UX - Templates should get a default name 


mareekuh:
When opening a template, the user will likely expect to immediately see that template. Instead, one gets a file-new kind of experience. This is fine, but it'd be more userfriendly to pre-name the file, starting with the template name and a number or other so that the user can just hit next/open and continue the process vs stop and think what to name a file they didnt create on their own nor have seen what it contains. 

![Screen Shot 2019-10-28 at 10 58 51 AM](https://user-images.githubusercontent.com/48503434/67706209-994d9880-f975-11e9-8444-eee59d01111d.png)

cwhitten:
Naming convention:

ExampleName-1
ExampleName-2
ExampleName-3
ExampleName-4

cwhitten:
addressed in https://github.com/microsoft/BotFramework-Composer/pull/1583

cwhitten:
addressed in https://github.com/microsoft/BotFramework-Composer/pull/1668

PR #1604, opened 2019-11-20
fix: Cannot enter custom name when creating a new bot

### LINKED ISSUE #1603: unable to enter custom name for new bot
## Describe the bug
#1583 introduced a bug in which a user is unable to type a custom name for a new bot because the input field cannot be cleared. It is always reset to the 'default' name.

## Version
21516678c2

liweitian:
## Description

Cannot add custom name when creating a new bot

## Task Item

Closes #1603 #1291 

## Type of change

## Checklist

## Screenshots
aaaa

### DIFF
--- a/Composer/packages/client/src/CreationFlow/DefineConversation/index.js
+++ b/Composer/packages/client/src/CreationFlow/DefineConversation/index.js
@@ -40,7 +40,7 @@ const validateForm = data => {
 export function DefineConversation(props) {
   const { state } = useContext(StoreContext);
   const { onSubmit, onGetErrorMessage, onDismiss, enableLocationBrowse } = props;
-  const { templateId, focusedStorageFolder } = state;
+  const { templateId, focusedStorageFolder, botName } = state;
   const [formData, setFormData] = useState({ errors: {} });
   const [location, setLocation] = useState('');
   const [disable, setDisable] = useState(false);
@@ -54,7 +54,8 @@ export function DefineConversation(props) {
       }
     });
     let i = 0;
-    let defaultName = `${templateId}-${i}`;
+    const bot = templateId || botName;
+    let defaultName = `${bot}-${i}`;
 
     while (
       botsInCurrentFolder.findIndex(bot => {
@@ -62,7 +63,7 @@ export function DefineConversation(props) {
       }) > -1
     ) {
       i = i + 1;
-      defaultName = `${templateId}-${i}`;
+      defaultName = `${bot}-${i}`;
     }
     updateForm('defaultName')(null, defaultName);
   }, [templateId, focusedStorageFolder]);
--- a/Composer/packages/client/src/CreationFlow/DefineConversation/index.js
+++ b/Composer/packages/client/src/CreationFlow/DefineConversation/index.js
@@ -1,25 +1,34 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-import React, { useState, Fragment } from 'react';
+import React, { useState, useContext, Fragment, useEffect } from 'react';
 import formatMessage from 'format-message';
 import { PrimaryButton, DefaultButton } from 'office-ui-fabric-react/lib/Button';
 import { DialogFooter } from 'office-ui-fabric-react/lib/Dialog';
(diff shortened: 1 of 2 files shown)


liweitian:
> The way this should work is the default name should be fetched on the initial render:
> 
> ```ts
> const [formData, setFormData] = useState({ name: getDefaultName(), errors: {} });
> ```
I do not think this will work. Our code structure is like
<A>
   update name based on location
   <B/> 
</A>

<B>
  will update location
</B>
We need to wait B finish updating location then the name can be calculated. And using two names 'name and defaultName' is to solve some other issues. For example, the defaultName need to be recalculated once the location changed.

a-b-r-o-w-n:
It is my opinion that the default name should only be generated once (on the first render). If it happens that the user has a bot already with that name and gets something like (EmptyBot-1), and then chooses to navigate to a different folder, it does not mean we should recalculate the name. Most likely, the user will provide their own name. This is just to populate the text field.

Composer/packages/client/src/CreationFlow/DefineConversation/index.js
```diff
@@ -34,33 +32,11 @@ const validateForm = data => {
 export function DefineConversation(props) {
   const { state } = useContext(StoreContext);
   const { onSubmit, onGetErrorMessage, onDismiss, enableLocationBrowse } = props;
-  const { templateId, focusedStorageFolder } = state;
-  const [formData, setFormData] = useState({ errors: {} });
+  const { templateId } = state;
+  const [formData, setFormData] = useState({ name: templateId, errors: {} });
```
a-b-r-o-w-n: This doesn't append with a number if a bot already exists with this name right? Isn't that what we want?

liweitian: I change the code back to the last version. Just like my last comment, our code structure is like
<A>
 update name based on location
 <B/>
</A>
<B>
  will update location
</B>
We need to wait B finish updating location then the number in name can be calculated. For this feature, a better code structure should be after B getting the location variable, then passing it to A as a prop. 
<B>
will update location
 <A location/>
</B>
<A>
   update name based on location
</A>
But for this small feature we do not need to refactor the code.

So use an useEffect(() => set default name, [location]) is necessary. And recalculation on name becomes a side effect. What do you think?

boydc2014:
> It is my opinion that the default name should only be generated once (on the first render). If it happens that the user has a bot already with that name and gets something like (EmptyBot-1), and then chooses to navigate to a different folder, it does not mean we should recalculate the name. Most likely, the user will provide their own name. This is just to populate the text field.

Agreed, offline chatted with @liweitian and clarified a few things
1 Only calculate the default name at first render totally make sense, we should not trying to re-calculate if user choose another folder, not only it's more effort, but also it's a weird experience. 
2 If the information is not enough to support us to do the calculation correctly at first render, we should fix the root cause. 
   * I know that if we created a bot, the `focusedStorageFolder` is not refreshed, we should refresh that

Besides that, looks like this `<LocationSelectContent>` component is holding too much information (default folder, children) inside it, we should make this `<LocationSelectContent>` component more pure and more state-driven, so that all the location information is stored in state and easy access by any component. But this might be in the scope of this PR.

Composer/packages/client/src/CreationFlow/DefineConversation/index.js
a-b-r-o-w-n: I see 2 options, both remove the need for the `defaultName` state variable:

1. if `templateId` and `focusedStorageFolder` are set on the first render, we do not need the `useEffect` hook.
2. If they are not set, we can use the `useEffect` hook with a guard. 
```ts
useEffect(() => {
  if (!name) {
    // calculate name
   setFormData({ name: '...' });
  }
}, [templateId, focusedStorageFolder]);
```

liweitian: A wrong name will be set on the first render. Then focuedStorageFolder will be updated. But the name cannot be updated due to if(!name). I think I need to refactor code. Update on focusedStorageFolder should be put on the upper layer rather than locationSelectContent component just like what Dong Lei said.

liweitian:
Need to refactor code

PR #1608, opened 2019-11-20
feat: Convert ludown to bf-lu

lei9444:
## Description

use bf-lu to update lubuild and @bfc/lib.

save invalid data to lu #1089
## Task Item

closes #1556

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Code refactor (non-breaking change which improve code quality, clean up, add tests, etc)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Doc update (document update)

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [x] I have functionally tested my change

## Screenshots

Please include screenshots or gifs if your PR include UX changes.
image

PR #1667, opened 2019-11-26
feat: Deep linking for the notification page

### LINKED ISSUE #1582: Deep linking for the notification page

As a composer user, If there are errors in dialog/lu/lg files, all the errors/warnings will be rolled up to the notification page. If user click the errors in notification page, it will navigate back to the specific location. 
- For dialog errors (eg. property field ), it will navigate back to the dialog and the error property field will be highlighted with red rectangle. 
- For lu errors, it will navigate back to the dialog and luis editor will be in red rectangle
- For lg errors, it will navigate to the lg all up view in edit mode and focus on the error line

**Additional context**

lei9444:
## Description
- [x] For dialog errors (eg. property field ), it will navigate back to the dialog and the error property field will be highlighted with red rectangle.
- [x] For lu errors, it will navigate back to the dialog and luis editor will be in red rectangle
- [x] For lg errors, it will navigate to the lg all up view in edit mode and focus on the error line

## Task Item

refs #1582 

## Type of change

## Checklist

## Screenshots
dialog

lg
lu


Composer/packages/client/src/ShellApi.ts
```diff
@@ -132,8 +133,13 @@ export const ShellApi: React.FC = () => {
     if (id === undefined) throw new Error('must have a file id');
     const file = lgFiles.find(file => file.id === id);
     if (!file) throw new Error(`lg file ${id} not found`);
+    let templates: LGTemplate[] = [];
+    try {
+      templates = lgUtil.parse(file.content);
+    } catch (error) {
+      console.error(error);
```
compulim: I am worried about data loss due to bad `lgUtil.parse`.

Composer/packages/client/src/pages/language-generation/code-editor.tsx
```diff
@@ -8,17 +8,19 @@ import get from 'lodash/get';
...
 import { CodeRange, LgFile } from '@bfc/shared';
+import { editor } from '@bfcomposer/monaco-editor/esm/vs/editor/editor.api';
 
 import * as lgUtil from '../../utils/lgUtil';
 
 interface CodeEditorProps {
   file: LgFile;
   onChange: (value: string) => void;
-  codeRange?: Partial<CodeRange> | null;
+  codeRange?: Partial<CodeRange> | undefined;
```
compulim: I think you don't need `| undefined`, because `?` implies that.

```suggestion
  codeRange?: Partial<CodeRange>;
```

Composer/packages/client/src/pages/language-generation/index.tsx
```diff
@@ -37,12 +38,24 @@ const LGPage: React.FC<RouteComponentProps> = props => {
   const { lgFiles, dialogs } = state;
   const [editMode, setEditMode] = useState(false);
   const [fileValid, setFileValid] = useState(true);
-  const [codeRange, setCodeRange] = useState<CodeRange | null>(null);
+  const [codeRange, setCodeRange] = useState<CodeRange | undefined>(undefined);
```
compulim: I am new to TypeScript in React Hooks, so I am unsure if the syntax is correct. But try to minimize the code like this.

```suggestion
  const [codeRange, setCodeRange] = useState<CodeRange?>();
```

Composer/packages/client/src/pages/language-generation/index.tsx
```diff
@@ -37,12 +38,24 @@ const LGPage: React.FC<RouteComponentProps> = props => {
...
+  const [lgEditor, setLgEditor] = useState<editor.IStandaloneCodeEditor | null>(null);
+  // const editorRef: MutableRefObject<editor.IStandaloneCodeEditor | null> = useRef(null);
```
compulim: Remove commented code.

Composer/packages/client/src/pages/language-generation/index.tsx
```diff
@@ -37,12 +38,24 @@ const LGPage: React.FC<RouteComponentProps> = props => {
   const { lgFiles, dialogs } = state;
   const [editMode, setEditMode] = useState(false);
   const [fileValid, setFileValid] = useState(true);
-  const [codeRange, setCodeRange] = useState<CodeRange | null>(null);
+  const [codeRange, setCodeRange] = useState<CodeRange | undefined>(undefined);
+  const [lgEditor, setLgEditor] = useState<editor.IStandaloneCodeEditor | null>(null);
```
compulim: Should we use `undefined` instead?

Composer/packages/client/src/pages/language-generation/index.tsx
```diff
@@ -37,12 +38,24 @@ const LGPage: React.FC<RouteComponentProps> = props => {
...
 
+  const hash = props.location ? props.location.hash : '';
   const subPath = props['*'];
   const isRoot = subPath === '';
   const activeDialog = dialogs.find(item => item.id === subPath);
 
+  useEffect(() => {
+    if (hash && lgEditor) {
+      const matched = hash.match(/line=(\d+)/g);
+      if (matched && matched.length > 0) {
```
compulim:
```suggestion
      if (matched && matched.length) {
```

Composer/packages/client/__tests__/components/notificationList.test.js
```diff
@@ -8,9 +8,9 @@ import { NotificationList } from '../../src/pages/notifications/NotificationList
 
 describe('<NotificationList/>', () => {
   const items = [
```
cwhitten: why have id be a string but reference a number?

```javascript
{
  id: '1'
}
```

also, if any of these strings render to the user, they need to be wrapped in `formatMessage`

Composer/packages/client/src/pages/language-generation/index.tsx
cwhitten: is this necessary?

Composer/packages/client/src/pages/language-generation/index.tsx
```diff
@@ -113,15 +126,15 @@ const LGPage: React.FC<RouteComponentProps> = props => {
...
+      navigateTo('/language-generation/');
     } else {
       navigateTo(`language-generation/${id}`);
     }
   }
 
   function onToggleEditMode() {
     setEditMode(!editMode);
-    setCodeRange(null);
+    setCodeRange(undefined);
```
cwhitten: why the need for this change?

Composer/packages/client/src/pages/language-generation/index.tsx
```diff
@@ -37,12 +38,24 @@ const LGPage: React.FC<RouteComponentProps> = props => {
...
+  // const editorRef: MutableRefObject<editor.IStandaloneCodeEditor | null> = useRef(null);
 
+  const hash = props.location ? props.location.hash : '';
   const subPath = props['*'];
   const isRoot = subPath === '';
   const activeDialog = dialogs.find(item => item.id === subPath);
 
+  useEffect(() => {
+    if (hash && lgEditor) {
+      const matched = hash.match(/line=(\d+)/g);
```
compulim: Prefer this approach.

```suggestion
      const match = /line=(\d+)/g.exec(hash);
```

Then you can look up `match[1]`, it will be the number.

If it is no-match, `match` will be falsy.

So the final code looks like:

```js
      const match = /line=(\d+)/g.exec(hash);

      if (match) {
        lgEditor.revealLine(+match[1]);
      }
```

Also, consider using `+` instead of `parseInt` or `Number()`. All of them are different, but `+` is fastest and work in you case. Some of them deal with string, undefined, etc, differently.

Composer/packages/client/src/pages/language-generation/index.tsx
```diff
@@ -215,7 +229,14 @@ const LGPage: React.FC<RouteComponentProps> = props => {
           <div css={contentEditor}>
             {editMode ? (
               <Suspense fallback={<LoadingSpinner />}>
-                <CodeEditor file={lgFile} codeRange={codeRange} onChange={onChange} />
+                <CodeEditor
+                  file={lgFile}
+                  codeRange={codeRange}
+                  onChange={onChange}
+                  editorDidMount={editor => {
```
cwhitten: can you just use

```jsx
editorDidMount={setLgEditor}
```

Composer/packages/extensions/obiformeditor/src/Form/fields/RecognizerField/index.tsx
```diff
@@ -142,12 +154,9 @@ export const RecognizerField: React.FC<FieldProps<MicrosoftIRecognizer>> = props
           {() => {
             if (selectedFile && isLuFileSelected) {
               const updateLuFile = (newValue?: string): void => {
-                shellApi
-                  .updateLuFile({ id: selectedFile.id, content: newValue })
-                  .then(() => setErrorMsg(''))
-                  .catch(error => {
-                    setErrorMsg(error);
-                  });
+                shellApi.updateLuFile({ id: selectedFile.id, content: newValue }).catch(error => {
```
cwhitten: Would the following work?

```javascript
 shellApi.updateLuFile({ id: selectedFile.id, content: newValue }).catch(setErrorMsg)
```

Composer/packages/client/src/pages/language-generation/index.tsx
compulim: I am unsure if `useState` allow empty args for setter, but it will be cleaner if we can use `setCodeRange()` instead.

Composer/packages/client/src/pages/language-generation/index.tsx
compulim: Use `useCallback`.

Can we name this `onMount`?

Composer/packages/client/src/pages/notifications/NotificationList.tsx
```diff
@@ -23,7 +24,7 @@ const columns: IColumn[] = [
     minWidth: 30,
     maxWidth: 30,
     onRender: (item: INotification) => {
-      return <FontIcon iconName={icons[item.type].iconName} css={typeIcon(icons[item.type])} />;
+      return <FontIcon iconName={icons[item.severity].iconName} css={typeIcon(icons[item.severity])} />;
```
compulim: Cache `icon[item.severity]` so we don't need to access it twice.

Composer/packages/client/src/pages/language-generation/index.tsx
lei9444: The codeRange' type was defined in lgEditor as Partial<ICodeRange> | -1 | undefined. So I use undefined to replace null.

lei9444: I find that the setCodeRange doesn't support empty args.

lei9444:
Hi, @cwhitten @compulim Thanks for your review. I have resolved some of these comments. And I will add dialog's error link in this PR.

Composer/packages/client/src/pages/language-generation/index.tsx
```diff
@@ -215,7 +228,7 @@ const LGPage: React.FC<RouteComponentProps> = props => {
           <div css={contentEditor}>
             {editMode ? (
               <Suspense fallback={<LoadingSpinner />}>
-                <CodeEditor file={lgFile} codeRange={codeRange} onChange={onChange} />
+                <CodeEditor file={lgFile} codeRange={codeRange} onChange={onChange} onMount={setLgEditor} />
```
zhixzhan: how about  do this in code-editor.tsx? it read url hash,  let index.tsx  aware `#line1`  logic from editor may not be necessary ?

Composer/packages/client/src/ShellApi.ts
boydc2014: here the templates is used to generate a "view" for form editor and\or other editors, the original data is not touched and can be accessed via LG all up view. 

so this `parsing` solely won't cause data loss, there is a potential loss if the upper layer (editors) think there is no templates and try to reset the whole file.  This is a typical risk, it's like you open a word document, due to bad parse the data is not loaded into your view, you choose to clean up and re-type your content.  The fix is to make sure the parse is solid.

lei9444:
update the UI 
if bot has errors, the start button will disabled
image

Composer/packages/lib/indexers/src/dialogIndexer.ts
```diff
@@ -38,14 +38,22 @@ function ExtractLgTemplates(dialog): string[] {
...
       targets.forEach(target => {
-        templates.push(...extractLgTemplateRefs(target).map(x => x.name));
+        // match a template name match a temlate func  e.g. `showDate()`
+        // eslint-disable-next-line security/detect-unsafe-regex
+        const reg = /\[([A-Za-z_][-\w]+)(\(.*\))?\]/g;
+        let matchResult;
+        while ((matchResult = reg.exec(target)) !== null) {
+          const templateName = matchResult[1];
+          templates.push(templateName);
+        }
```
a-b-r-o-w-n: Why don't we want to use `extractLgTemplateRefs`?

Composer/packages/lib/shared/src/types/shell.ts
```diff
@@ -1,38 +1,10 @@
...
-  isRoot: boolean;
-  lgFile: string;
-  lgTemplates: string[];
-  luFile: string;
-  luIntents: string[];
-  referredDialogs: string[];
-  relativePath: string;
-  triggers: ITrigger[];
-}
+export * from '@bfc/indexers/lib/type';
```
a-b-r-o-w-n: Is this here just reduce churn in this PR? Otherwise, why not have the consuming packages just import the types directly?

Composer/packages/server/__tests__/models/bot/botProject.test.ts
```diff
@@ -228,11 +228,14 @@ describe('lu operation', () => {
...
-  it('should throw error when lu content is invalid', async () => {
+  it('should update diagnostics when lu content is invalid', async () => {
     const id = 'root';
     const content = 'hello \n hello3';
-
-    await expect(proj.updateLuFile(id, content)).rejects.toThrow();
+    const luFiles = await proj.updateLuFile(id, content);
+    const result = luFiles.find(f => f.id === id);
+    if (result !== undefined) {
+      await expect(result.diagnostics.length).toBeGreaterThan(0);
```
a-b-r-o-w-n: This does not need to be `await`ed.

a-b-r-o-w-n: You can also drop the `result !== undefined` and do `expect(result?.diagnostics?.length).toBeGreaterThan(0)` (typescript 3.7)

Composer/packages/server/src/models/bot/botProject.ts
```diff
@@ -3,9 +3,9 @@
 
 import fs from 'fs';
 
-import isEqual from 'lodash/isEqual';
 import { FileInfo, DialogInfo, LgFile, LuFile, getNewDesigner } from '@bfc/shared';
-import { dialogIndexer, luIndexer, lgIndexer } from '@bfc/indexers';
+import { dialogIndexer, lgIndexer } from '@bfc/indexers';
+import { luIndexer } from '@bfc/indexers/lib/luIndexer';
```
a-b-r-o-w-n: Why does this need to be imported via lib, and the others just through the package root?

Composer/packages/client/src/styles.ts
```diff
@@ -81,7 +81,18 @@ export const bot = css`
 `;
 
 export const botButton = css`
-  margin-left: 15px;
+  margin-left: 5px;
+`;
+
+export const errorButton = css`
+  color: #a80000;
```
a-b-r-o-w-n: Can we use a value that is exported from the fabric package? Like `SharedColors` or something?

Composer/packages/client/src/pages/language-generation/index.tsx
```diff
@@ -112,7 +122,7 @@ const LGPage: React.FC<RouteComponentProps> = props => {
 
   function onSelect(id) {
     if (id === '_all') {
-      navigateTo('/language-generation');
+      navigateTo('/language-generation/');
```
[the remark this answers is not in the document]
lei9444: Hi, andy. The left nav button use 
`        getProps={(props: LinkGetProps) => {
          const isActive = exact ? props.isCurrent : props.isPartiallyCurrent;
          setActive(isActive);
          return {};
        }}`
to set active style. The path should keep the some. But I think I should edit the links in App.tsx.
I want to fix this bug:
image

Composer/packages/extensions/obiformeditor/src/Form/fields/RecognizerField/index.tsx
```diff
@@ -142,12 +154,7 @@ export const RecognizerField: React.FC<FieldProps<MicrosoftIRecognizer>> = props
           {() => {
             if (selectedFile && isLuFileSelected) {
               const updateLuFile = (newValue?: string): void => {
-                shellApi
-                  .updateLuFile({ id: selectedFile.id, content: newValue })
-                  .then(() => setErrorMsg(''))
-                  .catch(error => {
-                    setErrorMsg(error);
-                  });
+                shellApi.updateLuFile({ id: selectedFile.id, content: newValue }).catch(setErrorMsg);
```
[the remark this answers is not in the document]
lei9444: I have put all error message to file.diagnostics and use `useEffect(()=>{}, [selectedFile])` to check error. Do you think it's reasonable to do like that?

Composer/packages/lib/indexers/src/dialogIndexer.ts
lei9444: I have synced with @yeze322 @zhixzhan. I can implement string => name function separately in indexer.
We want to make indexer independent. If use `extractLgTemplateRefs`, indexer will depend on shared.

@boydc2014 Do you have any comments about this.

Composer/packages/server/__tests__/models/bot/botProject.test.ts
lei9444: I will upgrade typescript' version to 3.7.2 in package.json

lei9444: I found that you have already upgraded the version

Composer/packages/lib/indexers/src/dialogIndexer.ts
boydc2014: My suggestion is move some utils in shared into indexers if don't want indexers depends on shared, and re-export in shared for any upper layer using shared. 

To clarify, the reason we don't want indexers depends on shared, we want the vice-versa, is that indexers (including the validation inside) can be used by command line tool like "DialogLint". It's would be good we keep the pure "Declartive" world's concept in indexers. and let "shared" package be a shared package in Composer's world.

boydc2014: Given "extractLgTemplateRefs" is only used here, you can just move it here without re-export, then you need to move "LgTemplatRef" interface here and re-export.

Composer/packages/server/src/models/bot/botProject.ts
lei9444: Hi, @a-b-r-o-w-n , the ludown use `#!/usr/bin/env node` in the code. It can't be used in the client. so I separate them. In #1608 , I will convert ludown to bf-lu. This will solve this behavior.

lei9444:
Hi, @cwhitten @a-b-r-o-w-n @boydc2014, I have added some e2e test in this PR now. The lufile's diagnostic is a string list. I will update it in PR #1608.

Composer/packages/lib/indexers/src/dialogIndexer.ts
lei9444: I have moved all lgutils to indexers.

cwhitten:
hey @lei9444, when I have an error in the Condition for the Intent form that requires an expression, I don't see it show up as a line item in the Notifications page

lei9444:
> hey @lei9444, when I have an error in the Condition for the Intent form that requires an expression, I don't see it show up as a line item in the Notifications page

Hi, @cwhitten only these types' expression are checked now:
image

I can add more tomorrow.

cwhitten:
> > hey @lei9444, when I have an error in the Condition for the Intent form that requires an expression, I don't see it show up as a line item in the Notifications page
> 
> Hi, @cwhitten only these types' expression are checked now:
> image
> 
> I can add more tomorrow.

we should plum the schema in to this logic now  and let us drive which validation rules we apply to a field.

lei9444:
> > > hey @lei9444, when I have an error in the Condition for the Intent form that requires an expression, I don't see it show up as a line item in the Notifications page
> > 
> > 
> > Hi, @cwhitten only these types' expression are checked now:
> > image
> > I can add more tomorrow.
> 
> we should plumb the schema in to this logic now and let us drive which validation rules we apply to a field.

@cwhitten @a-b-r-o-w-n @boydc2014 I have added function to scan schema and all $role=expression will do expression check now

Composer/packages/lib/shared/tsconfig.json
```diff
@@ -3,5 +3,5 @@
   "compilerOptions": {
     "outDir": "lib"
   },
-  "include": ["./src/**/*", "./__tests__/**/*"]
+  "include": ["./src/**/*", "./__tests__/**/*", "../indexers/__test__/lgUtils"]
```
a-b-r-o-w-n: This shouldn't be included in this package's tsconfig.

cwhitten:
hey @lei9444,

Lets make the following adjustments to wrap this up:

- the Notifications table should be paginated, 10 row items per page
- the cursor:pointer style is currently only on text, it should be on the entire row item
- A single click on the row item should navigate to the appropriate page. Currently it takes a double-click.
- the dropdown selection should not be filterable by file, it should be filterable by type: `All`, `Errors`, `Warnings`
- In the row item text, we should not show the `Microsoft.TextInput` label, but use the `ConceptLabels` map from `shared`
- There are many "Warnings" associated with empty fields that are not required. These should not be warnings. On empty fields that are **required**, we should be rendering an Error. Generally, we should have very little to no "Warnings".
- Item overflow is adjusting the table header layout slightly, if you have many items it will push the table component up. It should be fixed in place.

Composer/packages/lib/indexers/src/dialogUtils/types.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export class ValidateFields {
+  constructor(expressions: ISearchResult) {
+    this.expressions = expressions;
+  }
+
+  public expressions: ISearchResult;
+
+  getDialogChecker(): { [key: string]: CheckerFunc[] } {
```
boydc2014: is mean? getFieldsCheckers?

lei9444:
-[ ] the Notifications table should be paginated, 10 row items per page
-[ ] the cursor:pointer style is currently only on text, it should be on the entire row item
-[ ] A single click on the row item should navigate to the appropriate page. Currently it takes a double-click.
-[ ] the dropdown selection should not be filterable by file, it should be filterable by type: `All`, `Errors`, `Warnings`
-[ ] In the row item text, we should not show the `Microsoft.TextInput` label, but use the `ConceptLabels` map from `shared`
-[ ] There are many "Warnings" associated with empty fields that are not required. These should not be warnings. On empty fields that are **required**, we should be rendering an Error. Generally, we should have very little to no "Warnings".
-[ ] Item overflow is adjusting the table header layout slightly, if you have many items it will push the table component up. It should be fixed in place.

lei9444:
image
Pagination is added to notification list now

cwhitten:
@lei9444 check out the following video: https://share.getcloudapp.com/Z4uwPKyg

I think we're still over-reporting empty fields as errors. The video shows me supplying incorrect expression syntax, but 4 errors being reported for it instead of 1. The 3 additional errors are the empty field errors that shouldn't be errors at all.

lei9444:
> @lei9444 check out the following video: https://share.getcloudapp.com/Z4uwPKyg
> 
> I think we're still over-reporting empty fields as errors. The video shows me supplying incorrect expression syntax, but 4 errors being reported for it instead of 1. The 3 additional errors are the empty field errors that shouldn't be errors at all.

Hi, @cwhitten , It's a bug in checker function, I have fixed it now. I forget to clean the message and use the error message for all properties.

PR #1668, opened 2019-11-27
fix: location select Content component

### LINKED ISSUE #1291: UX - Templates should get a default name

### LINKED ISSUE #1603: unable to enter custom name for new bot

liweitian:
## Description
Refactor the locationSelectContent part to make it more pure. 

The overall idea/code flow becomes. Every time we changed the location(the last accessed path), it will save to data.json, then the storages in stores(global state) will be changed too. This will trigger updating the focuedStorageFolder.

## Task Item
closes #1603 
closes #1291

## Type of change

## Checklist


Composer/packages/client/src/CreationFlow/DefineConversation/index.js
```diff
@@ -59,6 +75,22 @@ export function DefineConversation(props) {
...
+    let i = 0;
+    const bot = templateId || botName;
+    let defaultName = `${bot}-${i}`;
+
+    while (
+      bots.findIndex(bot => {
+        return bot.name === defaultName;
+      }) > -1
+    ) {
+      i = i + 1;
```
compulim:
```suggestion
      i++;
```

Composer/packages/client/src/CreationFlow/DefineConversation/index.js
```diff
@@ -59,6 +75,22 @@ export function DefineConversation(props) {
...
 
+  function getDefaultName() {
+    let i = 0;
+    const bot = templateId || botName;
+    let defaultName = `${bot}-${i}`;
+
+    while (
+      bots.findIndex(bot => {
+        return bot.name === defaultName;
+      }) > -1
```
compulim: The better way for determining whether `index === -1` or not, is using `~index`. It's called bitwase-negate. Since number in JavaScript is "two's complement", negating -1 will be 0.

```js
~(-1) === 0 //falsy
~(0) is truthy
~(1) is truthy, etc
```

Composer/packages/client/src/CreationFlow/DefineConversation/index.js
```diff
@@ -59,6 +75,22 @@ export function DefineConversation(props) {
     });
   };
 
+  function getDefaultName() {
+    let i = 0;
+    const bot = templateId || botName;
+    let defaultName = `${bot}-${i}`;
+
+    while (
+      bots.findIndex(bot => {
```
compulim: Add a maximum retry for defensive programming reason. I.e. if index > 10000, we shouldn't proceed and throw an error. Otherwise, we will hang the app.

Composer/packages/client/src/CreationFlow/DefineConversation/index.js
```diff
@@ -59,6 +75,22 @@ export function DefineConversation(props) {
     });
   };
 
+  function getDefaultName() {
+    let i = 0;
+    const bot = templateId || botName;
```
compulim: We are constructing the bot name in two path.

You should use a do-while loop, which should eliminate constructing twice.

Composer/packages/client/src/CreationFlow/DefineConversation/index.js
```diff
@@ -109,7 +136,14 @@ export function DefineConversation(props) {
             />
           </StackItem>
         </Stack>
-        {enableLocationBrowse && <LocationSelectContent onChange={updateLocation} allowOpeningBot={false} />}
+        {enableLocationBrowse && (
+          <LocationSelectContent
+            updateCurrentPath={updateCurrentPath}
+            allowOpeningBot={false}
```
compulim: If we don't have path for `true`, let's remove this prop and remove unneeded code paths.

Composer/packages/client/src/CreationFlow/LocationBrowser/LocationSelectContent.tsx
```diff
@@ -2,60 +2,19 @@
...
-
 import { jsx } from '@emotion/core';
-import { Fragment, useEffect, useState, useContext, useRef } from 'react';
+import { Fragment, useContext, useRef } from 'react';
 
 import { FileSelector } from './FileSelector';
 import { StoreContext } from './../../store';
 import { FileTypes } from './../../constants';
 
 export function LocationSelectContent(props) {
```
compulim: You can write it like this to simplify the code.

```suggestion
export function LocationSelectContent({ onOpen, allowOpeningBot = true, focusedStorageFolder, updateCurrentPath, currentPath }) {
```

Composer/packages/client/src/CreationFlow/index.js
```diff
@@ -21,8 +21,27 @@ export function CreationFlow(props) {
...
+    openBotProject,
+    createProject,
+    saveProjectAs,
+    saveTemplateId,
+    fetchStorages,
+    fetchFolderItemsByPath,
+  } = actions;
+  const { templateId, templateProjects, focusedStorageFolder, storages } = state;
+  const currentStorageIndex = useRef(0);
+  const currentStorageId = storages[currentStorageIndex.current] ? storages[currentStorageIndex.current].id : 'default';
```
compulim: Cache `storages[currentStorageIndex.current]` so we don't need to get it twice.

In JavaScript, every member accessor is very expensive, unlike C#. This is because JavaScript is prototypical object-oriented model, there are no virtual tables. In contrast, C# is inheritance OO model. So when writing code, try to reduce number of `.` (dots), `[]` (indexer), etc.

Composer/packages/client/src/CreationFlow/index.js
```diff
@@ -21,8 +21,27 @@ export function CreationFlow(props) {
...
+  const [currentPath, setCurrentPath] = useState('');
+  useEffect(() => {
+    if (storages && storages.length > 0) {
```
compulim:
```suggestion
    if (storages && storages.length) {
```

Composer/packages/client/src/CreationFlow/index.js
```diff
@@ -21,8 +21,27 @@ export function CreationFlow(props) {
...
+      const storageId = storages[currentStorageIndex.current].id;
```
compulim: Cache `currentStorageIndex.current`.

Composer/packages/client/src/CreationFlow/index.js
```diff
@@ -21,8 +21,27 @@ export function CreationFlow(props) {
...
+      const path = storages[currentStorageIndex.current].path;
+      const formatedPath = Path.normalize(path);
```
compulim:
```suggestion
      const formattedPath = Path.normalize(path);
```

Composer/packages/client/src/CreationFlow/index.js
```diff
@@ -34,6 +53,9 @@ export function CreationFlow(props) {
     });
 
     setBots(botsInCurrentFolder);
+    if (Object.keys(focusedStorageFolder).length > 0) {
```
compulim:
```suggestion
    if (Object.keys(focusedStorageFolder).length) {
```

Composer/packages/client/src/CreationFlow/index.js
```diff
@@ -66,6 +88,18 @@ export function CreationFlow(props) {
     }
   };
 
+  const updateCurrentPath = async (newPath, storageId) => {
+    if (!storageId) {
+      storageId = currentStorageId;
+    }
+    if (newPath) {
+      // const formatedPath = Path.normalize(newPath.replace(/\\/g, '/'));
```
compulim: Remove commented code.

Composer/packages/client/src/CreationFlow/index.js
```diff
@@ -66,6 +88,18 @@ export function CreationFlow(props) {
     }
   };
 
+  const updateCurrentPath = async (newPath, storageId) => {
+    if (!storageId) {
+      storageId = currentStorageId;
+    }
```
compulim: Looks like `storageId` is no longer being used.

Composer/packages/client/src/CreationFlow/index.js
```diff
@@ -66,6 +88,18 @@ export function CreationFlow(props) {
...
+    if (newPath) {
+      // const formatedPath = Path.normalize(newPath.replace(/\\/g, '/'));
+      const formatedPath = Path.normalize(newPath);
+      //await fetchFolderItemsByPath(storageId, formatedPath);
```
compulim: Remove commented code.

Composer/packages/client/src/store/action/storage.ts
```diff
@@ -95,5 +95,15 @@ export const fetchFolderItemsByPath: ActionCreator = async ({ dispatch }, id, pa
 };
 
 export const updateCurrentPath: ActionCreator = async ({ dispatch }, path) => {
-  await httpClient.put(`/storages/currentPath`, { path: path });
+  try {
+    const response = await httpClient.put(`/storages/currentPath`, { path: path });
```
compulim:
```suggestion
    const response = await httpClient.put(`/storages/currentPath`, { path });
```

Composer/packages/client/src/store/action/storage.ts
```diff
@@ -95,5 +95,15 @@ export const fetchFolderItemsByPath: ActionCreator = async ({ dispatch }, id, pa
...
+    dispatch({
+      type: ActionTypes.GET_STORAGE_SUCCESS,
+      payload: {
+        response,
+      },
+    });
+  } catch (err) {
+    dispatch({ type: ActionTypes.GET_STORAGE_FAILURE, payload: null, error: err });
```
compulim: You can make it simpler by writing it like this.

Also, is `payload: null` required? If we leave it out, it will be more like `payload: undefined`.

```js
  } catch (error) {
    dispatch({ type: ActionTypes.GET_STORAGE_FAILURE, payload: null, error });
```

Composer/packages/server/src/controllers/storage.ts
```diff
@@ -16,8 +16,8 @@ function createStorageConnection(req: Request, res: Response) {
 }
 
 function updateCurrentPath(req: Request, res: Response) {
-  StorageService.updateCurrentPath(req.body.path);
-  res.status(200).json('success');
+  const storages = StorageService.updateCurrentPath(req.body.path);
```
compulim: Prefer not to keep one-off result in a constant. You can just pass it in.

But depends on the team or other code, they may like this one.

Composer/packages/server/src/services/storage.ts
```diff
@@ -96,6 +96,7 @@ class StorageService {
   public updateCurrentPath = (path: string) => {
     this.storageConnections[0].path = path;
     Store.set(this.STORE_KEY, this.storageConnections);
+    return this.storageConnections;
```
compulim: I am okay to return this.storageConnections.

But I think you don't need to return this, because the caller should be able to grab it by themselves.

The better approach, which may not applies to this code, is that. If you have an update function that modify a value and return it out. It is better to treat the value immutable. That means, the function get passed in, will not be changed. The returned value will be a new one with modified content.

Composer/packages/client/src/CreationFlow/DefineConversation/index.js
```diff
@@ -29,11 +30,26 @@ const validateForm = data => {
...
+    enableLocationBrowse,
+    focusedStorageFolder,
+    updateCurrentPath,
+    currentPath,
+    bots,
+  } = props;
+
+  const { state } = useContext(StoreContext);
+  const { storages, storageFileLoadingStatus, templateId, botName } = state;
+  const [formData, setFormData] = useState({ name: getDefaultName(), errors: {} });
```
compulim: Can we split this into `formDataName` and `formDataErrors`?

Composer/packages/client/src/CreationFlow/DefineConversation/index.js
```diff
@@ -29,11 +30,26 @@ const validateForm = data => {
...
   const [disable, setDisable] = useState(false);
 
+  useEffect(() => {
```
compulim: Put this `useEffect` call below `updateForm` function.

Composer/packages/client/src/CreationFlow/DefineConversation/index.js
```diff
@@ -109,7 +136,14 @@ export function DefineConversation(props) {
             />
           </StackItem>
         </Stack>
-        {enableLocationBrowse && <LocationSelectContent onChange={updateLocation} allowOpeningBot={false} />}
+        {enableLocationBrowse && (
+          <LocationSelectContent
+            updateCurrentPath={updateCurrentPath}
```
compulim: Instead of passing a function to child component to use. Consider passing a function that the child component can "fire event to".

Although the code is almost the same, probably just name changes, the philosophy of sending events make your code less prone to bugs, compare to passing a function. This is because less coupling between two components.

Composer/packages/client/src/CreationFlow/index.js
```diff
@@ -21,8 +21,27 @@ export function CreationFlow(props) {
...
+  const currentStorageId = storages[currentStorageIndex.current] ? storages[currentStorageIndex.current].id : 'default';
+  const [currentPath, setCurrentPath] = useState('');
+  useEffect(() => {
+    if (storages && storages.length > 0) {
+      const storageId = storages[currentStorageIndex.current].id;
+      const path = storages[currentStorageIndex.current].path;
+      const formatedPath = Path.normalize(path);
+      fetchFolderItemsByPath(storageId, formatedPath);
+    }
+  }, [storages]);
```
compulim: As `currentStorageIndex` is a ref, if it get changed, it will not run `fetchFolderItemsByPath`.

And also, I scanned the code and it seems `currentStorageIndex` never get changed, it is always `0`.

Composer/packages/client/src/CreationFlow/DefineConversation/index.js
```diff
@@ -11,46 +13,77 @@ import { TextField } from 'office-ui-fabric-react/lib/TextField';
...
+      );
+    }
+
+    if (
+      name &&
+      ~bots.findIndex(bot => {
+        const path = Path.join(focusedStorageFolder.parent, focusedStorageFolder.name, name);
+        return bot.path === path;
+      })
+    ) {
```
a-b-r-o-w-n: const newBotPath = Path.join(focusedStorageFolder.parent, focusedStorageFolder.name, name);

```ts
if (name && bots.find(bot => bot.path === newBotPath) {
```

It's confusing to do a double negation.

Composer/packages/client/src/CreationFlow/DefineConversation/index.js
```diff
@@ -11,46 +13,77 @@ import { TextField } from 'office-ui-fabric-react/lib/TextField';
...
+    focusedStorageFolder,
+    onCurrentPathUpdate,
+    currentPath,
+    bots,
+  } = props;
+
+  const { state } = useContext(StoreContext);
+  const { templateId, botName } = state;
+  const [formData, setFormData] = useState({ name: getDefaultName() });
+  const [formDataErrors, setFormDataErrors] = useState({ errors: {} });
```
a-b-r-o-w-n: `{ errors: {} }` is a different type than what `setFormDataErrors` is being invoked with. You should convert this file to typescript to catch this kind of thing.

Composer/packages/client/src/CreationFlow/DefineConversation/index.js
```diff
@@ -59,28 +92,22 @@ export function DefineConversation(props) {
...
+    const bot = templateId || botName;
+    let defaultName = '';
+
+    do {
+      i++;
+      defaultName = `${bot}-${i}`;
+    } while (
+      ~bots.findIndex(bot => {
+        return bot.name === defaultName;
+      }) &&
```
a-b-r-o-w-n: Again, this double negation is confusing. Just use `find` instead.

Composer/packages/client/src/CreationFlow/DefineConversation/index.js
```diff
@@ -59,28 +92,22 @@ export function DefineConversation(props) {
...
+      i < MAXTRYTIMES
+    );
+    return defaultName;
```
a-b-r-o-w-n: So if `MAXTRYTIMES` is reached without finding a unique name, the default name will be the last computed name and invalid right?

Composer/packages/client/src/CreationFlow/LocationBrowser/LocationSelectContent.tsx
```diff
@@ -2,69 +2,28 @@
...
 import { jsx } from '@emotion/core';
-import { Fragment, useEffect, useState, useContext, useRef } from 'react';
+import { Fragment, useContext, useRef } from 'react';
+
+import { CreationFlowStatus } from '../../constants';
 
 import { FileSelector } from './FileSelector';
 import { StoreContext } from './../../store';
 import { FileTypes } from './../../constants';
+export function LocationSelectContent({ onOpen, focusedStorageFolder, onCurrentPathUpdate, currentPath }) {
```
a-b-r-o-w-n: This should have its props interface defined.

Composer/packages/client/src/CreationFlow/LocationBrowser/LocationSelectContent.tsx
```diff
@@ -2,69 +2,28 @@
...
-
 import { jsx } from '@emotion/core';
-import { Fragment, useEffect, useState, useContext, useRef } from 'react';
+import { Fragment, useContext, useRef } from 'react';
+
+import { CreationFlowStatus } from '../../constants';
 
 import { FileSelector } from './FileSelector';
 import { StoreContext } from './../../store';
 import { FileTypes } from './../../constants';
```
a-b-r-o-w-n: Please add new line between imports and exports.

Composer/packages/client/src/CreationFlow/index.js
a-b-r-o-w-n: This was marked resolved, but was not addressed.

Composer/packages/server/src/services/storage.ts
```diff
@@ -93,9 +93,13 @@ class StorageService {
     }
   };
 
-  public updateCurrentPath = (path: string) => {
-    this.storageConnections[0].path = path;
-    Store.set(this.STORE_KEY, this.storageConnections);
+  public updateCurrentPath = (path: string, storageId: string) => {
+    const index = this.storageConnections.findIndex(s => s.id === storageId);
+    if (~index) {
```
a-b-r-o-w-n: Double negation here again.

Composer/packages/client/src/CreationFlow/DefineConversation/index.js
liweitian: Previously, it will be an endless loop if we cannot find an unique name(probably due to a bug). According to @compulim, we should throw an error when the MAXTIMES reached. But I feel even if this do happen, we should not decrease user's experience. This is a small feature. So I keep the code this way just like what you said. If the code generate a duplicated name, an error message will show.

Issue #1669, opened 2019-11-27
Running docker-composer up currently fails


benbrown:
## Describe the bug
Currently, docker-compose up results in a failure with a SIGKILL during the build process.

<img width="821" alt="Screen Shot 2019-11-27 at 12 07 45 PM" src="https://user-images.githubusercontent.com/729873/69748772-adb8b880-110e-11ea-8ce3-9b653ca2518b.png">

## Version
current stable

## OS 
What operating system are you using?

- [x] macOS

## To Reproduce

1. Clone repo
2. Amend Composer/Dockerfile with `ENV NODE_OPTIONS "--max-old-space-size=4096"`
3. Run docker-composer up
4. Error!

## Expected behavior
I would expect the build to complete and for Composer to run.

## Additional context
This has been reported by a customer running Windows locally as well.

cwhitten:
addressed in https://github.com/microsoft/BotFramework-Composer/pull/1709

PR #1670, opened 2019-11-27
fix: Provide more memory to Node in Docker

benbrown:
## Description

Add `ENV NODE_OPTIONS "--max-old-space-size=4096"` to the Dockerfile to prevent Node from running out of memory inside the container

## Task Item

refs #1669 

## Type of change

## Checklist


boydc2014:
I did some investigation on this previously, and i try to set --max-old-space-size to `2048` because that's the  default size of the allocated memory for docker-compose builder container. and i even don't have a way to change that. 

Any size bigger than that, like 4096, will result in a OOF,  because the container is asking more, but the system refuse to provide.  

So @benbrown can you test with another machine? see if this will work?

benbrown:
@boydc2014 Are you suggesting that I need to test this on a machine with more RAM?

boydc2014:
> @boydc2014 Are you suggesting that I need to test this on a machine with more RAM?

I was suggesting testing with a machine with less RAM, :) . I think i figured out 
image

benbrown:
I don't understand how computers work anymore.

PR #1674, opened 2019-11-29
chore: add startup script to check for oudated versions

cwhitten:
refs #1674

This change introduces a script to parse the local git branch and use it to do sha inspection of the corresponding remote branch. I've added a `RELEASE_BRANCH` guard to avoid unnecessary checks in development mode. This seems appropriate for now.

I've added execution of the script inside both `build` and `startall`.

Screen Shot 2019-11-29 at 12 54 09 PM

Screen Shot 2019-11-29 at 12 53 48 PM

### DIFF
--- a/Composer/package.json
+++ b/Composer/package.json
@@ -19,7 +19,7 @@
     "packages/tools/language-servers/*"
   ],
   "scripts": {
-    "build": "node scripts/begin.js && node scripts/update.js && yarn build:prod",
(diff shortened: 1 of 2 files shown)


boydc2014:
> 💯 Only thought is why not prompt and update on the user's behalf?

because it might break user's current bots? :)

cwhitten:
> > 💯 Only thought is why not prompt and update on the user's behalf?
> 
> because it might break user's current bots? :)

Yes, this was my thinking.

PR #1677, opened 2019-12-02
test: add unit test

liweitian:
## Description
1. We have already test shellAPI by register and call the registered function, The only difference is: in our code the function is an action. So I think we should test action if needed rather than shellAPI.
2. In order to test createProject function in controllers/project, I changed code base a little bit, including two paths pass into assetManager and some checking code in assetManager.ts. The issues are interesting there.
    2.1 We use Path.resolve to change a relative path to an absolute path, something like Path.solve(./assets). We were expecting the path is calculated from node server directory: composer/package/server. However, when we use jest. The server's directory is depend on where you run yarn test, which cause the templates and runtimeTemplate fail to load.
   2.2 the assetManager.ts constructor is an async function but we can await it. In jest, the createProject function will be called before the constructor finish, which causes the templates and runtimeTemplate fail to load.
   
## Task Item

Closes #1677 

## Type of change

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works

## Screenshots


Composer/packages/client/__tests__/components/home.test.js
```diff
@@ -0,0 +1,88 @@
...
+    ];
+    const onClickTemplate = jest.fn(item => item);
+    const { container, getByText } = render(<ExampleList onClick={onClickTemplate} examples={templates} />);
+    expect(container).toHaveTextContent('Echo Bot');
+    const link = getByText('Echo Bot');
+    fireEvent.click(link);
+    expect(onClickTemplate.mock.results[0].value).toBe('EchoBot');
+  });
+
+  it('should render the tool bar', () => {
```
compulim:
```suggestion
  it('should call onClick handler when clicked', () => {
```

Composer/packages/client/__tests__/components/home.test.js
```diff
@@ -0,0 +1,88 @@
...
+      <RecentBotList onSelectionChanged={onSelectionChanged} recentProjects={recentProjects} />
+    );
+    expect(container).toHaveTextContent('a');
+    expect(container).toHaveTextContent('b');
+    const link = queryByLabelText('a');
+    fireEvent.click(link);
+    expect(onSelectionChanged.mock.results[0].value).toBe('path1');
+  });
+
+  it('should render the Example List', () => {
```
compulim:
```suggestion
  it('should dispatch onClick event when clicked on an ExampleList', () => {
```

Composer/packages/client/__tests__/components/home.test.js
```diff
@@ -0,0 +1,88 @@
...
+// Licensed under the MIT License.
+
+import * as React from 'react';
+import { fireEvent, render } from 'react-testing-library';
+
+import { RecentBotList } from '../../src/pages/home/RecentBotList';
+import { ExampleList } from '../../src/pages/home/ExampleList';
+import { ToolBar } from '../../src/components/ToolBar/index';
+describe('<Home/>', () => {
+  it('should render the recent bot list', () => {
```
compulim:
```suggestion
  it('should dispatch onSelectionChanged event when clicked on a link on <RecentBotList>', () => {
```

Composer/packages/server/__tests__/controllers/project.test.ts
```diff
@@ -0,0 +1,149 @@
...
+    Store: {
+      get: (key: string) => {
+        return data[key];
+      },
+      set: (key: string, value: any) => {
+        data[key] = value;
+      },
+    },
+  };
+});
```
a-b-r-o-w-n: This should be extracted to a mock file instead. https://jestjs.io/docs/en/manual-mocks#mocking-user-modules

Composer/packages/server/__tests__/mocks/samplebots/bot2/ComposerDialogs/Main.dialog
```diff
@@ -0,0 +1,39 @@
+{
```
a-b-r-o-w-n: Are these files generated during tests? If they are, they should be cleaned up as part of the test run as well as ignored from git. 

If they are not generated, we should move them to a `fixtures` directory instead of `mocks`.

PR #1707, opened 2019-12-05
fix: visual editor lg template don't show

### LINKED ISSUE #1700: lg content does not show in **Send a response** action node
## Describe the bug
I authored lg inline, but the content does not show in **Send a response** action node. 

## Screenshots
image

zhixzhan:
## Description

In #1700 , visual editor lg templates were not shown because of the `getLgTemplate` api in ShellApi was broken. ~This PR fixes this issue by adding a mapper inside 'getLgTemplate' for converting the LgTemplate object with prototype to a plain object.~

~According to the [The structured clone algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm), Function objects cannot be duplicated accross windows; attempting to do so will throw a DATA_CLONE_ERR exception.~

fixes this by get lg templates from shellData.

## Task Item

Fix #1700

## Type of change

## Checklist

## Screenshots


Composer/packages/client/src/ShellApi.ts
```diff
@@ -132,8 +132,14 @@ export const ShellApi: React.FC = () => {
...
-
-    return lgUtil.parse(file.content);
+    return lgUtil.parse(file.content).map(template => {
+      const { name, parameters, body } = template;
+      return {
+        name,
+        parameters,
+        body,
+      };
+    });
```
a-b-r-o-w-n: This is not clear to me why this would make any difference?

zhixzhan: Raw parse result contains complicate class `LGTemplate` not only just `{ name, parameters, body }` but also has some prototype methods cannot be cloned and send by browser post message api. And data consumer may don't need it, so here filter them.

zhixzhan: By the way, `templates` data is already provided by `getData` api, maybe there do not have a very strong reason to add this `getLgTemplates`  method in shellApi. @yeze322

boydc2014: Good catch @zhixzhan , please confirm this, if not necessary, we can just remove it.

yeze322: @a-b-r-o-w-n https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm here is the reference zhixiang sent to me about copy behavior in postMessage

zhixzhan:
@yeze322 I removed `getLgTemplates` method from shellApi, then get `templates` from existed `shellData`, can you take a look, what if i missed something ?

yeze322:
> @yeze322 I removed `getLgTemplates` method from shellApi, then get `templates` from existed `shellData`, can you take a look, what if i missed something ?

Looks good to me. `ShellData` in visual editor is already outdated, it's a good change.

PR #1713, opened 2019-12-06
build: add ability to configure runtime path

a-b-r-o-w-n:
refs #1669 

* Fixes issue when creating bots in docker.
* Adds a very simple e2e health check that starts docker compose and hits localhost:3000

I attempted to run the build on all platforms but mac and windows had errors (unrelated to the docker-compose bit itself).


boydc2014:
Looks good, verified on my side and on @liweitian.

PR #1714, opened 2019-12-06
build: add eslint-plugin-node rulesets to yarn lint

hailiu2586:
## added eslint-plugin-node ruleset to yarn lint

1. This allows yarn lint to scan for missing entry in package.json from use of 'import ...' in the source code (rule 'node/no-extraneous-import')

1. fixed the reported rule violations in tools and extensions by looking up the missing entry from yarn.lock and add those the corresponding package.json

1. added inline suppression of no-extraneous-require for absh specific command

The net result is that 'npm install --production' will work against packages/server/package.json, (after transform '@bfc/...: *' to '@bfc/..: {path to local .tgz}'

refs #1778

## Type of change

- [x ] New feature (non-breaking change which adds functionality)

## Checklist

- [x ] I have performed a self-review of my own code
- [x ] I have functionally tested my change

## Screenshots


a-b-r-o-w-n:
@hailiu2586 I recently removed the need for `prepare-prod.js` and instead stopped fighting against yarn workspaces. Checkout the updated Dockerfile if you are interested. That being said, since we want to support the ABS-H deploy, we will be happy to accept some platform changes. I'll need to consider the botbuilder-expression change from a semver to a uri resolution. As for what to do with `prepare-prod.js`, is it possible for this not to be included in the Composer repo?

boydc2014:
> @hailiu2586 I recently removed the need for `prepare-prod.js` and instead stopped fighting against yarn workspaces. Checkout the updated Dockerfile if you are interested. That being said, since we want to support the ABS-H deploy, we will be happy to accept some platform changes. I'll need to consider the botbuilder-expression change from a semver to a uri resolution. As for what to do with `prepare-prod.js`, is it possible for this not to be included in the Composer repo?
> 
> EDIT:
> I forgot to mention that the botbuilder-expression package should be published to npm as part of the R7 release, so maybe we can hold on that change until that occurs and the problem is solved?

Yes, expression-library will goes into npm in 1 day or 2. Npm will be used as release version, but i'm expecting composer will somehow still need daily package from myget, and the correct package name for that is "botframework-expression" (so the dependency in indexers needs to be corrected) and the suggested way to refer a daily package, is through yarn.lock, as we did previously, and as Andy copied into docker.

zhixzhan:
> > @hailiu2586 I recently removed the need for `prepare-prod.js` and instead stopped fighting against yarn workspaces. Checkout the updated Dockerfile if you are interested. That being said, since we want to support the ABS-H deploy, we will be happy to accept some platform changes. I'll need to consider the botbuilder-expression change from a semver to a uri resolution. As for what to do with `prepare-prod.js`, is it possible for this not to be included in the Composer repo?
> > EDIT:
> > I forgot to mention that the botbuilder-expression package should be published to npm as part of the R7 release, so maybe we can hold on that change until that occurs and the problem is solved?
> 
> Yes, expression-library will goes into npm in 1 day or 2. Npm will be used as release version, but i'm expecting composer will somehow still need daily package from myget, and the correct package name for that is "botframework-expression" (so the dependency in indexers needs to be corrected) and the suggested way to refer a daily package, is through yarn.lock, as we did previously, and as Andy copied into docker.

fixed in https://github.com/microsoft/BotFramework-Composer/pull/1737

Composer/.eslintrc.js
```diff
@@ -105,5 +117,11 @@ module.exports = {
         'security/detect-unsafe-regex': 'off',
       },
     },
+    {
+      files: ['__tests__/**/*'],
+      rules: {
+        'node/no-missing-import': 'off'
+      }
+    }
```
a-b-r-o-w-n: This should move to the block above. Already takes care of test overrides.

Composer/packages/client/.eslintrc.js
```diff
@@ -4,4 +4,15 @@ module.exports = {
...
     tsconfigRootDir: __dirname,
   },
+  rules: {
+    'node/no-extraneous-import': 'off',
+  },
+  overrides: [
+    {
+      files: ['__tests__/**/*.js'],
+      rules: {
+        'node/no-missing-import': 'off',
```
a-b-r-o-w-n: This could be removed.

Composer/.eslintrc.js
hailiu2586: merged above

Composer/packages/client/.eslintrc.js
hailiu2586: removed and added override about 'jest.d.ts' to the .eslintrc.js at the root

Composer/packages/server/src/server.ts
```diff
@@ -136,13 +127,21 @@ function launchLanguageServer(socket: rpc.IWebSocket) {
   server.start();
 }
 
-attachLSPServer(wss, server, `${BASEURL}/lg-language-server`, webSocket => {
-  // launch language server when the web socket is opened
-  if (webSocket.readyState === webSocket.OPEN) {
-    launchLanguageServer(webSocket);
-  } else {
-    webSocket.on('open', () => {
+configAppInsightsAsNeeded().then(() => {
```
hailiu2586: @a-b-r-o-w-n  does this refactor meet your requirement on how app insights get configured?

a-b-r-o-w-n: I was thinking we could add that in a separate pull request.

a-b-r-o-w-n: ping @hailiu2586

PR #1756, opened 2019-12-11
chore: refactor lg page route & url

### LINKED ISSUE #1681: navigate back action for LG table view edit

navigate from dialog flow to LG all up view, do not change the focus. now "All" is selected in lg all up view.
click 3 dot to edit the specific template
click 'back' in browser, it will still in edit mode
click 'back' again, navigate back to dialog flow 
Actually I am expecting to back to the table view first and then back to the dialog flow.
so we'd better refine the navigation path for lg all up view

**Additional context**

zhixzhan:
## Description

URL design:

```
/language-generation/[fileId]/edit?t=[template]#L=[line]
                    \   1   / \ 2 / \    3    / \  4  /

1. lg(dialog) file id, if empty, redirect to /common.
2. edit mark, if exist, go to editor, if not, go to list view.
3. template in current lg file, go to inline editor   
4. go to editor at specify line
```

URL examples:

- All up list view
```/language-generation/common```

- Edit common.lg
```/language-generation/common/edit```

- Main.lg  list view
```/language-generation/Main```
- Edit Main.lg
```/language-generation/Main/edit```

- Edit Main.lg on #template-name
```/language-generation/Main/edit?t=template-name```

- Edit common.lg#line 25
```/language-generation/common/edit#L=25```

- Edit Main.lg#line 5
```/language-generation/Main/edit#L=5```

## Task Item

Close #1681 

## Type of change

## Checklist

## Screenshots

preview

Forward & backward inside page:
preview

### CODE REVIEW

**Comment by boydc2014** [comment]
It's good to encode state into URL, so that we can to deep link & jump. 

A few suggestions, not sure all valid, just my preferences
1. I prefer more explicit url than "if something not present then default to xx", which means
    * I would like to see common in lg url, if we want to encode fileid in url, no default to common behavior
    * I would prefer to use /language-generation/<fileid>/view to distinguish from "/edit", no default to view.
2. What happened if i both specific t=templateName and #L34?

**Comment by zhixzhan** [comment]
> 2. What happened if i both specific t=templateName and #L34?

@boydc2014 according to URL schema, `?search parameters` have more weight than `#fragment`， so it will go to inline template editing page, then this page do not take `#L34`, then it would be ignored.

File: Composer/packages/client/src/pages/language-generation/code-editor.tsx [hunk 1]
```diff
@@ -7,51 +7,79 @@ import { LgEditor, LGOption } from '@bfc/code-editor';
...
+  const { actions, state } = useContext(StoreContext);
+  const { lgFiles } = state;
+  const file = lgFiles.length ? lgFiles[0] : null;
   const [diagnostics, setDiagnostics] = useState(get(file, 'diagnostics', []));
   const [content, setContent] = useState('');
   const [errorMsg, setErrorMsg] = useState('');
   const [lgEditor, setLgEditor] = useState<editor.IStandaloneCodeEditor | null>(null);
 
-  const fileId = file && file.id;
+  const search = props.location ? props.location.search : '';
```
a-b-r-o-w-n: Could use optional chaining: `props.location?.search || ''`

Composer/packages/client/src/pages/language-generation/code-editor.tsx
```diff
@@ -7,51 +7,79 @@ import { LgEditor, LGOption } from '@bfc/code-editor';
...
+  const templateMatched = /^\?t=([-\w]+)/.exec(search);
```
a-b-r-o-w-n: I think we already include the `query-string` package which can be used instead.

Composer/packages/client/src/pages/language-generation/code-editor.tsx
```diff
@@ -7,51 +7,79 @@ import { LgEditor, LGOption } from '@bfc/code-editor';
...
+  const templateId = templateMatched ? decodeURIComponent(templateMatched[1]) : undefined;
+  const template = templateId && file ? file.templates.find(({ name }) => name === templateId) : undefined;
+
+  const hash = props.location ? props.location.hash : '';
```
a-b-r-o-w-n: optional chaining

Composer/packages/client/src/pages/language-generation/code-editor.tsx
```diff
@@ -7,51 +7,79 @@ import { LgEditor, LGOption } from '@bfc/code-editor';
...
+  const lineMatched = /L(\d+)/g.exec(hash);
```
a-b-r-o-w-n: `query-string`

Composer/packages/client/src/pages/language-generation/code-editor.tsx
```diff
@@ -7,51 +7,79 @@ import { LgEditor, LGOption } from '@bfc/code-editor';
...
+  const line = lineMatched ? +lineMatched[1] : undefined;
+
+  const fileId = (file && file.id) || 'common';
```
a-b-r-o-w-n: optional chaining

Composer/packages/client/src/pages/language-generation/code-editor.tsx
```diff
@@ -7,51 +7,79 @@ import { LgEditor, LGOption } from '@bfc/code-editor';
...
+        : diagnostics;
+
+    const isInvalid = !isValid(currentDiagnostics);
+    const text = isInvalid ? combineMessage(currentDiagnostics) : '';
     setErrorMsg(text);
   }, [diagnostics]);
 
+  const editorDidMount = useCallback((lgEditor: editor.IStandaloneCodeEditor) => {
+    setLgEditor(lgEditor);
+  }, []);
```
a-b-r-o-w-n: Any reason this is `useCallback`? If I recall, this is only called once anyways. What's the perf impact of doing this?

Composer/packages/client/src/pages/language-generation/code-editor.tsx
```diff
@@ -7,51 +7,79 @@ import { LgEditor, LGOption } from '@bfc/code-editor';
...
+
   useEffect(() => {
-    if (lgEditor) {
-      lgEditor.revealLine(line);
+    if (lgEditor && line !== undefined) {
+      window.requestAnimationFrame(() => {
+        lgEditor.revealLine(line);
+        lgEditor.focus();
+        lgEditor.setPosition({ lineNumber: line, column: 1 });
+      });
```
a-b-r-o-w-n: Can you explain why we need raf?

Composer/packages/client/src/pages/language-generation/index.tsx
```diff
@@ -35,27 +31,10 @@ const CodeEditor = React.lazy(() => import('./code-editor'));
...
-      if (match) setLine(+match[1]);
-    }
-  }, [hash]);
-
-  // for now, one bot only have one lg file by default. all dialog share one lg
-  // file.
+  const path = props.location ? props.location.pathname : '';
+  const matched = /^\/language-generation\/(\w+)/.exec(path.replace(/edit(\/)*$/, ''));
+  const fileId = matched ? matched[1] : '';
+  const edit = /edit(\/)*$/.test(path);
```
a-b-r-o-w-n: See previous comments.

Composer/packages/client/src/pages/language-generation/index.tsx
```diff
@@ -202,13 +154,14 @@ const LGPage: React.FC<RouteComponentProps> = props => {
...
-              <Suspense fallback={<LoadingSpinner />}>
-                <CodeEditor file={lgFile} template={inlineTemplate} line={line} />
-              </Suspense>
-            ) : (
-              <TableView file={lgFile} activeDialog={activeDialog} onClickEdit={onTableViewClickEdit} />
-            )}
+            <Suspense fallback={<LoadingSpinner />}>
+              <Router primary={false} component={Fragment}>
+                <CodeEditor path="edit" />
+                <CodeEditor path=":fileId/edit" />
```
a-b-r-o-w-n: We probably just want to wrap the code editor paths with suspense. This may require moving Suspense into the CodeEditor component

Composer/packages/client/src/pages/language-generation/code-editor.tsx
a-b-r-o-w-n: This package probably doesn't work with the fragment format (`Lxx`). Question: why do it that way instead of `L=xx`?

zhixzhan: `#L5` is learned from url fragment, and some github link like 

```https://github.com/microsoft/BotFramework-Composer/blob/7119916f67466ee70f5c48dc36c2071587c5d629/Composer/packages/client/src/pages/language-generation/code-editor.tsx#L40```

```
https://github.com/microsoft/BotFramework-Composer/pull/1756#discussion_r358477500
```
it maybe sampler than `L=xx`

File: Composer/packages/client/src/pages/language-generation/code-editor.tsx (hunk 7 above)
Comment (zhixzhan) [reply Composer/packages/client/src/pages/language-generation/code-editor.tsx]: the purpose here is set cursor to the error line, notice user **fix at here**.

>lgEditor.revealLine(line);

this method scroll specify line into view port, but don't focus on editor

> lgEditor.focus();

this method focus on editor, but no cursor

>lgEditor.setPosition({ lineNumber: line, column: 1 });

this set cursor at error line start.

File: Composer/packages/client/src/pages/language-generation/index.tsx (hunk 9 above)
Comment (zhixzhan) [reply Composer/packages/client/src/pages/language-generation/index.tsx]: Yes, the only component need to be lazy loaded is `<LgEditor >` . but i meet some problem when i fellow those step
 
1. remove `Suspense` wrap in index.tsx, so lazy load shouldn't appear here.
```
// const CodeEditor = React.lazy(() => import('./code-editor'));
import CodeEditor from './code-editor';
```

2. add  `Suspense`  wrap in code-editor.tsx,` and lazy load `<LgEditor >`.
```
// import { LgEditor } from '@bfc/code-editor';
const { LgEditor } = React.lazy(() => import('@bfc/code-editor'));
```
this caused an import error,  after some digging, I found it's `React.lazy` haven't support named  module import. we have to create an intermediate module for reexporting.
I guess it cost more than the benefits.

**Comment by zhixzhan** [comment]
> It's good to encode state into URL, so that we can to deep link & jump.
> 
> A few suggestions, not sure all valid, just my preferences
> 
> 1. I prefer more explicit url than "if something not present then default to xx", which means
>    
>    * I would like to see common in lg url, if we want to encode fileid in url, no default to common behavior

Make sense, now `/` would redirect to `/common`

>    * I would prefer to use /language-generation/fileid/`view` to distinguish from "/edit", no default to view.

I prefer `/fileid` as default page rather than `/fileid/view`, cause `/` already expressed the meaning of view, it can also be thought as the parent route of `/fileid/edit`

> 
> Make the url strict at first would help us at this stage when we are not stable enough to design for "shortcuts". We can have a catch all fallback at high-level, but better not have fallback treatment at too many levels.
> 
> 1. What happened if i both specific t=templateName and #L34?

thought again, i think allow `?t=xx` and `#L` both exist also make sense.
e.g 
```?t=Hello#L3``` can explain as`  open Hello with inline editor, then go to L3 ` . 
if Hello max-line< 3, go to last line.

File: Composer/packages/client/src/pages/language-generation/index.tsx (hunk 9 above)
Comment (a-b-r-o-w-n) [reply Composer/packages/client/src/pages/language-generation/index.tsx]: My comment doesn't make any sense anyways. We are still lazy importing the code editor route component.

File: Composer/packages/client/src/pages/language-generation/code-editor.tsx [hunk 10]
```diff
@@ -3,67 +3,97 @@
...
+  const { lgFiles } = state;
+  const { fileId } = props;
+  const file = lgFiles?.find(({ id }) => id === 'common');
   const [diagnostics, setDiagnostics] = useState(get(file, 'diagnostics', []));
   const [content, setContent] = useState('');
   const [errorMsg, setErrorMsg] = useState('');
   const [lgEditor, setLgEditor] = useState<editor.IStandaloneCodeEditor | null>(null);
 
-  const fileId = file && file.id;
+  const search = props?.location?.search ?? '';
```
a-b-r-o-w-n: Just a nitpick here: `props` is always defined, so no need to optional chain off of it.

Composer/packages/client/src/pages/language-generation/code-editor.tsx
```diff
@@ -3,67 +3,97 @@
...
+  const searchTemplateName = querystring.parse(search).t;
+  const templateId = Array.isArray(searchTemplateName) ? searchTemplateName[0] : searchTemplateName;
+  const template = templateId && file && file.templates.find(({ name }) => name === templateId);
+
+  const hash = props?.location?.hash;
+  const lineMatched = hash && /L(\d+)/g.exec(hash);
```
a-b-r-o-w-n: Can we make the fragment that can be parsed by the query-string package? ex: `#L=50`

Composer/packages/client/src/pages/language-generation/code-editor.tsx
```diff
@@ -115,36 +119,38 @@ const CodeEditor: React.FC<CodeEditorProps> = props => {
...
 
-  const lgOption: LGOption = {
-    inline: inlineMode,
-    content: get(file, 'content', ''),
-    template,
-  };
+  const lgOption = template
+    ? {
+        inline: inlineMode,
+        content: file?.content ?? '',
```
a-b-r-o-w-n: Nice!

Composer/packages/client/src/pages/language-generation/index.tsx
```diff
@@ -35,7 +35,12 @@ const LGPage: React.FC<RouteComponentProps> = props => {
   const matched = /^\/language-generation\/(\w+)/.exec(path.replace(/edit(\/)*$/, ''));
   const fileId = matched ? matched[1] : '';
   const edit = /edit(\/)*$/.test(path);
-  const lgFile = lgFiles.length ? lgFiles[0] : null;
+  const file = lgFiles?.find(({ id }) => id === 'common');
+
+  useEffect(() => {
+    // fileId not found, redirect to common
+    if (!fileId) navigateTo('/language-generation/common');
```
a-b-r-o-w-n: Couple of things:

1. Could this be done with the router instead?
2. If not, we should use something like `replaceState` instead so this uri is not in the history stack.

Composer/packages/client/src/pages/language-generation/code-editor.tsx
zhixzhan: Nice catch, thanks

zhixzhan:
> Only real request is to use a fragment that can be parsed using query-string.

Updated, now use `#L=5` insteadof `#L5` for parsing consideration

boydc2014:
Looks good to me now, #L=5 is a little bit different than GitHub style, but it's OK, it's explicit enough and friendly to parsing. 

Anyway, seems that other comments are also resolved, @a-b-r-o-w-n  can you confirm again?

PR #1777, opened 2019-12-16
feat: Add dynamic choices to Choice Prompt

### LINKED ISSUE #1644: Ability to dynamically set Mulitple Choice options
Currently multiple choice options can only be set in a static format. I manually have to write each option and cannot set the options to an array from an http call.

I would like the ability to make an http call, get back a list or array of values and then ask the user a question in which they would be able to select the option from that result. For example

HTTP call returns ["Red", "Blue", "Green"]
Bot asks user: What is your favorite color?
Options are Red, Blue, and Green

We have considered displaying a string format of options with ID values and requesting the user to input the ID value next to the option they wish.

tdurnford:
## Description
Added dropdown in multiple choice prompt to allow the user to choose between static and dynamic choices.

## Task Item
Closes #1644

## Type of change

## Checklist

## Screenshots

image

image


Composer/packages/extensions/obiformeditor/src/Form/fields/PromptField/ChoiceInput/Choices.tsx
```diff
@@ -3,239 +3,74 @@
...
-  onEdit: (idx: number, value?: IChoice) => void;
-}
+import { StaticChoices } from './StaticChoices';
+import { DynamicChoices } from './DynamicChoices';
 
 interface ChoicesProps {
   id: string;
   schema: JSONSchema6;
-  formData?: IChoice[];
+  formData?: IChoice[] | string;
```
a-b-r-o-w-n: This maybe should be defined in the shared types:

```ts
type IChoice = ObjectChoice[] | string;
```

Choose a better name for `ObjectChoice`, but you get the idea.

Composer/packages/extensions/obiformeditor/src/Form/fields/PromptField/ChoiceInput/DynamicChoices.tsx
```diff
@@ -0,0 +1,26 @@
...
+}
+
+export const DynamicChoices: React.FC<DynamicChoicesProps> = props => {
+  const {
+    formData = '',
+    onChange,
+    schema: { description },
+  } = props;
+  return (
+    <TextField onChange={(_, value = '') => onChange && onChange(value)} value={formData} placeholder={description} />
```
a-b-r-o-w-n: This should be an Expression Widget.

Composer/packages/extensions/obiformeditor/src/Form/fields/PromptField/ChoiceInput/Choices.tsx
```diff
@@ -3,239 +3,74 @@
...
-            root: { margin: '7px 0 7px 0' },
-          }}
-          onBlur={handleBlur}
+    <React.Fragment>
+      <div style={{ display: 'flex', justifyContent: 'space-between', padding: '5px 0' }}>
+        <WidgetLabel
+          label={label}
+          description={formatMessage(
+            "A list of options to present to the user. Synonyms can be used to allow for variation in a user's response."
+          )}
```
a-b-r-o-w-n: This may need to correlate with the choice type because this doesn't make sense with an expression.

Composer/packages/lib/shared/src/types/sdk.ts
```diff
@@ -111,7 +113,7 @@ export interface AttachmentInput extends Partial<InputDialog> {
 export interface ChoiceInput extends Partial<InputDialog> {
   /** The output format. */
   outputFormat?: 'value' | 'index';
-  choices?: IChoice[];
+  choices?: IChoice;
```
a-b-r-o-w-n: Just a thought: What if we had an `AnyOf` type alias for this? I think it would make this easier to connect with the schema.

PR #1786, opened 2019-12-18
feat: update new trigger modal according to design

### LINKED ISSUE #1641: [Inline LU] Create an intent with utterances when creating tigger
When creating a trigger, allow the user to provide a name for the trigger and sample utterances.

https://www.figma.com/file/04DymSq4xElhOBdAyutvvTbq/Composer-SH-06.19-?node-id=1799%3A488

liweitian:
## Description
Update according to new trigger design. Will replace the inline editor after it is done.

## Task Item
Closes #1641 
## Type of change

## Checklist

## Screenshots

Syntax Highlight & Completion
image


Composer/packages/client/src/components/ProjectTree/TriggerCreationModal.tsx
```diff
@@ -71,7 +88,7 @@ export const TriggerCreationModal: React.FC<TriggerCreationModalProps> = props =
   const { state } = useContext(StoreContext);
   const { dialogs, luFiles } = state;
   const luFile = luFiles.find(lu => lu.id === dialogId);
-  const dialogFile = dialogs.find(dialog => dialog.id === dialogId);
+  // const dialogFile = dialogs.find(dialog => dialog.id === dialogId);
```
a-b-r-o-w-n: remove

Composer/packages/client/src/components/ProjectTree/TriggerCreationModal.tsx
```diff
@@ -83,38 +100,39 @@ export const TriggerCreationModal: React.FC<TriggerCreationModalProps> = props =
       });
       return;
     }
+    const newContent = get(luFile, 'content', '') + '\n\r' + '# ' + formData.intent + '\r' + formData.triggerPhrases;
```
a-b-r-o-w-n: Let use the unix line endings instead of carriage return.

cwhitten:
@liweitian I'd like to hold on this until we have the inline monaco editor to provide syntax highlighting and validation.

Composer/packages/client/src/pages/notifications/types.ts
```diff
@@ -1,13 +1,89 @@
...
+  constructor(id: string, location: string, diagnostic: Diagnostic) {
+    this.id = id;
+    this.severity = DiagnosticSeverity[diagnostic.severity] || '';
+    this.message = `In ${replaceDialogDiagnosticLabel(diagnostic.path)} ${diagnostic.message}`;
+    this.diagnostic = diagnostic;
+    this.location = location;
+    this.type = NotificationType.DIALOG;
+    this.dialogPath = diagnostic.path;
+  }
+}
```
yeze322: Seems you need to define a base class `Notification` which implements `INotification` here, and use inheritance to build `Dialog/Lg/Lu Notification`.

Composer/packages/client/src/pages/notifications/types.ts
```diff
@@ -1,13 +1,89 @@
...
+  dialogPath?: string;
+  constructor(id: string, location: string, diagnostic: Diagnostic, luFile: LuFile, dialogs: DialogInfo[]) {
+    this.id = id;
+    this.severity = DiagnosticSeverity[diagnostic.severity] || '';
+    this.message = createSingleMessage(diagnostic);
+    this.diagnostic = diagnostic;
+    this.location = location;
+    this.type = NotificationType.LU;
+    this.dialogPath = this.findDialogPath(luFile, dialogs, diagnostic);
+  }
```
yeze322: Follows my previous comment, with a base class, your constructor will construct base class with `super()`, only one field `dialogPath` needs to be handled here.

Composer/packages/client/src/pages/notifications/types.ts
```diff
@@ -1,13 +1,89 @@
...
+
+  private findDialogPath(luFile: LuFile, dialogs: DialogInfo[], d: Diagnostic) {
+    const intentName = luFile.intents.find(intent => {
+      const { range } = intent;
+      if (!range) return false;
+      return d.range && d.range.start.line >= range.startLineNumber && d.range.end.line <= range.endLineNumber;
```
yeze322: To avoid checking and accessing `d`, `d.range.start.line` (and `end.line`) should live in `findDialogPath`'s context.
Do not put it inside your loop, it's hard to read.

Composer/packages/client/src/pages/notifications/useNotifications.tsx
```diff
@@ -2,14 +2,12 @@
...
 import { useContext, useMemo } from 'react';
-import { createSingleMessage } from '@bfc/indexers';
 import get from 'lodash/get';
 import { LgNamePattern } from '@bfc/shared';
 
 import { StoreContext } from '../../store';
-import { replaceDialogDiagnosticLabel } from '../../utils';
 
-import { INotification, DiagnosticSeverity } from './types';
+import { INotification, DialogNotification, LuNotification, LgNotification } from './types';
```
yeze322: Seems `INotification` is never referneced

Composer/packages/client/src/utils/luUtil.ts
```diff
@@ -20,6 +20,12 @@ export interface LuIntent {
 
 const NEWLINE = '\n';
 
+function isValid(diagnostics: any[]) {
+  return diagnostics.every(item => {
+    item.Severity !== 'ERROR';
```
yeze322: You have defined a `DiagnosticSeverity` constant under `notification/types`.
Use that constant here.

Composer/packages/client/src/utils/luUtil.ts
```diff
@@ -69,8 +75,10 @@ export function updateIntent(content: string, intentName: string, intent: LuInte
   let targetSectionContent;
   const updatedSectionContent = textFromIntent(intent);
   const resource = luParser.parse(content);
-  const { Sections } = resource;
-
+  const { Sections, Errors } = resource;
+  const updateSectionParsed = luParser.parse(updatedSectionContent);
+  // if has error, do nothing.
+  if (isValid(updateSectionParsed.Errors) === false || isValid(Errors) === false) return content;
```
yeze322: if `isValid` returns a boolean, you don't need explicity `=== false` here

Composer/packages/lib/indexers/src/utils/diagnosticUtil.ts
```diff
@@ -44,6 +46,22 @@ export function filterTemplateDiagnostics(diagnostics: Diagnostic[], template: L
   });
 }
 
+export function filterSectionDiagnostics(diagnostics: Diagnostic[], section: LuIntentSection): Diagnostic[] {
+  const { range } = section;
+  if (!range) return diagnostics;
+  const filteredDiags = diagnostics.filter(d => {
+    return d.range && d.range.start.line >= range.startLineNumber && d.range.end.line <= range.endLineNumber;
```
yeze322: I've seen this code before, it's duplicated with some of my previous comments.
Avoid repeating it.

Composer/packages/extensions/obiformeditor/src/Form/widgets/LuEditorWidget.tsx
```diff
@@ -0,0 +1,73 @@
...
+  formContext: FormContext;
+  name: string;
+  height?: number | string;
+  onChange: (template?: string) => void;
+}
+
+export const LuEditorWidget: React.FC<LuEditorWidgetProps> = props => {
+  const { formContext, name, height = 250 } = props;
+  const luFileId = formContext.currentDialog.id;
+  const luFile: LuFile | null = formContext.luFiles.find(f => f.id === luFileId);
```
yeze322: This component has several problems. The usage of `useEffect` hook is anti-pattern meanwhile the data flow is not well designed (as an example, the `onChange` handler submits data to both local states and external API).

The props and local states are mixed together, makes your component hard to maintain.

You really need to rethink what's your scenario is and carefully define the shape of your input props to make sure you can fully leverage change detection mechanism.

Before your modification on this file, do read these 2 docs:
1. https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops
1. https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#preferred-solutions

yeze322: I know state management of Shell has historical problems, consider using a Class Component with lifecycle methods before Leilei's fix on Shell states.

Composer/packages/client/src/utils/luUtil.ts
liweitian: not the same cannot use it here

boydc2014:
Inline lu lsp is in, please merge master and resolve all the comments

Composer/packages/lib/indexers/src/utils/diagnosticUtil.ts
```diff
@@ -28,10 +30,22 @@ export function offsetRange(range: Range, offset: number): Range {
...
+  return diagnosticStartLine >= sectionStartLine && diagnosticEndLine <= sectionEndLine;
+}
+
 export function filterTemplateDiagnostics(diagnostics: Diagnostic[], { range }: { range?: CodeRange }): Diagnostic[] {
   if (!range) return diagnostics;
   const filteredDiags = diagnostics.filter(d => {
-    return d.range && d.range.start.line >= range.startLineNumber && d.range.end.line <= range.endLineNumber;
+    return (
+      d.range &&
+      isDiagnosticWithInRange(d.range.start.line, d.range.end.line, range.startLineNumber, range.endLineNumber)
```
yeze322: Needs a better interface
```ts
function isDiagnosticWIthinRange(diagnostic: Diagnostic, range: CodeRange) {
    // extract line number inside
}
```

Composer/packages/client/src/pages/notifications/useNotifications.tsx
```diff
@@ -53,26 +37,20 @@ export default function useNotifications(filter?: string) {
...
-            id,
-          });
+          notifactions.push(new LgNotification(id, lgTemplateName, location, diagnostic, dialogs));
         });
     });
     return notifactions;
   }, [dialogs, luFiles, lgFiles]);
 
-  const notifications: INotification[] = !filter ? memoized : memoized.filter(x => x.severity === filter);
+  const notifications: Notification[] = !filter ? memoized : memoized.filter(x => x.severity === filter);
```
yeze322: `filter ? memoized.filter() : memoized`

Composer/packages/lib/indexers/src/utils/diagnosticUtil.ts
```diff
@@ -46,6 +60,25 @@ export function filterTemplateDiagnostics(diagnostics: Diagnostic[], { range }:
   });
 }
 
+export function filterSectionDiagnostics(diagnostics: Diagnostic[], section: LuIntentSection): Diagnostic[] {
+  const { range } = section;
+  if (!range) return diagnostics;
+  const filteredDiags = diagnostics.filter(d => {
+    return (
+      d.range &&
+      isDiagnosticWithInRange(d.range.start.line, d.range.end.line, range.startLineNumber, range.endLineNumber)
```
yeze322: update interface

Composer/packages/client/src/pages/notifications/types.ts
```diff
@@ -1,13 +1,89 @@
...
+    this.message = createSingleMessage(diagnostic);
+  }
+
+  private findDialogPath(luFile: LuFile, dialogs: DialogInfo[], d: Diagnostic) {
+    const intentName = luFile.intents.find(intent => {
+      const { range } = intent;
+      if (!range) return false;
+      return (
+        d.range &&
+        isDiagnosticWithInRange(d.range.start.line, d.range?.end.line, range.startLineNumber, range.endLineNumber)
```
yeze322: update interface

yeze322:
All my comments resolved! great 👍

Composer/packages/extensions/obiformeditor/src/Form/widgets/LuEditorWidget.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { FormContext } from '../types';
+
+interface LuEditorWidgetProps {
+  formContext: FormContext;
+  name: string;
+  height?: number | string;
+  onChange: (template?: string) => void;
+}
+
+export class LuEditorWidget extends React.Component<LuEditorWidgetProps> {
```
cwhitten: Let's convert this to a function component and use hooks instead of class methods. 

I believe the `getDerivedStateFromProps` implementation is causing some unexpected behavior in editing the form. For example, data is getting overwritten at times, and I'm unable to add newlines to the bottom of an intent definition like I can in the inline LG editor. The behavior should be almost identical to the inline LG editor. See `LgEditorWidget.tsx` for its implementation. It should be used as a strong reference.

alanlong9278: As @yeze322 said, 
>This component has several problems. The usage of useEffect hook is anti-pattern meanwhile the data flow is not well designed (as an example, the onChange handler submits data to both local states and external API).

>The props and local states are mixed together, makes your component hard to maintain.

>You really need to rethink what's your scenario is and carefully define the shape of your input props to make sure you can fully leverage change detection mechanism.
> Though state management of Shell has historical problems, you can try the Class Component with lifecycle methods before Leilei's fix on Shell states.
> (I was also trapped in this problem before)

There is a bug about luEditor value in `luEditorWidget`: state `localVale` and props `luIntent.body` are mixed. So I try to use class component to fix it. 
I fixed the bug: unable to add newlines to the bottom.

cwhitten: Understood. I cut this: https://github.com/microsoft/BotFramework-Composer/issues/2138 to follow-up.

yeze322: Under current data flow, `getDerivedStateFromProps` can fully support weitian's scenario though it's still not a good pattern.

I think the best practice of this scenario is to use a fully uncontrolled component with a `key` (according to [official doc](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops)). A key here might be the data version of current local state, requires shell or indexer work

PR #1790, opened 2019-12-19
feat: Update LG and built in expression functions in LSP

### LINKED ISSUE #1791: Should update LG and built in expression functions in LSP
Should update LG and built in expression functions in LSP

Danieladu:
## Description

Update LG and built in expression functions in LSP

## Task Item
Closes #1791 

## Type of change

## Checklist

### DIFF
--- a/Composer/packages/tools/language-servers/language-generation/src/builtinFunctionsMap.ts
+++ b/Composer/packages/tools/language-servers/language-generation/src/builtinFunctionsMap.ts
@@ -18,7 +18,7 @@ export class FunctionEntity {
   public Introduction: string;
 }
 
-// todo should add params info and other info og the function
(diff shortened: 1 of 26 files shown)


Composer/packages/tools/language-servers/language-generation/src/builtinFunctionsMap.ts
```diff
@@ -699,6 +684,7 @@ export const buildInfunctionsMap: Map<string, FunctionEntity> = new Map<string,
       'Return the first non-null value from one or more parameters. Empty strings, empty arrays, and empty objects are not null.'
     ),
   ],
+  /* C# only
```
boydc2014: We can comment out those functions out for the reason that, those are not likely to be used, but not for c# only.  
because 1. C# works in Composer, because the runtime composer leveraged today is C#.  2. We should figure out in Expression side, why we are having C# only functions.

PR #1792, opened 2019-12-19
feat: Enable showTypingMiddleware in BotProject

### LINKED ISSUE #1781: We need to be able to tell the user that we are busy with SendActivity of type "typing"
We do not know how to adapt the Bot template generated by Bot composer so that in each action that the Bot performs, a method that sends the typing message event can be called.
We try this on **ComposerBot.cs**:

```
 public override async Task OnTurnAsync(ITurnContext turnContext, CancellationToken cancellationToken = default(CancellationToken))
        {
            **await turnContext.ShowTypingIndicator(100);**

```
But it seems that is not working when one if the actions is busy (for example an Http get request).

We believe that best solution would be to have a method that we could overwrite in the core of the composer, so that each time it was to process an action, this method would be called first. In it we could add "typing" by code.

Another alternative was having an especial Emit Event named "Bot Is Typing" or similar.

**Additional context**

luhan2017:
## Description

In BotProject, enable showTypingMiddleware by default.

## Task Item

fixes #1781 

## Type of change

## Checklist

## Screenshots

### DIFF
--- a/BotProject/CSharp/BotManager.cs
+++ b/BotProject/CSharp/BotManager.cs
@@ -75,6 +75,7 @@ namespace Microsoft.Bot.Builder.ComposerBot.Json
             var userState = new UserState(storage);
             var conversationState = new ConversationState(storage);
             var inspectionState = new InspectionState(storage);
+            var showTypingMiddleware = new ShowTypingMiddleware();
 
             // manage all bot resources
             var resourceExplorer = new ResourceExplorer().AddFolder(botDir);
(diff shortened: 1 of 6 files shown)


a-b-r-o-w-n:
Do users need to be able to opt out of this by some other means rather than changing the runtime?

cwhitten:
@luhan2017 it should also be possible to render the following lg template in a bot response to achieve the same behavior, right?
```
# SendTypingIndicator
[Activity
  Type = "typing"
]
```

luhan2017:
> @luhan2017 it should also be possible to render the following lg template in a bot response to achieve the same behavior, right?
> 
> ```
> # SendTypingIndicator
> [Activity
>   Type = "typing"
> ]
> ```

@cwhitten , it is not that easy, the typing hint shows up depending on when there is an action need to be waited, for example: httprequest, qna request and luis request, it is not so straight forward on how to set that. So the recommendation is setting it in ShowTypingMiddleware, it will automatically take effect on all the responses. Make sense?

luhan2017:
If we don't want to enable it by default. Another option could be we have a setting called "enableShowTypingMiddleware" in the composer setting. and our runtime can decide whether to enable it by that setting.

cwhitten:
@luhan2017 Im aware of the constraint in order for the Activity to show, but I cannot get the Activity driven typing indicator to work at all, and I'd like to see that working. Can you confirm that it does indeed work on a HttpAction and add it to the sample?

luhan2017:
Close this PR because we don't want to enable it by default, had commented in the issue on how to enable the showtyping feature.

PR #1793, opened 2019-12-19
chore: share server's resource to lsp server

### LINKED ISSUE #1678: [LSP]  auto-suggestion for template name/body should be meaningful

Editing in LG all up view, when you type template name/body, it will auto suggest lots of irrelevant info .

- For template name, should not auto-suggest or should auto-suggest based on the context 
   eg. list out the template name during template editing , this could help user to avoid duplicate name definition across different dialog. 
- For body editing, only auto suggest  pre-built function and the predefined template name to make the LG editing more productive.

**Additional context**

zhixzhan:
## Description

some code refine on LSP server.
1. share server resource to LSP server, so the lgFile content don't need pass to LSP server through client, it would improve the consistency of all lg resource.

2. more precisely suggestion according to current editing context
```
> To learn more about the LG file format...       --- COMMENTS
# Greeting                                        --- TEMPLATENAME
- Hello                                           --- TEMPLATEBODY
- Hi, @{name}, what's the meaning of 'state'
         |
         +------------------------------------------- EXPRESSION
```

- COMMENTS, TEMPLATENAME suggest nothing,
- TEMPLATEBODY  suggest keywords (IF ELSEIF etc.) at line start.
- EXPRESSION suggest template & prebuilt-function

## Task Item

close #1678 

## Type of change

## Checklist

## Screenshots

preview


Composer/packages/tools/language-servers/language-generation/src/LGServer.ts
```diff
@@ -14,37 +14,44 @@ import {
...
-
-  constructor(protected readonly connection: IConnection) {
+  /**
+   * LG documents infomation store.
+   * The connection between lgOption and botProjectService.
+   * Help do inline template editing and server resource passing.
+   */
+  protected LGDocuments: LGDocument[] = [];
+
+  constructor(protected readonly connection: IConnection, protected readonly botProjectService?) {
```
boydc2014: I understand the intention is to leverage the API server to get LG file content, instead of passing the whole content directly through LSP protocol. 

Although this is a specific change for Composer, we should seek for a generic solution based on a generic interface, to make sure the modules are better decoupled.

In this case, the suggestion is to let LGServer accept a generic delegate\callback\interface which can get a LGFile given a file id, then implement this delegate using BotProjectService instead of passing it in. 

For example, let's say "FileResolver(id) => string or LGFile". Creating such an interface will force you to think, what's the precise and minimal requirement from LSP to the external world.

zhixzhan: ok, it's a good idea, i will rewrite this part.

boydc2014:
@cwhitten  looks good to me on this one, can you help review again?

PR #1794, opened 2019-12-19
feat: Update OnConversationUpdate actions

### LINKED ISSUE #1632: Expand the OnConversationUpdate trigger template to include a ForEach loop based action
Today, we use a simple template for `OnConversationUpdate` trigger in composer which uses a condition expression to eliminate `Bot` being added to the conversation. We should update this template to include a ForEach loop so each member added tot he conversation can be uniquely identified and greeted.

### LINKED ISSUE #1780: Rewrite greeting activity trigger in order to work with web chat and directline
We found that if we use a bot created using Bot Framework Composer in Web Chat control using DirectLine channel, the greeting is displayed twice. The second time is after user sends first message to the bot. The reason is clear, is related to the double event OnConversationUpdateActivity that is fired when the conversation between bot and user starts, the first when bot is added to conversation and the second when is added the user.
We know that it is a "knowed problem" on BotFramework with directline that is solved in most of templates with this two lines of code:
foreach (var member in membersAdded)
  if (member.Id != turnContext.Activity.Recipient.Id)

I'd like that Bot Framework Composer editor Greeting works by default like in DialogAndWelcome Bot example of Bot Framework SDK 4.

Our first try was to use a distinct value on:
**OnConversationUpdateActivity.condition**
toLower(turn.Activity.membersAdded[0].name) != 'bot'
->
substring(toLower(turn.Activity.membersAdded[0].name), 0, 3) != 'bot'

Because we detect that name is not always "bot" in our case was "bot-xxxx..."
But it seems that condition on OnConversationUpdateActivity is not really working: we print the condition on message to user and it send 'True' and 'False' but is showing twice.

We found a workarround following this steps:

```
· blank trigger OnConversationUpdateActivity condition (because is not working)
· adding "Loop: for each item" with itemsProperty = turn.Activity.membersAdded
· adding a "Branch: if/else" with condition = string(dialog.foreach.value.id) != string(turn.Activity.Recipient.id)
· Send welcome message only if condition is True
```
```
  {
      "$type": "Microsoft.OnConversationUpdateActivity",
      "$designer": {
        "id": "376720"
      },
      "actions": [
        {
          "$type": "Microsoft.Foreach",
          "$designer": {
            "id": "887482",
            "name": "Loop: for each item"
          },
          "itemsProperty": "turn.Activity.membersAdded",
          "actions": [
            {
              "$type": "Microsoft.IfCondition",
              "$designer": {
                "id": "129743",
                "name": "Branch: if/else"
              },
              "condition": "string(dialog.foreach.value.id) != string(turn.Activity.Recipient.id)",
              "actions": [
                {
                  "$type": "Microsoft.SendActivity",
                  "$designer": {
                    "id": "789374",
                    "name": "Send a response"
                  },
                  "activity": "- Welcome, I'm Bot Composer!"
                }
              ]
            }
          ]
        }
      ]
    }
```
This workarround fixes the bug and we think is the equivalent to Bot Framework standard Welcome Bot:
foreach (var member in membersAdded)
  if (member.Id != turnContext.Activity.Recipient.Id)

· We use IDs instead of Names.
· Foreach is better than single condition, because we thinks this will work on other channels with more than user at conversation with the bot like in Teams.

**Additional context**

luhan2017:
## Description

Update OnConversationUpdate conditions, and add a foreach step to loop on MemberAdded.

## Task Item

fixes #1780 
fixes #1632 

## Type of change

## Checklist

## Screenshots

### DIFF
--- a/Composer/packages/lib/shared/src/dialogFactory.ts
+++ b/Composer/packages/lib/shared/src/dialogFactory.ts
@@ -8,7 +8,7 @@ import { appschema } from './appschema';
 import { copyAdaptiveAction } from './copyUtils';
 import { deleteAdaptiveAction, deleteAdaptiveActionList } from './deleteUtils';
 import { MicrosoftIDialog } from './types';
-
+import { SDKTypes } from './types';
 interface DesignerAttributes {
   name: string;
   description: string;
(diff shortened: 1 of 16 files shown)


Composer/packages/lib/shared/src/dialogFactory.ts
```diff
@@ -36,7 +36,36 @@ const initialDialogShape = {
...
   'Microsoft.OnConversationUpdateActivity': {
     $type: 'Microsoft.OnConversationUpdateActivity',
-    condition: "toLower(turn.Activity.membersAdded[0].name) != 'bot'",
+    actions: [
+      {
+        $type: 'Microsoft.Foreach',
+        $designer: {
+          id: nanoid('1234567890', 6),
+          name: 'Loop: for each item',
+        },
```
a-b-r-o-w-n: These should use a combination of `getNewDesigner` and the `ConceptLabels`.

PR #1796, opened 2019-12-23
fix: Create bot from scratch may create a non-empty bot

### LINKED ISSUE #1814: create a new bot
## Describe the bug
1. create a echoBot(will set global state templateId = echoBot)
2. create a new bot from scratch (should create a new empty bot but create a echoBot)

liweitian:
## Description
when the state templateId is set. Create a bot from scratch will use that templateId. But it should be emptyBot.

## Task Item

Closes #1814 
## Type of change

## Checklist

## Screenshots

### DIFF
--- a/Composer/packages/client/src/CreationFlow/CreateOptions/index.js
+++ b/Composer/packages/client/src/CreationFlow/CreateOptions/index.js
@@ -72,10 +72,10 @@ export function CreateOptions(props) {
 
   const handleJumpToNext = () => {
     if (option === 'Create from template') {
-      if (template === null) onNext(templates[0]);
+      if (template === null) onNext(templates[1].key);
       else onNext(template.key);
     } else {
-      onNext(null);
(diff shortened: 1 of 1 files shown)


Composer/packages/client/src/CreationFlow/CreateOptions/index.js
```diff
@@ -72,10 +72,10 @@ export function CreateOptions(props) {
 
   const handleJumpToNext = () => {
     if (option === 'Create from template') {
-      if (template === null) onNext(templates[0]);
+      if (template === null) onNext(templates[1].key);
```
boydc2014: let's make it clear by not using a magic number. 

probably explicitly say `emptyBotId = templates[1].key`

boydc2014: And please investigate, when this template is null, why not we default having a choice set for user.

PR #1797, opened 2019-12-23
feat: setError to notify user trigger name mismatch

### LINKED ISSUE #1643: [Inline LU] handle intent name changes from all up LU editor
If a user edits an intent name in the all up LU editor, update the dialog data (`triggers[OnIntent].trigger`) to reference the new name.

liweitian:
## Description
see task description and the screenshot. It will notify user the trigger name mismatch with the intent name

## Task Item
Closes #1643 

## Type of change

## Checklist

## Screenshots
WeChat Image_20191223180208


boydc2014:
Put this onhold in case it's accidentally merged, because this needs a systematic fix than an ad-hoc fix.

liweitian:
Need to move lu/lg to front-end just like dialogs. Then continue on this PR.

PR #1798, opened 2019-12-23
fix: Check all expressions in dialog

### LINKED ISSUE #1783: [Linting & Validation] expression fields in arrays do not get validated
For instance, `DeleteProperties.items.property` does not get validated as an expression.

### LINKED ISSUE #1822: Small bug in Set properties description
## Describe the bug
In the **Properties panel** of `Set properties` action, the description shows `delete properties`.

## Expected behavior
The description should show `set properties`. 

## Screenshot
image

lei9444:
## Description

Check array in the dialog.
add specific check function for some types: 'Microsoft.ChoiceInput'.

## Task Item

Closes #1783 
Closes #1822 
## Type of change

## Checklist

## Screenshots

image


Composer/packages/lib/indexers/src/dialogIndexer.ts
```diff
@@ -127,27 +127,26 @@ function ExtractReferredDialogs(dialog): string[] {
...
-    // it's a valid schema dialog node.
-    if (has(value, '$type') && has(expressionProperties, value.$type)) {
-      const diagnostics = IsExpression(path, value, { ...expressionProperties[value.$type] });
-
-      if (diagnostics) {
-        errors.push(...diagnostics);
-      }
+    if (has(value, '$type')) {
+      checkerFuncs.forEach(checkerFunc => {
+        const result = checkerFunc(path, value, value.$type, schema.definitions[value.$type]);
```
boydc2014: Here, you are applying every checker on every node regardless of the $type. Then inside a special checker like "checkChoices", you check the $type, and stop if it's not a certain type. 

This works, but very readable, because if you think about we have more and more special checkers later, i have to go into every checker to see how this type of node is getting checked. 

Instead, if i use a map of [type=>checkers or list[checkers]], i would know every precisely, what type of checking is performed for a given type. 

What do you think?

Composer/packages/client/src/utils/convertUtils/parsePathToFocused.ts
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { parsePathToSelected } from './parsePathToSelected';
+
+export function parsePathToFocused(path: string): string {
+  //path is like main.trigers[0].actions[0]
```
cwhitten: s/trigers/triggers

Composer/packages/client/src/utils/convertUtils/parsePathToFocused.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  const trigger = parsePathToSelected(path);
+
+  const list = path.split('.');
+
+  const matchActions = list.filter(x => x.startsWith('actions') || x.startsWith('elseActions'));
```
cwhitten: `actions` and `elseActions` should come off of the schema and not hardcoded here

Composer/packages/client/src/utils/convertUtils/parsePathToFocused.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  if (matchActions.length > 0) {
+    return trigger + '.' + matchActions.join('.');
```
cwhitten: prefer interpolation over concatenation

Composer/packages/server/schemas/editor.schema
```diff
@@ -270,7 +270,7 @@
       "helpLinkText": "Learn more"
     },
     "Microsoft.SetProperties": {
-      "title": "Delete Properties",
+      "title": "Set Properties",
```
cwhitten: Nice

Composer/packages/lib/indexers/src/dialogUtils/dialogChecker.ts
```diff
@@ -11,30 +11,94 @@ import { CheckerFunc } from './types';
...
-    } else {
-      try {
-        ExpressionParser.parse(exp);
-      } catch (error) {
-        message = `${formatMessage('must be an expression:')} ${error})`;
-      }
+const createPath = (path: string, type: string): string => {
+  const steps = ['triggers', 'actions', 'elseActions'];
+  let list = path.split('.');
+  const matchs = list.filter(x => !steps.every(step => !x.startsWith(step)));
```
cwhitten: s/matchs/matches

Composer/packages/lib/indexers/src/dialogUtils/dialogChecker.ts
```diff
@@ -11,30 +11,94 @@ import { CheckerFunc } from './types';
...
-    if (!exp && requiredTypes[property]) {
-      message = formatMessage(`is missing or empty`);
-    } else {
-      try {
-        ExpressionParser.parse(exp);
-      } catch (error) {
-        message = `${formatMessage('must be an expression:')} ${error})`;
-      }
+const createPath = (path: string, type: string): string => {
+  const steps = ['triggers', 'actions', 'elseActions'];
```
cwhitten: these values should come off the schema. i'm not exactly sure how, though.

Composer/packages/lib/indexers/src/dialogUtils/dialogChecker.ts
```diff
@@ -11,30 +11,94 @@ import { CheckerFunc } from './types';
...
+    diagnostic.path = createPath(path, type);
+    return diagnostic;
+  }
+
+  return null;
+};
+
+function findAllRequiredType(schema: any): { [key: string]: boolean } {
+  if (!schema) return {};
+  const types = schema.anyOf?.filter(x => x.title === 'Type');
```
cwhitten: Where is this `Type` coming from?

Composer/packages/lib/indexers/src/dialogUtils/dialogChecker.ts
```diff
@@ -11,30 +11,94 @@ import { CheckerFunc } from './types';
...
+}
+
+export const IsExpression: CheckerFunc = (path, value, type, schema) => {
+  if (!schema) return [];
+  const diagnostics: Diagnostic[] = [];
+  const requiredTypes = findAllRequiredType(schema);
+  Object.keys(value).forEach(key => {
+    const property = value[key];
+    if (Array.isArray(property)) {
+      const itemsSchema = schema.properties[key].items;
```
cwhitten: will `items` always exist on `properties[index]` ?

Composer/packages/lib/indexers/src/dialogUtils/dialogChecker.ts
```diff
@@ -11,30 +11,94 @@ import { CheckerFunc } from './types';
...
+      if (itemsSchema?.$role === 'expression') {
```
cwhitten: If we want to use the TS optional chaining operator we'll need to add a babel transform to support the browsers that Composer is to run in, right?

Composer/packages/lib/indexers/src/dialogUtils/dialogChecker.ts
```diff
@@ -11,30 +11,94 @@ import { CheckerFunc } from './types';
...
+  if (typeof choices === 'string') {
+    const diagnostic = checkExpression(choices, false, `${path}.choices`, type);
+    if (diagnostic) return [diagnostic];
+  }
   return null;
 };
+
+export const checkerFuncs: { [type: string]: CheckerFunc[] } = {
+  '.': [IsExpression], //this will check all types
+  'Microsoft.ChoiceInput': [checkChoices],
```
cwhitten: can we pull this valued off the SDKTypes object? `[SDKTypes.ChoiceInput]`

Composer/packages/lib/indexers/src/dialogUtils/dialogChecker.ts
lei9444: Hi, @a-b-r-o-w-n has already updated the typescript's version to 3.7.2. This project use tsc to build. I think we don't need to use babel to transform.

Composer/packages/lib/indexers/src/dialogUtils/dialogChecker.ts
lei9444: Hi @cwhitten. I have synced with @luhan2017 and checked the schema. The original schema in [microsoft/botbuilder-dotnet](https://github.com/microsoft/botbuilder-dotnet/tree/master/libraries/Microsoft.Bot.Builder.Dialogs.Adaptive/Schemas/Actions) use required to set object is required. But in our sdk.schema, use anyof to set.
Microsoft.DeleteProperties:
`
    "required": [
        "properties"
    ],`

`            "anyOf": [
                {
                    "title": "Reference",
                    "required": [
                        "$copy"
                    ]
                },
                {
                    "title": "Type",
                    "required": [
                        "properties",
                        "$kind"
                    ]
                }
            ]`

Microsoft.EditActions:
`
    "required": [
        "changeType",
        "actions"
    ],`

`            "anyOf": [
                {
                    "title": "Reference",
                    "required": [
                        "$copy"
                    ]
                },
                {
                    "title": "Type",
                    "required": [
                        "changeType",
                        "actions",
                        "$kind"
                    ]
                }
            ]`
Microsoft.EditArray:
`
    "required": [
        "changeType",
        "itemsProperty"
    ],`

`            "anyOf": [
                {
                    "title": "Reference",
                    "required": [
                        "$copy"
                    ]
                },
                {
                    "title": "Type",
                    "required": [
                        "changeType",
                        "itemsProperty",
                        "$kind"
                    ]
                }
            ]`

Composer/packages/lib/indexers/src/dialogUtils/dialogChecker.ts
lei9444: Hi, @cwhitten We don't want to make indexers depends on shared because the shared has a lot of composer's util function. We want to make indexers be used outside the composer.

Maybe we can create a library called schema in the future. 
@boydc2014 do you have any comments about these?

cwhitten: The schema needs to be injected in the running application, it needs to be "stateless" and exported from one place. Pulling it from `shared` is the right thing to do, and we can extract this further to come from a less heavy dependency. Making the change now will make that effort much easier.

boydc2014: Pull this from one single place instead of  hard coding it is definitely a good practice, and the concern of "this shared package is too big" is also valid. I agree @cwhitten that we should  pull it now, and seek to split this "shared" into less heavy modules later.  The name `shared` already indicate it's not something we thought too much about :)

Composer/packages/lib/indexers/src/dialogUtils/dialogChecker.ts
boydc2014: So, in short, the schema merging tool will turn the standardized "required" filed in json schema, into a "anyof" field with a hard-coded key "Type", right?

PR #1806, opened 2019-12-25
chore: update lu/lg all up view

### LINKED ISSUE #822: Short-term fix the nav in LG\LU all up view

liweitian:
## Description
update lu/lg all up view UX

## Task Item
Closes #822 

## Type of change

## Checklist


boydc2014:
@zhixzhan  can you help review this first?

Composer/packages/client/src/pages/language-generation/index.tsx
```diff
@@ -35,54 +34,31 @@ interface LGPageProps extends RouteComponentProps<{}> {
 const LGPage: React.FC<LGPageProps> = props => {
   const { state } = useContext(StoreContext);
   const { lgFiles, dialogs } = state;
+  const [selectedDialogItem, setSelectedDialogItem] = useState('common');
```
zhixzhan: it's not true with the initial state `common`, if I visit the url like:
`http://localhost:3000/language-generation/Main`, 
`Main` is current active dialog id. suggest:
1. useState('common')  --> useState( ...  init with path .... )
2. change naming from `selectedDialogItem` to `dialogId`;

ps: I prefer compute dialogId instead of useState
```
const dialogId  = useMemo( ..., [ path, dialogs ] )
```

Composer/packages/client/src/pages/language-generation/index.tsx
```diff
@@ -35,54 +34,31 @@ interface LGPageProps extends RouteComponentProps<{}> {
...
-            name: 'All',
-            url: '',
-            isExpanded: true,
-            links: [subLinks],
-          },
-        ],
-      },
-    ];
+  const navLinks = useMemo(() => {
+    const newDialogLinks = dialogs.map(dialog => {
```
zhixzhan: we should also put `Main.dialog` at this sequences's second place
```
-All
-Main
-TheRest
...

 ```

File: Composer/packages/client/src/pages/language-generation/index.tsx [hunk 3]
```diff
@@ -35,54 +34,31 @@ interface LGPageProps extends RouteComponentProps<{}> {
...
+    });
+    return newDialogLinks;
   }, [dialogs]);
 
   const onSelect = useCallback(
     id => {
       let url = `/language-generation/${id}`;
       if (edit) url += `/edit`;
       navigateTo(url);
+      setSelectedDialogItem(id);
```
zhixzhan: if we compute dialogId from path, we don't need maintain a state value here.

Composer/packages/client/src/pages/language-understanding/index.tsx
```diff
@@ -154,33 +120,19 @@ const LUPage: React.FC<DefineConversationProps> = props => {
...
-                    selectors: {
-                      'ul>li>ul button.ms-Nav-chevronButton:after': {
-                        borderLeft: 'none',
-                      },
-                    },
-                  },
-                  chevronButton: {
-                    backgroundColor: 'transparent',
-                  },
+          {navLinks.map(dialog => {
```
zhixzhan: can we reuse the leftNav component defined in lg page ? they looks like  have the same constructure.

Composer/packages/client/src/pages/language-understanding/styles.js
```diff
@@ -125,3 +125,15 @@ export const consoleStyle = css`
   padding: 15px;
   margin-bottom: 20px;
 `;
+
+export const dialogItem = selected => css`
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  background: ${selected ? '#f2f2f2' : '#ffffff'};
```
zhixzhan: can we replaced it with a fabric theme color  like : 
```
import { NeutralColors } from '@uifabric/fluent-theme';

background: ${selected ? NeutralColors.gray20 : NeutralColors.white};
```

Composer/packages/client/src/pages/language-understanding/styles.js
```diff
@@ -125,3 +125,15 @@ export const consoleStyle = css`
...
+  font-weight: ${selected ? 600 : 400};
```
zhixzhan: and this 
```
import {FontWeights } from 'office-ui-fabric-react/lib/Styling';
```

Composer/cypress/integration/LGPage.spec.ts
```diff
@@ -29,6 +29,9 @@ context('LG Page', () => {
     cy.get('@switchButton').click();
 
     // nav to Main dialog
-    cy.get('.dialogNavTree a[title="__TestTodoSample.Main"]').click();
+    //cy.findAllByText('__TestTodoSample.Main').click({ multiple: true });
```
zhixzhan: we can remove unused code

zhixzhan:
To keep `/microsoft/BotFramework-Composer` clean,  we should develop and PR from our own forking repository, like Andy previously mentioned here. 
https://github.com/microsoft/BotFramework-Composer/blob/stable/CONTRIBUTING.md#forking

lei9444:
all
There is a bug when refresh or go back. we should get the id from the URL.

Composer/packages/client/src/pages/language-generation/index.tsx
lei9444: why not use fileId directly?

Composer/packages/client/src/pages/language-generation/index.tsx
```diff
@@ -35,47 +34,30 @@ interface LGPageProps extends RouteComponentProps<{}> {
 const LGPage: React.FC<LGPageProps> = props => {
   const { state } = useContext(StoreContext);
   const { lgFiles, dialogs } = state;
+
   const path = props.location?.pathname ?? '';
   const { fileId = 'common' } = props;
   const edit = /edit(\/)*$/.test(path);
```
yeze322: a confusing name and a confusing regex here (not introduced by this PR but still a problem)
1. variable `edit` is a boolean, needs a more descriptive name
2. this regex is strange, input `edit/` or  `edit////` or `xxxxedit/` will return true, is that expected?

Composer/packages/client/src/pages/language-generation/index.tsx
```diff
@@ -35,47 +34,30 @@ interface LGPageProps extends RouteComponentProps<{}> {
...
-          isExpanded: true,
-        };
-      } else {
-        result.links = result.links || [];
-        result.links.push(item);
-      }
-      return result;
-    }, {} as INavLink);
+    if (mainDialog) {
+      newDialogLinks.splice(mainDialogIndex, 1);
```
yeze322: you can use the return value of `Array.prototype.splice()`, it's exactly the `mainDialog` object you need (wrapped in an array)

Composer/packages/client/src/pages/language-generation/index.tsx
```diff
@@ -35,47 +34,30 @@ interface LGPageProps extends RouteComponentProps<{}> {
 const LGPage: React.FC<LGPageProps> = props => {
   const { state } = useContext(StoreContext);
   const { lgFiles, dialogs } = state;
+
   const path = props.location?.pathname ?? '';
   const { fileId = 'common' } = props;
   const edit = /edit(\/)*$/.test(path);
   const file = lgFiles.find(({ id }) => id === 'common');
+  const navLinks = useMemo(() => {
```
yeze322: Do we need a `useMemo` here?

1. `navLinks` is computed from `dialogs`.
2. Component will be re-rendered whenever `navLinks` changes
3. Calculation inside `useMemo` will be invoked whenever `dialogs` changes

It's a traditional 'computed value' scenarios.
In a redux-style component, there will be a `mapStateToProps` handler for transforming data;
In a React FC, just call the transformer function per render.

Composer/packages/client/src/pages/language-understanding/index.tsx
```diff
@@ -29,57 +27,31 @@ const LUPage: React.FC<DefineConversationProps> = props => {
...
+      return { id: dialog.id, url: dialog.id, key: dialog.id, name: dialog.displayName };
+    });
+    const mainDialogIndex = newDialogLinks.findIndex(link => link.id === 'Main');
+    const mainDialog = newDialogLinks.find(link => link.id === 'Main');
+
+    if (mainDialog) {
+      newDialogLinks.splice(mainDialogIndex, 1);
+      newDialogLinks.splice(0, 0, mainDialog);
+    }
+
```
yeze322: We've seen this logic in LG page. DRY, needs to be refine.

Composer/packages/client/src/pages/language-understanding/index.tsx
```diff
@@ -29,57 +27,31 @@ const LUPage: React.FC<DefineConversationProps> = props => {
...
-            key: '_all',
-            name: 'All',
-            url: '_all',
-            isExpanded: true,
-            links: subLinks,
-          },
-        ],
-      },
-    ];
+  const navLinks = useMemo(() => {
```
yeze322: `useMemo`, the same problem

Composer/packages/client/src/pages/language-understanding/index.tsx
```diff
@@ -154,33 +126,19 @@ const LUPage: React.FC<DefineConversationProps> = props => {
...
-                  chevronButton: {
-                    backgroundColor: 'transparent',
-                  },
+          {navLinks.map(dialog => {
+            return (
+              <div
+                css={dialogItem(activeDialog ? activeDialog.id === dialog.id : dialog.id === '_all')}
+                key={dialog.id}
+                onClick={() => {
+                  onSelect(dialog.id);
```
yeze322: duplicate logic with LG. You need a shared component, `<NavLinks />`

yeze322: `LU` and `LG` page could be treated as containers. They just transform `dialogs` to `navLinks`, so transformation logic lives here.

Rendering work could be delegated to a new shared component `<NavLinks navLinks={[.....]} />`.

yeze322:
@liweitian Left some comments on code quality.

Composer/packages/client/src/pages/language-generation/index.tsx
boydc2014: @zhixzhan any idea on this? if this is not intented, can we fix in this PR

boydc2014:
> @liweitian Left some comments on code quality, we can make small improvements
> 
> (sorry I clicked on the 'closed' button)

Thanks @yeze322 , I think @zhixzhan should also take a look for those comments and advises, see if we can optimize some internal structure in this PR.  If some points is valid but requires a bigger change, it's OK to mark an issue and solve it in next batch.

Composer/packages/client/src/pages/language-generation/index.tsx
zhixzhan: `/\/edit(\/)?$/` is more precisely, here we expect match `/edit` `/edit/`

Composer/packages/client/src/pages/language-generation/index.tsx
zhixzhan: I don't get it, like you said "it's a typical 'computed value' scenarios" , we expect `navLinks` re-calculate only when `dialogs` changes. can you give some code change suggestion?

Composer/packages/client/src/pages/language-understanding/index.tsx
zhixzhan: It has some differences between LG and LU page.
1. LG page nav tree root is `common.lg` file 
2. LU page nav tree root is a virtual `merged *.lu` file.
we can't unified this.

The rest (dialog tree) in LG/LU is same, we could wrap it as a component.

Composer/packages/client/src/pages/language-generation/index.tsx
yeze322: > I don't get it, like you said "it's a typical 'computed value' scenarios" , we expect `navLinks` re-calculate only when `dialogs` changes. can you give some code change suggestion?

I got the difference. In your case, `dialogs` comes from StoreContext.state (didn't notice that, I thoguht it's from props, my bad).
The render process will be executed whenever store state changes, which is more frequently than we expect.

yeze322: we need the `useMemo` hook here, just share the data transformer across LG and LG page

Composer/packages/client/src/pages/language-understanding/index.tsx
yeze322: ditto, no problem on the `useMemo`, share the same func with LG

Composer/packages/client/src/pages/language-generation/index.tsx
boydc2014: Let's fix this in this PR? @liweitian  @zhixzhan

PR #1808, opened 2019-12-26
fix: undo/redo does not reflect in form

### LINKED ISSUE #1784: [Form Editor] undo/redo does not reflect in form
## Describe the bug
When undoing or redoing, the dialog files and visual editor reflect the changed data, but the form does not.

## Version
7594a264

## Browser
What browser are you using?

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [x] Edge (chromium)

## OS 

- [x] macOS
- [ ] Windows
- [ ] Ubuntu

## Expected behavior
The form should update when undoing or redoing.

lei9444:
## Description

The data in the form editor was controlled as a local state.
update the local data when data from shell changed. 

## Task Item

closes #1784 

## Type of change

## Checklist

### DIFF
--- a/Composer/packages/extensions/obiformeditor/src/FormEditor.tsx
+++ b/Composer/packages/extensions/obiformeditor/src/FormEditor.tsx
@@ -3,7 +3,7 @@
 
 /** @jsx jsx */
 import { Global, jsx } from '@emotion/core';
-import React, { useState, useMemo } from 'react';
+import React, { useState, useMemo, useEffect } from 'react';
 import { Dropdown } from 'office-ui-fabric-react/lib/Dropdown';
 import { JSONSchema6Definition, JSONSchema6 } from 'json-schema';
 import merge from 'lodash/merge';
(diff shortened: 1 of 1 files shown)


zhixzhan:
It still doesn't work for `Undo` in inline-editor, which also keep local state value, or other child component take the same strategy. And we can't require all child component must not have their own local state.

so I proposal we clarify the message type, when shell says `undo` or `reset` instead of `update`, the form editor realizes the changes is beyond itself can maintain, then re-mount itself, and all sub-localData will be clean.

ps: I found the global localData is added in PR #1106 for improving form performance.

lei9444:
> It still doesn't work for `Undo` in inline-editor, which also keep local state value, or other child component take the same strategy. And we can't require all child component must not have their own local state.
> 
> so I proposal we clarify the message type, when shell says `undo` or `reset` instead of `update`, the form editor realizes the changes is beyond itself can maintain, then re-mount itself, and all sub-localData will be clean.
> 
> ps: I found the global localData is added in PR #1106 for improving form performance.

Yes, This is only for input. inline editor undo will be fixed in another PR. I will use undoable to wrap  the update lg function

PR #1813, opened 2019-12-28
fix: Undo / redo behavior on LG resources

### LINKED ISSUE #1212: Undo / redo behavior on LG resources
## Version
57baf4c8

## Describe the bug
LG resouces not covered under undo / redo scenarios.

## To Reproduce
1. Add a SendActivity action, input some text
2. Delete it
3. Undo, no activity
4. (If undo is fixed, should redo delete the recovered activity again?)

## Expected behavior

lei9444:
## Description

wrap the updatelg function as undoable

## Task Item

closes #1212

## Type of change

## Checklist

- [x] I have commented my code, particularly in hard-to-understand areas

## Screenshots

test


Composer/packages/lib/indexers/src/type.ts
```diff
@@ -54,8 +54,9 @@ export interface LuFile {
   id: string;
   relativePath: string;
   content: string;
-  parsedContent?: IParsedObject;
   diagnostics: Diagnostic[];
+  parsedContent?: IParsedObject;
+  forceUpdate?: boolean;
```
boydc2014: do we need this forceUpdate?

Composer/packages/client/src/constants/index.ts
```diff
@@ -89,6 +89,7 @@ export enum ActionTypes {
   GET_ENDPOINT_SUCCESS = 'GET_ENDPOINT_SUCCESS', // remote publish
   ONBOARDING_ADD_COACH_MARK_REF = 'ONBOARDING_ADD_COACH_MARK_REF',
   ONBOARDING_SET_COMPLETE = 'ONBOARDING_SET_COMPLETE',
+  RECORD_EXTERNAL_UPDATE = 'RECORD_EXTERNAL_UPDATE',
```
boydc2014: SET_UPDATE_STATUS? maybe more descriptive?

Composer/packages/lib/shared/src/types/shell.ts
```diff
@@ -16,6 +18,11 @@ export interface BotSchemas {
   diagnostics?: any[];
 }
 
+export interface ExternalUpdate {
```
boydc2014: Can you put a little bit of the design and intention here into comments maybe.

I know you are tracking some data is updated outside externally to Form or Visual.

Composer/packages/client/src/store/action/dialog.ts
```diff
@@ -112,20 +114,25 @@ export const debouncedUpdateDialog = debounce(async (store, id, content) => {
   }
 }, 500);
 
-export const updateDialogBase: ActionCreator = (store, { id, content }) => {
+export const updateDialogBase: ActionCreator = (store, { id, content }, externalUpdate?: ExternalUpdate) => {
```
boydc2014: So the design is when some content is updated in shell, we mark the "externalUpdate" to notify any component (like Form) which may holds an local data to refresh, right

boydc2014: I think we just need to give a better name for this

boydc2014:
@cwhitten  @a-b-r-o-w-n  this one looks good to me now. 

The root cause of LG undo\redo behavior not reflected in Form is that form use local data which only init once when first render, after first render the data flow is single-direction (local => global). So, when a global redo/undo updated the store, the local data is not synced. 

The reason we choose a single-direction data flow is for consistency and efficiency, which is good. So the fix here @lei9444  did is to track the state is updated by other data source other than Form (so far it's only undo/redo, if i understand correctly), then Form will decide to re-init local data once it knows this piece of data is updated `externally`.

Composer/packages/extensions/obiformeditor/src/FormEditor.tsx
```diff
@@ -30,15 +30,16 @@ export interface FormEditorProps extends ShellData {
...
   const type = getType(localData);
 
   useEffect(() => {
-    if (!isEqual(localData, data)) {
+    // undo and redo will update the localData
+    if (externalUpdate?.scope === UpdateScope.DialogFile && !isEqual(localData, data)) {
       setLocalData(data);
     }
-  }, [data]);
+  }, [data, externalUpdate]);
```
a-b-r-o-w-n: I'm confused why the `!isEqual(localData, data)` wasn't sufficient. In general, I'm not a big fan of this `externalUpdate` api. My view is that `localData` is an internal (FormEditor) performance optimization and should be made to work when the global data is updated outside of the form. @lei9444 can you revisit this and see if we can make it work without the `externalUpdate`?

lei9444: Hi, @a-b-r-o-w-n @boydc2014 The purpose I want to add external status is that if the update is from the editor and we update the local value will cause a bug like this:
test1

So I add a status to tell the extensions there is an update from the shell.

Do you have any better ways to solve the problem?

boydc2014: @a-b-r-o-w-n my understanding is that:
1. using a local data is essentially making the data flow single-direction, very similar to how client save data to server today.   And this "externalUpdate" is a little bit like servery says "your data is not the latest, you need to refresh". 
2. the !Equals is not good enough because this !Equals check is actually will always be true when the local data is changing very quickly, like the "typing" scenario @lei9444  metioned above, which caused the "eating byte" bug, right？
3. The `externalUpdate` check would be good enough, i would expect this would replace this "!Equals" check.

is this make sense?

zhixzhan: I prefer force remount formEditor from ExtensionContainer.tsx when we want some force flush. 
keep the concept of  `externalUpdate` in shell, then formEditor do not have to aware any `externalUpdate` and can keep it's internal value as well.

a-b-r-o-w-n: Sorry I was out sick yesterday so haven't had a chance to respond. 

I understand the race-condition that using a simple `isEqual` causes, but since the shell save is debounced, we should have enough time between a keypress and the store being updated for it not to matter.

The option @zhixzhan mentioned is interesting. My only concern would be the need to maintain any local extension state when a flush event occurs.

boydc2014: yes, @a-b-r-o-w-n if the store updated quick enough before another key stoke, it would work, and that's the case for today. the only thing missing is, we need to notify the form (in Shell side) be be refreshed, then ExtensionContainer layer will remount form (like @zhixzhan). actually that's the first iteration of Leilei's solution, and in that solution, leilei registered a new api instead of create and keep a new state.

The pros of that one are Form doesn't requires any changes, and all the works in done by Shell.  The cons  are that overall a little more effort and change, compared to the current solution that directly let Form to refresh itself based on this shell state. 

My previous standpoint was that since anyhow this kind of signal that "data is updated elsewhere" is must be aware of the extension (ExtensionContainer+Form), i prefer a simple change that let Form react to that signal.    Based on @a-b-r-o-w-n  and @zhixzhan, i think prefer the changes been made in Shell and ExtensionContainer level, is also a very conceptually clean approach that keep things decoupled.

@lei9444 can you switch to the previous approach, see that more resonates with folks?

lei9444: @a-b-r-o-w-n @boydc2014 @zhixzhan I have added a key to force update the extension if the update is from the shell. Maybe we can add the status to the props in the future, and editors can get the status and judge for themself.

lei9444:
@a-b-r-o-w-n @boydc2014 @zhixzhan I have added a key to force update the extension if the update is from the shell. Maybe we can add the status to the props in the future, and editors can get the status and judge for themself.

Composer/packages/client/src/extension-container/ExtensionContainer.tsx
```diff
@@ -157,7 +169,9 @@ function ExtensionContainer() {
 
   const RealEditor = shellData.data ? getEditor() : null;
 
-  return RealEditor && <RealEditor {...shellData} onChange={shellApi.saveData} shellApi={shellApi} />;
+  return (
+    RealEditor && <RealEditor key={forceUpdateCount} {...shellData} onChange={shellApi.saveData} shellApi={shellApi} />
+  );
```
a-b-r-o-w-n: So this will completely reset all internal state in the extension right? I'm not sure we want that.

lei9444: I have removed it now.

Composer/packages/client/src/extension-container/ExtensionContainer.tsx
```diff
@@ -121,6 +121,9 @@ const shellApi: ShellApi = {
 
 function ExtensionContainer() {
   const [shellData, setShellData] = useState<ShellData>({} as ShellData);
+  const [forceUpdateCount, setForceUpdateCount] = useState<number>(0);
+  const forceUpdateCountRef = useRef(forceUpdateCount);
+  forceUpdateCountRef.current = forceUpdateCount;
```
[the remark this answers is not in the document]
lei9444: I just want to use the state of count in the created function. I have removed these all now.

lei9444:
@a-b-r-o-w-n @boydc2014 @zhixzhan. I have tried the approach @a-b-r-o-w-n gave and It works well.
This way can avoid the text input swallow characters and we don't need to add the update external API.

PR #1819, opened 2020-01-02
fix: Add support of auto suggestion memory variable

### LINKED ISSUE #1807: autosuggestion for memory variables in LG editor
When user is authoring in LG editor and typing dialog.xxx, they would like to get the auto suggested info for memory variables.

Following scenarios need to be taken care.
- User defined
eg. user.name; dialog.api_response, user.address.city
- prebuilt func 
eg. this.value, this.turnCount, turn.dialogevent.value
- luis related
eg. #intent.score, @entity,turn.DialogEvent.Value.Entities.xxx

cosmicshuai:
fixes #1807 

## Description
Support of auto suggestion memory variable. 

memoryVariable


Composer/packages/tools/language-servers/language-generation/resources/memoryVariables.json
```diff
@@ -0,0 +1,13 @@
+{
+    "this" : [{"value": ""}, {"turnCount": ""}],
```
boydc2014: this structure is a little wired, why not

"this": {
  "value": ""
  "turnCount":""
}

Composer/packages/tools/language-servers/language-generation/src/utils.ts
```diff
@@ -132,3 +134,11 @@ export function updateTemplateInContent(content: string, { name, parameters = []
   const resource = LGParser.parse(content);
   return resource.updateTemplate(name, name, parameters, body).toString();
 }
+
+export function loadMemoryVariavles(path: string): object {
```
boydc2014: Can you work with @lei9444 and @zhixzhan to figure out a path to make sure we can load from "pre-defined structure" + "the structure which indexer discovered"? 

we may not implement indexer right now in this PR to discover memory paths, but we need make sure the interface is there to use, we can't too static like this

Composer/packages/tools/language-servers/language-generation/resources/memoryVariables.json
boydc2014: Also, can you work with @luhan2017 to make sure this list is as rich as possible.

Composer/packages/tools/language-servers/language-generation/src/LGServer.ts
```diff
@@ -43,6 +45,7 @@ export class LGServer {
   protected readonly documents = new TextDocuments();
   protected readonly pendingValidationRequests = new Map<string, number>();
   protected LGDocuments: LGDocument[] = []; // LG Documents Store
+  private readonly memoryVariables: object;
```
boydc2014: this should be able to be updated. 

this PR's goal is figure out a way to dynamic set and update this stuff and auto-suggest based on this data source.

Composer/packages/tools/language-servers/language-generation/resources/memoryVariables.json
```diff
@@ -0,0 +1,13 @@
+{
```
boydc2014: the format here is really confusing, can you use some sort of linting tool on json?

Composer/packages/tools/language-servers/language-generation/src/utils.ts
cosmicshuai: Currently, pre-defined structure and user-defined structure will be stored separately. And the completion suggestions will base on both of them. 
pre-defined structure is stored in an object, imported from lib.
user-defined structure will be updated by a trigger event. A fileResolver will handle reading the file contains the content.

boydc2014: Let's get sometimes today to discuss this, looks like i was not sure about what the static vs dynamic means.

Composer/packages/tools/language-servers/language-generation/src/LGServer.ts
```diff
@@ -43,8 +44,13 @@ export class LGServer {
   protected readonly documents = new TextDocuments();
   protected readonly pendingValidationRequests = new Map<string, number>();
   protected LGDocuments: LGDocument[] = [];
+  private memoryVariables: Record<string, any> = {};
 
-  constructor(protected readonly connection: IConnection, protected readonly resolver?: FileResolver) {
+  constructor(
+    protected readonly connection: IConnection,
+    protected readonly resolver?: FileResolver,
+    protected readonly memoryResolver?: FileResolver
```
boydc2014: First, this memoryResolver don't have to FileResolver, why this is a "FileResolver"??

Composer/packages/tools/language-servers/language-generation/src/LGServer.ts
```diff
@@ -84,13 +91,52 @@ export class LGServer {
           this.validate(textDocument);
         }
       }
+
+      if (UpdateUserDefinedMemoryVariablesMethod === method) {
```
boydc2014: What this is for?

Composer/packages/tools/language-servers/language-generation/src/LGServer.ts
boydc2014: Second, you should also enable the E2E workflow that a set of pre-defined variables through your defined interface here, to prove you interface work.

boydc2014: By "focus on interface", i'm not referring to "removing anything besides", the use case of this interface is part of interface design.

Composer/packages/tools/language-servers/language-generation/src/LGServer.ts
```diff
@@ -35,6 +35,7 @@ const { check, indexOne } = lgIndexer;
```
boydc2014: Let me restate my expectation, 

first, you created a completion experience based on a fixed data structure.  That's been done. 
then what we just have to do, is replace this fixed data structure with a delegate than passed in constructor, that's it, no more, no less.

Composer/packages/tools/language-servers/language-generation/src/utils.ts
```diff
@@ -125,3 +132,8 @@ export function checkTemplate(template: Template): LGDiagnostic[] {
     return diagnostic.message.includes('does not have an evaluator') === false;
   });
 }
+
+export type MemoryVaribleCompletionResult = {
+  endWithDotFlag: boolean;
```
boydc2014: endWithDot is good enough

and what's this flag used for?

boydc2014: This flag looks like a global state, but here is a local state in the `result`, looks weird

Composer/packages/server/src/services/project.ts
```diff
@@ -38,6 +38,11 @@ export class BotProjectService {
     return BotProjectService.currentBotProject?.files.find(file => file.name === name);
   }
 
+  public static staticMemoryResolver(name: string): string[] | undefined {
+    BotProjectService.initialize();
```
boydc2014: you don't have to initialize here

Composer/packages/server/src/services/project.ts
```diff
@@ -38,6 +38,11 @@ export class BotProjectService {
...
+    return ['this.value', 'this.turnCount', 'turn.DialogEvent.value', 'intent.score'];
```
boydc2014: Please work with @luhan2017  to enrich this.

PR #1835, opened 2020-01-08
fix: RepeatDialog not shown

### LINKED ISSUE #1831: Repeat this dialog node visual disappeared
Complete the necessary portions of this template and delete the rest.

## Describe the bug
When you add a Repeat this dialog node, it would add a unclickable blank space instead of an action node.

## Version
What version of the Composer are you using? Paste the build SHA found on the about page (`/about`).

## Browser
What browser are you using?

- [O ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS 

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

## To Reproduce
Create a Repeat this dialog node anywhere.

## Screenshots
Screen Shot 2020-01-07 at 1 20 08 PM

yeze322:
fixes #1831 

## Description
If a $type is not covered by uischema, it should fallback to a default element. However, the default elelement was set to `<></>` which is a blank node (it should show something).

Now the fallback renderer was set to a grey box
image

changes on uischema:
image

### DIFF
--- a/Composer/packages/extensions/visual-designer/src/schema/uischema.ts
+++ b/Composer/packages/extensions/visual-designer/src/schema/uischema.ts
@@ -8,6 +8,9 @@ import { ActionCard } from '../widgets/ActionCard';
 import { UISchema } from './uischema.types';
 
 export const uiSchema: UISchema = {
+  default: {
+    'ui:widget': ActionCard,
+  },
   [SDKTypes.EditArray]: {
     'ui:widget': ActionCard,
     content: data => `${data.changeType} {${data.itemsProperty || '?'}}`,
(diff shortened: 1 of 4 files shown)


boydc2014:
Thanks for the quick fix, can you share some context, root cause, things like that? so that we can understand more and avoid similar issues

yeze322:
> Thanks for the quick fix, can you share some context, root cause, things like that? so that we can understand more and avoid similar issues

If a $type is not covered by uischema, it should fallback to a default element. However, the default elelement was set to `<></>` which is a blank node (it should show something).

yeze322:
Since we are working on schema-driven visual editror, commit more changes to fix this issue from schema side

Issue #1870, opened 2020-01-17
Need to support persistent navigation back for LU error in notification page


hibrenda:
**Describe the solution you'd like**
If there is an error in lu format, all the errors will be rolled up into notification page. Once you click the lu errors , it will navigate back to the dialog or all up view for further editing. (now navigate back to inline editor)
- utterance errors for specific intent, it will navigate back to the hoist view in property view. (either trigger intent or user inputs)
- create entity errors, it will navigate back to the LU all up view

hibrenda:
addressed in #1872

PR #1872, opened 2020-01-17
feat: [Form Editor]inline lu editor in trigger

alanlong9278:
## Description

This pr has done two actions.

### Actions
- Add `inline editor` in `Intent recognize` trigger
image

- Remove `lu editor` in dialog.   
image

The `lu editor` in dialog was removed so that users would not be confused by the fact that they could change content in lu file in both places. And the diagnostics for each trigger intent will follow below the `inline editor` .

## Task Item

Refs #1870 

## Type of change

## Checklist

## Screenshots
lutrigger


Composer/packages/lib/shared/src/types/shell.ts
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 import { LGTemplate as LgTemplate } from 'botbuilder-lg';
+import { LuIntentSection } from '@bfc/indexers';
```
zhixzhan: @bfc/indexers depends on shared library, this would cause a build error

PR #1887, opened 2020-01-21
feat: Enable external plugins to provide storage, publishing and auth mechanisms

benbrown:
## Description

This PR includes 3 interrelated bits:

* Adjustments to the server to enable it to use a drop-in replacement for the file system storage
* A new plugin loader that evaluates and ingests plugin packages
* ability to plugin to storage, auth and publishing via the plugin
* A storage plugin that implements MongoDB backed storage

## Still To Do

- [ ] plumb auth info down into storage layer
- [ ] Rationalize differences between my changes to csharpBotconnector and changes already in master
- [ ] Create a page in the app for managing Composer-wide settings such as DB connect string
- [ ] Code review and revisions
- [ ] Review of plugin manifest structure and capabilities 

## Type of change

- [X] New feature (non-breaking change which adds functionality)

## Checklist

## Screenshots


boydc2014:
Great to see, @benbrown i was working on a LocalBotManager service which implements "publish\history\rollback", it's ready now, so i was trying to implement a IPublisher in composer to hook up.  I saw you already have a publisher, I will take a look see how i integrate with this. Also left some comments based on my learning and investigation.

Composer/packages/server/src/controllers/publisher.ts
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import pluginLoader from '../services/pluginLoader';
+
+export const PublishController = {
```
boydc2014: Probably we should consider make this a REST api managing "publishers" which means:
1. Get /publishers, return a bunches of publisher plugins with ID
2. POST /publishers/:pid/publish, publishing using one publisher
3. POST /publishers/:pid/(rollback/history/status)

Composer/packages/server/src/server.ts
```diff
@@ -19,115 +19,118 @@ import { apiRouter } from './router/api';
...
-      'Content-Security-Policy',
-      CS_POLICIES.concat([
-        `script-src 'self' 'nonce-${req.__nonce__}';`,
-        // TODO: use nonce strategy after addressing issues with monaco-editor pacakge
-        "style-src 'self' 'unsafe-inline'",
-        // `style-src 'self' 'nonce-${req.__nonce__}';`,
-      ]).join(' ')
-    );
-  }
+pluginLoader.loadPluginsFromFolder(__dirname + '/../../../plugins', app).then(() => {
```
boydc2014: loadPluginsFromFolders is great.  Can you support also loading from code, by that, i was facing such a scenario:

I was trying to create a HttpPublisherGenerator which takes a config (name, host, etc) and return a IPublisher, so that for different slot, i will register different Publisher hopefully like this 
```typescript 
var httpPublishers:  IPublisher = [
    HttpPublisherGenerator(config1), // config 1 with product slot url
    HttpPublisherGenerator(config2), // config 2 with testing slot url  
]
```

Composer/packages/server/src/controllers/publisher.ts
benbrown: 100% agree, this is part of the work I've done already in another branch!

Composer/packages/server/src/server.ts
benbrown: Good call!

benbrown: Having seen your implementation, I don't 100% agree with this.

I think you would register ONE instance of the HttpPublisher, and then pass in the appropriate config, which is a bot-specific setting rather than a Composer-wide setting.

boydc2014: Thanks, saw you comments there, i haven't thought about whether the publisher configuration is per-bot or per-composer, my gut feeling is making the config per bot would be more flexible, on the other hand making the configuration in composer will be easier to manage and a little clean.  Let me mark that as a open question and think about that. cc 
two chris @christopheranderson  @cwhitten 

One of my focus on prototype was trying to setup a structure that code-based publisher and config-based plugin (throw HttpPublisher) can work well together. So that, average users don't have to code to enable a plugin (i believe that's part of intention of your original design doc) and we still keep the possibility to let user code when user want to or have to, for features that can't achieve throw by configuring HttpPublisher, which meet @christopheranderson's point at that meeting.

Anyway, either way it's OK to me, and we need to make sure we surface the configuration out, become multiple plugins or multiple profiles (for one plugin).

benbrown:
I am now exploring what it will take to make the storage layer identity aware -- that is, passing in user information from the auth system all the way down to each storage read/write operation.

benbrown:
Superceded by #2023

PR #1889, opened 2020-01-22
fix: LU LSP suggest empty string entities and roles in entering utterance

cosmicshuai:
## Description
Add a filter for remove the empty string entities and roles when calling suggestions in entering utterance.

For example, when user entered: 
#Intent Name
- the weather is {

LSP will generate the result of the entities user already defined.

## Screenshots
suggest Entity


Composer/packages/tools/language-servers/language-understanding/src/matchingPattern.ts
```diff
@@ -130,7 +130,7 @@ export function getSuggestionEntities(luisJson: any, suggestionEntityTypes: stri
     suggestionEntityTypes.forEach(entityType => {
       if (luisJson[entityType] !== undefined && luisJson[entityType].length > 0) {
         luisJson[entityType].forEach(entity => {
-          if (entity) {
+          if (entity && entity.name.length > 0) {
```
compulim:
```suggestion
          if (entity && entity.name) {
```

Composer/packages/tools/language-servers/language-understanding/src/matchingPattern.ts
```diff
@@ -177,7 +177,9 @@ export function getSuggestionRoles(luisJson: any, suggestionEntityTypes: string[
         luisJson[entityType].forEach(entity => {
           if (entity.roles !== undefined && entity.roles.length > 0) {
             entity.roles.forEach(role => {
-              suggestionRolesList.push(role);
+              if (role.length > 0) {
```
compulim: If `role` is a string, we could just:

```suggestion
              if (role) {
```

Or even,

```js
role && suggestionRolesList.push(role);
```

PR #1939, opened 2020-02-04
fix: update framework and package version in template botproject

### LINKED ISSUE #1927: Unable to execute because the specified command or file could not be found error when running Echo bot example
When I run the Echo Bot example I get this error message : Unable to execute because the specified command or file could not be found. Possible reasons: * You misspelled an integrated dotnet command. * You intended to run a .NET Core program, but dotnet-bin / Debug / netcoreapp2.1 / BotProject.dll does not exist. * You intended to run a global tool, but the executable with the name indicated above dotnet was not found in the PATH.

I have all the prerequisites:

Node.js: version 12.13.0
Yarn: latest stable version
Bot Framework Emulator: latest stable version
.NET Core SDK 2.2: required to test your bot

I have also .Net Core v3.1 & v 3.0 installed in my machine

VanyLaw:
## Description

Cause: The original version package reference was remove from repo, and caused sample bot project build failed.
Solution: 
- update package reference in template botproject
- update schema to match the new version sdk
- change the script to throw error sothat the process can catch it.
- update netcoreapp form 2.1 to 3.1

And @alanlong9278 will help to fix the UI to match schema change(number to expression).

FYI. those sdk main changes:
1. New recognizers set:
- Microsoft.AdaptiveCardRecognizer
- Microsoft.QnAMakerRecognizer
- Microsoft.CrossTrainedRecognizerSet
- Microsoft.RecognizerSet
- Microsoft.OnChooseIntent (trigger)
- Microsoft.OnQnAMatch (trigger)

2. FormDialog related triggers and actions
- Microsoft.Ask
- Microsoft.OnAssignEntity
- Microsoft.OnChooseEntity
- Microsoft.OnChooseProperty
- Microsoft.OnClearProperty

3. Other new actions and triggers
- Microsoft.GotoAction
- Microsoft.ContinueLoop
- Microsoft.BreakLoop
- Microsoft.UpdateActivity
- Microsoft.DeleteActivity
- Microsoft.GetActivityMembers
- Microsoft.GetConversationMembers
- Microsoft.SignOutUser
- Microsoft.OnEndOfActions

4. minor updates
- Add disabled property on actions
- Change includeActivity to activityProcessed (only this one have a breaking change, but we don't enabled it in Composer UX yet, so this is fine.)

There is only one breaking change in the schema, others are all additions. 

## Task Item

fixes #1927


InDieTasten:
I've tried this on my local environment, and it did not resolve the issue.
I find it strange, that this is still running on target framework netcoreapp2.1. Wouldn't it make sense, to update to netcoreapp2.2 ?

aalmada:
[The documentation does specify `.NET Core SDK 2.2` as prerequisite](https://github.com/microsoft/BotFramework-Composer/blob/stable/docs/setup-yarn.md#prerequisites) and not `2.1`.

VanyLaw:
> I've tried this on my local environment, ~and it did not resolve the issue.~
> 
> I find it strange, that this is still running on target framework netcoreapp2.1. Wouldn't it make sense, to update to netcoreapp2.2 ?
> 
> **EDIT:** This resolves the issue, but I was getting build errors from a configured nuget source that failed. It's kind of annoying though, that the build errors can only be validated from running the build manually, and that composer wouldn't log the build errors to console. Makes reproducing and fixing a bit difficult.

If you want to update to netcoreapp2.2, you need to change the package references and add some extra reference, like this:
image
I am not sure will it cause other compatible issues. Need to investigate it.

aalmada:
I applied the changes from this PR to my local version and I confirm that it fixes the 'Start Bot' issue.

boydc2014:
@VanyLaw  Can we take this chance to upgrade to 3.1 if possible, 3.1 is a LTS version, and  i hope that could save us efforts a lot of time.

luhan2017:
@VanyLaw, could you also update the sdk.schema in the template and Composer folder, and test whether the template bots work correctly?

VanyLaw:
> @VanyLaw, could you also update the sdk.schema in the template and Composer folder, and test whether the template bots work correctly?

Some sample bots will throw "must be an expression" error. Seems like number need to be changed to expression after schema updated. According to @alanlong9278 , It will cause hige UI change. Maybe we need a discussion about it.

BotProject/Templates/CSharp/Scripts/build_runtime.ps1
```diff
@@ -6,4 +6,12 @@ if ((dotnet --version) -lt 3) {
 # This command need dotnet core more than 3.0
 dotnet user-secrets init
 
-dotnet build
+# Merge all streams into stdout
```
boydc2014: at the top of this file, the error check should also be 3.1

VanyLaw: done

Issue #1945, opened 2020-02-04
Expose publish functionality as Express endpoints


benbrown:
Part of #1928 

Create endpoints allowing the Composer web client to interact with user-configured publishing plugins.

A draft version of this work is in progress [here in the publisher controller](https://github.com/microsoft/BotFramework-Composer/blob/benbrown/plugins/Composer/packages/server/src/controllers/publisher.ts)  and [here in the api router](https://github.com/microsoft/BotFramework-Composer/blob/benbrown/plugins/Composer/packages/server/src/router/api.ts#L49).

Publishing methods are registered via a plugin mechanism. They will follow the form laid out [evolving format defined here](https://github.com/microsoft/BotFramework-Composer/blob/benbrown/plugins/Composer/packages/server/src/services/pluginLoader.ts#L40)

## Describe the solution you'd like

Implement the following routes and their companion features in the controller.

These routes should return JSON and use standard HTTP status codes to indicate success, failure, and other statuses.

- [x] /publish/types - return a list of all currently supported publishing methods as defined by plugins 
- [x] /publish/:projectId/publish/:target - using a specific publishing method and a configuration provided by the user, publish a given bot using the method
- [x] /publish/:projctId/status/:target - using a specific publishing method and a configuration, return the status of the latest publish action. Use http status codes to indicate completed, in progress, error
- [x] /publish/:projectId/history/:target - using a specific publishing method and a configuration, return a list of historical publish events. should accept an offset or pagination indicator.
- [x] /publish/:projectId/rollback/:target - provided a configuration and an identifier from the publish history, execute the rollback function of given method

benbrown:
There is actually one last API to implement, I will do so today and re-close.

benbrown:
this is done now

Issue #1928, opened 2020-01-31
Add support for loading plugin extensions into Composer


benbrown:
Composer must support connectivity to outside services, and allow outside services to plugin and extend the capabilities of the Composer editor. This will be achieved by exposing a series of APIs to external plugins that can be loaded dynamically into the application and, by calling the APIs, hook into key areas of Composer.

The first round of plugin functionality is focused on the low level plugin architecture as well as extension points for storage, authentication and publishing.

This is a tracking issue for a bunch of sub-tickets that will define the work necessary to implement this feature.

## Scenarios

* Allow Composer to store/load content in a database or remote storage
* Allow Composer to use AAD for authentication
* Allow Composer to publish a bot to a remote runtime
* Allow Composer to publish a bot directly to a Git repo
* Allow Composer to use a custom database or API for authentication

### Finish Plugin Loader

An in progress version of the plugin loader is in the `benbrown/storage branch` [specifically in this file](https://github.com/microsoft/BotFramework-Composer/blob/benbrown/storage/Composer/packages/server/src/services/pluginLoader.ts).

Basically the way it works is that there is a single instance of the plugin loader that has access to the Express webserver instance. When the app launches, the plugin loader looks for plugins in the plugins/ folder and loads everything it finds. It will require in the main file, then look for an initialize function. 

When loading the plugin, Composer gives the plugin code limited access to Composer and a special plugin namespace. Plugins can currently set up a storage provider, an auth provider, or add custom web routes and middleware.

From within the main Composer app, features access the plugin namespace via the plugin singleton. For example, [use of a custom storage provider is implemented here](https://github.com/microsoft/BotFramework-Composer/blob/benbrown/storage/Composer/packages/server/src/models/storage/storageFactory.ts)

Additional work to finish the plugin loader implementation:

- [x] Refine APIs used to expose Composer features @benbrown 
- [x] Add an instance of PassportJs and necessary accessor features so that plugins can access auth and identity @benbrown 
- [ ] Expose a settings page that controls Composer-wide settings that will be made available to plugins, such as a database connect URI used with storage or a oauth client id used during authentication.
- [x] Remove the prototype plugins from this repo @benbrown 
- [ ] Publish a package of classes and interfaces usable by plugin developers 

### Publishing Plugins

Ben has implemented a v1 of publish plugin loader. This allows a plugin to register methods for publish, get status of publish, get publish history, and rollback.

[...]

PR #1952, opened 2020-02-05
draft: Remote publish E2E prototype

boydc2014:
### Description
A follow up of the remote publish doc https://microsoft.sharepoint-df.com/:w:/r/teams/Composer/_layouts/15/Doc.aspx?sourcedoc=%7B0B524AA7-23E4-44BC-A0FC-9BD258232DA7%7D&file=Remote%20Publish%20from%20Composer.docx&wdOrigin=OFFICECOM-NTP.MAIN.REC&action=default&mobileredirect=true

This PR consists 3 parts
- [x] a local bot manager service that take over the runtime management and bot hosting
  - accept the http protocol and manage runtime in it's own space
  - support publish\history\rollback
  - support multiple bots at same time, port management, fast reload
- [x]  some publisher management in composer
   - a IPublisher interface
   - a HttpPublisher which creates publisher given a config
- [x] UX hook up in composer

### Architecture diagram
image

### Goals
1. Verify the approach that sperate the concern of runtime from composer, into dedicated runtime service would work
2. Figure out the details of the http publisher api
3. Figure out how much effort we need to have a runtime hosting service that implements those api
4. See how we can build a non-code(config-based) publisher system on top of a generic coding interface
5. Figure out how much we can\need to clean the core, how much abs-h specific code we've built
 
### Non-goal
- This PR touched some parts of how a publisher is registered and generally how a plugin get loaded and surfaced out through API. But not tended to be the standard.  

#### Stage1: replace BotProject1 with BotManagerService
- [ ] pass LUIS settings when publishing
- [ ] clean BotProject
- [ ] remove the copying runtime logic from composer core 
#### Stage2:  integrated in plugin system 
- [ ] hook in with plugin system
#### Stage3: cleanup and polish
- [ ] clean IEnvironment, IConnector, IEnvSettings
- [ ] publisher controller polish


benbrown:
This looks great and is roughly in line with what I've been doing in the plugin loader.

@boydc2014 in my implementation, the loaded publishing mechanisms actually live in the PluginLoader service, rather than in the publishing service itself. It may be more appropriate for them to live in the publisher service -- but we will need to make sure it can be exposed to the plugin loader so that plugins can add new methods.

One key difference is that, in my current design, the configuration for each publishing endpoint will actually live in the bot settings. We may need a special case for local publish to test runtime, but in general, I believe these settings should be part of the bot configuration and NOT handled by the plugin or environment variables.

What I did in my prototype is, the user provides a configuration object via settings page that is bound to a specific method of publishing. Then, when the publish action is requested, Composer passes that configuration into the plugin to fulfill the request.  This enables a user to set up multiple publishing targets of the same type without installing multiple plugins -- for example, a dev, staging and production instance that differ only in their configuration.

benbrown:
See also, #1945 and #1928

boydc2014:
> This looks great and is roughly in line with what I've been doing in the plugin loader.
> 
> @boydc2014 in my implementation, the available publishing mechanisms live in the PluginLoader service, rather than in the publishing service itself. It may be more appropriate for them to live in the publisher service -- but we will need to make sure it can be exposed to the plugin loader so that plugins can add new methods.
> 
> One key difference is that, in my current design, the configuration for each publishing endpoint will actually live in the bot settings. We may need a special case for local publish to test runtime, but in general, I believe these settings should be part of the bot configuration and NOT handled by the plugin or environment variables.
> 
> What I did in my prototype is, the user provides a configuration object via settings page that is bound to a specific method of publishing. Then, when the publish action is requested, Composer passes that configuration into the plugin to fulfill the request. This enables a user to set up multiple publishing targets of the same type without installing multiple plugins -- for example, a dev, staging and production instance that differ only in their configuration.

I think we are on the same page even for the key difference you mentioned, my thought is we have a IPublisher *coding* interface for plugin, then I have a HttpPublisher to create HttpPublisher via config,  there is another type of IPublisher that may load from local plugin folder.

In the case of HttpGenerator, i use multiple config will generate multiple HttpPublisher, see the code in PublisherService.init(), which allow user to publish with different target without have to modify config, user are creating multiple plugin(Ipublisher) in my case, by it's created very lightweight, the `HttpPublisher` serve as a "PublisherGenerator" in this case. 

It's not ready yet, i plan to complete it to some degree, and reach out to you and folks to review it later, probably next week, and the implementation is not to be production, mainly focus exploration on the protocol.

boydc2014:
@benbrown  I read through your comments here and in your PR, and re-clarify the goals and non-goals of this PR in the description above, to make sure this PR focusing on run time decoupling, publishing protocol evaluation and implementation and other relate refinement on the composer core ifself. Leave out the plugin loading and managing as a non-goal, and let us seek for a good approach works for all scenarios we are targeting now.

lei9444:
update some ux
ui

boydc2014:
Close this, since this PR reach its goal, the follow up will be a local plugin which separate the concern of runtime from Composer core

PR #1960, opened 2020-02-06
style: left nav and dialog tree align

### LINKED ISSUE #1908: Left Nav and Dialog Nav Alignment
This issue - I'm not sure the best solve. Maybe in one of the stand-ups we can decide based on the amount of work vs the reward if in needs to happen.

The spacing on the Left Nav and the Dialog Nav don't line up - could be as simple as tightening up the spacing on the dialog list to make it match. That could make some odd looking issues too perhaps making (LU and Delete Intent look connected) perhaps we need a different selected and hoover colors just to make sure we aren't confusing things.

Anyway - this makes me a little crazy, but that could just be me.

Screen Shot 2020-01-27 at 11 18 21 AM

liweitian:
## Description
Align the left nav and dialog tree.
Do we still need to update the color described in the task?
## Task Item
Closes #1908 
## Screenshots
image


DesignPolice:
yes I think that tightens things up a bit... I think we will need do some scrolling around to make sure it doesn't need a color difference... @cwhitten or @a-b-r-o-w-n - getting these aligned will remove the visual tension of all the boxes being out of alignment, but it could need make feel like the left nav and the dialog nav are connected - thoughts?

a-b-r-o-w-n:
@DesignPolice I think this looks ok. The one thing I think is a little too subtle is the gray color of the selected dialog / trigger.

image

I wonder if adding something with a little more contrast would help like in the left nav.

DesignPolice:
@a-b-r-o-w-n I was looking at the Teams, they have the same "nav next to a nav" and they do a color shift on the left vs the right- the current colors are just the Fabric Fluent standard rollovers... so maybe a rollover or focus change?

DesignPolice:
I think we could do a different hover color and maybe selected color on the Dialogs.  @a-b-r-o-w-n @cwhitten - but I would need to mock that up... Do you want me to do that now on this issue or make a mock up and make it another # 

The only instance I have found in existing Fluent/Fabric deck is a Split Button that shows two layers of gray next to each other.  or the Teams model that adds a shade of their Purple over one menu. We could do a shade of blue over the gray? 

Screen Shot 2020-02-06 at 3 16 57 PM

liweitian:
Hi @DesignPolice @cwhitten @a-b-r-o-w-n  is there any update on the design?

DesignPolice:
I will mock one up today @liweitian - I got distracted to other tasks this week. 

m

DesignPolice:
Okay - @mareekuh @cwhitten @sangwoohaan 

I had an idea about this - It will help distinguish our Left Nav from the Dialog Menu, esp now that they line up - This just changes the rollover color - only on the left nav. and the Pressed Selected background color and the icons to white - This is a change from the normal Fluent only on the selected state being solid vs the highlight on the left, but similar to some others. 

I don't know if this is harder than it sounds, but here is a comp to look at. Then leaving the Dialog Nav alone it should give a subtle but obvious difference.

Screen Shot 2020-02-12 at 2 50 51 PM

Also a link to the Figma file.

https://www.figma.com/file/Ux0vucOvaLL2xH22WpFgV65X/MarcB_Composer-UI?node-id=4652%3A691

liweitian:
@DesignPolice I updated the screenshots. Let me know if there is anything else I need to do.

DesignPolice:
Hi @liweitian we going to make some different changes... This version had some problems I didn't expect... will upload a change tomorrow.

DesignPolice:
Hi @liweitian -

Here is a Marieke approved update to this thread.

Hamburger Menu - revert to current build - no blue - ( so no change from Stable today ) 

Only change will be Darker Shade on Hover - 

and instead of **Pressed Selected** having a darker gray we are going to do a white background with the existing Blue Tab. 

The other addition will be a 1px line on the far right of the nav - to help separate the Nav and the Dialog List. 

screen shot 

Screen Shot 2020-02-17 at 3 22 29 PM


Figma link.

https://www.figma.com/file/Ux0vucOvaLL2xH22WpFgV65X/MarcB_Composer-UI?node-id=4652%3A691

liweitian:
> Hi @liweitian -
> 
> Here is a Marieke approved update to this thread.
> 
> Hamburger Menu - revert to current build - no blue - ( so no change from Stable today )
> 
> Only change will be Darker Shade on Hover -
> 
> and instead of **Pressed Selected** having a darker gray we are going to do a white background with the existing Blue Tab.
> 
> The other addition will be a 1px line on the far right of the nav - to help separate the Nav and the Dialog List.
> 
> screen shot
> 
> Screen Shot 2020-02-17 at 3 22 29 PM
> 
> Figma link.
> 
> https://www.figma.com/file/Ux0vucOvaLL2xH22WpFgV65X/MarcB_Composer-UI?node-id=4652%3A691

done! I updated the screenshot.

DesignPolice:
Thanks @liweitian @a-b-r-o-w-n  - small change, but I like this one. Appreciate all the help. M

PR #2005, opened 2020-02-18
feat: support multi-line node block in Visual Editor

### LINKED ISSUE #1701: Action nodes redesign to improve visual canvas' readability
### Description
Currently, all the action nodes have the same format(Label-sublabel) which works for few, but not for the others. This is demanding a redesign sooner or later.
The redesign's goal is to improve the overall readability of visual canvas through a flexible design system – you should be able to quickly scan through it and understand the flow, where to find what. 

### Execution plan
- [ ] (in PR #2005 ) Layout: smart layouter, support size-sensitive node blocks
- [ ] (in PR #2017 ) Widget: enhance the appearance of `SetProperties` as a multi-line node, extract shared elements
- [ ] Widget: implement multi-section node widget which is split by '-------v-------', apply to specific $types
- [ ] LG: Syntax hightlighting lib of LanguageGeneration(LG) text 
- [ ] Layout: node size adjustment and node styles fine tune

### Other context
Design link: https://www.figma.com/file/04DymSq4xElhOBdAyutvvTbq/Composer-SH-06.19?node-id=4585%3A1049

yeze322:
## Description

fixes #1701 

Make the node block size-sensitive in Visual Editor: action flow's layout will change along with a single node block's size automatically.
(leverages the [react-measure](https://github.com/souporserious/react-measure) lib as a new dependency)

demo-multiline-node

### Why we need this functionality?
In new action design, some $types have dynamic sizes. For example, when the prompt content inside a SendActivity node becomes very long, the node will have 2 or more lines to show full content.
image
image

This is a challenge to our layouter, we need to make the layout size-sensitive.

### How it works?
1. Wrap a child node with the 'react-measure' lib, it will fire an `onResize` event whenever its DOM size changes;
2. Visual Editor maintains a layout computing chain, layout will be updated bottom-up via the `onResize` chain;
3. A size-sensitive parent node do several things:
   - remembers child nodes' size and position
   - observes child nodes' size & recompute the layout once it receives new size
   - recompute its layout (node positions, edge length) & report its new size to higher level
   - redraw itself with the new layout (saved as states)
4. After a node is measured, its size will be cached in a global cache. This will improve the flickering problem significantly.

### What's changed?
1. style: update the css of `FormCard` to make basic node blocks can show multiline text
2. fix: reconnect the boundary propagation chain (`onResize` chain), it had been outdated for a while
2. refactor: extract common layout logic as a new hook `useSmartLayout`, used to observe node sizes and recompute new layout when needed
3. refactor: migrate widgets `StepGroup`, `IfCondition`, `SwitchCondition`, `Foreach`, `Prompt` to the new hook to make them size-sensitive
5. feat: implement a global `DesignerCache` class for caching calculated node sizes

### What's next?
1. Alan is working on new node design, we will have a PR for new version of visual nodes
2. I will keep optimizing UISchema especially the schema of ChoiceInput, hopefully in my next PR.
3. lg hook is stll causing the flickering problem, waiting for Leilei's indexer work.

## Task Item
#1701 

## Screenshots

image


Composer/packages/extensions/visual-designer/src/components/nodes/templates/FormCard.tsx
```diff
@@ -124,12 +125,13 @@ export const FormCard: FunctionComponent<NodeProps> = ({
...
-              whiteSpace: 'nowrap',
-              textOverflow: 'ellipsis',
-              overflow: 'hidden',
+              whiteSpace: 'initial',
               fontSize: '12px',
               lineHeight: '19px',
               fontFamily: 'Segoe UI',
+              overflowWrap: 'break-word',
+              wordBreak: 'break-all',
+              display: 'contents',
```
a-b-r-o-w-n: Can you make sure this has the browser support necessary? https://developer.mozilla.org/en-US/docs/Web/CSS/display

yeze322: I changed it to 'inline-block' and confirmed no other usage of `display: contents`

cwhitten:
Is there a limit to the vertical/horizontal growth potential of a node? It
shouldn’t be unbounded.

yeze322:
@cwhitten No there isn't. Maybe we can add a `max-height` css property to limit the bounds?

cwhitten:
Yes that sounds like a good idea. And we may need to truncate text on the
final line when we hit that boundary. Thoughts?

yeze322:
@cwhitten that's right, let me update some css

PR #2013, opened 2020-02-19
fix: the diagnostic path error

### LINKED ISSUE #1972: [Linting & Validation] some diagnostics do not get reported in form due to path format
## Describe the bug

Fields in deep paths do not show errors in the form.

## Version

22685455

## To Reproduce

1. Open up the `Dialog Actions` sample.
2. Go to the `SwitchConditon` dialog.
3. In the `SwitchConditon` action, add an action that includes a condition
4. Add an invalid condition, ex. `foo = bar`
5. Notice the error appears in the notifications page, but not in the form.

## Expected behavior

Error should also appear in the form.

## Screenshots

image

## Additional context

LG errors still show in the form.

lei9444:
## Description
the diagnostic path is wrong
the  format is like 'path#type#property'

## Task Item

Closes #1972 
## Screenshots

image

image

### DIFF
--- a/Composer/packages/lib/indexers/__tests__/dialogUtils/dialogChecker.test.ts
+++ b/Composer/packages/lib/indexers/__tests__/dialogUtils/dialogChecker.test.ts
@@ -1,0 +1,19 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+import { createPath } from './../../src/dialogUtils/dialogChecker';
+
+describe('create right diagnostic path', () => {
+  it('should check if the diagnostics have errors', () => {
+    expect(createPath('Main.triggers[7].actions[0].cases[1].actions[0].condition', 'type')).toBe(
+      'Main.triggers[7].actions[0].cases[1].actions[0]#type#condition'
+    );
+    expect(createPath('Main.triggers[7].actions[0].condition', 'type')).toBe(
+      'Main.triggers[7].actions[0]#type#condition'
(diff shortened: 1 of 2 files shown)


lei9444:
> This is fine but I have a question why we need to do this? Shouldn't the json path be enough to route the diagnostics to fields in the form?

Hi @a-b-r-o-w-n , I have talked about this with @alanlong9278 ,If I can do this in shell, Alan can find the fields more easier.

alanlong9278:
> This is fine but I have a question why we need to do this? Shouldn't the json path be enough to route the diagnostics to fields in the form?

Hi, @a-b-r-o-w-n It was done for get formErrors，the filter condition needs to split path, type and props from diagnostics in form editor.
```
const formErrors = useMemo(() => {
    if (props.currentDialog && props.currentDialog.diagnostics) {
      const currentPath = props.focusPath.replace('#', '');
      const diagnostics = get(props.currentDialog, 'diagnostics', [] as Diagnostic[]);

      return diagnostics.reduce((errors, d) => {
        const [dPath, dType, dProp] = d.path?.split('#') || [];
        const dPropName = dProp
          ?.replace('[', '')
          ?.replace(']', '')
          ?.replace('.', '');

        if (dPath === currentPath && dType === type && dPropName) {
          errors[dPropName] = d.message;
        }

        return errors;
      }, {});
    }

    return {};
  }, [props.currentDialog]);
```

PR #2072, opened 2020-02-25
chore: remove `OAuth` action from 'External Resource' group in the dialog menu

### LINKED ISSUE #2064: Node creation menu: duplicated `Oauth Login` entires
## Describe the bug

`Oauth Login` appears twice in a node creation menu:
image

image

## Browser

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

yeze322:
## Description
Closes #2064 

We have two entries for the `OauthInput` type.
As we discussed in the issue, it might be cleaner to just be in ask a question section.

## Task Item
#2064 

## Screenshots
image

### DIFF
--- a/Composer/packages/lib/shared/src/viewUtils.ts
+++ b/Composer/packages/lib/shared/src/viewUtils.ts
@@ -91,7 +91,6 @@ export const dialogGroups: DialogGroupsMap = {
     types: [
       SDKTypes.HttpRequest,
       SDKTypes.EmitEvent,
-      SDKTypes.OAuthInput,
       SDKTypes.QnAMakerDialog,
       //  SDKTypes.CodeStep
     ],


a-b-r-o-w-n:
Was this the decision we landed on? I think @benbrown would argue that this should exist in the external resources grouping. I sort of agree.

benbrown:
> Was this the decision we landed on? I think @benbrown would argue that this should exist in the external resources grouping. I sort of agree.

Yes, I agree with this assertion. Oauth is a special beast and doesn't fit neatly into "ask a question." It is in both places right now because I wanted to make sure people could find it if they are thinking "this is something I need to collect" OR if they are thinking "I need to connect into my external login flow"

yeze322:
That's reasonable, I've quoted Ben's comments in #2064 .
Mark this PR as on hold, let's just close it if we don't need a cleaup

mareekuh:
Lol this is exactly the discussion Vishwac and I had before Ignite and why there are 2 entries.
Let's leave it at the 2 entries. For some it is associated with an external resource, for others it is to seek a particular type of input from the user.

Leave as is, both entry points remain.

PR #2118, opened 2020-02-27
feat: add adaptive-form and extension packages

### LINKED ISSUE #1974: Aligning the array UI implementation
The implementations of array UI seem to be different in some action nodes.

Branch : Switch - has an unnecessary 'value' label between 'Case' and the array.
Screen Shot 2020-02-10 at 9 55 13 AM

Intent recognized, Delete properties - missing the stroke below the title
Screen Shot 2020-02-10 at 9 56 27 AM

In prompt nodes - indented, and missing the stroke
Screen Shot 2020-02-10 at 9 54 15 AM

### LINKED ISSUE #1975: Move 'prompt settings' section into Other tab in prompts
The prompt setting sub-section always appears on the bottom of prompt property editors. Let's move it into the Other tab.

Desired design : https://www.figma.com/file/LlTlh5vXwq91zjGnFtrUUR6l/Composer-UI-Master-library?node-id=180%3A167

### LINKED ISSUE #2119: Use bot project schema to drive ui experience

a-b-r-o-w-n:
## Reasoning

Currently, the app is mainly driven by a single static schema file (`appschema.ts`). This has worked well up until now by giving us a fixed point of reference to build against, but in order to work with the current form library, we have had to make modifications that deviate from the true sdk schema.

As we approach GA, we need to start supporting multiple versions of the sdk schema as well as schemas that have additional custom components, which means that the bot project's schema needs to drive the UI experience. This set of changes enables that, as well as introducing some new concepts as well as modifying some existing patterns.

#### Extension

The extension package is a shared set of utilities that provide apis to create composer extensions / ui-plugins. It currently has a small surface area, but the intention is that it can grow as composer's capabilities expand. 

It exports a few utilities so far:
- `<Extension />` component which is meant to be rendered in Composer and provides a context for extensions to consume.
- `useShellApi` is a hook that provides shell data and shell apis to any component in an extension or ui-plugin.
- Typescript interfaces that are shared across extensions and ui-plugins.

#### Adaptive Form

This is a replacement for the `obiformeditor` package that re-implements the underlying json schema form renderer, with modifications and customization to support the composer forms UI/UX. It borrows many ideas from the json schema form library but has a smaller interface that is more tailored to our form usage at this point in time. The new pattern this enables is the usage of a ui plugin.

#### UI Plugin

A UI plugin is a framework that developers can use to change / extend composer form functionality. It is heavily based on the ui schema idea from the json schema form library, but adds some additional capabilities such as registering new recognizer types and providing a way to customize the form based on the schema `$role` property.

This should be integrated with the visual designer's ui schema in the future so that the whole editing surface can be customized within a single plugin architecture.

#### @monaco-editor/react

This change also switches the monaco react editor package to one that does not require a webpack plugin to work with our custom language servers. It was a drop-in replacement for the most part, with some minor api changes.

## Consolidation of types into `@bfc/shared`
In an effort to make sharing of types more effective, I've moved types from `@bfc/indexers` into `@bfc/shared`. This allows other packages to take dependencies on the types defined in the indexers without the dependency on the indexers themselves. This probably introduced more churn than necessary, sorry!

## Future

I attempted to minimize the churn in the Composer core while laying the ground work for future refactors. I'll jot down some ideas below:

- Refactor visual designer to use extension pattern (#2120)
- Remove the extension container / iframe embedding. Hoist extension rendering to composer core. -- This idea has not been fruitful to us and by eliminating it we are freed to make more performance enhancements. (#2121)
- Make the extensions themselves not require props -- just use the extension context to get data, etc. This will require some work in the extension to re-render when the shell data / apis change.

### Links

Closes #2119
Closes #1974
Closes #1975

#664
#2120
#2121

## Authors
@a-b-r-o-w-n 
@tdurnford


Composer/packages/extensions/adaptive-form/src/components/FormRow.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  if (propSchema) {
+    return (
+      <SchemaField
+        key={`${id}.${row}`}
+        className={className}
+        definitions={definitions}
+        depth={depth}
+        id={`${id}.${row}`}
+        label={label === false ? false : undefined}
```
tdurnford: Should this just be `undefined`? If it's false, it will hide every child property's label.

a-b-r-o-w-n: This is probably just a copy / paste hold over.

cwhitten:
seeing the following:

1. selecting the LUIS recognizer throws an alert?

Screen Shot 2020-03-24 at 7 40 48 PM

2. deselecting and reselecting a response node doesn't render the content in the form

Screen Shot 2020-03-24 at 7 42 34 PM

3. Switch condition, adding cases renders Unsupported field

Screen Shot 2020-03-24 at 7 44 01 PM

4. Choice input, can't see full option text (should we default to something other than expression)?

Screen Shot 2020-03-24 at 7 45 07 PM

PR #2144, opened 2020-03-02
fix: some known issues in lu page

liweitian:
## Description
1. disable edit button in lu all up view page
2. use \r\n during creating a new trigger modal


liweitian:
will reopen

PR #2146, opened 2020-03-02
fix: remove edit button in lu all up view

liweitian:
## Description

1. disable edit button in lu all up view page
2. use \r\n during creating a new trigger modal

fix #2146 

## Screenshots


Composer/packages/client/src/components/ProjectTree/TriggerCreationModal.tsx
```diff
@@ -132,7 +132,7 @@ export const TriggerCreationModal: React.FC<TriggerCreationModalProps> = props =
 
   const onTriggerPhrasesChange = (body: string) => {
     const errors = formData.errors;
-    const content = '#' + formData.intent + '\n' + body;
+    const content = '#' + formData.intent + '\n\r' + body;
```
a-b-r-o-w-n: Why is this necessary? It seems platform dependent at the very least.

PR #2153, opened 2020-03-03
fix: deployment script, decouple debugging and deployment settings

### LINKED ISSUE #2007: Response status code does not indicate success: 403 (Forbidden) after Azure deploym,e
#1316  Describe the bug

I created a new Composer bot (LUIS & QnA) and just added an Adaptive Card as a greeting. The bot works totally fine if I test it locally, but after the deployment to Azure I just get the greeting card and then I always get "Response status code does not indicate success: 403 (Forbidden)." as a message by the bot.

image

Am I missing something after the deployment?

## Version

test

## Browser

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [x] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

zidaneymar:
## Description

fix #2007 

1. Fix several bugs of deployment script.
2. Decouple appsettings to appsettings.Devlopment.json(for composer) and appsettings.Production.json(for azure deployment)


boydc2014:
@zidaneymar can you elaborate the design of decoupling appsetttings.json into dev.json and prod.json? a few questions is 
1. is this just for publish or will composer need to write to a appsettings.{env}.json instead of appsettings.json?
2. is this a breaking change? what's the impact here? at least i saw now we need a new arugment to start runtime
3. i see you hard code development.json and prod.json, what if other env?

anyhow, let's get time to address those, and i put an onhold on this one in case accidentally merged.

zidaneymar:
> @zidaneymar can you elaborate on the design of decoupling appsetttings.json into dev.json and prod.json? a few questions is
> 
> 1. is this just for publish or will composer need to write to a appsettings.{env}.json instead of appsettings.json?
> 2. is this a breaking change? what's the impact here? at least i saw now we need a new arugment to start runtime
> 3. i see you hard code development.json and prod.json, what if other env?
> 
> anyhow, let's get time to address those, and i put an onhold on this one in case accidentally merged.

1. For now composer will update the customized settings file which is in **ComposerDialogs\settings\appsettings.json**, I think the **root appsettings.json** will not be changed by composer, it has many startup parameters for botruntime.
2. This is to fix #2007, previously the deployment script will write the app id, app password, blob storage and other config to the **root appsettings.json,** if user want to start up the local debugging environment, it will get an 401 unauthorized since the runtime has appid and pwd now.
So we need to decouple the composer dubugging env(deployment) and deployment env(production) to different files.
3. I'm not sure about the other env, in my understanding we have only two environments for debugging(users could set it as 'composer' but it is still used for debugging) and deployment.

BotProject/Templates/CSharp/Program.cs
```diff
@@ -24,12 +24,25 @@ public static void Main(string[] args)
...
+
+                builder.AddJsonFile($"ComposerDialogs/settings/appsettings.json", optional: true, reloadOnChange: true)
+                    .AddJsonFile("appsetting.json", optional: true, reloadOnChange: true)
+                    .UseLuisConfigAdaptor()
+                    .UseLuisSettings();
+
                 if (env.IsDevelopment())
+                {
+                    // Local Debug
+                    builder.AddJsonFile("appsettings.Development.json", optional: true, reloadOnChange: true);
```
boydc2014: s/Development/development

Composer/packages/server/src/models/connector/csharpBotConnector.ts
```diff
@@ -152,6 +152,8 @@ export class CSharpBotConnector implements IBotConnector {
           `bin/Debug/${envSettings.runtimeFrameworkVersion}/BotProject.dll`,
           `--urls`,
           this.endpoint,
+          `--environment`,
```
boydc2014: do we need to update the protocol to add a new parameter

zidaneymar: > do we need to update the protocol to add a new parameter

dotnet receive the env parameter to determine whether it's under development

zidaneymar: this param will be sent to 'dotnet run' iteself, not a botproject level param. 
https://stackoverflow.com/questions/50821504/asp-net-core-set-hosting-environment-in-build-process

PR #2156, opened 2020-03-03
feat: Prevent old content from overwriting newer changes

### LINKED ISSUE #1916: Add support for multi-user features
This is an umbrella issue for multi-user features that will link to more specific tasks.

It is currently not possible for multiple users to simultaneously use a single shared instance of Composer. This prevents users from hosting Composer on a website, and makes it difficult to collaborate.

## Scenarios

Composer should be updated to allow multiple simultaneous sessions that will allow the following scenarios (all assume a single shared instance of Composer running on a shared web resource):

* a developer wants to open 2 dialogs from same bot in different tabs
* a developer wants to open 2 different bots in different tabs
* a developer wants to open a dialog in one tab, and an LG file in another tab.
* Developer wants to open a dialog in one tab while a second developer edits the LG from that dialog
* 2 developers want to open the same bot 
* 2 developers want to open different bots
* A developer wants to send a link to a specific dialog to another developer using the same instance of Composer
* A developer is prevented from overwriting changes to a document made in another tab or by another user

## Required Changes

The work required to implement this functionality can largely be described as allowing multiple instances of the browser "client" to be served by a single instance of the Node-based "server."  These clients may be a single user with multiple tabs open, or multiple users on different machines.

Broadly, this will require that the client specify the identity of the bot being modified as part of (most) API requests, and for the server to be updated to use this information to make modifications to the appropriate resources.

### Basic Multi-user support
- [x] #1917: Update the server to generate and store a unique bot ID for every bot project that is opened. This bot ID must point to the actual "path" to the bot in storage. 
- [x] #1918: Update client code to receive and pass back the bot ID
- [x] #1918: Update client code to include bot ID in the URLs such that `/dialogs/main` is now `/bot/:ID:/dialogs/Main` and ensure that, given a bot id in the url, the application is instantiated with the proper configuration and parameters (in other words, enable deep links)
- [x] Track down other code in the server that might be storing remote user state unnecessarily

### Prevent content overwrite

The application should prevent a user from overwriting changes by another user (or by same user in a different tab). If changes have occurred since the user initially loaded the document, changes should be discarded and an error message should be presented informing the user they should reload the page.

This will prevent key issues we've already heard from customers including overwriting yourself by editing in the wrong tab, and coworker overwriting newer changes with an old instance of an editor.

For reference, see a previous ticket on this subject #1330 

- [x] When the server provides content from a bot project, it should include the last modified time of each piece of content
- [x] When a write operation is made from the client, the last modified time should be passed back down to the server.
- [ ] Before any write operation is undertaken, check the provided last modified time against the current last modified time from the document in storage. If provided < current, reject the write. If provided >= current, accept it and return a new last modified time to the client.
- [x] The client should update its internal last modified time for a given document any time a successful write operation occurs.
    * this is slightly tricky, as in our current implementation, we are debouncing the save action. however, since according to docs, "subsequent calls to the debounced function return the result of the last func invocation." it is feasible that each save action can pass its updated timestamp down to the next.
- [x] The client should present a notification and disable the editing UI when a write is rejected. Present a button to reload the page or at least reload the current document to enable editing.
### [ON HOLD] Support in language servers

These changes will now not be necessary because the language servers are moving into the client.

- [x] Ensure language servers have access to composer config data where mapping is stored
- [x] Update the language servers to support multiple simultaneous bot projects
- [x] Update client calls to language servers to include the bot project ID
- [x] Store indexed information for each bot project in a memory cache along with a last accessed timestamp
- [x] Purge information from memory cache that has not been accessed for > 5 minutes

benbrown:
## Description

This PR introduces a mechanism that will prevent one instance of Composer from overwriting a newer version of that content.  It does this by using the lastModified date provided by the storage layer.

Every write operation now includes the document's last modified date provided at read time. The server checks the date against the current date before accepting the write. If the provided date is older than the latest date "on disk," the write is rejected and the client presents a popup message.

## Task Item

Fixes #1916


Composer/packages/client/src/store/action/lu.ts
```diff
@@ -9,25 +9,41 @@ import { ActionCreator, State } from '../types';
...
+      },
+    });
   } catch (err) {
     setError(store, {
       message: err.response && err.response.data.message ? err.response.data.message : err,
       summary: 'UPDATE LU ERROR',
     });
     //if update lu error, do a full refresh.
-    fetchProject(store);
+    // fetchProject(store);
```
cwhitten: what's the plan here?

Composer/packages/server/src/controllers/project.ts
```diff
@@ -188,9 +188,15 @@ async function updateDialog(req: Request, res: Response) {
...
-    await currentProject.updateDialog(req.body.id, req.body.content);
-    res.status(204);
-    // res.status(200).json({ timestamp: new Date() });
+    // prevent content overwrite
+    const dialog = currentProject.dialogs.find(d => d.id === req.body.id);
+    if (dialog && new Date(dialog.lastModified).getTime() > new Date(req.body.lastModified).getTime()) {
+      res.status(409).json({ message: 'This content was updated since it was loaded. Please reload.' });
+    } else {
+      const lastModified = await currentProject.updateDialog(req.body.id, req.body.content);
+      // res.status(204);
```
cwhitten: remove?

Composer/packages/server/src/controllers/project.ts
```diff
@@ -236,8 +242,15 @@ async function updateLgFile(req: Request, res: Response) {
...
-    const lgFiles = await currentProject.updateLgFile(req.body.id, req.body.content);
-    res.status(200).json({ lgFiles });
+    const file = currentProject.lgFiles.find(d => d.id === req.body.id);
+    if (file && new Date(file.lastModified).getTime() > new Date(req.body.lastModified).getTime()) {
+      res.status(409).json({ message: 'This content was updated since it was loaded. Please reload.' });
+    } else {
+      const lastModified = await currentProject.updateLgFile(req.body.id, req.body.content);
+      res.status(200).json({ lastModified: lastModified });
+    }
+    // const lgFiles = await currentProject.updateLgFile(req.body.id, req.body.content);
```
cwhitten: seems the behavior of this fucction has changed with this commented out

Composer/packages/server/src/controllers/project.ts
```diff
@@ -236,8 +242,15 @@ async function updateLgFile(req: Request, res: Response) {
 
   const currentProject = await BotProjectService.getProjectById(projectId, user);
   if (currentProject !== undefined) {
-    const lgFiles = await currentProject.updateLgFile(req.body.id, req.body.content);
-    res.status(200).json({ lgFiles });
+    const file = currentProject.lgFiles.find(d => d.id === req.body.id);
+    if (file && new Date(file.lastModified).getTime() > new Date(req.body.lastModified).getTime()) {
+      res.status(409).json({ message: 'This content was updated since it was loaded. Please reload.' });
```
cwhitten: We need to allow translation for this string. Typically we can send a status code to the client and have the client own the copy.

Composer/packages/server/src/controllers/project.ts
```diff
@@ -282,8 +295,13 @@ async function updateLuFile(req: Request, res: Response) {
   const currentProject = await BotProjectService.getProjectById(projectId, user);
   if (currentProject !== undefined) {
     try {
-      const luFiles = await currentProject.updateLuFile(req.body.id, req.body.content);
-      res.status(200).json({ luFiles });
+      const file = currentProject.luFiles.find(d => d.id === req.body.id);
+      if (file && new Date(file.lastModified).getTime() > new Date(req.body.lastModified).getTime()) {
+        res.status(409).json({ message: 'This content was updated since it was loaded. Please reload.' });
```
cwhitten: Same status code/translation concern here

Composer/packages/server/src/models/bot/botProject.ts
```diff
@@ -297,16 +294,16 @@ export class BotProject {
...
       throw new Error(`no such lu file ${id}`);
     }
 
-    await this._updateFile(luFile.relativePath, content);
+    const lastModified = await this._updateFile(luFile.relativePath, content);
     await this.luPublisher.onFileChange(luFile.relativePath, FileUpdateType.UPDATE);
-
-    return this.mergeLuStatus(this.luFiles, this.luPublisher.status);
+    return lastModified;
+    // return this.mergeLuStatus(this.luFiles, this.luPublisher.status);
```
cwhitten: remove?

Composer/packages/server/src/models/bot/botProject.ts
```diff
@@ -415,12 +412,17 @@ export class BotProject {
     debug('Creating file: %s', absolutePath);
     await this.fileStorage.writeFile(absolutePath, content);
 
+    // TODO: we should get the lastmodified from the writeFile operation
```
cwhitten: s/lastmodified/lastModified

Composer/packages/server/src/models/bot/botProject.ts
```diff
@@ -433,12 +435,21 @@ export class BotProject {
...
-
     const absolutePath = `${this.dir}/${relativePath}`;
-    await this.fileStorage.writeFile(absolutePath, content);
+
+    // only write if the file has actually changed
+    if (this.files[index].content !== content) {
+      await this.fileStorage.writeFile(absolutePath, content);
+    }
+
+    // TODO: we should get the lastmodified from the writeFile operation
```
cwhitten: s/lastmodified/lastModified

cwhitten: speaking to the TODO comment, perhaps we can _memoize_ `stat` to avoid redundant storage operations

Composer/packages/client/src/store/action/lu.ts
benbrown: Right now, the call to setError causes a popup to appear that tells the user to reload the page to get the latest version of the content.

I think we could specialize that to catch the 409 error that is returned in this instance and show a dialog that *actually does do the fetchProject* for the user instead of having them manually reload the page. This would also take care of the static error message being priovided by the server (comments below)

Composer/packages/server/src/controllers/project.ts
benbrown: It has changed. We no longer return the content of tall the LG files. As far as I can tell, this has no impact as the return value wasn't actually being used on the client side. I will review this again to make sure.

PR #2162, opened 2020-03-04
update name initialization

liweitian:
## Description
Update name initialization which compares to all files not just bots in current path. 
## Screenshots
aaaaa


liweitian:
merged to wenyi's branch

PR #2163, opened 2020-03-04
fix: Double scroll bars in dialog's properties pane

### LINKED ISSUE #2157: Double scroll bars in dialog's properties pane
## Describe the bug

image

## To Reproduce

1. Select a dialog in the navigation pane.
2. See error.

## Expected behavior

Remove one scroll bar.

alanlong9278:
## Description
fix: Double scroll bars in dialog's properties pane.

### How it works?
- hidden editor container scroll
- set editor root div `overflow: auto` to support scroll if editor content is longer than iframe height

## Task Item
closes #2157 

## Screenshots

scroll


yeze322:
I noticed there is anothoer unneessary scroll bar
image

PR #2190, opened 2020-03-06
fix: use nightly build to replace the private bf-lu package

lei9444:
## Description
closes #2190


Composer/packages/client/package.json
```diff
@@ -20,7 +20,7 @@
     "@bfc/extensions": "*",
     "@bfc/indexers": "*",
     "@bfc/shared": "*",
-    "@bfcomposer/bf-lu": "^1.2.9",
+    "@microsoft/bf-lu": "4.8.0-preview.110925",
```
feich-ms: Is it possible not to make it a fixed version? Automatically use latest version.

Issue #2143, opened 2020-03-02
[Accessibility][Bypass Blocks - Property Pane]: Ensure frame must have tittle attribute.


ashish315:
**User Experience** 
Frame title should be define properly so that user can get the information of control.

**Test Environment**
Browser: Microsoft Edge Canary Version: 82.0.425.0 (Official build)
OS build: 2004 (19564.1005)
URL: [Bot Framework Composer](http://localhost:3000/dialogs/Main)
[**Prerequisite Here**](https://github.com/microsoft/BotFramework-Composer)

**Repro Steps:** 
1. Open the above mentioned URL in EDGE.
2. "Bot Framework Composer" homepage will gets opened.
3. Select any of the previously created bot from "Recent Bots" or create one from the available Bots under 'Examples' and activate it.
4. Navigate to already created dialog tree item and activate it. Such as- Greetings, Actions, EndDialog etc.
5. Click on "+" and select 'Ask a question'. Then select any of the available options except Oauth login.
6. Navigate to property pane appeared at right side of the screen and select 'Bot Asks'.
7. Run Accessibility insight and observe the issue.

**Actual:** 
Adaptive property frame doesn't contain any tittle attribute. 

**Expected:**
Ensures <iframe> and <frame> elements contain a non-empty title attribute.

**Note:**
This accessibility issue was found using Accessibility Insights for Web 2.14.1 (axe-core 3.3.2), a tool that helps find and fix accessibility issues. Get more information & download this tool at [http://aka.ms/AccessibilityInsights](http://aka.ms/AccessibilityInsights).

**Element path**
.css-vaviek

**Snippet**
`<iframe name="FormEditor" src="/extensionContainer.html" class="css-vaviek"></iframe>`

**How to fix**
    Fix any of the following:
      aria-label attribute does not exist or is empty
      aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty
      Element has no title attribute or the title attribute is empty
      Element's default semantics were not overridden with role="presentation"
      Element's default semantics were not overridden with role="none"

**Note**
Same issue observed throughout the website for every property pane as well as the flow design.

**MAS Impacted:** [MAS2.4.1](https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={20bce2aa-9952-4a10-af79-3376bcc8766d})

**Attachment:**
[...]

Issue #2123, opened 2020-02-28
[Accessibility][Name, Role, Value - Notification]: Ensures every ARIA input field has an accessible name


ShellySoni:
**User Experience:**
This issue should be resolved for good coding practice as all ARIA input field must has an accessible name.

**Test Environment:**
Environment: Chromium Edge (Anaheim) + Narrator
Build Version: 2004 (OS Build 19564.1005)​​
Microsoft Edge Version: 82.0.425.0 (Official build) canary (64-bit)​
Test URL: http://localhost:3000/home

**Repro Steps:**
1. Open URL: http://localhost:3000/home
2. "Bot Framework Composer" homepage will gets opened.
3. Select any of the previously created bot from "Recent Bots" or create one from the available Bots under 'Examples' and activate it.
4. Select 'Notification' from left navigation.
5. Run K4W tool on the page and Observe the issue. 

**Actual Result:**
ARIA input field i.e. 'All' does not has an accessible name. Screen reader read only like : All selected, selection between three items'.

This accessibility issue was found using Accessibility Insights for Web(keros), a tool that helps debug and find accessibility issues earlier. Get more information & download this tool at http://aka.ms/Keros.

**Expected Result:**
Ensures every ARIA input field has an accessible name. Screen reader should read like : Notification_All button selected, selection between three items'.

**MAS Reference:**
https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={248054a6-5e68-4771-9e1e-242fb5025730}

**Attachment:**
[MAS4.1.2_Notification_All Button.zip](https://github.com/microsoft/BotFramework-Composer/files/4266321/MAS4.1.2_Notification_All.Button.zip)

corinagum:
![MAS4 1 2_Notification_All Button](https://user-images.githubusercontent.com/14900841/75577176-4d3f5d00-5a16-11ea-9fe5-87c49d1070c6.png)
![MAS4 1 2_Notification_All Button_K4W](https://user-images.githubusercontent.com/14900841/75577177-4dd7f380-5a16-11ea-990a-e17920127223.png)

ashish315:
As checked and verified on below environment, the issue is fixed now. So, we are closing the issue.

**Test Environment**
Browser: Microsoft Edge Dev {Version 83.0.467.0 (Official build) dev (64-bit)}
OS build: 2004 (19588.1000)
Screen Reader: Narrator
[...]

PR #2197, opened 2020-03-08
a11y: Accessibility issue

liweitian:
## Task Item
refs #2143 
refs #2123 

## Screenshots
微信图片_20200308230924

### DIFF
--- a/Composer/packages/client/src/pages/design/index.tsx
+++ b/Composer/packages/client/src/pages/design/index.tsx
@@ -389,6 +389,7 @@ function DesignPage(props) {
                     hidden={triggerButtonVisible || !selected}
                     src={`${rootPath}/extensionContainer.html`}
                     ref={addRef}
+                    title={'visual editor'}
                   />
                   {!selected && onRenderBlankVisual(triggerButtonVisible, openNewTriggerModal)}
                 </div>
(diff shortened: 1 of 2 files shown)


Composer/packages/client/src/pages/design/index.tsx
```diff
@@ -389,6 +389,7 @@ function DesignPage(props) {
                     hidden={triggerButtonVisible || !selected}
                     src={`${rootPath}/extensionContainer.html`}
                     ref={addRef}
+                    title={formatMessage('visual editor')}
```
corinagum: @DesignPolice As a team, do we want to enforce copy rules?
`Visual editor`
also for the changes below: 
`Form editor`,
`Show all notifications`,
`Show ${item} notifications`

corinagum: For reference: https://docs.microsoft.com/en-us/windows/win32/uxguide/text-ui

corinagum: *Looks like it might actually be "Visual Editor". I remember the rules incorrectly :/

DesignPolice: Thanks @corinagum - I think we should be following any and all copy rules. My guess is that we have a mix of going to fast and not one source for copy... which we should have. I will tag @mareekuh on this and maybe we can get Yan to take a pass, unless you are wanting the practice... but, I think copy probably needs some help outside of just the rules.

corinagum: Either way, happy to help out. lmk!

Issue #2178, opened 2020-03-05
git pull origin stable (breaks working version)


ADefWebserver:
Running git pull origin stable breaks working version

I run:
C:\Users\Michael\Source\Repos\BotFramework-Composer\Composer>yarn startall

and get this:

> C:\Users\Michael\Source\Repos\BotFramework-Composer\Composer>yarn startall
> yarn run v1.22.0
> $ node scripts/update.js && concurrently --kill-others-on-fail "npm:runtime" "npm:start"
> Checking for updates...
> 
> From https://github.com/microsoft/BotFramework-Composer
>    81232c5f..c0c3aa0a  benbrown/plugins        -> origin/benbrown/plugins
>  * [new branch]        chrande/electron2       -> origin/chrande/electron2
>  * [new branch]        cwhitten/left-align-flow -> origin/cwhitten/left-align-flow
>  * [new branch]        cwhitten/update-checksums -> origin/cwhitten/update-checksums
>  * [new branch]        donglei/plugins         -> origin/donglei/plugins
>  * [new branch]        julong/activity         -> origin/julong/activity
>  * [new branch]        julong/doubleScrollinForm -> origin/julong/doubleScrollinForm
>  * [new branch]        julong/headerTag        -> origin/julong/headerTag
>  * [new branch]        julong/nameForComponent -> origin/julong/nameForComponent
>  * [new branch]        luhan/updateHttpSample  -> origin/luhan/updateHttpSample
>    6ff0da3f..781f3877  master                  -> origin/master
>  * [new branch]        qika/FixBotProject      -> origin/qika/FixBotProject
>  * [new branch]        qika/deployupdate       -> origin/qika/deployupdate
>  * [new branch]        qika/deployupdate2      -> origin/qika/deployupdate2
>    1a6839d6..322e74c4  vishwac/master          -> origin/vishwac/master
>    1a750e58..62ed4719  vishwac/stable          -> origin/vishwac/stable
>  * [new branch]        wenyi                   -> origin/wenyi
>  * [new branch]        wenyluo/plugins         -> origin/wenyluo/plugins
> An update to Composer is available.
> 
> Run 'git pull origin stable' or fetch from origin/stable using another git-based tool.

I then run this:
**git pull origin stable**

and see this:

>  create mode 100644 Composer/packages/server/assets/projects/AskingQuestionsSample/ComposerDialogs/TextInput/TextInput.lg
[...]

PR #2202, opened 2020-03-09
Clone 'master' because stable has issues

sammydeprez:
# Description
#2178
## Problem:
When user follow tutorial to build Composer, it fails
Stable branch can't be used to try out the Composer

## Root cause:
https://github.com/microsoft/BotFramework-Composer/issues/2178#issuecomment-596041394

## Solution:
Clone master branch

### DIFF
--- a/docs/setup-yarn.md
+++ b/docs/setup-yarn.md
@@ -17,6 +17,11 @@ Bot Framework Composer is designed to be a hosted web app. Currently, you need t
     ```
     git clone https://github.com/microsoft/BotFramework-Composer.git
     ```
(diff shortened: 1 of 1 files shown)


a-b-r-o-w-n:
@sammydeprez thanks for the PR. Since we have addressed this issue in master, we are going to be releasing a new stable version today.

a-b-r-o-w-n:
@sammydeprez there is now a new version of the stable branch which includes a fix to the yarn install error.

sammydeprez:
great! thx.

PR #2206, opened 2020-03-10
feat: Move

cwhitten:
closes #2206 

https://share.getcloudapp.com/eDuxRoXn

### DIFF
--- a/Composer/packages/client/src/pages/design/index.tsx
+++ b/Composer/packages/client/src/pages/design/index.tsx
@@ -11,8 +11,9 @@ import formatMessage from 'format-message';
 import { globalHistory } from '@reach/router';
 import get from 'lodash/get';
 import { PromptTab } from '@bfc/shared';
-import { getNewDesigner, seedNewDialog } from '@bfc/shared';
+import { seedNewDialog } from '@bfc/shared';
 import { DialogInfo } from '@bfc/indexers';
+import set from 'lodash/set';
 
 import { VisualEditorAPI } from '../../messenger/FrameAPI';
 import { TestController } from '../../TestController';
@@ -324,8 +325,14 @@ function DesignPage(props) {
   }, [dialogs, breadcrumb]);
 
   async function onSubmit(data: { name: string; description: string }) {
-    const content = { ...getNewDesigner(data.name, data.description), generator: `${data.name}.lg` };
-    const seededContent = seedNewDialog('Microsoft.AdaptiveDialog', content.$designer, content, state.actionsSeed);
+    const seededContent = seedNewDialog(
+      'Microsoft.AdaptiveDialog',
+      { name: data.name, description: data.description },
+      {
+        generator: `${data.name}.lg`,
+      }
+    );
+    set(seededContent, 'triggers[0].actions', state.actionsSeed || []);
     await actions.createDialog({ id: data.name, content: seededContent });
   }
 
--- a/Composer/packages/client/src/pages/design/index.tsx
+++ b/Composer/packages/client/src/pages/design/index.tsx
@@ -11,7 +11,7 @@ import formatMessage from 'format-message';
 import { globalHistory } from '@reach/router';
 import get from 'lodash/get';
 import { PromptTab } from '@bfc/shared';
-import { seedNewDialog } from '@bfc/shared';
+import { seedNewDialog, SDKTypes } from '@bfc/shared';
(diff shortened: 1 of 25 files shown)


Composer/packages/lib/shared/src/dialogFactory.ts
```diff
@@ -146,7 +148,7 @@ export const seedNewDialog = (
       ...designerAttributes,
     },
     ...seedDefaults($type),
-    ...(initialDialogShape[$type] || {}),
+    ...(initialDialogShape[$type] ? initialDialogShape[$type](seededActions) : {}),
```
yeze322: maybe we can check the type of initial shape, such as
```
if (typeof initialShape === 'function') {
  return initialShape(...params);
} else {
  return initialShape;
}
```
Then we don't need to wrap every type with arrow function

cwhitten: ah i like that

yeze322:
After moving SendActivty or Input types to new dialog, lg templates will be lost.
Relies on #2236 to fix that problem.

Composer/packages/client/src/pages/design/index.tsx
```diff
@@ -312,8 +325,14 @@ function DesignPage(props) {
   }, [dialogs, breadcrumb]);
 
   async function onSubmit(data: { name: string; description: string }) {
-    const content = { ...getNewDesigner(data.name, data.description), generator: `${data.name}.lg` };
-    const seededContent = seedNewDialog('Microsoft.AdaptiveDialog', content.$designer, content);
+    const seededContent = seedNewDialog(
+      'Microsoft.AdaptiveDialog',
```
a-b-r-o-w-n: nit: use `SDKTypes` enum here.

Composer/packages/client/src/pages/design/index.tsx
```diff
@@ -312,8 +325,14 @@ function DesignPage(props) {
...
+      { name: data.name, description: data.description },
+      {
+        generator: `${data.name}.lg`,
+      }
+    );
+    set(seededContent, 'triggers[0].actions', state.actionsSeed || []);
```
a-b-r-o-w-n: Why not just add this in the `seedNewDialog` call?

Composer/packages/extensions/visual-designer/src/utils/jsonTracker.ts
```diff
@@ -152,18 +152,22 @@ export function deleteNodes(inputDialog, nodeIds: string[], callbackOnRemovedDat
 }
 
 export function insert(inputDialog, path, position, $type) {
-  const dialog = cloneDeep(inputDialog);
-  const current = get(dialog, path, []);
   const newStep = {
     $type,
     ...seedNewDialog($type, { name: generateSDKTitle({ $type }) }),
```
a-b-r-o-w-n: ❓Should we move the name generation into `seedNewDialog`?

yeze322: I think so, we need a cleanup on all usages of seedNewDialog. Let me handle that in another PR.
image
image
@a-b-r-o-w-n do you think it's necessary to rename `seedNewDialog` to `seedAdaptiveElement` or some other names?

yeze322: https://github.com/microsoft/BotFramework-Composer/issues/2244

Composer/packages/client/src/pages/design/index.tsx
yeze322: Chris had a solution that we can use function inside `initialDialogShape`
```ts
[SDKTypes.AdaptiveDialog]: (initialActions) => ({
    triggers: [
      { actions: initialActions }
    ]
})
```
I just noticed that it's lost after my commit, I will recover that

Composer/packages/client/src/pages/design/index.tsx
yeze322: resolved

Composer/packages/client/src/pages/design/index.tsx
yeze322: resolved

Composer/packages/extensions/visual-designer/src/utils/jsonTracker.ts
a-b-r-o-w-n: Thanks @yeze322

PR #2236, opened 2020-03-11
refactor: support lg templates cross-file copy during Visual Editor copy / paste

### LINKED ISSUE #2207: Cross-file copy of LU / LG resources
## Describe the bug

If a user copies a SendActivity action to another dialog file, the LG content inside that action won't be copied along with the action.

We should also cover the copy behavior of LU resources.

## Browser

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

## Additional context
Follows #2199

### LINKED ISSUE #2210: Multi nodes deletion should also delete the lg template
single nodes worked well but select multi nodes and click delete, the lg templates still exist in lg file.

yeze322:
## Description

The old copy logic relied on a `copyLgTemplate` api provided by @zhixzhan , however, as reported in a recent issue #2207 , after lg templates were saved in separated files, it's not capable for cross-file copy behavior anymore.

In this PR, I made a refactoring on the copy-paste flow to make the copy of actions independent from lg file structure and lg operation api.

### Comparasion
Compared to the old solution that having a heavy 'paste' logic and external api dependency (`copyLgTemplate`):
```ts
// 1 - copy
const copy = actions => _.deepClone(actions);

// 2 - paste
const paste = actions => {
  const clonedActions = copyConstructor(actions);
  walkAdaptiveActions(
    clonedActions,
    action => {
      action.$designer = getDesigner(action.$designer.name, action.$designer.description);
      if (action.$type === 'SendActivity') {
        action.activity = copyLgTemplate(from=action.activity, to=`bfdactivity-${action.$designer.id}`);
      }
    }
  insert(dialog, actions);
}
```
The new solution is more similar to a real clipboard in which copied/cut actions keep real lg text:
```ts
// 1 - copy
const copy = actions => {
  const copiedActions = copyConstructor(actions);
  const dereferenceLgTemplate = lgStr => getLgTemplateBody(lgStr);
  walkAdaptiveActions(
    copiedActions,
    action => {
      if (action.$type === 'SendActivity') action.activity = dereferenceLgTemplate(action.activity);
    }
  )
}

// 2 - paste
const paste = actions => {
  const copiedActions = copyConstructor(actions);
  const buildLgReference= lgStr => updateLgTemplate(id, lgStr);
  walkAdaptiveActions(
    copiedActions,
    action => {
      if (action.$type === 'SendActivity') action.activity = buildLgReference(action.activity);
    }
  )
  insert(copiedActions);
}
```
The most significant difference here is when does it handle lg template references.
- in the old solution, it handles lg references during `paste`; it's buggy if the source actions were deleted before they were pasted.
- in the new solution
  - lg references will be dereferenced during 'copy' which means that a `'${bfdactivity-1234}'` string will be copied as its real value `'hello'` (lgTemplate.body)
  - (**Not included in this PR**) Correspondingly, during 'paste', lg body string `'hello'` will be transformed to a reference again by the api `createLgTemplate`

### Benefits
1. It supports cross-file lg copy since lg body string is kept in clipboard
2. It fixes a rare user behavior: copy nodes -> delete old nodes -> paste nodes
3. It will potentially support dumping copied composer actions to plain text files & copying json in plain text file into composer

### Todos in the future:
- [ ] create necessary lg templates during `paste` as described above (need a new createLgTemplate api @zhixzhan )
- [ ] `ObiEditor.tsx` is exploding, consider to extract copy/paste logic out of it.

## Task Item
closes #2207 
closes #2210

### DIFF
--- a/BotProject/Templates/CSharp/Scripts/deploy.ps1
+++ b/BotProject/Templates/CSharp/Scripts/deploy.ps1
@@ -117,7 +117,13 @@ if ($luisAuthoringKey -and $luisAuthoringRegion) {
 	if (Get-Command bf -errorAction SilentlyContinue) {
 		$customizedSettings = Get-Content $(Join-Path $remoteBotPath settings appsettings.json) | ConvertFrom-Json
 		$customizedEnv = $customizedSettings.luis.environment
-		bf luis:build --in .\ --botName $name --authoringKey $luisAuthoringKey --dialog --out .\generated --suffix $customizedEnv -f
+		
+		# create generated folder if not exists
+		if (!(Test-Path generated)) {
+			New-Item -ItemType Directory -Force -Path generated
+		}
+		
+		bf luis:build --in .\ --botName $name --authoringKey $luisAuthoringKey --dialog --out .\generated --suffix $customizedEnv -f --region $luisAuthoringRegion
 	}
 	else {
 		Write-Host "bf luis:build does not exist, use the following command to install:"
--- a/Composer/packages/client/src/pages/setting/deployment/deployWizard.js
+++ b/Composer/packages/client/src/pages/setting/deployment/deployWizard.js
@@ -57,9 +57,9 @@ export const DeployWizard = props => {
       children: <DeployWizardStep2 botValues={botValues} nextStep={completeStep2} closeModal={resetModal} />,
     },
     {
-      title: formatMessage('Deploy your Bot'),
+      title: formatMessage('Publish your Bot'),
       subText:
-        'Deploy your bot code and Composer assets to Azure using the command below. Run this any time you want to update your bot on Azure.',
+        'Publish your bot code and Composer assets to Azure using the command below. Run this any time you want to update your bot on Azure.',
       children: <DeployWizardStep3 botValues={botValues} closeModal={resetModal} />,
     },
   ];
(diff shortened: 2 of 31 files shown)


yeze322:
@zhixzhan I need a new LG api whose interface is
```ts
const createLgTemplate = (templateName: string, templateBody: string, templateParams?: any[]) => Promise<string>; // returns either the new template name or the whole template
```
Do you have any idea on that?

yeze322:
@cwhitten this PR will fix the problem in your 'Move' PR that moving SendActivity to new dialog will lose lg string. Current version works but the value of 'prompt' field in SendActivity will be the lg body string.
For example:
```ts
// Action to be moved
{
  $type: 'SendActivity',
  activity: '${bfdactivity-1234()}' // the real content is '- hello'
}
// Action will appear in new dialog 
{
  $type: 'SendActivity',
  activity: '- hello'
}
```
I will solve that problem when lg api is ready.

zhixzhan:
> @zhixzhan I need a new LG api whose interface is
> 
> ```ts
> const createLgTemplate = (templateName: string, templateBody: string, templateParams?: any[]) => Promise<string>; // returns either the new template name or the whole template
> ```
> 
> Do you have any idea on that?

Did you miss the `id` of file? And what's the difference compare to existed shell api?
```
updateLgTemplateHandler({ id, templateName, template }, event)
```

yeze322:
@zhixzhan I used the `updateLgTemplate` you told me, it could solve my problem!

PR #2246, opened 2020-03-12
chore: sync oauth input property in form editor

### LINKED ISSUE #2260: OAuth Login null reference exception
When using the OAuth login component in the Composer, i get an exception with the following message: 

> "Object reference not set to an instance of an object."

image

I'm using the OAuth component exaclty as i was using it with the previous release and it was working perfectly.

## Version

image

## Browser

- [X] Chrome
- [ ] Safari
- [ ] Firefox
- [X] Edge

## OS

- [ ] macOS
- [X] Windows
- [ ] Ubuntu

alanlong9278:
## Description

sync oauth input property in form editor. Replace `tokenProperty` to `property` for `OAuthInput` in form editor appschema.

## Task Item

fixes #2260

### DIFF
--- a/Composer/packages/lib/shared/src/appschema.ts
+++ b/Composer/packages/lib/shared/src/appschema.ts
@@ -1960,7 +1960,7 @@ export const appschema: OBISchema = {
           description: 'Time out setting for the OAuth signin card.',
           default: '900000',
         },
-        tokenProperty: {
+        property: {
           $role: 'expression',
           title: 'Token property',
           description: 'Property to store the OAuth token result.',


Composer/packages/lib/shared/src/appschema.ts
```diff
@@ -1960,7 +1960,7 @@ export const appschema: OBISchema = {
           description: 'Time out setting for the OAuth signin card.',
           default: '900000',
         },
-        tokenProperty: {
+        property: {
           $role: 'expression',
           title: 'Token property',
```
a-b-r-o-w-n: Do the other properties need to change? Like title, description, etc?

luhan2017: There are some additional schema like allowInteruption etc, let's don't change it for R8, because it needs UX update which is too heavy.

PR #2233, opened 2020-03-11
feat: enable composer to call a remote skill 

luhan2017:
## Description

In this PR, I updated the bot project and enable composer to connect to a remote skill. 
Note: A remote skill right now could not be a normal skill, It should have the ability to send and handle EndOfConversation type of Activity.  Refer to https://github.com/microsoft/botbuilder-dotnet/tree/master/FunctionalTests/Skills/SimpleBotToBot/EchoSkillBot to create a remote skill.

How to connect to a remote skill:

Step1: configure settings. to call a remote skill, the bot must be registered with a MicrosoftAppID and MicrosoftPassword, and also SkillHostEndpoint, which is the skill endpoint of the current bot. 

  "MicrosoftAppPassword": "",
  "MicrosoftAppId": "",
  "SkillHostEndpoint": "http://localhost:3978/api/skills/" // http://localhost:3978 is the endpoint of the caller not callee. 

Step2: add SkillDialog in the bot, for example:
{
  "$kind": "Microsoft.SkillDialog",
  "skillAppId": "3083d5b8-a3b6-4d50-82c7-af10aa96b20c",   // the remote skill MicrosoftAppId
  "skillEndpoint": "http://localhost:1205/api/messages",  // the remote skill endpoint
  "activity": "Hi"   
}

Note: step2 can be added in Composer UX after the UX part is ready. 

Step3: Click StartBot and Test in Emulator.

This PR is only to enable Composer to call a remote skill. It doesn't include creating a "remote skill" using composer. 

## Task Item

fixes #696 

## Screenshots

image

PR #2249, opened 2020-03-12
feat: Support creating SkillDialog in 'New Dialog...'

yeze322:
## Description

follows #2233 

(has a style issue)
image


cwhitten:
@yeze322 I don't think this is the UX direction we want, let's follow-up offline.

yeze322:
As discussed with Chris, another solution was already integrated in Lu's PR. Close this one

PR #2266, opened 2020-03-13
chore: add default alwaysPrompt value in OAuthInput step

alanlong9278:
## Description

add default alwaysPrompt value in OAuthInput step.

There is a bug in SDK, that requires alwaysPrompt property must have value, otherwise it will throw an exception called : Object reference to null.

This PR is to hotfix that issue in Composer side to work around that bug. 

## Task Item

fixes #2266


luhan2017:
This is a hotfix for the OAuth issue.

PR #2274, opened 2020-03-16
fix: add disk names

### LINKED ISSUE #2257: Load project from X: hard drive
Hi all,
I need to load project from D drive!
From the GUI, I cannot select a different drive than C:

image

liweitian:
## Description
1. The bug only exists on windows because of hard disk partition.
2. The parent of root disk will be This PC. Go to 'This PC', it will show all existing disk names.
3. Note that this is a corner case, Path module does not recognize 'This PC' as a valid path.
## Task Item
Closes #2257 
## Screenshots
aaaaa

### DIFF
--- a/Composer/packages/lib/shared/src/constant.ts
+++ b/Composer/packages/lib/shared/src/constant.ts
@@ -1,5 +1,6 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
+import formatMessage from 'format-message';
 
 export const SensitiveProperties = ['MicrosoftAppPassword', 'luis.authoringKey', 'luis.endpointKey'];
 export const FieldNames = {
@@ -10,3 +11,5 @@ export const FieldNames = {
   DefaultCase: 'default',
   Cases: 'cases',
 };
+
+export const thisPC = formatMessage('This PC');
--- a/Composer/packages/lib/shared/src/constant.ts
+++ b/Composer/packages/lib/shared/src/constant.ts
@@ -1,6 +1,5 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
-import formatMessage from 'format-message';
 
 export const SensitiveProperties = ['MicrosoftAppPassword', 'luis.authoringKey', 'luis.endpointKey'];
 export const FieldNames = {
@@ -11,5 +10,3 @@ export const FieldNames = {
   DefaultCase: 'default',
   Cases: 'cases',
 };
-
-export const thisPC = formatMessage('This PC');
--- a/Composer/packages/server/src/services/storage.ts
+++ b/Composer/packages/server/src/services/storage.ts
@@ -70,6 +70,20 @@ class StorageService {
   // return connent for file
   // return children for dir
   public getBlob = async (storageId: string, filePath: string, user?: UserIdentity) => {
+    if (filePath === 'This PC') {
+      const diskNames = ['C:/', 'D:/', 'E:/', 'F:/', 'G:/', 'H:/', 'I:/', 'J:/', 'K:/'].filter(n => fs.existsSync(n));
+      return {
+        name: '',
+        parent: 'This PC',
+        children: diskNames.map(d => {
+          return {
+            name: d,
+            type: 'folder',
+            path: d,
+          };
+        }),
+      };
+    }
     const storageClient = this.getStorageClient(storageId, user);
     const stat = await storageClient.stat(filePath);
     if (stat.isFile) {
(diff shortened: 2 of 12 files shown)


Composer/packages/server/src/settings/index.ts
```diff
@@ -18,8 +19,12 @@ interface Settings {
   runtimeFrameworkVersion: string;
   botsFolder: string;
   appDataPath: string;
+  isWindows: boolean;
+  validDiskNames: string[];
 }
 
+const diskNames = ['C:/', 'D:/', 'E:/', 'F:/', 'G:/', 'H:/', 'I:/', 'J:/', 'K:/'];
```
boydc2014: let's talk about this later, see if we can avoid a hard code here

boydc2014: The suggestion is, no matter what the implementation is, it's would be necessary to update the protocol, adding new state or settings, we should use existing abstraction, and treat disk part in windows just as normal folders.

boydc2014:
Looks way better now.

Composer/packages/server/src/services/storage.ts
```diff
@@ -68,6 +68,20 @@ class StorageService {
   // return connent for file
   // return children for dir
   public getBlob = async (storageId: string, filePath: string) => {
+    if (filePath === 'This PC') {
+      const diskNames = ['C:/', 'D:/', 'E:/', 'F:/', 'G:/', 'H:/', 'I:/', 'J:/', 'K:/'].filter(n => fs.existsSync(n));
```
yeze322: have you tested the behavior if I rename 'This PC' to another name?

yeze322: image

liweitian: This PC is a sample name. I think we just need a way for users to switch disk here.

yeze322: At least we need to verify 2 scenarios to know your code's limitation:
1. does it work after user renamed the PC name?
2. does it work for users in different regions whose system language is not English?

Even if we don't care about these two scenarios, the string 'This PC' shouldn't be hard-coded and appears in two different files.

liweitian: I updated code to take scenarios 2 into consideration and translated 'this PC'. It should be good now.

cwhitten:
@liweitian please address conflicts

liweitian:
> @liweitian please address conflicts

done

Composer/packages/lib/shared/src/constant.ts
```diff
@@ -10,3 +11,5 @@ export const FieldNames = {
   DefaultCase: 'default',
   Cases: 'cases',
 };
+
+export const thisPC = formatMessage('This PC');
```
a-b-r-o-w-n: This can't be statically export using `formatMessage`. You should see this in the browser console as a warning. Any usage of `formatMessage` must be at runtime because it uses runtime settings for determining the locale.

Composer/packages/server/src/services/storage.ts
```diff
@@ -70,6 +72,20 @@ class StorageService {
   // return connent for file
   // return children for dir
   public getBlob = async (storageId: string, filePath: string, user?: UserIdentity) => {
+    if (filePath === thisPC) {
+      const diskNames = ['C:/', 'D:/', 'E:/', 'F:/', 'G:/', 'H:/', 'I:/', 'J:/', 'K:/'].filter(n => fs.existsSync(n));
```
a-b-r-o-w-n: Couldn't disk names be arbitrary?

Composer/packages/lib/shared/src/constant.ts
a-b-r-o-w-n: Have you considered using `/` instead of `This PC`? I'm concerned that by translating this string at runtime, there may be some hidden bugs due to the translation. Also, the server is not configured for i18n at this point, so it would not get translated on the server.

boydc2014: yes, i was also thinking about "/" don't require any extra explanation

Composer/packages/server/src/services/storage.ts
boydc2014: We should call some system api, if not, use some commands should also be fine.

Composer/packages/lib/shared/src/constant.ts
liweitian: switched to /.

Composer/packages/server/src/services/storage.ts
liweitian: Googled about this problem. Actually we cannot have an arbitrary names but we can change the label of disks. The Path C:/.../... won't change to Label:/.../... if we give it a label. So do we really need to show something like C:(label)/..?
Or are you suggesting I should remove the hard code here? I added a cmd script in env.ts which will get all disks names.

PR #2275, opened 2020-03-16
fix: use ':' instead of '=' for prefix

### LINKED ISSUE #2222: Drop adding `=` prefix if property value starts with `=`
Consider dropping double `=` if property value already has `=` prefix.

image

alanlong9278:
## Description
 Use `:` instead of `=` for prefix

## Task Item
closes #2222 

## Screenshots

image


cwhitten:
@alanlong9278 we decided we wanted to keep the expression = sign but replace the assignment operator in the node with `:`, so:

`somval: =somefunc`

Can you update this?

alanlong9278:
> @alanlong9278 we decided we wanted to keep the expression = sign but replace the assignment operator in the node with `:`, so:
> 
> `somval: =somefunc`
> 
> Can you update this?

Sure.

Composer/packages/extensions/visual-designer/src/utils/prefixOperation.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+export function addEqualPrefix(first: string, second: string): string {
```
boydc2014: Since we just move from = to :, do we still need this function?

alanlong9278: > Since we just move from = to :, do we still need this function?

revert code for the new logic

PR #2284, opened 2020-03-16
feat: Managed LU sample - updates to TodoBotWithLuis

### LINKED ISSUE #806: [Managed LU] Implement complicated sample bot with managed LU
Please reference 
https://github.com/microsoft/BotFramework-Composer/issues/733

https://microsoft.sharepoint.com/:w:/r/teams/ConversationalAI785/_layouts/15/Doc.aspx?sourcedoc=%7BD4BE6F56-9089-45C2-8815-192E66786A0C%7D&file=Language%20Understanding%20in%20Bot%20Framework%20Designer.docx&action=default&mobileredirect=true&cid=0d3c9409-ede4-4fac-8da0-7fcd9545b58a

vishwacsena:
## Description

Fixes #806 
- Updated TodoBotWithLUIS to use `Managed LU` constructs
- Updated LU to the new `@` entity definition syntax
- Configured all inputs to allow interruption with expressions controlling `Value` and `AllowInterruption` properties.

## Documentation 

# Advanced dialog and language understand concepts

Conversations do not always progress is a linear fashion. Users might want to over specify information or present information out of order or would like to make corrections etc. 

Bot Framework composer relies on the advanced dialog capabilities offered by Adaptive dialogs as well as relies on managing one or more backing LUIS applications to support these advanced scenarios. 

This topic looks at each one of these in a bit more detail. For this doc, snippets/ concepts covered here are available in the `TestManagedLU` sample.

## Flexible entity extraction
Let's take this simple example - 
```
user: Hi
Bot: Hello, what is your name?
```
To this prompt from the bot, the user can say one of the following
```
user: vishwac
user: my name is vishwac
user: you can call me v
```
In several of these examples, we need the bot to not just verbatim take everything the user said as input but detect `vishwac` or `v` as the user name. 

Within composer, you can achieve this by **setting the recognizer on the dialog to LUIS** and subsequently using LU notation to define user response to the specific `what is your name` input. 

Take a look at the input and all its configured properties in the `TestManagedLU` under `UserProfile` dialog, `BeginDialog` trigger, `AskForName` input.

The input has the following configuration - 

```
    "property": "user.name"
    "value": "=coalesce(@userName, @personName)"
    "allowInterruptions": "!@userName && !@personName"
```
And the following `Expected response` LU configuration

```
- my name is {@userName = vishwac}
- I'm {@userName = tom}
- you can call me {@userName = chris}
- I'm {@userName = scott} and I'm {@userAge = 36} years old
> add few patterns
- my name is {@userName}

> add entities
@ prebuilt personName hasRoles userName
```

Two key properties of interest are `value` and `allowInterruptions`. The expression specified in `value` property will be evaluated on every single time user responds to the specific input. 

In this case, the expression `=coalesce(@userName, @personName)` attempts to take the first non null entity value `userName` or `personName` and assigns it to `user.name`. The input will issue a prompt if the property `user.name` is null even after the value assignment unless `always prompt` evaluates to `true`.

The next property of interest is `allowInterruptions`. This is set to the following expression - `!@userName && !@personName`. This literally means what this expression reads - _allow an interruption if we did not find a value for entity `userName` or entity `personName`_

Notice that you can just focus on things the user can say to respond to this specific input in the `Expected replies`. With these capabilities, you get to provide labelled examples of the entity and use it no matter where/ how it was expressed in the user input. 

If a specific user input does not work, simply try adding that utterance to the `Expected response`.

## Out of order entity extraction

Consider this example

```
user: add an item to the list
bot: sure, what is the title? 
user: buy milk
bot: ok. pick the list type - todo | shopping
user: shopping list
bot: ok. i've added that.
```

The user could have pre-emptively answered multiple questions in the same resposne. Here is an example

```
user: add an item to the list
bot: sure, what is the title?
user: add buy milk to the shopping list
bot: ok. I've added that.
```

You can see how this is all wired up in the `TestManagedLU` under `Add` dialog, `BeginDialog` trigger, `AskForTitle` and `AskForListType` inputs.

By including the `value` property on each of these inputs, we can pick up any entities recognized by the recognizer even if it was specified out of order.

## Interruption

Interruptions can be handled at two levels - locally within a dialog as well as re-routing as a global interruption. By default adapive dialog does this for any inputs - 

1. On every user response to an input action's prompt, 
2. Run the recognizer configured on the parent adaptive dialog that holds the input action
3. Evaluate the `allowInterruption` expression. 
    a. If it evaluates to `true`, evaluate the triggers that are tied to the parent adaptive dialog that holds the input action. If any triggers match, execute the actions associated with that trigger and then issue a re-prompt when the input action resumes.
    b. If it evaluates to `false`, evaluate the `value` property and assign it as a value to the `property`. If `null` run the internal entity recognizer for that input action (e.g. number recognizer for number input etc) to resolve a value for that input action.

### Handling interruptions locally
With this, you can add contextual responses to inputs via `OnIntent` triggers within a dialog. Consider this example: 

```
user: hi
bot: hello, what is your name?
user: why do you need my name?
bot: I need your name to address you correctly. 
bot: what is your name?
user: I will not give you my name
bot: Ok. You can say "My name is <your name>" to re-introduce yourself to me. 
bot: I have your name as "Human"
bot: what is your age? 
```

See `TestManagedLU` under `UserProfile` dialog, `Why`, `NoValue` or `Cancel` triggers.

### Handling interruptions globally

Adaptive dialogs have a consultation mechanism which propagates a user message up the parent dialogs until a dialog has a trigger that fires. If no dialog's triggers fired upon consultation then the active input action gets the user utterance back for its own processing. Consider this example - 

```
user: hi
bot: hello, what is your name? 
user: what can you do?
bot: I'm a demo bot. I can manage todo or shopping lists. 
bot: what is your name? 
```

Notice that the bot understood interruption and presented the help response. See `TestManagedLU` `UserProfile` as well as `Help` dialogs.


Composer/packages/server/assets/projects/ToDoBotWithLuisSample/ComposerDialogs/Main/Main.dialog
```diff
@@ -1,18 +1,18 @@
...
-  "autoEndDialog": true,
+  "autoEndDialog": "true",
   "defaultResultProperty": "dialog.result",
-  "recognizer": "Main.lu",
   "triggers": [
     {
       "$type": "Microsoft.OnConversationUpdateActivity",
       "$designer": {
-        "id": "376720",
-        "name": "WelcomeUser"
```
cwhitten: This is breaking an End to End test. Can you give this a name of `WelcomeUser`?

vishwacsena: Yes. Added.

xieofxie:
how about add scores to intents of main dialog? Or todo like 'play basketball' will be routed to Delete

vishwacsena:
Yes. The original sample had this and I missed this in the rewrite. I'll update the PR.

Composer/packages/server/assets/projects/ToDoBotWithLuisSample/dialogs/AddItem/AddItem.dialog
```diff
@@ -0,0 +1,114 @@
...
+        {
+          "$type": "Microsoft.SetProperties",
+          "$designer": {
+            "id": "811190",
+            "name": "Set properties"
+          },
+          "assignments": [
+            {
+              "property": "dialog.itemTitle",
+              "value": "=coalesce(@itemTite, $itemTitle)"
```
xieofxie: Do we need coalesce $itemTitle here? I have tested and find that dialog.xx variables will always be null when the dialog is created.

vishwacsena: @xieofxie this is good practice to show in our samples that help users understand how to build sub-dialogs that can be composed easily with other dialogs. This sets the stage up for a sub-dialog to be called in both an intent trigger as well as for any re-use invocation where some properties might be passed in via `Options` to BeginDialog

PR #2285, opened 2020-03-17
fix: Some editting bugs in Lu editor

### LINKED ISSUE #2141: inline editing for lu errors and inline editing refresh for notification navigation
Please reference following video :
https://recordit.co/7c4apRcaLx

- Editing in all up view with error
- (lsp issue) press enter after one utterance (expect the new line starts with '-', but '-' has been added to the next line and this new added starts with nothing)
- Notification page showed the errors 
- navigation back to "design flow" page
- (error) form does not refreshed (as no error showed) 
- refresh manually and inline form reloaded
- (error) inline editing in form to eliminate the error (but the lu file does not updated)
-so notification error does not change and back to all up view still can see the error

cosmicshuai:
fixes #2141 

1. Incorrect syntax highlight for entity definition. For example:
image

2. Wrong place of inserting '-' in auto-completion of utterance
docFormat

3. trigger error of lebeling experience
labeling

### DIFF
--- a/Composer/packages/lib/code-editor/src/languages/lu.ts
+++ b/Composer/packages/lib/code-editor/src/languages/lu.ts
@@ -36,12 +36,12 @@ export function registerLULanguage(monaco: typeof monacoEditor) {
         [/^\s*>\s*[\s\S]*$/, { token: 'comments' }],
         [/^\s*-/, { token: 'utterrance-indentifier', next: 'utterrance' }],
         [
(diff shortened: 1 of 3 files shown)


zhixzhan:
Can you record a gif or screenshot? that would be very nice.

cosmicshuai:
> Can you record a gif or screenshot? that would be very nice.

Sure. You can see the gif now.

Issue #2261, opened 2020-03-12
Handle empty lg file better


a-b-r-o-w-n:
When creating a new empty bot, all of the lg fields give this error. Should we suppress this?

![image](https://user-images.githubusercontent.com/3619060/76570669-18e68a80-6473-11ea-94f7-6cf33c791221.png)

Danieladu:
It is actually a warning diagnostic, the editor needs to distinguish between warning and error diagnostics.

zhixzhan:
Sure, will update it

cwhitten:
addressed via https://github.com/microsoft/BotFramework-Composer/pull/2289

PR #2289, opened 2020-03-17
feat: lg/lu editor support warning message style

zhixzhan:
## Description

Distinguish warning message and error message.

## Task Item

refs #2261 

## Screenshots

Untitled


zhixzhan:
> There are some duplicate code paths here that I'm concerned about. Is it possible to rethink how the shell shares errors to the LU/LG fields instead of each component needing to parse/format the diagnostics?
> 
> One solution is moving the knowledge of the diagnostics into the code editors themselves because their domain is already aware of the lg/lu diagnostic.

I also thought about that, accept diagnostics instead of errorMsg or warningMsg can simplify a lot usage scenarios. only if we don't want customize the message, directly show the message from diagnostics. For example, same diagnostic in inline editor show less message compare to all up view, cause form have less space.
If it's not so important to customize message, we should pass diagnostic, or further more do diagnosing in editor.

Composer/packages/client/src/pages/language-generation/code-editor.tsx
```diff
@@ -59,9 +60,10 @@ const CodeEditor: React.FC<CodeEditorProps> = props => {
   useEffect(() => {
     const currentDiagnostics = inlineMode && template ? filterTemplateDiagnostics(diagnostics, template) : diagnostics;
 
-    const isInvalid = !isValid(currentDiagnostics);
-    const text = isInvalid ? combineMessage(currentDiagnostics) : '';
-    setErrorMsg(text);
+    const errors = findErrors(currentDiagnostics);
+    const warnings = findWarnings(currentDiagnostics);
+    setErrorMsg(combineMessage(errors));
+    setWarningMsg(combineMessage(warnings));
```
yeze322: Not introduced by this PR and we don't need to fix, the problem here is we are setting states inside `useEffect`.
It indicates that either our component props is not well designed and broke the change detection or it could be computed data from props.

yeze322: Correspondingly, I checked the LU editor below and found that these two `code-editor` are sharing a lot of code logic but LU editor is using computed value instead of states, that's confusing.
Seems we just copied one editor to build the other one.

Composer/packages/extensions/obiformeditor/src/Form/widgets/LgEditorWidget.tsx
```diff
@@ -74,11 +74,20 @@ export const LgEditorWidget: React.FC<LgEditorWidgetProps> = props => {
     },
   };
 
-  const diagnostic = lgFile && filterTemplateDiagnostics(lgFile.diagnostics, template)[0];
+  const combineMessage = (diagnostics: Diagnostic[]): string => {
```
yeze322: Is the `combineMessage` here the same as the one inside bfc/indexer? If so we can use the one in bfc/indexer, if not we should rename it.

yeze322: Another thing I'm thinking about is should those dialognostics logic belong to LG editor lib? The LGEditor could just receive the `diagnostics` as an input prop.
You know that the whole FormEditor are under refactoring by Andy, if we insert so many diagnostics logic inside form/LgEditorWidget, they are hard to be merged.

Composer/packages/extensions/obiformeditor/src/Form/widgets/LuEditorWidget.tsx
```diff
@@ -78,11 +78,19 @@ export class LuEditorWidget extends React.Component<LuEditorWidgetProps> {
   render() {
     const { height = 250 } = this.props;
     const { luFile, luFileId, luIntent, name } = this;
-    const diagnostic = luFile && filterSectionDiagnostics(luFile.diagnostics, luIntent)[0];
 
-    const errorMsg = diagnostic
-      ? diagnostic.message.split('error message: ')[diagnostic.message.split('error message: ').length - 1]
-      : '';
+    const combineMessage = (diagnostics: Diagnostic[]): string => {
```
yeze322: ditto. 1. LU & LG are too similar to be separated components. 2. dialogstic logic might belong to LU/LG editor lib

Composer/packages/extensions/obiformeditor/src/Form/widgets/LgEditorWidget.tsx
zhixzhan: Yeah, so I guess everybody likes diagnostic more than errorMsg, then I should refactor this.

Composer/packages/client/src/pages/language-generation/code-editor.tsx
zhixzhan: The difference is lg save in real time, but lu inline editor only do save when content is valid, that's an issue from luParser may be fixed in future. It decides the error message comes from different place, useState is more flexible here.

Composer/packages/extensions/obiformeditor/src/Form/widgets/LuEditorWidget.tsx
zhixzhan: The CRUD/parse logic of lg/lu are not  100% same. Abstract them into one component would bring more maintenance cost, so  keep two standalone components at this moment is more flexiable. except the copied code are not that looks fancy.

zhixzhan:
I worked out a solution to move to pass diagnostics:
1. editor both accept diagnostics and errorMsg/warningMsg. but the display priority  is errorMsg>diagnostics.errorMsg>warningMsg>diagnostics.warningMsg

2. simplify the message, use `L2-L3` replace `line 2 - line 3`, remove the `source: Main `. Then the message would not be too long for inline editor. but also includes necessary information.

Composer/packages/client/src/pages/language-generation/code-editor.tsx
yeze322: I got it

PR #2295, opened 2020-03-17
a11y: make dialog name required

### LINKED ISSUE #2073: [Accessibility][Keyboard -New Dialog]: Mandatory sign is not defined for name in New dialog.
**User Experience** 
User with vision disability will not get any information about the required field. He might be confused with this.

**Test Environment**
OS build: 2004 (19564.1005)
Screen Reader: Narrator

**Repro Steps:** 
3. Now, navigate to 'New Dialog' button available in the filter dialog pane and activate it..
4. Type the name of the dialog which is already created ex- Actions.
5. Observer the issue.

**Actual:** 
The name is mandatory fields but no visual or audio information is conveyed to the user about it.

**Expected:**
Their should be astreak symbol for the mandatory field as well as the screen reader should inform about it, so that the user will aware about that.

**MAS Impacted:** [MAS3.3.4](https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={5f773cee-c1c7-4292-b330-cb2f4aad00c4})

**Attachment:**
[MAS3.3.4_Mandatory sign is not defined for name in New dialog.zip](https://github.com/microsoft/BotFramework-Composer/files/4249012/MAS3.3.4_Mandatory.sign.is.not.defined.for.name.in.New.dialog.zip)
MAS3 3 4_Mandatory sign is not defined for name in New dialog

beyackle:
## Description

One-line fix to make it clear that this field is mandatory

## Task Item

Closes #2073 

## Screenshots

name-required

### DIFF
--- a/Composer/packages/client/src/pages/design/createDialogModal.tsx
+++ b/Composer/packages/client/src/pages/design/createDialogModal.tsx
@@ -89,6 +89,7 @@ export const CreateDialogModal: React.FC<CreateDialogModalProps> = props => {
               onChange={updateForm('name')}
               errorMessage={formDataErrors.name}
               data-testid="NewDialogName"
+              required
             />
           </StackItem>
           <StackItem grow={0} styles={wizardStyles.halfstack}>


corinagum:
@boydc2014  This should also include `aria-required = true`. This helps with the AT reading out the information. Could you add that in as well?

beyackle:
Adding an aria prop to a Fabric component is something the Fabric team should know about (and would help anyone else adding the required prop to their TextFields); I'm just taking advantage of the API they give us.

corinagum:
🤦‍♀️ Checked the docs properly after your message. Sorry about that. I'll go back and fix my own code later in my next PR.

corinagum:
Looks like they don't plan on implementing it. 
https://github.com/OfficeDev/office-ui-fabric-react/issues/11585#issuecomment-569999664

Thanks for the correction, @beyackle

PR #2296, opened 2020-03-18
feat: Update BotProject.csproj package version to nuget 4.8.0 and 4.8.0-rc0

luhan2017:
closes #2296


boydc2014:
Looks good, just a question, why only those 3 packages need to be updated to rc, what about other packages?

PR #2306, opened 2020-03-19
fix: manage LU resources in Visual Editor

yeze322:


yeze322:
I'm thinking about that we may not need a copy LU mechanism

PR #2310, opened 2020-03-19
a11y: remove extra Link stuff from settings Nav

### LINKED ISSUE #2102: [Accessibility][Focus Order - Settings]: Extra focus is provided on buttons available under 'Configuration'.
**User Experience:**
Keyboard users will get confuse if extra focus is provided on buttons available under 'Configuration' when user navigate using up/down arrow keys.

**Test Environment:**
Environment: Chromium Edge (Anaheim) + Narrator
Build Version: 2004 (OS Build 19564.1005)​​
Microsoft Edge Version: 82.0.425.0 (Official build) canary (64-bit)​
Test URL: http://localhost:3000/home

**Repro Steps:**
1. Open URL: http://localhost:3000/home
2. "Bot Framework Composer" homepage will gets opened.
3. Select any of the previously created bot from "Recent Bots" or create one from the available Bots under 'Examples' and activate it.
4. Select 'Settings' from left navigation.
5. Navigate on buttons available under 'Configuration' using up/down arrow keys.
6. Observe the issue.

**Actual Result:**
Extra focus is provided on buttons available under 'Configuration' when user navigate using up/down arrow keys. First tab focus is provided on whole control, after this when user navigate using down arrow key then focus is not visible. Then on again pressing down arrow key, focus gets visible to next button.

**Note:**
When user check with screen reader, then it is observed that along with the extra focus, screen reader read two different roles on the same button i.e when focus is on whole button, then its role as button is announced & when user press down arrow key then its role as link is announced. Please ( Refer:Attachment ).

**Expected Result:**
Extra focus should not be provided on buttons available under 'Configuration'. On pressing down arrow key, focus should directly jump to next button.

**MAS Reference:**
https://microsoft.sharepoint.com/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={0de7fbe1-ad7e-48e5-bcbb-8d986691e2b9}

**Attachment:**
[MAS2.4.3_Settings_Screen reader_Incorrect announcement of role.zip](https://github.com/microsoft/BotFramework-Composer/files/4255385/MAS2.4.3_Settings_Screen.reader_Incorrect.announcement.of.role.zip)

beyackle:
## Description

The Nav component provided by Fabric is good enough at managing its own state, and we don't need a lot of the extra machinery we had to make Links do the right thing; it's that machinery that was causing the extra keyboard-navigation targets.

## Task Item

Closes #2102 

## Screenshots

Just to demonstrate that it still looks the same:

image

### DIFF
--- a/Composer/packages/client/src/pages/setting/index.tsx
+++ b/Composer/packages/client/src/pages/setting/index.tsx
@@ -3,10 +3,10 @@
 
 /** @jsx jsx */
 import { jsx } from '@emotion/core';
-import { Fragment, useContext, useState } from 'react';
+import { Fragment, useContext } from 'react';
 import formatMessage from 'format-message';
-import { Nav } from 'office-ui-fabric-react/lib/Nav';
-import { Link, RouteComponentProps } from '@reach/router';
+import { Nav, INavLinkGroup } from 'office-ui-fabric-react/lib/Nav';
+import { RouteComponentProps } from '@reach/router';
 
 import { StoreContext } from '../../store';
 import { ToolBar } from '../../components/ToolBar';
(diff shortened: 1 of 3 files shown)


liweitian:
Tried on my side. It is good.

PR #2316, opened 2020-03-20
fix: auto re-init data.json when schema updated

### LINKED ISSUE #2315: Open bot from history hit errors
- In composer home page, open bot from history list , hit following errors.
image

- If opened the new created bot, it worked well. 
 so this only happened for bot created without the botID updates.

boydc2014:
## Description

When the data.json used  (the server data store) doesn't align with the new data.json template, we consider composer has been updated, and simply re-init data.json. later we may consider do a gracefully update. 

Given our multi-user work, which introduced a new key in data.json, this will help user clean recent history and do a fresh start. 

I also clean up ABS-related logic of initializing data store.

## Task Item

fixes #2315


a-b-r-o-w-n:
@boydc2014 what if instead we just wrote a version into the data?

boydc2014:
> @boydc2014 what if instead we just wrote a version into the data?

if we don't reuse same key for different purpose, comparing keys won't be good enough. Adding a explicit version means any time we want to update the schema of data.json, developer needs to update this version manually.  I don't know, i feel like it would be a little bit error-prone, "keys" here implicitly indicate a version, somehow.

boydc2014:
> > @boydc2014 what if instead we just wrote a version into the data?
> 
> if we don't reuse same key for different purpose, comparing keys might be good enough. Adding a explicit version means any time we want to update the schema of data.json, developer needs to update this version manually in a correct format. I don't know, i feel like it would be a little bit error-prone, "keys" here implicitly indicate a version, somehow. what do you think

Offline synced with @a-b-r-o-w-n , i'm convinced let's start add a version here and leverage the generic data migration mechanism to achieve this.

Issue #2092, opened 2020-02-26
[Accessibility][Usable - Bot Responses-Edit]: No Keys Information is provided for keyboard users to navigate out from the edit field.


ShellySoni:
**User Experience:**
For better usability purpose and for keyboard users it will be good if proper information about keys to be used is provided either in terms of tooltip or some pop-up as provided using screen reader.

**Test Environment:**
Environment: Chromium Edge (Anaheim) + Narrator
Build Version: 2004 (OS Build 19569.1000)​​
Microsoft Edge Version: 82.0.436.0 (Official build) canary (64-bit)​
Test URL: http://localhost:3000/home

**Repro Steps:**
1. Open URL: http://localhost:3000/home
2. "Bot Framework Composer" homepage will gets opened.
3. Select any of the previously created bot from "Recent Bots" or create one from the available Bots under 'Examples' and activate it.
4. Note: Better to create new "Dialog Actions" bot from example.
5. Select 'Bot Responses' from left navigation.
6. Navigate throughout the page when "Edit toggle button is enabled".
7. Observe the issue when user reach within the edit field.

**Actual Result:**
No Information is provided i.e. what keys should be used for keyboard users that will help to navigate out from the edit field. Normally 'esc' key is used but here esc key is also not working.

**Note:**
Same issue is repro within 'User Input - Edit Mode'.

**Observation:**
By using screen reader, user is able to know the keys i.e. Press 'Alt+F1' to open the dialog, then use 'Ctrl+M' key to navigate out.

**Expected Result:**
Proper Information should be provided i.e. what keys should be used for keyboard users that will help to navigate out from the edit field. Information either in terms of Tooltip or some pop-up should be provided.

**MAG Guidelines:**
https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/Doc.aspx?sourcedoc=%7bF98F4235-B6BD-47D3-8DA4-F28B9B6FBFA8%7d&file=Guideline%20101%20%E2%80%93%20Recommended%20Assistive%20Technology%20Tools%20for%20MAS%20Testing%20of%20Microsoft%20Digital%20Properties.docx&action=default&mobileredirect=true

**Attachment:**
[Usable_Bot Responses_Edit Mode.zip](https://github.com/microsoft/BotFramework-Composer/files/4253682/Usable_Bot.Responses_Edit.Mode.zip)

corinagum:
Can we also get a11y tags added to this issue? Including severity :)

[...]

PR #2329, opened 2020-03-22
a11y: Add esc key listener to jump out of edit mode in lu/lg page

liweitian:
## Description
Add esc key listener to jump out of edit mode in lu/lg page
## Task Item
Refs #2092
## Screenshots
aaaaa


Composer/packages/client/src/pages/language-generation/index.tsx
```diff
@@ -64,6 +64,23 @@ const LGPage: React.FC<LGPageProps> = props => {
...
+    const handleUserKeyPress = event => {
+      const { keyCode } = event;
+      //when press esc and in edit mode
+      if (keyCode === 27 && edit) {
+        const url = `/bot/${projectId}/language-generation/${fileId}`;
+        navigateTo(url);
+      }
+    };
+
+    document.addEventListener('keydown', handleUserKeyPress);
```
zhixzhan: I guess Andy means move EventListenerTarget from document to CodeEditor itself ,then pop up an event like ' editorWantClose ' somehow let page to handle it.

liweitian: @a-b-r-o-w-n  @zhixzhan moved eventHandler down to baseEditor

a-b-r-o-w-n: This is not what I mean. I mean that instead of the document having the event listener, some element closer to the editor should have it. For example:

```ts
<div onKeydown={handleKeydown}>
  <CodeEditor {...props} />
</div>
```

I do not think this behavior belongs in the code-editor package.

a-b-r-o-w-n: Another option here is to add a custom key command upon the editor mounting like in https://github.com/microsoft/BotFramework-Composer/blob/master/Composer/packages/lib/code-editor/src/LuEditor.tsx#L122

liweitian: applied option 2 now

a-b-r-o-w-n: > I do not think this behavior belongs in the code-editor package.

See my comment below.

a-b-r-o-w-n:
To be very clear, the only 2 files that need to change are `language-generation/code-editor.tsx` and `language-understanding/code-editor.tsx`

liweitian:
> To be very clear, the only 2 files that need to change are `language-generation/code-editor.tsx` and `language-understanding/code-editor.tsx`

it's done....

Composer/packages/client/src/pages/language-generation/code-editor.tsx
```diff
@@ -2,14 +2,18 @@
 // Licensed under the MIT License.
 
 /* eslint-disable react/display-name */
-import React, { useState, useEffect, useMemo, useContext, useCallback } from 'react';
+import path from 'path';
```
a-b-r-o-w-n, line 5: We should not be using the path module in the client. I know that there is precedent for doing so, but we should not expand the usage.

Composer/packages/client/src/pages/language-generation/code-editor.tsx
```diff
@@ -57,8 +60,18 @@ const CodeEditor: React.FC<CodeEditorProps> = props => {
...
+  const onInit = (monaco: Monaco) => {
+    monacoRef.current = monaco;
+  };
+
   const editorDidMount: EditorDidMount = (_getValue, lgEditor) => {
     setLgEditor(lgEditor);
+    const m = monacoRef.current;
+    if (lgEditor && m && currentPath) {
+      lgEditor.addCommand(m.KeyCode.Escape, function() {
+        navigate(path.resolve(currentPath, '../'));
```
a-b-r-o-w-n, line 72: See my comment about about not using path.

a-b-r-o-w-n: Another way to approach this is not use navigate at all and instead create two actions, something like `startEditMode` and `endEditMode`. Then the business logic about what to do for those events exists in the actions layer.

liweitian:
@a-b-r-o-w-n Should we close this pr? And only updating the doc to let the user know Ctrl-M can exit the edit mode.

a-b-r-o-w-n:
We probably should get some guidance from a11y folks first.

a-b-r-o-w-n:
@corinagum do you have guidance here or should we reach out to the a11y team?

corinagum:
@liweitian @a-b-r-o-w-n Taking a look - will get back to you asap.

a-b-r-o-w-n:
Just found this:

Screen Recording 2020-04-08 at 10 20 AM

corinagum:
I'm closing this PR and circling discussion back to the  original convo. I agree linking to documentation (though, I'd like confirmation on how we do it) is a better alternative to changing the Monaco editor behavior.

PR #2333, opened 2020-03-23
chore: '$type' -> '$kind'

### LINKED ISSUE #2301: update app to read/write $kind instead of $type
In 4.8, the SDK has begun deprecating the `$type` in favor of `$kind`. Our app currently only knows about `$type`. We need to read & write `$kind` instead.

yeze322:
# ⚠️ WARNING: BREAKING CHANGE

This will break existing bots that have `$type` written. To fix, update all occurrences of `$type` to `$kind`. But note that if the bot is on an older version of the runtime (early 4.7 release), the `$kind` property does not work.

## Description

Converts all usage of `$type` to `$kind`.

## Task Item

closes #2301


yeze322:
UT failed because of '@bfc/shared' not updated

a-b-r-o-w-n:
@yeze322 thanks for opening this. As soon as I get the form work merged, I will tackle this.

Issue #2247, opened 2020-03-12
Shell - debouncedUpdateLg cannot handle paralleled lg update requests


yeze322:
## Describe the bug
In a parralleled scenario, `deboundcedUpdateLg` api will lose changes.

Scenario:
Moves lg template A from 'D1.lg' to 'D2.lg', fires 2 paralleled api calls:
1. create template A' in 'D2.lg'
2. delete template A from 'D1.lg'

In this case, the first request will be cancelled.

## Browser

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

yeze322:
Possible solutions:
1. use well-managed immutable states + observer
2. event stream

yeze322:
Another hack I found in composer codebase (1y ago hacked by Andy)
![image](https://user-images.githubusercontent.com/8528761/76845473-e6290180-6879-11ea-8adf-79ad6b704a95.png)
It should also be addressed here

cwhitten:
addressed via https://github.com/microsoft/BotFramework-Composer/pull/2348

Issue #2103, opened 2020-02-26
move all indexer operation from server to client


lei9444:
1. server will only support file's CRUD
2. The client store will get the project data from indexer.
3. All dialog, lu, lg operation logic will move to client.
4. new api may be added to support lu publish.

cwhitten:
addressed via https://github.com/microsoft/BotFramework-Composer/pull/2348

PR #2348, opened 2020-03-24
refactor: Move indexer to client and add file persistence layer

lei9444:
## Description
1. move indexer from server to client
2. add file persistence layer to do data synchronizing between client and server
image

       File persistence will operation the change for every file
       So I remove the http request and denounce in file operation action. 

## Task Item
refs #2103
refs #2247


lei9444:
> This breaks the scenario of initiating the create dialog flow from the from in a BeginDialog action.

Hi, @a-b-r-o-w-n I have reverted some code about create dialog, and it works now.

Composer/packages/client/src/store/action/lg.ts
```diff
@@ -1,52 +1,45 @@
...
-  const file = state.lgFiles.find(l => l.id === id);
-  if (file) {
-    store.dispatch({
-      type: ActionTypes.UPDATE_LG_SUCCESS,
-      payload: { id, projectId, content, lastModified: file.lastModified },
-    });
-    debouncedUpdateLg(store, id, projectId, content, file.lastModified);
+  // slot with common.lg import
+  let lgInitialContent = '';
+  const lgCommonFile = lgFiles.find(({ id }) => id === 'common');
```
zhixzhan: lgFile's id is included locale, e.g
lgFiles:
[
'common.en-us',
'common.zh-ch',
...
]

Composer/packages/client/src/store/action/dialog.ts
```diff
@@ -1,159 +1,136 @@
...
-  const dialog = state.dialogs.find(dialog => dialog.id === id);
-  let dialogs = clonedeep(state.dialogs);
-  if (!isStackEmpty) {
-    dialogs = reject(dialogs, ['id', id]);
+export const createDialogBase: ActionCreator = async (store, { id, content }) => {
+  const fixedContent = autofixReferInDialog(id, content);
+  const onCreateDialogComplete = store.getState().onCreateDialogComplete;
+  if (typeof onCreateDialogComplete === 'function') {
+    setTimeout(() => onCreateDialogComplete(id));
   }
```
yeze322: dispatch an `onCreateDialogComplete` action at somewhere handles the payload of CREATE_DIALOG

Composer/packages/client/src/store/action/dialog.ts
```diff
@@ -1,159 +1,136 @@
...
+export const removeAllRelatedFiles: ActionCreator = async (store, id) => {
+  removeDialogBase(store, id);
+  //remove dialog should remove all locales lu and lg files
+  const { luFiles, lgFiles } = store.getState();
+  const luIds = luFiles.filter(file => getBaseName(file.id) === id);
+  const lgIds = lgFiles.filter(file => getBaseName(file.id) === id);
+  luIds.forEach(({ id }) => removeLuFile(store, id));
+  lgIds.forEach(({ id }) => removeLgFile(store, id));
+
+  navTo(store, 'Main');
```
yeze322: it's not an `action`, belongs to the reducer logic

Composer/packages/client/src/store/action/dialog.ts
```diff
@@ -1,159 +1,136 @@
...
-      payload: {
-        response,
-      },
-    });
-  } catch (err) {
-    setError(store, {
-      message: err.response && err.response.data.message ? err.response.data.message : err,
-      summary: 'CREATE DIALOG ERROR',
-    });
+export const createAllRelatedFiles: ActionCreator = async (
```
yeze322: ditto, shouldn't be part of the action file

yeze322: other 'actions' have the same problem, we need to differentiate actions (data payload) from reducer logic (bussiness logic with HTTP call or complicated transformation), otherwise the core store state will be unpredictable

(we are not far from that, just move them to another file will help)

Composer/packages/client/src/store/action/lu.ts
```diff
@@ -1,49 +1,37 @@
...
-    store.dispatch({ type: ActionTypes.UPDATE_LU_SUCCESS, payload: { id, content } });
-    debouncedUpdateLu(store, id, projectId, content, file.lastModified);
-  }
+export const updateLuFile: ActionCreator = async (store, { id, content }) => {
+  luFileStatusStorage.updateFile(store.getState().botName, id);
+  store.dispatch({
+    type: ActionTypes.UPDATE_LU,
+    payload: { id, content },
+  });
+};
```
yeze322: Correct pattern:
```ts
// Action
const action = updateLuFile('a.lu'); // got { type: 'UPDATE_LU', payload: { id: 'a.lu' } }

// Reducer
const reducer = (store, action) => {
   if (action.type === 'UPDATE_LU') {
      const newLuFile = update(action.id);
      const luFiles = replaceOldLuFileWithNewLuFile(store.luFiles, 'a.lu', newLuFile);
      return { ...store, luFiles };
   }
}
```
Reason: Decouple your core domain events from bussiness logic.
`Action` is your core domain events, it connects UI and store; meanwhile `Reducer` is the place you handle the bussiness logic, it varies.

yeze322: Though we prefer `useReducer` over `redux`, the redux pattern should still be followed.

boydc2014:
> image
> data flow IMO

yes, agree, we are not trying to a "side way", we are trying to make the persistence layer transparent to client actions.

boydc2014:
Can you solve the conflicts?

Composer/packages/client/src/store/action/dialog.ts
boydc2014: I don't know exactly what we do in this onCreateDialogComplete, it seems not aligns with the pattern we are using here, but if possible, it it's just related changes like lg,lu files creation, we can just pre-compute the changes and commit to store once.

lei9444:
> Is it possible to add some test coverage since the coverage dropped ~1%?

Yes, I will add more tests to increase the test coverage

Composer/packages/client/src/store/action/dialog.ts
lei9444: This is used to create a dialog. we have three actions: createBegin, create and createCancel.  onCreateDialogComplete is used to wait the createBegin complete.
createBegin is used to open 
image

lei9444:
@a-b-r-o-w-n @boydc2014 @cwhitten If this refactor is Ok, I think next PR I can refactor the undo/redo. It's too expensive to wrapper the actions now. I want to change it to @cwhitten mentioned before https://github.com/microsoft/BotFramework-Composer/pull/729#issuecomment-524638737

Composer/packages/client/src/store/action/dialog.ts
a-b-r-o-w-n: This is required because the extensions do not have direct access to the reducer store, as well as different behavior upon completion of a dialog creation. But the creation flow is async and has disjointed steps. So the initiator of the create dialog flow needs to be able to determine what to do when it is complete.

If there is a better way to model this in the store, I'd be happy to discuss.

Composer/packages/client/src/store/reducer/index.ts
```diff
@@ -97,44 +107,42 @@ const removeRecentProject: ReducerFunc = (state, { path }) => {
...
-  state.dialogs = response.data.dialogs;
-  state.luFiles = updateLuFilesStatus(state.botName, response.data.luFiles);
-  state.lgFiles = response.data.lgFiles;
-  state.showCreateDialogModal = false;
-  state.actionsSeed = [];
-  delete state.onCreateDialogComplete;
+const removeLgFile: ReducerFunc = (state, { id }) => {
+  state.lgFiles = state.lgFiles.reduce((result: LgFile[], file) => {
+    if (getBaseName(file.id) === id || file.id === id) {
+      filePersistence.notify(FileChangeType.DELETE, file.id, FileExtensions.Lg, '');
```
boydc2014, line 140: We should not require each reducer to manually add a notify. The notification should be done by your framework.

Composer/packages/client/src/store/reducer/index.ts
```diff
@@ -97,44 +107,42 @@ const removeRecentProject: ReducerFunc = (state, { path }) => {
...
 
-const createDialogCancel: ReducerFunc = state => {
-  state.showCreateDialogModal = false;
-  delete state.onCreateDialogComplete;
+  const { parse } = lgIndexer;
+  const lgImportresolver = importResolverGenerator(state.lgFiles, '.lg');
+  const { templates, diagnostics } = parse(content, id, lgImportresolver);
+  const lgFile = { id, templates, diagnostics, content };
+  state.lgFiles.push(lgFile);
+  filePersistence.notify(FileChangeType.CREATE, id, FileExtensions.Lg, content);
```
boydc2014, line 133: The whole effort of decouple file persistence from actions is to make it transparent to upper layer, so requiring a manually notify in reducer is just moving the problem from actions into reducers, didn't actually make it transparent.

Composer/packages/client/src/store/action/dialog.ts
boydc2014: get it, @a-b-r-o-w-n i talked to Leilei and dig it a little bit, it was a 2 step actions and extension needs to get the result inorder to further update the action node, i don't have a good proposal here, maybe we can do better by having a more generic approach for extensions to get notification instead of directly inject functions into store, but that's would reduce this essentially complex, and will likely be a overkill. 

let's leave it for now, as far as in this PR, the goal is decouple the concern of persistence from the upper layer (the editors and other components that emit actions), which they can assume the store is always available and update correctly and fast.

boydc2014:
Offline talked with Leilei, to clarify that ideally we don't want actions and reducers to care about or even know about the persistence layer exists. We want to make it totally transparent, so that upper layer will perform on a stable, fast local store. 

Since we managed the middle-ware chaining, so the perfect place to do this, is have a middle ware injected *after the reducer computed the store update* and notify persistence layer.

lei9444:
@a-b-r-o-w-n @cwhitten @boydc2014 I have changed the notification mechanism for persistence layer. I wrapped the reducer, so the layer can get the action, stale state and the new state. I can get the processed file content from state and use action type and action.payload.id to find which file is updated. The reducer itself doesn't known if the persistence layer exists.

PR #2352, opened 2020-03-24
a11y: adjust format of error messages in form editor

### LINKED ISSUE #2114: [Accessibility][Usable - Property Pane_Ask a question (Bot Asks)]: Focus goes on 'Here' present into the error message without expanding the error message.
**User Experience** 
Low vision users who rely on keyboard will face difficulty and easily confused if the focus moves to the link in error message without expanding whole error message.

**Test Environment**
OS build: 2004 (19564.1005)
Screen Reader: Narrator

**Repro Steps:** 
1. Open the above mentioned URL in EDGE and turn on narrator.
2. Navigate to already created dialog tree item and activate it. Such as- Greetings, Actions, EndDialog etc.
3. Click on "+" and select 'Ask a question'. Then select any of the available options except Oauth login.
4. Navigate to property pane appeared at right side of the screen.
5. Now, navigate to 'Prompt' edit field and delete all the content & press space.
6. Now, navigate forward with tab key and Observe the issue.

**Actual:** 
On entering blank data in the prompt edit field, an error message appeared with 'Here' as link which gets the focus with Tab key but the whole error message is not expanding. Due to which it looks like the focus goes to the hidden element.

**Expected:**
As soon as the interactive element gets the keyboard focus the error message should be expended.

**Note:**
Same issue is observed for all the Property Pane which have prompt or any syntax edit fields appeared at right side of the screen.

**MAG Reference:** [MAG](https://microsoft.sharepoint.com/teams/msenable/_layouts/15/Doc.aspx?sourcedoc=%7BD0F24E02-1D9C-4DA2-8F5B-2A3C1176F3BF%7D&file=Guideline%2017%20A%20-%20Contrast%20(Minimum)%20When%20Conveying%20Information.docx&action=edit&mobileredirect=true)

**Attachment:**
[Usable_focus goes into the error message without expanding.zip](https://github.com/microsoft/BotFramework-Composer/files/4261750/Usable_focus.goes.into.the.error.message.without.expanding.zip)
focus goes into the error without expanding them

beyackle:
## Description

This changes the props of the MessageBar components we use in order to work around the problem navigation has with moving to a link inside a message when the message is collapsed.

## Task Item

Closes #2114

## Screenshots

image

### DIFF
--- a/BotProject/Templates/CSharp/ComposerBot.cs
+++ b/BotProject/Templates/CSharp/ComposerBot.cs
@@ -22,7 +22,7 @@ namespace Microsoft.Bot.Builder.ComposerBot.Json
         private readonly IStatePropertyAccessor<DialogState> dialogState;
         private readonly string rootDialogFile;
 
-        public ComposerBot(ConversationState conversationState, UserState userState, ResourceExplorer resourceExplorer, BotFrameworkClient skillClient, SkillConversationIdFactoryBase conversationIdFactory)
+        public ComposerBot(ConversationState conversationState, UserState userState, ResourceExplorer resourceExplorer, BotFrameworkClient skillClient, SkillConversationIdFactoryBase conversationIdFactory, string rootDialog)
         {
             HostContext.Current.Set(skillClient);
             HostContext.Current.Set(conversationIdFactory);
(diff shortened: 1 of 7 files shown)


beyackle:
There we go. Somewhere they changed the format of the error message object, and now we can read the new ones too. 
image

beyackle:
We didn't have a way to display warning-level messages before, but I've fixed that too.

image

beyackle:
Note: this will need to be changed once PR#2118 goes through. Don't merge this yet.

beyackle:
Going to close this for a clean redo in the new adaptive-editor directory.

Issue #2357, opened 2020-03-25
multi-locale/file serialization regressions


cwhitten:
1. serialization has a couple minor issues:
   - when a template is selected for creation, the main/entry dialog is the name of the template, when it should be the name the user gave for the bot.
   - sub-dialogs in the /dialogs folder are not lower-cased.
2. Configuring LUIS as recognizer appears to break. (500's from emulator)

boydc2014:
in 3#, there is no "generated" folder, looks like lubuild isn't up to date with multi-locale change, co-assign Leilei for this.

zhixzhan:
Hi @cwhitten for 
> sub-dialogs in the /dialogs folder are not lower-cased

did you mean previously  existed  bot? created by composer or not?

ps: current migration script use `Main.dialog` as a flag of old structure bot and apply serialization

cwhitten:
addressed via https://github.com/microsoft/BotFramework-Composer/pull/2360

PR #2358, opened 2020-03-25
fix: can't find referred lu files for publishing

lei9444:
## Description
The referred lu id should add locale before search

## Task Item
refs #2357
-> 2. Configuring LUIS as recognizer appears to break. (500's from emulator)

## Screenshots
for the multi locale publish result will be like:
imgos
imgos (2)
imgo

### DIFF
--- a/Composer/packages/server/src/models/bot/botProject.ts
+++ b/Composer/packages/server/src/models/bot/botProject.ts
@@ -933,10 +933,7 @@ export class BotProject {
 
   private isReferred = (LUFile: LuFile) => {
     const dialogs = this.dialogs;
-    if (dialogs.findIndex(dialog => dialog.luFile === LUFile.id) !== -1) {
-      return true;
-    }
-    return false;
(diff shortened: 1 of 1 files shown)


Composer/packages/server/src/models/bot/botProject.ts
```diff
@@ -933,10 +933,7 @@ export class BotProject {
 
   private isReferred = (LUFile: LuFile) => {
     const dialogs = this.dialogs;
-    if (dialogs.findIndex(dialog => dialog.luFile === LUFile.id) !== -1) {
-      return true;
-    }
-    return false;
+    return !!~dialogs.findIndex(dialog => dialog.luFile === Path.basename(LUFile.id, `.${this.locale}`));
```
zhixzhan: So [alllocale].*.lu would be publish at this moment?

lei9444: update: publish all locale files now

PR #2362, opened 2020-03-25
fix: lg resolver with locale

### LINKED ISSUE #2361: lg can't resolve some imported lg
## Describe the bug
if an import case like:

```
a.en-us.lg
[import](b.lg) 
 
b.en-us.lg
[import](c.lg) 
```
[EchoBot-3.zip](https://github.com/microsoft/BotFramework-Composer/files/4379998/EchoBot-3.zip)

it would failed to resolve c.lg

## Browser

- [x] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [x] macOS
- [ ] Windows
- [ ] Ubuntu

## Screenshots

image

zhixzhan:
## Description

Enable resolve file with or without locale.

```
   *  for example:
   *  in todosample.en-us.lg:
   *   [import](common.lg)
   *
   *  would resolve to common.en-us.lg || common.lg
   *
```

## Task Item

close #2361

### DIFF
--- a/Composer/packages/client/src/store/index.tsx
+++ b/Composer/packages/client/src/store/index.tsx
@@ -5,9 +5,9 @@ import React, { useReducer, useRef } from 'react';
 import once from 'lodash/once';
 import { ImportResolverDelegate, LGParser } from 'botbuilder-lg';
 import { LgFile, LuFile } from '@bfc/indexers';
+import { importResolverGenerator } from '@bfc/shared';
 
 import { prepareAxios } from '../utils/auth';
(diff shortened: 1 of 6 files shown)


zhixzhan:
> This looks like roughly the same code in each of these places. Can we extract out to a shared utility? It probably belongs in the the shared package.

Great idea, I implement a  function  called `importResolverGenerator` in shared lib, by feeding it different resources, now  client/server, lg/lu all could reuse it.

PR #2367, opened 2020-03-25
a11y: have screen-reader extract card titles from headers

### LINKED ISSUE #2058: [Accessibility][Info & Relation -Flow Chart] : Narrator didn't announce anything when the narrator focus goes on the whole node.
**User Experience** 
The screen reader dependent user will not be able to know the presence of the node which will confuse the AT user.

**Test Environment**
OS build: 2004 (19564.1005)
Screen Reader: Narrator

**Repro Steps:** 
3. Now, navigate to the flow chart.
4. Observe the narrator narration on the node name like "Send a response".

**Actual:** 
The narrator didn't announce anything when the focus lands on the node name.

**Expected:**
The narrator should announce the name of the node so that user will know the current position of the focus.

**Note:**
When the focus reaches on the node, the property pane gets appeared on the right side of window. The narrator should inform about the pane appearance.
Same issue found with the closing node. The narrator remains silent on the closing node.

**Observation:**
Same issue observed with NVDA.

**MAS Impacted:** [MAS1.3.1](https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={54f28d1f-a2d1-4dcd-84e1-5c9b87e8aba4})

**Attachment:**
[MAS1.3.1_Narrator didn't inform about the node name.zip](https://github.com/microsoft/BotFramework-Composer/files/4241231/MAS1.3.1_Narrator.didn.t.inform.about.the.node.name.zip)

beyackle:
## Description

The label that gets read by the screen reader turns out to come from the ElementWrapper component, so this adds some mechanisms to that code so it can reach in and extract the title of what it's rendering. I couldn't use a callback because ElementWrapper just has generic "children" with no guarantee of that having any shape in particular, and I didn't want to add a whole new "ariaLabel" field to our schema when it would just duplicate the same information already in "title".

There are a lot more screen-reader-related issues going on in this codebase, but this at least fixes one of them.

## Task Item

Closes #2058

### DIFF
--- a/Composer/packages/extensions/visual-designer/src/components/renderers/ElementWrapper.tsx
+++ b/Composer/packages/extensions/visual-designer/src/components/renderers/ElementWrapper.tsx
@@ -3,8 +3,9 @@
 
 /** @jsx jsx */
 import { jsx, css } from '@emotion/core';
-import { FC, useContext } from 'react';
+import { FC, useContext, ReactNode, ReactElement } from 'react';
 import classnames from 'classnames';
+import { generateSDKTitle } from '@bfc/shared';
 
 import { AttrNames } from '../../constants/ElementAttributes';
 import { NodeRendererContext } from '../../store/NodeRendererContext';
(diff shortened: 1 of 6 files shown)


Composer/packages/extensions/visual-designer/src/components/renderers/ElementWrapper.tsx
```diff
@@ -26,10 +27,34 @@ const nodeBorderDoubleSelectedStyle = css`
...
   onEvent: (eventName: NodeEventTypes, eventData: any) => any;
 }
 
-export const ElementWrapper: FC<ElementWrapperProps> = ({ id, tab, onEvent, children }): JSX.Element => {
+function checkHasProps(node: ReactNode): node is ReactElement {
+  return (node as ReactElement).props != null;
+}
+
+function extractNodeTitle(node: ReactNode, titleInHeader: boolean): string {
+  if (node == null || typeof node !== 'object') {
```
corinagum: `===` for `node==null`? Or does it not matter for `null`?

corinagum:
LGTM :)

Composer/packages/extensions/visual-designer/src/components/renderers/ElementWrapper.tsx
beyackle: Checks for `null` and `undefined` still just use `==` by convention.

PR #2377, opened 2020-03-26
fix: Supporting suggest cards in inline LG editing

### LINKED ISSUE #2330: LG LSP: Add auto-complete/auto-suggest for Activity and Cards
Add auto-complete/auto-suggest into the LG LSP to provide support when authoring Activities and Cards. See RespondingWithCards sample for list of Cards to suggest auto-complete.

cosmicshuai:
closes #2330 

Now card suggestion are enabled in inline LG.
Also, formating of auto complete is improved. The indent before ending '[' will be deleted automatically. 
card suggestion in line


Composer/packages/lib/code-editor/src/BaseEditor.tsx
```diff
@@ -23,6 +23,7 @@ const defaultOptions: monacoEditor.editor.IEditorConstructionOptions = {
   glyphMargin: false,
   folding: false,
   renderLineHighlight: 'none',
+  formatOnType: true,
```
a-b-r-o-w-n: Let's make sure this doesn't cause weirdness in the other places that the code-editor is used.

cosmicshuai: All editor instances inherited from baseEditor are enabled for document formating command as a client. But here only in LG editor will have the ability to do so since the LG server send 
      documentOnTypeFormattingProvider: {
                 firstTriggerCharacter: '\n',
              }
at initialization. 
Moreover, the code at server side will make sure the  documentOnTypeFormatting will only affect typing in structure LG.

I will test all other editor.

cosmicshuai: > Let's make sure this doesn't cause weirdness in the other places that the code-editor is used.

I tested the editors in composer. Only the LG editor will have the ability to format structure lg input.

cwhitten: thanks for investigating @cosmicshuai

PR #2405, opened 2020-03-27
fix: use custom schema when creating new dialogs

### LINKED ISSUE #2344: Issue while using QnA Maker integration
I am getting error "Object reference not set to an instance of an object.", when I use QnA maker with unknown intent trigger as per [documentation ](https://docs.microsoft.com/en-us/composer/how-to-add-qna-to-bot)

image

Kindly check and let me know what is the issue.

### LINKED ISSUE #2370: use bot project schema in dialog factory to set default values

a-b-r-o-w-n:
## Description

Removes any dependency on appschema.

## Task Item

closes #2370
closes #2344


Composer/packages/lib/shared/src/dialogFactory.ts
```diff
@@ -1,40 +1,50 @@
...
+  return {
+    $designer: {
+      name,
+      description,
+      id: generateUniqueId(6),
+    },
+  };
+}
+
+const initialDialogShape = () => ({
```
a-b-r-o-w-n: This needs to be a function in order to get unique ids every time this is called.

PR #2440, opened 2020-03-30
feat: publishing page

### LINKED ISSUE #2355: Implement New Publish page
# Implement New Publish page

Implement a new page inside the settings that will be used to configure
and view the status of one or more publishing targets.

Publishing "methods" are enabled in Composer by installing a plugin.  An example of a publish method is "publish to azure bot service" or "publish to PVA." Composer does not need to know what happens inside the plugin - it hands the bot project content to the plugin, and the plugin takes _some action_ and reports back a status.

A publishing "target" is the combination of a publish method and a configuration profile. The configuration profile is a set of configuration in the form of a JSON object. This configuration
contains all the information necessary to use the given method to publish the content to a specific target.  An example of htis would be "Publish this bot to my production ABS slot" or "Publish this to HR bot on PVA."

The act of publishing is defined as taking read only copy of the current bot (as seen in Composer), and passing it to the publishing plugin.  The publishing plugin will take some action, and return a status. The status can be either succeeded, failed, or in progress.

New targets can be created from the publish page. To add a new target, the user must specify the method (a list is available via `/api/publish/types`), a user-specified name ("My Production slot" or "Remote testing instance") and an optional configuration object. This object must be valid JSON.

A rough prototype of the publish page is already in master in `/client/src/pages/settings/publisher/`

Depends on the APIs built in #1945

## Publish page:

- [x] Create a new page inside the Settings menu for the new publish page
- [x] Implement the ability to add a new publishing target
  - [x] User clicks "+ add new"
  - [x] User selects a publishing method from a list of configured publishing plugins
  - [x] User provides a free-form text name for the target
  - [x] user provides valid JSON configuration
  - [x] Validate JSON and present a validation error if it does not successfully parse
  - [x] Prevent duplicate names from being used to describe multiple targets and provide a validation error if the name is already in use
  - [x] Save the target configuration to the bot's settings under `publishTargets` key. `publishTargets` should be an array of objects in the form `{name, type, configuration, lastPublish}`
  - [x] Update the UI to represent the new target
- [x] For each configured target in the bot's `publishTargets` key, render an item in the publish page navigation. This item should include the name, description, method name, and 
- [x] When a user selects a publish target, show the detail view in the main pane
  - [x] update the URL to represent which target is selected. 
  - [x] make it possible to "permalink" to a specific target. "/bots/(mybotid)/settings/publish/(targetName)"
  - [x] if `lastPublish` is set on target, make a call to the `/api/publish/:projectId/status/:target` url and load latest status. Display status message in main pane.
  - [x] If statuscode is 200, indicate success (green dot)
  - [x] if statuscode is 500, indicate error (red dot)
  - [x] if statuscode is 202, indicate publish in progress. Poll every X seconds until a 200 or 500 is reached - increase X by 10 seconds each time. (spinner)
- [x] Implement context menu
  - [x] Delete - allow target to be removed. should have confirmation.
  - [x] Edit - show edit view, allow modification of name, method and configuration
- [x] Use icon from design instead of http status code
- [x] Format the date in the time column to M/D/Y h:m
- [x] Implement "all" view that shows recent history of all configured profiles
- [x] Finish implementation of "show log" 
- [x] RELATED: Need to add log field to payload 

## Default Publishing Profile:

Currently, there is a system-provided "default" profile used to update the local test runtime. There are no user configurable optons for this at current time. The behavior of this "default" profile may change over time - IE it may become user configurable. This profile should be excluded from the UI.

## Publish Button:

- [x] Add a publish button to top menu bar of publish page (see design)
- [x] The publish button should be DISABLED unless there are 1 or more publish targets configured NOT including the "default" local test target.
- [x] If publish button is enabled, clicking the button leads to a modal dialog where the publish action is taken.
  - [x] User provides an optional title and description for the revision being published. title should default to a friendly timestamp, such as "2020-03-24 4:06pm" These values are provided to the publishing plugin for internal use such as providing a publishing history.
  - [x] Primary action button is publish. When clicked, perform publish and close modal. The status of the publish should be displayed in the main pane.

## Plugin Loader:

- [x] Extend publish plugin definition to include a description. This should default to the description available in the package.json file, but should also be setable by the plugin.  Example:  "Upload this bot to a remote Azure web host"
- [x] Include description in the results of `/api/publish/types`

benbrown:
## Description
Implementation of the new publish page UI.
Somethings different from the design:
1.  **Hide the test in emulator button now.**  Because publishing does not necessarily result in an emulator-friendly link being present and it is much more complex than the local scenario. Including this button on this page will create confusion around the other "Start bot" button.
2. Add a rollback button.
3. Add side menu about edit and delete in each profile. 

## Task Item
closes #2355 

## Screenshots

za24e-oz60e


Composer/packages/client/src/pages/publish/index.tsx
```diff
@@ -0,0 +1,242 @@
...
+    props.onDismiss();
+    await props.onSubmit(comment);
+  };
+  return props.target ? (
+    <Dialog hidden={props.hidden} onDismiss={props.onDismiss} dialogContentProps={publishDialogProps}>
+      <Fragment>
+        <div css={publishDialogText}>{props.target.name}</div>
+        <form onSubmit={submit}>
+          <TextField
+            placeholder="Provide a bried description of this publish. It will appear on the publish history list"
```
benbrown, line 228: There is a typo here - "brief"

Composer/packages/client/src/pages/publish/index.tsx
```diff
@@ -0,0 +1,242 @@
...
+      </div>
+      <div css={ContentStyle} data-testid="Publish">
+        <div css={projectContainer}>
+          {publishTarget.length > 0 ? (
+            <Fragment>
+              <div css={targetListTiTle}>All profiles</div>
+              <TargetList list={settings.publishTargets} onSelect={setSelectedTarget} selectedTarget={selectedTarget} />
+            </Fragment>
+          ) : (
+            <div>This bot does not have a profile to publish yet</div>
```
benbrown, line 171: This div is currently unstyled.  See design.

benbrown:
I added a new `mockRemotePublish` plugin, which provides publish, status and history functionality.  This will work with multiple bots and multiple profiles -- IE, you can set up a dev/test/staging profile and publish to them separately.

It is my hope that we can use this mock endpoint to complete the UI build-out before proceeding to the *real* publishing targets.

With that in mind, I think the next steps would be (as listed in #2355)

* Implement the error handling in the profile creation flow
* Update the UI after a user has done a publish with resulting status/messages
* Poll for status if latest publish resulted in a 202 (or if history indicates a 202, or if no history is available but lastPublish is set)

benbrown:
Updates from Ben today:

* tightened up the definition of the incoming and outgoing payloads for the APIs and plugins.
* updated the localPublish and mockRemotePublish to use new payloads
* Updated client so it can maintain multiple parallel histories (one for each profile)
* Updated client so that publish command will update the history with results
* Updated client so that getStatus will update history as well 
* added logic to call getStatus if publish or history returns a 202 status

Composer/packages/client/src/pages/publish/index.tsx
VanyLaw: change it

benbrown:
After this is code reviewed, we need to do one last task:

- [ ] Disable the mockRemotePublish plugin

Composer/packages/client/src/pages/setting/publisher/index.js
```diff
@@ -187,9 +189,11 @@ const CreatePublishTarget = props => {
 
   const submit = () => {
     try {
+      // validate that JSON parses correctly
       JSON.parse(config);
       return props.onSave(name, targetType, config);
     } catch (err) {
+      console.error(err);
       alert('Error parsing configuration');
```
cwhitten, line 197: this alert isn't in the diff, however do we really plan to use the vanilla `alert` control?

Composer/packages/client/src/store/reducer/index.ts
```diff
@@ -417,14 +417,25 @@ const setUserSessionExpired: ReducerFunc = (state, { expired } = {}) => {
   return state;
 };
 
-const setPublishTypes: ReducerFunc = (state, { response }) => {
-  state.publishTypes = response;
+const setPublishTypes: ReducerFunc = (state, { typelist }) => {
+  // const types: string[] = [];
```
cwhitten, line 421: remove?

Composer/packages/client/src/store/reducer/index.ts
```diff
@@ -417,14 +417,25 @@ const setUserSessionExpired: ReducerFunc = (state, { expired } = {}) => {
...
+  // typelist.map(item => types.push(item.name));
+  state.publishTypes = typelist;
   return state;
 };
 
 const publishSuccess: ReducerFunc = (state, payload) => {
-  state.botEndpoints[state.projectId] = `${payload.results?.result?.endpoint || 'http://localhost:3979'}/api/messages`;
-  state.botStatus = BotStatus.connected;
+  if (payload.target.name === 'default' && payload.endpointURL) {
+    state.botEndpoints[state.projectId] = `${payload.endpointURL || 'http://localhost:3979'}/api/messages`;
```
cwhitten, line 429: I'm not sure we can count on 'http://localhost:3979' any longer now that the runtime will find and use an available port.

Composer/packages/client/src/store/reducer/index.ts
```diff
@@ -435,12 +446,39 @@ const publishFailure: ReducerFunc = (state, { error }) => {
...
     state.botStatus = BotStatus.connected;
+    state.botEndpoints[state.projectId] = `${payload.endpointURL || 'http://localhost:3979'}/api/messages`;
+  }
+
+  // if no history exists, create one with the latest status
+  // otherwise, replace the latest publish history with this one
+  if (!state.publishHistory[payload.target.name] && payload.status !== 404) {
+    state.publishHistory[payload.target.name] = [payload];
+  } else if (payload.status !== 404) {
+    // TODO: this should only happen if they actually represent the same item...
```
cwhitten, line 461: I don't understand the todo comment.

Composer/packages/client/src/store/reducer/index.ts
```diff
@@ -435,12 +446,39 @@ const publishFailure: ReducerFunc = (state, { error }) => {
...
+    state.publishHistory[payload.target.name][0] = payload;
   }
   return state;
 };
 
+const getPublishStatusFailed: ReducerFunc = (state, payload) => {
```
cwhitten, line 467: This looks like the same code from L458. Can you combine?

Composer/packages/client/src/store/reducer/index.ts
```diff
@@ -519,6 +557,9 @@ export const reducer = createReducer({
   [ActionTypes.PUBLISH_SUCCESS]: publishSuccess,
   [ActionTypes.PUBLISH_FAILED]: publishFailure,
   [ActionTypes.GET_PUBLISH_STATUS]: getPublishStatus,
+  [ActionTypes.GET_PUBLISH_STATUS_FAILED]: getPublishStatusFailed,
+
```
cwhitten, line 561: nit: remove newline

Composer/packages/server/src/controllers/publisher.ts
```diff
@@ -8,45 +8,76 @@ import merge from 'lodash/merge';
...
+    const { metadata, sensitiveSettings } = req.body;
     const projectId = req.params.projectId;
     const currentProject = await BotProjectService.getProjectById(projectId, user);
 
-    // find publish config by name.
-    const configs = currentProject.settings?.publishTargets?.filter(t => t.name === target) || [defaultPublishConfig];
-    const config = configs.length ? configs[0] : undefined;
-    const method = config ? config.type : undefined;
+    // deal with publishTargets not exist in settings
+    const publishTargets = currentProject.settings?.publishTargets || [];
```
cwhitten, line 45: we could also provide `publishTargets` as a default in the settings object when a bot is created? Won't the local publish target come out of the box?

Composer/plugins/mockRemotePublish/src/index.ts
```diff
@@ -0,0 +1,130 @@
+// Copyright (c) Microsoft Corporation.
```
cwhitten, line 1: If wondering if instead of removing this we can keep it around as a test fixture?

Composer/packages/client/src/pages/publish/createPublishTarget.tsx
```diff
@@ -0,0 +1,89 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+/** @jsx jsx */
+import { jsx } from '@emotion/core';
+// import { useState, useContext } from 'react';
```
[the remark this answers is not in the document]
VanyLaw: done

Composer/packages/client/src/pages/publish/createPublishTarget.tsx
```diff
@@ -0,0 +1,89 @@
...
+    if (!targetType || !name || errorMessage) {
+      return true;
+    } else {
+      return false;
+    }
+  };
+
+  const submit = async () => {
+    await props.onSave(name, targetType, JSON.stringify(config, null, 2) || '{}');
+    props.onCancel();
```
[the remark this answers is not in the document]
VanyLaw: Nope, onSave just call setSettings action to update new settings, it have no return value. We inspect success or failure in action and reducer. And onCancel means close the popupDialog. I update the function name to make it more reasonable.

Composer/packages/client/src/pages/publish/index.tsx
```diff
@@ -0,0 +1,476 @@
...
+          setSelectedTarget(undefined);
+          onSelectTarget('all');
+        }
+      }
+    },
+    [settings.publishTargets, projectId, botName]
+  );
+
+  return (
+    <div>
```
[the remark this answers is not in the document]
VanyLaw: Change it to Fragment

Composer/packages/client/src/pages/setting/publisher/index.js
```diff
@@ -82,32 +82,34 @@ export const Publisher = () => {
...
     console.log('publish targets changed');
-    if (
-      publishTargets.filter(target => {
-        return target.statusCode === 202;
-      }).length
-    ) {
-      actions.getPublishStatus();
-      console.log('NEED TO LOAD PUBLISH STATUS');
-    }
+    // if (
```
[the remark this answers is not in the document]
VanyLaw: I think it's the previous version of publish page, I will remove it.

Composer/packages/client/src/store/reducer/index.ts
VanyLaw: Removed it

Composer/packages/server/src/controllers/publisher.ts
VanyLaw: Yes, I think we can add it, but in order to prevend the older bot, I think we still need to add this check.

benbrown: the local publish target is handled by Composer and does not have to exist in the settings.
That said, yes, this field _should_ exist in the default settings shape.

Composer/plugins/mockRemotePublish/src/index.ts
benbrown: Works for me but we will have to disable it so it doesn't appear in the dropdown.

Composer/packages/client/src/pages/setting/publisher/index.js
VanyLaw: It's outdated publish page, I remove this file.

Composer/packages/client/src/pages/setting/publisher/index.js
```diff
@@ -144,7 +146,7 @@ export const Publisher = () => {
     <div style={styles.page}>
       <h1 style={styles.header}>Publish your bot to a remote</h1>
       <PrimaryButton text="Add Destination" onClick={addDestination} styles={styles.button} />
-      {publishTypes && publishTypes.length && <div>publish to types: {publishTypes.join(',')}</div>}
+      {publishTypes && publishTypes.length && <div>publish to types: {publishTypes.map(t => t.name).join(',')}</div>}
```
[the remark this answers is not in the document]
VanyLaw: Outdated file, removed.

Composer/plugins/mockRemotePublish/src/index.ts
VanyLaw: disable in package.json

Composer/packages/server/src/controllers/publisher.ts
VanyLaw: I add the publishTargets as a field in default settings object.

Composer/packages/client/src/pages/publish/index.tsx
```diff
@@ -0,0 +1,476 @@
...
+  );
+};
+
+export default Publish;
+const LogDialog = props => {
+  const logDialogProps = {
+    title: 'Publish Log',
+  };
+  return (
+    <Dialog
```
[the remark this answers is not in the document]
VanyLaw: Dialog has a default hidden field to controll if shown or not. Why do we put the hidden field outside the Dialog Component?

Composer/packages/client/src/pages/setting/publisher/index.js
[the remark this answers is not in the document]
VanyLaw: outdated page, I remove it

Composer/packages/client/src/pages/setting/publisher/index.js
VanyLaw: outdated pages, I remove it

Composer/packages/client/src/store/reducer/index.ts
VanyLaw: I think use status as a part of history when 
1. history not exist, but the profile has lastPublishTime field. It means this bot has published in this profile before, but the bot still publishing and not response. So it call getStatus and use the status as history.
2. history exist, when bot publish to profile, then get 202 response and save in history, which means bot still publishing.  After 10 seconds, Frontend will call getStatus api to get the publish response(success or fail) to update the history. So, maybe the todo comment is this situation.

VanyLaw: Ben wrote a doc about those api before. https://microsoft-my.sharepoint-df.com/:w:/r/personal/benbro_microsoft_com1/_layouts/15/guestaccess.aspx?e=UadX7x&share=EQUB-_H36KxOu68O_-VujGoBAze7u-3TiyNx38sBogrnww

Composer/packages/client/src/store/reducer/index.ts
VanyLaw: Combined

Composer/packages/client/src/store/reducer/index.ts
VanyLaw: done

Composer/packages/client/src/store/reducer/index.ts
benbrown: the point of the TODO comment is, there should be some validation that the returned status item actually represents the  0th item in the history array, which is currently being replaced.

there is a scenario where, for example, someone else started a publish just seconds after yours, and the status that gets returned does not match and your publish would just vanish from the list.

I can address this...

Composer/packages/client/src/pages/publish/index.tsx
```diff
@@ -0,0 +1,476 @@
...
+  const [editTarget, setEditTarget] = useState();
+
+  const isRollbackSupported = useMemo(
+    () => (target, version): boolean => {
+      if (version.id && version.status === 200 && target) {
+        const type = publishTypes?.filter(t => t.name === target.type)[0];
+        if (type?.features?.rollback) {
+          return true;
+        } else {
+          console.log('rollback not supported on', type);
```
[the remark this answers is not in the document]
benbrown: i will remove these - they were just there to validate the logic.

Composer/packages/client/src/pages/publish/index.tsx
```diff
@@ -0,0 +1,476 @@
...
+        },
+        undefined
+      );
+    },
+    [settings.publishTargets, projectId, botName, editTarget]
+  );
+
+  useEffect(() => {
+    setDialogProps({
+      title: 'Add a publish profile',
```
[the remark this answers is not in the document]
benbrown: fixed

Composer/packages/client/src/pages/publish/index.tsx
```diff
@@ -0,0 +1,476 @@
...
+    setEditDialogHidden(false);
+  };
+
+  const onDelete = useMemo(
+    () => async (index: number) => {
+      const result = await OpenConfirmModal(
+        formatMessage('This will delete the profile. Do you wish to continue?'),
+        null,
+        {
+          confirmBtnText: formatMessage('Yes'),
```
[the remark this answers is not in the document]
benbrown: Are you talking about `confirmBtnText`? this is defined way up inside the confirm modal that is used throughout the app...

Composer/packages/client/src/pages/publish/index.tsx
```diff
@@ -0,0 +1,476 @@
...
+        hidden={addDialogHidden}
+        onDismiss={() => setAddDialogHidden(true)}
+        dialogContentProps={dialogProps}
+        modalProps={{ isBlocking: true }}
+        minWidth={350}
+      >
+        {dialogProps.children}
+      </Dialog>
+      <Dialog
+        hidden={editDialogHidden}
```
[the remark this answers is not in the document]
benbrown: done

Composer/packages/client/src/pages/publish/index.tsx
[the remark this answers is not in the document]
benbrown: done

Composer/packages/client/src/store/reducer/index.ts
benbrown: addressed.

PR #2462, opened 2020-04-01
feat: Setup infrastructure to launch and package Composer as an Electron app

### LINKED ISSUE #2224: Refactor& Update Electron Branch
- [x] Get rid of duplicate server code (use workspace dependencies)
- [x] Make sure dev workflow works in client / server config as well as client / electron-server config
- [x] Make sure distributed binary works for core scenarios

tonyanziano:
## Description

This PR piggybacks off of @christopheranderson's previous work to launch and package Composer as an Electron application.

Highlights of this PR:

- Introduction of the `electron-server` package which is a lightweight wrapper that starts up the Composer web app inside of an Electron shell
- Build infrastructure allowing the Composer web app into a cross-platform (Win / Mac / Linux) desktop application which will then be leveraged by our DevOps pipelines to produce stable / nightly releases

Please take a look at `electron-server/README.md` for information on how to build / run / package Composer Electron locally.

## Task Item

closes #2224 

## Screenshots

composer-icons-win


Composer/packages/electron-server/package.json
```diff
@@ -0,0 +1,150 @@
...
+    "source-map-explorer": "^2.1.0",
+    "ts-node": "^8.4.1"
+  },
+  "dependencies": {
+    "@bfc/server": "*",
+    "debug": "4.1.1",
+    "fix-path": "^3.0.0",
+    "lodash": "^4.17.15"
+  },
+  "build": {
```
a-b-r-o-w-n, line 61: is it possible for this to live outside of the package.json?

Composer/packages/electron-server/src/main.ts
```diff
@@ -0,0 +1,104 @@
...
+    browserWindowOptions.icon = join(getUnpackedAsarPath(), 'resources/composerIcon_1024x1024.png');
+  }
+  const win = new BrowserWindow(browserWindowOptions);
+
+  // and load the index.html of the app.
+  const CONTENT_URL = isDevelopment ? 'http://localhost:3000/' : 'http://localhost:5000/';
+  console.log('Loading project from: ', CONTENT_URL);
+
+  // TODO: get rid of this timeout
+  setTimeout(() => {
```
a-b-r-o-w-n, line 37: when and why can this be removed?

Composer/packages/server/src/server.ts
```diff
@@ -26,92 +26,92 @@ import pluginLoader from './services/pluginLoader';
...
-        CS_POLICIES.concat([
-          `script-src 'self' 'nonce-${req.__nonce__}';`,
-          // TODO: use nonce strategy after addressing issues with monaco-editor pacakge
-          "style-src 'self' 'unsafe-inline'",
-          // `style-src 'self' 'nonce-${req.__nonce__}';`,
-        ]).join(' ')
-      );
-    }
+export async function start(pluginDir?: string) {
+  const clientDirectory = path.resolve(require.resolve('@bfc/client'), '..');
```
a-b-r-o-w-n, line 30: Can you explain why this is necessary?

Composer/packages/electron-server/src/main.ts
tonyanziano: This apparently was an issue on Mac. @srinaath can provide more context. This was not required on Windows.

srinaath: So this needs to be removed hopefully soon. So, the reason is we `await` for the `server` to spin up. Electron builds a shell around `http://localhost:5000`. Server start() resolves the promise yet it takes a couple of seconds before `http://localhost:5000` is available. Hence, if i dont put that timeout, electron waits for the promise, the promise is resolved and immediately it spins up the wrapper window around the URL which is still empty as `http://localhost:5000` is not available yet (takes a few seconds for it to load on Mac). So, we have to  refresh contents after the window is loaded for us to get the correct view. I have tried a few electron events that indicate if the url is loaded or not etc. None seem to work. But, I figured this is unblocking our distribution pipeline for now  and I can work on removing it in the next few PR's

Composer/packages/server/src/server.ts
tonyanziano: I initially did this to back out of what Chris Anderson did in his branch, and also get rid of the need for the `/client/` build script to copy the build files over to the `/server/` package.

It seemed more straightforward to just take a dependency on `/client/` from `/server/`, because technically it does have a hard dependency on `/client/`, even though it's implicit (client must copy its built files over for server to work properly).

Also, in the Composer Electron meeting that started the hand-off from Chris to me, we discussed a potential future where we would have `composer-client`, `composer-server` (core that mounts the routes), and `composer-electron` (or whatever your wrapper would be -- Restify, Express, etc.) packages that would be pieced together to start Composer. In this model, copying files over from client to server doesn't really fit.

I'm ok with reverting this change and re-enabling the file copying if that's what we want. I was just trying to think ahead and set this up for a model where these packages aren't always local workspaces.

Composer/packages/electron-server/package.json
tonyanziano: Yes, I can extract this out into a json 👍

Composer/packages/server/src/server.ts
a-b-r-o-w-n: I'm not too sure about this approach. I agree that there is an implicit dependency on the client, but really anything can inject static assets to the public directory and have express serve them. The reason I'm questioning this is the changes to the dockerfile. It pretty clearly shows the additional complexity that is required on the host machine (by needing to copy client, extension, ui-plugins, etc). I don't want to block this PR on that so we can discuss more later.

Composer/packages/electron-server/electron-builder-config.json
```diff
@@ -0,0 +1,96 @@
...
+        "x": 380,
+        "y": 244,
+        "type": "link",
+        "path": "/Applications"
+      }
+    ]
+  },
+  "publish": null,
+  "remoteBuild": false
+}
```
a-b-r-o-w-n, line 96: Thanks!

Composer/packages/electron-server/src/main.ts
a-b-r-o-w-n: Ah. I think I see why. In `server.ts` just return a promise and resolve when the express app is ready. Line 129.

Composer/packages/server/src/server.ts
tonyanziano: Yeah, the additions to the docker file made me a little hesitant as well. Happy to discuss it further once this is in.

Composer/packages/electron-server/src/main.ts
srinaath: Thanks Andy.. Tony has made the fix to remove it

PR #2463, opened 2020-04-01
feat: add js version deployment script

### LINKED ISSUE #2461: Feature: Make Azure Deploy Plugin

- [x] Make Azure Deploy plugin by create and deploy script. Support publish method.

1. - [x] make sure the plugin does not expect the declarative assets to be local
2. - [x] make sure the login and provision script out of composer, and save the result in settings.json
3. - [x] show log in log dialog.
4. - [x] deploy to azure 
- [x] Support BotStatus getting.

--- Out of scope for now ---
- [ ] Support histroy getting.
- [ ] Support rollback.
- [ ] show provision in composer frontend.

zidaneymar:
## Description

add js version deploy script, same as the powershell one.
closes #2461 

## Task Item

closes #2461 

part of the azure deploy plugin


BotProject/Templates/CSharp/Scripts/Node/botProjectDeploy.ts
```diff
@@ -0,0 +1,506 @@
+
```
boydc2014, line 1: extra line breaks

BotProject/Templates/CSharp/Scripts/Node/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "botframeworkdeploy",
+  "version": "1.0.0",
+  "description": "typescript version of bot deployment",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "create": "tsc create.ts && node create.js"
```
boydc2014, line 8: the convention is "start" instead of "create"

BotProject/Templates/CSharp/Scripts/Node/botProjectDeploy.ts
```diff
@@ -0,0 +1,506 @@
...
+    private notEmptyLuisModel(file: string) {
+        return fs.readFileSync(file).length > 0
+    }
+
+    public async deploy(name: string, environment: string, luisAuthoringKey?: string, luisAuthoringRegion?: string, logFile?: string, botPath?: string, language?: string) {
+        if (!logFile) {
+            logFile = 'deploy_log.txt'
+        }
+        if (!this.creds) {
+            this.creds = await msRestNodeAuth.interactiveLogin()
```
boydc2014, line 234: can this interactiveLogin been invoked in non-cli env? replace this with a token parameter?

BotProject/Templates/CSharp/Scripts/Node/botProjectDeploy.ts
```diff
@@ -0,0 +1,506 @@
...
+            await this.botPrepareDeploy(deployFilePath)
+        }
+
+        const zipPath = path.join(this.projFolder, 'code1.zip')
+        if (await fs.pathExists(zipPath)) {
+            await fs.remove(zipPath)
+        }
+
+        // dotnet publish
+        const publishFolder = path.join(this.projFolder, 'bin\\Release\\netcoreapp3.1')
```
boydc2014, line 253: this path can be passed in via config

BotProject/Templates/CSharp/Scripts/Node/botProjectDeploy.ts
```diff
@@ -0,0 +1,506 @@
...
+        await this.dotnetPublish(publishFolder, this.projFolder, botPath)
+        const settingsPath = path.join(this.projFolder, 'appsettings.deployment.json')
```
boydc2014, line 255: same here

BotProject/Templates/CSharp/Scripts/Node/botProjectDeploy.ts
```diff
@@ -0,0 +1,506 @@
...
+                    body: account,
+                    json: true,
+                    headers: { "Authorization": `Bearer ${token.accessToken}`, "Ocp-Apim-Subscription-Key": luisAuthoringKey },
+                } as rp.RequestPromiseOptions;
+                const response = await rp.post(luisAssignEndpoint, options)
+                console.log(response)
+            }
+            console.log('Luis Publish Success! ...')
+        }
+        console.log('Packing up the bot service ...')
```
boydc2014, line 353: library can't emit console.lg

BotProject/Templates/CSharp/Scripts/Node/README.md
```diff
@@ -0,0 +1,21 @@
...
+>> const bot = new BotProjectDeploy(subId, projFolder)<br>
+>> const name = 'Repalce With Your Bot Name, This should be unique to global'<br>
+>> const environment = 'Replace With Environment'<br>
+>> const location = 'Replace With App Location, for example, westus'<br>
+>> const luisAuthoringKey = null<br>
+>> const luisAuthoringRegion = null<br>
+>> const appId = null<br>
+>> const appPassword = 'Replace With Your Password'<br>
+
+> 3. npm run create
```
boydc2014, line 17: s/create/start

BotProject/Templates/CSharp/Scripts/Node/README.md
```diff
@@ -0,0 +1,21 @@
+# Node Deployment
+
+## Instructions
+> 1. npm install
+> 2. Repalce the parameters in create.ts
```
carlosscastro, line 5: typo: replace

BotProject/Templates/CSharp/Scripts/Node/botProjectDeploy.ts
carlosscastro, line 1: do we need copyright headers for these files?

carlosscastro:
P0 to remove console.logs and have well defined error conditions that are bubbled to library callers

carlosscastro:
The Azure library usage looks good, great that you are using ARM client libraries.

BotProject/Templates/CSharp/Scripts/Node/botProjectDeploy.ts
```diff
@@ -0,0 +1,506 @@
...
+            const data = `[config]\nproject = BotProject.csproj`
+            fs.writeFile(pathToDeploymentFile, data, (err) => {
+                reject(err)
+            })
+        })
+    }
+
+    private async dotnetPublish(publishFolder: string, projFolder: string, botPath?: string) {
+        const projectPath = path.join(projFolder, 'BotProject.csproj')
+        await exec(`dotnet publish ${projectPath} -c release -o ${publishFolder} -v q`)
```
carlosscastro, line 183: To be consistent, you could consider using ARM to deploy rather than dotnet publish. The nice thing is that web app zip deploy through ARM will be useful the day we have other, non-dotnet languages, since it just requires a zip file. 

Here more info: https://docs.microsoft.com/en-us/azure/app-service/deploy-zip

BotProject/Templates/CSharp/Scripts/Node/botProjectDeploy.ts
```diff
@@ -0,0 +1,506 @@
...
+            Object.assign(luisConfig, luisAppIds)
+
+            const deploymentSettingsPath = path.join(publishFolder, 'appsettings.deployment.json')
+            let settings: any = await fs.readJson(deploymentSettingsPath)
+            settings['luis'] = luisConfig
+
+            await fs.writeJson(deploymentSettingsPath, settings)
+            const token = await this.creds.getToken()
+
+            const getAccountUri = `${luisEndpoint}/luis/api/v2.0/azureaccounts`
```
carlosscastro, line 331: Is there a luis client that we could use for this instead of crafting urls and requests manually? So that if the luis API changes, the client does the changes for us :)

BotProject/Templates/CSharp/Scripts/Node/botProjectDeploy.ts
zidaneymar: > To be consistent, you could consider using ARM to deploy rather than dotnet publish. The nice thing is that web app zip deploy through ARM will be useful the day we have other, non-dotnet languages, since it just requires a zip file.
> 
> Here more info: https://docs.microsoft.com/en-us/azure/app-service/deploy-zip

hi carloss, I use dotnet publish to produce an executable dotnet project, and then use the zip deploy you mentioned to publish that folder

BotProject/Templates/CSharp/Scripts/Node/botProjectDeploy.ts
zidaneymar: > Is there a luis client that we could use for this instead of crafting urls and requests manually? So that if the luis API changes, the client does the changes for us :)

yes, I should use that client, and indeed there is an azure luis client inside the @azure-js-sdk, but it also have some issues, I have tried to use that client for luis accounts authoring but failed, so I turned to use the straight forward http request way..

cwhitten:
@zidaneymar can you please address conflicts?

zidaneymar:
> @zidaneymar can you please address conflicts?

resolved

Composer/packages/lib/bot-deploy/src/botProjectDeploy.ts
```diff
@@ -0,0 +1,597 @@
...
+  private zipPath: string;
+  private publishFolder: string;
+  private settingsPath: string;
+  private templatePath: string;
+  private dotnetProjectPath: string;
+  private generatedFolder: string;
+  private remoteBotPath: string;
+  private logger: (string) => any;
+
+  private readonly tenantId = '72f988bf-86f1-41af-91ab-2d7cd011db47';
```
cwhitten, line 41: whose tenant id is this? It's used globally across all instances of this library?

Composer/packages/lib/bot-deploy/src/botProjectDeploy.ts
```diff
@@ -0,0 +1,597 @@
...
+import {
+  Deployment,
+  DeploymentsCreateOrUpdateResponse,
+  DeploymentsValidateResponse,
+  ResourceGroup,
+  ResourceGroupsCreateOrUpdateResponse,
+} from '@azure/arm-resources/esm/models';
+import { GraphRbacManagementClient } from '@azure/graph';
+import * as msRestNodeAuth from '@azure/ms-rest-nodeauth';
+import * as fs from 'fs-extra';
```
cwhitten, line 15: I'm concerned about this library being bound to being used on the filesystem. We should hook into the storage layer of Composer to reuse the serialize configurations when writing .dialog, .lu, .lg, so when someone has CosmosDB as its storage layer, this library can still service their Azure deployment needs. 

@benbrown @carlosscastro Looking for your suggestions on priority. Do we P0 or P1 this? The work to pull this out is non-trivial and could take awhile. Perhaps we land Publishing & Runtime E2E and revisit.

Composer/packages/lib/bot-deploy/src/botProjectDeploy.ts
```diff
@@ -0,0 +1,597 @@
...
+      this.templatePath,
+      location,
+      resourceGroupName,
+      timeStamp,
+      deploymentTemplateParam
+    );
+    this.logger(validation);
+
+    if (validation.error) {
+      this.logger(`! Template is not valid with provided parameters. Review the log for more information.`);
```
cwhitten, line 523: the interface for this component is not exposed at a fine enough grain. Let's update it. I see at least three statuses, perhaps 4:

`INFO`, `ERROR`, `WARNING`, `SUCCESS`

So I recommend an object that takes at least two fields, `status` & `message`

```javascript
this.logger({
  status: 'ERROR'
  message: ```
    Operation failed. Template is not valid with provided parameters. Review the log for more information. 
    To delete this resource group, run 'az group delete -g ${resourceGroupName} --no-wait'
    ```
})
```

Composer/packages/lib/bot-deploy/src/botProjectDeploy.ts
benbrown: i need to get deeper into how all of this works, but as it is, we pull stuff *out of storage and onto disk* before doing the deploy -- so in that sense, it doesn't really matter.

Composer/packages/lib/bot-deploy/src/botProjectDeploy.ts
zidaneymar: it belongs to microsoft organization, and we need this tenant id to initialize a ms graph client to create an app registration based on the password provided by user.
```
    const graphClient = new GraphRbacManagementClient(credsForGraph, this.tenantId, {
      baseUri: 'https://graph.windows.net',
    });
```

for app registration, azure js sdk also provide an options for whether this application is available to other tenants, 

```
  /**
   * Whether the application is available to other tenants.
   */
  availableToOtherTenants?: boolean;
```

https://github.com/Azure/azure-sdk-for-node/blob/master/Documentation/Authentication.md#for-azure-graph-sdk

Composer/packages/lib/bot-deploy/src/botProjectDeploy.ts
zidaneymar: resolved

Issue #2078, opened 2020-02-25
[Accessibility][Screen reader - Property Pane_Ask a question]: Screen reader is not reading any information while navigating over the syntax suggestions available in "Prompt(?)" edit field.


ashish315:
**User Experience** 
If a screen reader will not provide any information about the syntax suggestions available in "Prompt(?)" edit field, then screen reader dependent users will never get to know about code creation ideas.

**Test Environment**
Browser: Microsoft Edge Canary Version: 82.0.425.0 (Official build)
OS build: 2004 (19564.1005)
Screen Reader: Narrator
URL: [Bot Framework Composer](http://localhost:3000/dialogs/Main)
[**Prerequisite Here**](https://github.com/microsoft/BotFramework-Composer)

**Repro Steps:** 
1. Open the above mentioned URL in EDGE and turn on narrator.
2. Navigate to already created dialog tree item and activate it. Such as- Greetings, Actions, EndDialog etc.
3. Click on "+" and select 'Ask a question'. Then select any of the available options except Oauth login.
4. Navigate to property pane appeared at right side of the screen and navigate t"Prompt(?)" edit field.
5 .After '-' type "if" and navigate over the syntax suggestions.
6. Observer the issue.

**Actual:** 
Screen reader is not reading any information while navigating over the syntax suggestions available in "Prompt(?)" edit field. Screen reader keeps silent.

**Expected:**
Screen reader should announce the syntax suggestions available in "Prompt(?)" edit field as "If suggestion 1 of 3"

**MAS Impacted:** [MAS1.3.1](https://microsoft.sharepoint.com/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={54f28d1f-a2d1-4dcd-84e1-5c9b87e8aba4})

**Attachment:**
[MAS1.3.1_Syntax suggestions.zip](https://github.com/microsoft/BotFramework-Composer/files/4249580/MAS4.2.1_Syntax.suggestions.zip)
![MAS4 2 1_Syntax suggestions](https://user-images.githubusercontent.com/60913131/75241216-b8591b80-57eb-11ea-8192-9da4cc58cf86.png)

beyackle:
We're using the React version of the Monaco editor as provided here: https://github.com/suren-atoyan/monaco-react. On its own demo page (https://monaco-react.surenatoyan.com/), the autocomplete suggestions just get read as the names themselves with no indication of which number the user is on or how many there are in the box.

[...]

PR #2467, opened 2020-04-01
fix: remove duplicate entries from dropdown

### LINKED ISSUE #2456: duplicate completions in LG editors
## Describe the bug

Swapping between tabs in the form editor can cause multiple copies of completions to stack up.

## Browser

Edge 80.0

## OS

Windows 10

## To Reproduce

1. Go to the Bot Composer, and open a bot.
2. Move to or create an "Ask a Question" box (of any type other than OAuth Login).
3. Start typing "- if" in the Bot Asks box, and notice that it brings up autocomplete suggestions.
4. Move to the "Other" tab in the form editor, then back to "Bot Asks".
5. Start typing "- if" in the box again, and notice that there are now 5 copies of the same one. Moving to Other adds 3 (one for each of the LGEditor boxes there), and then moving back to Bot Asks adds one more.
6. Note that continuing to switch tabs stacks more and more copies up.

## Expected behavior

There should only be one instance of every possible completion no matter how many times we switch tabs.

## Screenshots

image

beyackle:
## Description

The LgEditor and LuEditor kept re-registering their languages, which is what led to the duplicate entries in the autosuggest box.

## Task Item

Refs #2078 
Fixes #2456 

## Screenshots

image


a-b-r-o-w-n:
:shipit:

PR #2487, opened 2020-04-02
feat: create lg template when schema provides a default value

### LINKED ISSUE #2406: [dialog factory] create lg template when an lg field has a default value
## Describe the bug

Currently, when adding actions that include default values for lg, like QnA step, the editors complain that there are no templates in the lg file. To fix this, the form should create one if there is no value and a default is provided.

## To Reproduce

1. Create a new QNA maker step in a brand new bot without any lg (like EmptyBot).

## Expected behavior

No error should be shown.

## Screenshots

image

a-b-r-o-w-n:
## Description

Previously, the user had to interact with the editor in order to get the template saved.

Note: There is an issue right now with the lg api where if there are multiple parallel writes, the last write wins. So only the last lg field on the form will be created. There is an issue tracking this: #2247 

## Task Item

closes #2406

### DIFF
--- a/Composer/packages/ui-plugins/lg/src/LgField.tsx
+++ b/Composer/packages/ui-plugins/lg/src/LgField.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-import React, { useCallback, useState, useRef, useEffect } from 'react';
+import React, { useCallback, useState, useRef, useEffect, useMemo } from 'react';
 import { LgEditor } from '@bfc/code-editor';
 import { FieldProps, useShellApi } from '@bfc/extension';
 import { FieldLabel } from '@bfc/adaptive-form';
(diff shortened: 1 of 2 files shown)


Composer/packages/ui-plugins/lg/src/LgField.tsx
```diff
@@ -48,19 +48,39 @@ const LgField: React.FC<FieldProps<string>> = props => {
     [lgName, lgFileId]
   );
 
-  const template = (lgFile &&
-    lgFile.templates &&
-    lgFile.templates.find(template => {
+  const hasTemplate = useMemo(() => lgFile?.templates.find(t => t.name === lgName), []);
+  const template = (hasTemplate &&
```
zhixzhan, line 52: Here seems no need to find again, if hasTemplate is not a boolean value. it's equal to
```
const template = hasTemplate || {
     name: lgName,
     parameters: [],
     body: getInitialTemplate(name, value),
   };
```

Issue #2494, opened 2020-04-03
Template "todo with luis" not work


xhr0804:
## Describe the bug

When I create a bot from the template "todo with luis", and use the emulator to test, I got an error.
![image](https://user-images.githubusercontent.com/44257569/78336748-7aa59a80-75c2-11ea-8415-86a576939bee.png)
![image](https://user-images.githubusercontent.com/44257569/78336887-ae80c000-75c2-11ea-8f98-3c9b6be1b0bd.png)

cwhitten:
addressed via https://github.com/microsoft/BotFramework-Composer/pull/2497

PR #2497, opened 2020-04-03
fix: rename ToDoBotWithLuisSample assets to lowercase

### LINKED ISSUE #2495: need to lowercase the file and directory name for todowithluis Sample
file name and directory name under dialogs are not lowercase. eg. "AddItem"

zhixzhan:
## Description

rename ToDoBotWithLuisSample assets to lowercase

## Task Item

close #2495 
refs #2494


Composer/packages/server/assets/projects/ToDoBotWithLuisSample/dialogs/AddItem/AddItem.dialog
```diff
@@ -109,6 +109,6 @@
       ]
     }
   ],
-  "generator": "AddItem.lg",
-  "recognizer": "AddItem.lu"
+  "generator": "additem.lg",
```
luhan2017, line 112: do we also need to rename the folder and .dialog file?

zhixzhan: Yes,,, git commit did not catch the file rename, now updated.

PR #2521, opened 2020-04-06
feat: Create bot flow

### LINKED ISSUE #1728: UX | Create-new wizard extension and enhancement
Pick up where we left off with the Create New bot/dialog wizard.

Business justification:
Currently, Composer skips the step if Design/Plan and jumps right into building a dialog, assuming the user knows what they are creating and why, and that they are fluent with Composer and the SDK. The 'Created New' wizard attached below helps people getting started by asking them to input some information in a 'human' /conversational manner, and uses this to bootstrap the dialog. I believe this will lower the threshold for users to get started with Composer and hit the ground running and making it easier to understand concepts such as LG, LU, Events, Triggers etc. as they are already repopulated with the content provided in the Wizards.

Also 
- We are approaching the stage where we're allowing people to create multiple types of dialogs, so our Create New flow needs to scale to support this and make this easily discoverable:
- adaptive dialogs & sub-dialogs
- generated dialogs
- QnA dialogs
- Samples

https://www.figma.com/file/NqlONHOHjE0amtoHlLlopg9p/Composer-XY-07.16?node-id=1632%3A25465

Enable users to 'seed' their dialog with 
- Take full advantage of Managed LU

liweitian:
## Description
Applied new design for create bot option 3.

## Task Item
Closes #1728 
## Screenshots
aaaaa


cwhitten:
@liweitian can you please add some tests and also address the currently failing Unit and E2E tests? Thank you.

Composer/packages/client/src/CreationFlow/CreateOptions/index.js
```diff
@@ -27,83 +47,120 @@ export function CreateOptions(props) {
...
+      minWidth: 200,
+      maxWidth: 450,
+      isResizable: !disabled,
+      data: 'string',
+      styles: rowTitle(disabled),
+      onRender: item => item.description,
+    },
+    {
+      key: 'column3',
+      name: formatMessage('Created'),
```
cwhitten, line 92: This column isn't necessary. Please remove

Composer/packages/client/src/CreationFlow/CreateOptions/index.js
```diff
@@ -27,83 +47,120 @@ export function CreateOptions(props) {
...
+      fieldName: 'name',
+      minWidth: 150,
+      maxWidth: 200,
+      isResizable: !disabled,
+      data: 'string',
+      styles: rowTitle(disabled),
+      onRender: item => item.name,
+    },
+    {
+      key: 'column2',
```
cwhitten, line 80: can you give this field a key of "description"

Composer/packages/client/src/CreationFlow/CreateOptions/index.js
```diff
@@ -27,83 +47,120 @@ export function CreateOptions(props) {
...
+    if (option === 'CreateFromTemplate') {
       onNext(template);
     } else {
       onNext(emptyBotKey);
     }
   };
 
+  const tableColums = [
+    {
+      key: 'column1',
```
cwhitten, line 69: can you give this field a key of "name"

Composer/packages/client/src/CreationFlow/CreateOptions/index.js
liweitian: Removed. But the design has this column.

Composer/packages/client/src/CreationFlow/CreateOptions/index.js
```diff
@@ -27,83 +47,109 @@ export function CreateOptions(props) {
...
+      );
+    }
+    return null;
+  };
   return (
     <Fragment>
       <ChoiceGroup
         label={formatMessage('Choose how to create your bot')}
-        defaultSelectedKey="Create from scratch"
+        defaultSelectedKey="CreateFromScratch"
```
a-b-r-o-w-n, line 113: Its probably a good idea to make this and `CreateFromTemplate` a const value.

Composer/cypress/integration/CreateNewBot.spec.ts
```diff
@@ -21,7 +21,7 @@ context('Creating a new bot', () => {
 
   it('can create a bot from the ToDo template', () => {
     cy.findByTestId('Create from template').click();
-    cy.findByTestId('TodoSample').click();
+    cy.findByTestId('TodoSample').click({ force: true });
```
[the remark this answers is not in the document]
liweitian: https://github.com/cypress-io/cypress/issues/3450
https://github.com/cypress-io/cypress/issues/871

Issue #2555, opened 2020-04-08
luis publishing issue


KayMKM:
## Describe the bug

When I created a dialog with luis recognizer, but there is no .lu file in it. The luis publish progress doesn't work and doesn't throw any error.

## Browser

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

## To Reproduce

1. Create a dialog
2. Choose LUIS as recognizer
3. Click start

## Expected behavior

throw error if it couldn't publish luis

cwhitten:
addressed via https://github.com/microsoft/BotFramework-Composer/pull/2557

PR #2557, opened 2020-04-08
fix: catch the throw error message from the luis file check function

lei9444:
## Description
The luis check function before publishing will check empty file and referred lu files. The error message should show if the check fail.

Do some refactor about the publish luis and load bot logic:
The botStatus ={
  connected = 'connected',
  unConnected = 'unConnected',
  publishing = 'publishing',
  published = 'published',
  reloading = 'loading',
  pending = 'pending',
  failed = 'failed',
}

If the project is new the status = unConnected;
if there are some errors when starting the bot status = failed, the callout will show the error message, the the status will be set pedding.
if status = publishing or reloading, will show the loading component.
if publish luis succeed the status will be set published and then will call the loadBot function.
if finish the flow, the status will be set connected.

## Task Item
refs #2555


Composer/packages/client/src/store/action/lu.ts
```diff
@@ -83,8 +83,8 @@ export const publishLuis: ActionCreator = async ({ dispatch, getState }, authori
   } catch (err) {
     dispatch({
       type: ActionTypes.PUBLISH_LU_FAILED,
-      payload: { title: Text.LUISDEPLOYFAILURE, message: err.response.data.message },
+      payload: { title: Text.LUISDEPLOYFAILURE, message: err.message || err.response.data.message },
     });
-    throw new Error(err.response.data.message);
+    throw err;
```
a-b-r-o-w-n, line 88: This is an anti-pattern. Our ActionCreators should not not return anything or throw. That's what the dispatcher is for.

Can we keep line 86 and delete 88 and make sure no application code is relying on this behavior?

lei9444: Hi, @a-b-r-o-w-n I have do some refactor about these code and set botstatus to update the ui and logic in test flow.

Composer/packages/client/src/components/TestController/loading.tsx
```diff
@@ -0,0 +1,30 @@
...
+  botStatus: BotStatus;
+}
+
+export const Loading: React.FC<ILoadingProps> = props => {
+  const { botStatus } = props;
+  const publishing = botStatus === BotStatus.publishing;
+  const reloading = botStatus === BotStatus.reloading;
+
+  if (!publishing && !reloading) return null;
+  console.log(botStatus);
```
a-b-r-o-w-n, line 21: This should be removed.

PR #2586, opened 2020-04-09
feat: Visual Editor as an extension

### LINKED ISSUE #2120: refactor visual designer to use extension pattern

yeze322:
## Description
Apply the extension pattern to Visual Editor
fix #2120 

### Extension pattern's impact on Visual Editor
1. Visual Editor received verbose props before, now it consumes Extension context.
image

2. Visual Editor didn't support plugin before, now it supports plugin so we can put complicated $kinds out of visual's core lib.
image

3. Visual Editor's uischema (visualSchema) only supported direct import React Component before, now it supports register a widget and consume it by string.
image

### Main changes
1. naming pattern: `uiSchemaXXX` -> `visualSchemaXXX`
2. integrate Visual Editor to the plugin system
3. move several complicated $kinds to plugins (`SendActivity` -> plugin/lg, `TextInput` / `ChoiceInput` ... -> plugin/prompts)
4. create `@bfc/ui-shared` lib to share ui elements between Visual Editor and plugins (shared as a design language)

### Plugin schema that Visual Editor supports
Visual Editor requires two parts of config -- widgets and schema
```ts
{
  // register a React component as a visual editor widget named 'LgWidget'
  widgets: {
    LgWidget: (lgTemplateName) => <>Some React nodes</>,
  },
  schema: {
    'Microsoft.SendActivity': {
      // consume the widget
      widget: 'LgWidget',
      field: 'activity',
    }
  }
}
```
As an example, in `plugins/lg`, it declares a widget named 'LgWidget' and register the visual node definition of `Microsoft.SendActivity`.
image

## Task Item
closes #2120


Composer/packages/extensions/adaptive-form/src/PluginContext.ts
```diff
@@ -6,6 +6,7 @@ import { PluginConfig } from '@bfc/extension';
 
 const PluginContext = React.createContext<Required<PluginConfig>>({
   uiSchema: {},
+  visual: {},
```
cwhitten, line 9: these names here are odd. `uiSchema` ? `visual` ? Questions I have as a developer using this interface:

1. What's the difference? `uiSchema` corresponds to forms and `visual` corresponds to the flow editor? Not very self-descriptive.
2. Why the need for both?

a-b-r-o-w-n: What do you think would be most intuitive?

I think making a change like make sense.

```ts
interface uiSchema {
  formSchema: FormSchema;
  flowSchema: FlowSchema;
}
```

cwhitten: LGTM. @yeze322 can you make this change?

yeze322: @a-b-r-o-w-n @cwhitten Great! It's a good name, I will update it

yeze322: `visual` renamed to `flowSchema`, renamed `uiSchema` to `formSchema` in a separated PR https://github.com/microsoft/BotFramework-Composer/pull/2634

yeze322:
renamed `uiSchema` to `formSchema` in a separated PR https://github.com/microsoft/BotFramework-Composer/pull/2634

Issue #2596, opened 2020-04-10
LG syntax incompatible change


zhixzhan:
## Describe the bug

LG team recently  changed the naming policy. https://github.com/microsoft/botbuilder-dotnet/pull/3712
- Remove dash in identifier (template name)

It's an incompatible change, and previous LG file must change their template name, probably use _ instead.

e.g,  replace dash  `-` to underline `_`
```
bfdactivity-76HgfJ 
-> 
bfdactivity_76HgfJ
```

## Browser

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

zhixzhan:
We can upgrade all sample bot, and composer generate LG template pattern. One migration solution is before we switch to new lg parser (which would failed to parse), do replacement at current lg  parser ( which both support dash - & underline  _  ).

boydc2014:
@zhixzhan  your 3 work items are right. But i'm not sure about about the "migration solution" you mentioned, can we do a regex replace and match on every dialog and lg file, when loading bot will that work?

zhixzhan:
@boydc2014 Sure, regex replacement seems more reliable compared to use parser do update.

Issue #778, opened 2019-08-31
Default naming scheme, behavior for all concepts


vishwacsena:
# Default names
This specification outlines the default naming convention for all concepts in composer. 

All triggers, actions start with a default name but is user editable. 

## Bot
There is no default name for a Bot. User must explictly provide a name.

## Dialog
There is no default name for a dialog. User must explicitly provide a name.

## Trigger
- Only **Intent** trigger has a name property. 
- Default name for a trigger will be its type name - e.g. IntentTrigger
- In case there already exists a trigger with the name, add a number increment as suffix - e.g. IntentTrigger, IntentTrigger1, IntentTrigger2, ...
- All trigger names within a specific dialog must be unique.
- In case of intent trigger, the intent name will default to the trigger name.
- If user renames the trigger in the trigger's property editor, then the intent name is also updated.
- If user changes the intent name of a trigger in all-up view and/ or if this intent is used as a reference in another intent, then ask the user for the intended behavior -
    - Update all references to the new name
    - Keep reference as is
    
## Input/ prompt actions
- Default name for an input action will be its type name - e.g. NumberInput, TextInput, ChoiceInput, ConfirmInput, DateTimeInput, OAuthInput, AttachmentInput
- In case there already exists an input action with that name, add a number increment as suffix - e.g. NumberInput, NumberInput1, NumberInput2, ...
- All input names tied to a specific dialog must be unique.
- The intent name for an input default to the input name.
- The prompt LG template name defaults to the input name.
- The unrecognized prompt LG template name defaults to \<input-name>.unrecognized.prompt
- The invalid prompt LG template name defaults to \<input-name>.invalid.prompt
- If user renames the input in the input's property editor, then the intent name is also updated, LG template names are also updated.
- If user changes the intent name of an input in the all-up view or changes the LG template name tied to an input in the all-up view and/ or if the intent or lG template is used as a reference in another intent/ LG template, then ask the user for the intended behavior -
    - Update all references to the new name
    - Keep reference as is

## SendActivity
- Default name is sendActivity, add an increment as suffix to de-dupe - e.g. SendActivity, SendActivity1, SendActivity2, ...
- The default LG template name is the name of the SendActivity action
- If user renames the send activity in its property editor, then the LG template name is also updated.
- If user changes the LG template name in all-up view and/ or if this LG template is used as a reference in other intents or templates, then ask the user for the intended behavior -
    - Update all references to the new name
    - Keep reference as is

## All other actions
[...]

PR #2597, opened 2020-04-10
feat: lg naming migration

### LINKED ISSUE #2499: repeat this dialog not working
## Describe the bug

For this dialog, after ask, go back to main
image

## Version

Current master

## Browser

- [x] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

zhixzhan:
## Description

LG team recently  changed the naming policy. https://github.com/microsoft/botbuilder-dotnet/pull/3712
- Remove dash in identifier (template name)

It's an incompatible change, and previous LG file must change their template name, probably use _ instead.

e.g,  replace dash  `-` to underline `_`
```
bfdactivity-76HgfJ 
-> 
bfdactivity_76HgfJ
```

Likely there are 3 part work.

1. update all sample bot.
2. update generate LG template pattern. 
3. migrate user's bot

For 3, one migration solution is before we switch to new lg parser (which would failed to parse), do replacement at current lg  parser ( which both support dash - & underline  _  ).

#### LG
1. bfdactivity-1234 -> SendActivity_1234
2. bfdprompt-1234 -> TextInput_Prompt_1234
3. bfdinvalidPrompt-1234 -> TextInput_InvalidPrompt_1234
4. bfdunrecognizedPrompt-1234 -> TextInput_UnrecognizedPrompt_1234
5. bfddefaultValueResponse-1234 -> TextInput_DefaultValueResponse_1234

#### LU
6. TextInput.response-RuZhXe -> TextInput_Response_RuZhXe

In this PR, we've also updated the sdk packages and schems which includes some features and bug fixes:
1. Enable composer to create a skill. 
2. LanguagePolicy fix.
3. Schema update, Microsoft.Recognizer renamed to Microsoft.IRecognizer, which have some coresponding SDKKinds fixes. 
3. Some other SDK fix like, foreach, RepeatOption, some lg expression fixes, oauth fix....

fixes #2499, #2624

## Task Item

refs #2596 
refs #778 

## Screenshots

After load a old bot, replace results: 
*.dialog:
Screen Shot 2020-04-14 at 3 25 56 PM

*.lg:
Screen Shot 2020-04-14 at 3 11 12 PM

adaptiveCard.json
Screen Shot 2020-04-14 at 3 39 42 PM

### Won't auto-replace case:
1. reference is user customize defined lg template name, 
```
- ${my-template()}
```
technically can not distinguish it is an expression or a template reference.


cwhitten:
hey @zhixzhan. thanks for taking this one. I'd like to take the opportunity to change the LG template naming schema to be more inline with where we want to go. instead of `bfdactivity_${id}` let's write `SendActivity_${id}`. Can you include this change here?

Additionally, let's make sure to also update the following:

Inline LG naming conventions in the prompts
`# bfdprompt-RuZhXe()` -> `${inputType}_Prompt_${id}`

Inline LU naming conventions in the prompts
`# TextInput.response-RuZhXe` -> `${inputType}_Response_${id}`

cc @tdurnford do you have any comments here?

zhixzhan:
> hey @zhixzhan. thanks for taking this one. I'd like to take the opportunity to change the LG template naming scheme to be more inline with where we want to go. instead of `bfdactivity_${id}` let's write `SendActivity_${id}`. Can you include this change here?
> 
> Additionally, let's make sure to also update the following:
> 
> Inline LG naming conventions in the prompts
> `# bfdprompt-RuZhXe()` -> `${inputType}_Prompt_${id}`
> 
> Inline LU naming conventions in the prompts
> `# TextInput.response-RuZhXe` -> `${inputType}_Response_${id}`
> 
> cc @tdurnford do you have any comments here?

@cwhitten  Okay, will update it.

Composer/packages/ui-plugins/lg/src/LgField.tsx
```diff
@@ -35,11 +35,20 @@ const getInitialTemplate = (fieldName: string, formData?: string): string => {
 };
 
 const LgField: React.FC<FieldProps<string>> = props => {
-  const { label, id, description, value, name, uiOptions } = props;
+  const { label, id, description, value, name, uiOptions, schema } = props;
```
zhixzhan, line 38: Hi @tdurnford can you please help me take a look at here, I trying to get the `InputType` from `schema` property, (just like `LuisIntentEditor.tsx` did ), but data value passed to here seems different.

tdurnford: @zhixzhan  The `LgField` only has access to the schema for the particular field it is overriding - `IActivityTemplate` - while the `LuisIntentEditor` has access to the whole schema and can pull the `$kind` from it. Even though field overrides and intent editors behave differently, it looks like you can still get the `$kind` through the `useShellApi`.

```javascript
const { data } = useShellApi();
const kind = data?.$kind;
```

tdurnford:
@zhixzhan @cwhitten I think it's a good idea to standardize the naming conventions for lg and lu in the prompts, and `${inputType}_Prompt_${id}` and `${inputType}_Response_${id}` respectively makes sense to me.

cwhitten:
@zhixzhan I like your plan to migrate between LG parsers. Will you include that in this change?

zhixzhan:
> @zhixzhan I like your plan to migrate between LG parsers. Will you include that in this change?

@cwhitten you mean update the latest lgParser include this? I know the js version is on the way, probably ready at middle of this week. if so, yes we will include this.

cwhitten:
> > @zhixzhan I like your plan to migrate between LG parsers. Will you include that in this change?
> 
> @cwhitten you mean update the latest lgParser include this? I know the js version is on the way, probably ready at middle of this week. if so, yes we will include this.

No I mean the strategy to update old templates to the new naming convention.

Composer/packages/server/src/models/bot/botProject.ts
```diff
@@ -549,6 +549,91 @@ export class BotProject {
     }
   };
 
+  private _replaceDashInTemplateName = async () => {
```
zhixzhan, line 552: @cwhitten here implement a replace method at server, when load a bot, will update all old template name to new naming convention.

zhixzhan: But here only ensure old bot can be valid. completion renaming did covered at here, for example :
bfdprompt-1234 --->bfdprompt_1234 --//--> TextInput_Prompt_1234
Do you think we should implement a complete replacement here?

cwhitten: Yes, I would like to.

Composer/packages/ui-plugins/lg/src/LgField.tsx
zhixzhan: @tdurnford get it, thanks!

zhixzhan:
> > > @zhixzhan I like your plan to migrate between LG parsers. Will you include that in this change?
> > 
> > 
> > @cwhitten you mean update the latest lgParser include this? I know the js version is on the way, probably ready at middle of this week. if so, yes we will include this.
> 
> No I mean the strategy to update old templates to the new naming convention.

Sure, updated

Composer/packages/ui-plugins/lg/src/LgField.tsx
```diff
@@ -36,10 +36,21 @@ const getInitialTemplate = (fieldName: string, formData?: string): string => {
...
+  if (kind) {
+    const [, schemaType] = kind.split('.');
+    if (schemaType === 'SendActivity') {
+      lgType = schemaType;
+    } else {
+      const nameCapitalized = name.charAt(0).toUpperCase() + name.slice(1);
+      lgType = `${schemaType}_${nameCapitalized}`;
+    }
+  }
   const singleLgRefMatched = value && value.match(/\$\{([\w-]+)(\(.*\))\}/);
```
yeze322, line 52: we can leverage the `LgTemplateRef` util (though it's not touched)

Composer/packages/ui-plugins/lg/src/LgField.tsx
```diff
@@ -36,10 +36,21 @@ const getInitialTemplate = (fieldName: string, formData?: string): string => {
...
+
+  let lgType = name;
+  const kind = data?.$kind;
+  if (kind) {
+    const [, schemaType] = kind.split('.');
+    if (schemaType === 'SendActivity') {
+      lgType = schemaType;
+    } else {
+      const nameCapitalized = name.charAt(0).toUpperCase() + name.slice(1);
+      lgType = `${schemaType}_${nameCapitalized}`;
```
yeze322, line 49: Considering the 'schema-driven' pattern, we need to consume $kind string from `SDKKinds` instead of harcoding the `SendActivity` string.
```ts
if (kind === SDKKinds.SendActivity) {
  ...
}
```

yeze322: I'm also curious about why only `SendActivty` has a special naming pattern instead of naming it to `'SendActivity_activity_xxxx'`?

yeze322:
Just a discussion, shall we have an independent `composer-lg` library to handle LG naming pattern migrations in the future?

The lib will contain APIs such as `buildLgName()` and `parseLgName()` and evolve with lg lib, only the LG plugin imports it independently.

It will make the plugin system more healthy. (Ask for `composer-lu` in plugins/lu as well)

Composer/packages/ui-plugins/prompts/src/PromptField/UserInput.tsx
```diff
@@ -27,7 +27,7 @@ const UserInput: React.FC<PromptFieldProps<MicrosoftInputDialog>> = props => {
 
   const { const: kind } = (schema?.properties?.['$kind'] as any) || {};
   const [, promptType] = ((kind as string) || '').split('.');
-  const intentName = `${promptType}.response-${designerId}`;
+  const intentName = `${promptType}_Response_${designerId}`;
```
yeze322, line 30: LU name builder in plugins/prompts

Composer/packages/ui-plugins/luis/src/LuisIntentEditor.tsx
```diff
@@ -18,7 +18,7 @@ const LuisIntentEditor: React.FC<FieldProps<string>> = props => {
   if (typeof intentName === 'object') {
     const { $kind }: any = schema?.properties || {};
     const [, promptType] = $kind.const.split('.');
-    promptType && (intentName = `${promptType}.response-${designerId}`);
+    promptType && (intentName = `${promptType}_Response_${designerId}`);
```
yeze322, line 21: LU name builder in plugins/luis

yeze322: Shall we have a shared LU name builder?

Composer/packages/ui-plugins/lg/src/LgField.tsx
zhixzhan: Thanks remind!

zhixzhan:
> Just a discussion, shall we have an independent `composer-lg` library to handle LG naming pattern migrations in the future?
> 
> The lib will contain APIs such as `buildLgName()` and `parseLgName()` and evolve with lg lib, only the LG plugin imports it independently.
> 
> It will make the plugin system more healthy: a plugin holds its all necessary assets.
> 
> (Ask for `composer-lu` in plugins/lu as well)

I take a look at ui-plugins, they are depends on lib/shared and lib/indexers, for some types, interface or  utils. I guess it's ok to have a fat packages for those ui components.
Otherwise, we need to split those types, interface, and utils to very fine-grained level.

But before that, I agree implement lu name builder at lib/shared as a first step.

Composer/packages/ui-plugins/lg/src/LgField.tsx
zhixzhan: Seems no need to make change on LgTemplateRef, here just passed a new lgType to it.

zhixzhan:
updated

luhan2017:
I've tested this branch. All the luis bot are not working correctly because such settings are auto generated by lubuild, "=settings.luis.additem_en-us_lu", but we don't support - in expressions right now. 

Talked with @boydc2014 and @feich-ms , fei is going to make a change in lubuild to replace all the - with _ in expressions. We need to wait until fei's update is ready.

zhixzhan:
> Need to update bf lu package

@luhan2017 Luis package has been updated, can you have another try.

cwhitten:
do we need to migrate old luis settings that use the `'-'` character?

zhixzhan:
> do we need to migrate old luis settings that use the `'-'` character?

@luhan2017 if re-run lubuild or publish, can previous settings be erased?

luhan2017:
> @luhan2017 if re-run lubuild or publish, can previous settings be erased?

Yes they will be re-generated.

cwhitten:
@zhixzhan in the todowithluis sample I'm seeing broken behavior attempting to navigate between the all-up lg page and the design page, and also when trying to delete dialogs. I don't see the same behavior when using other samples.

Screen Shot 2020-04-18 at 11 21 21 AM

zhixzhan:
> @zhixzhan in the todowithluis sample I'm seeing broken behavior attempting to navigate between the all-up lg page and the design page, and also when trying to delete dialogs. I don't see the same behavior when using other samples.
> 
> Screen Shot 2020-04-18 at 11 21 21 AM
> 

> Also seeing the following in the console:
> 
> LU Language Client: The server is initialized.
> console-window.js:61 LU Language Client: Section todobotwithluissample-3.lu#newSection do not exist

@cwhitten updated

Issue #2610, opened 2020-04-10
Add down-sampling before publishing Lu files 


lei9444:
## Is your feature request related to a problem? Please describe.

1. If the number of intent's utterances more than 10(can be customized) times the least number of the intent's utterances, do a re-sampling process to keep the proportion;

2. If detect > 15000 utterances, downsize the utterances.

cwhitten:
https://github.com/microsoft/BotFramework-Composer/pull/2629

PR #2629, opened 2020-04-13
feat: add down sampling before luis publish

lei9444:
## Description
If the number of intent's utterances more than 10(can be customized) times the least number of the intent's utterances, do a re-sampling process to keep the proportion;

If detect > 15000 utterances, downsize the utterances.

add down sampling config to bot setting
image
 

## Task Item
refs #2610


Composer/packages/server/src/models/bot/sampler/BootstrapSampler.ts
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { BootstrapSampler } from '@microsoft/bf-dispatcher/lib/mathematics/sampler/BootstrapSampler';
+
+const SAMPLE_SIZE_CONFIGURATION = 2;
```
boydc2014, line 6: this should go through some config file

Composer/packages/server/src/models/bot/sampler/ReservoirSampler.ts
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+import { ReservoirSampler } from '@microsoft/bf-dispatcher/lib/mathematics/sampler/ReservoirSampler';
+
+import { IUtterance } from './BootstrapSampler';
+
+const MIN_SAMPLE_SIZE = 15000;
```
boydc2014, line 7: same here

boydc2014: if you don't make this configurable, i bet this limit would also never hit, this module will even never run?

Composer/packages/server/src/models/bot/sampler/BootstrapSampler.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export interface IUtterance {
+  text: string;
+  intent: string;
+  entities: any[];
+}
+
+export class ComposerBootstrapSampler extends BootstrapSampler<number> {
+  private _times = 10;
```
boydc2014, line 15: this ratio is also a key indicator, which should come through config, this is probably a ratio we adjust frequently

Composer/packages/server/src/models/bot/sampler/BootstrapSampler.ts
boydc2014: sorry, this one should probably not exposed to user.

boydc2014:
In terms of putting it into samples, we should give this a meaningful name

downsampling: {
    max_imbalance_ratio: 10, 
    max_utterance_allowed: 15000
}

lei9444:
> In terms of putting it into samples, we should give this a meaningful name
> 
> downsampling: {
> max_imbalance_ratio: 10,
> max_utterance_allowed: 15000
> }

Hi @boydc2014 I have add the config to client setting now. user can update it in setting page

boydc2014:
Looks good to me, can you ask @hibrenda or someone from skill team to have a try?

hibrenda:
I tried this for todo skill. 2 findings. 
- down sampling does not only  resizing the big ratio intent (eg. ratio is more than 1:10) but also resizing the small ratio intent (eg. 1:8) which should not be happen
- For bot developer's awareness, hope to get the notification after down sampling for some intents

lei9444:
@hibrenda @boydc2014 I have fix the first commit. For the second, I think we can have more discusses about this.

hibrenda:
Verified the the fix. now the resizing only for the utterance  number bigger than the ratio defined

cwhitten:
thank you @hibrenda

Issue #2016, opened 2020-02-19
[Accessibility][Info&Relationship-Filter Dialog]: Ensures elements under 'filter dialog' with an ARIA role that require child roles contain them.


ashish315:
**User Experience** 
Users who rely on AT will face difficulty to complete task if there is no child available for element.

**Test Environment**
Browser: Microsoft Edge Canary Version: 82.0.425.0 (Official build)
OS build: 2004 (19564.1005)
URL: [Bot Framework Composer](http://localhost:3000/dialogs/Main)
[**Prerequisite Here**](https://github.com/microsoft/BotFramework-Composer)

**Repro Steps:** 
1. Open the above mentioned URL in EDGE.
2. Select 'Design Flow' from left navigation.
3. Now, run the Accessibility Insight tools to find the issue.

**Actual:** 
Ensures elements under 'filter dialog' with an ARIA role that require child roles contain them.

**Expected:**
Every element with an ARIA role must have child roles contain them.

**Note:**
This accessibility issue was found using Accessibility Insights for Web 2.14.1 (axe-core 3.3.2), a tool that helps find and fix accessibility issues. Get more information & download this tool at [http://aka.ms/AccessibilityInsights](http://aka.ms/AccessibilityInsights).

**Element path:**
div[data-testid="DialogTreeItemEmitAnEvent"]

**Snippet:**
`<div data-testid="DialogTreeItemEmitAnEvent" class="ms-FocusZone css-117 ms-OverflowSet css-zqjw3r root-176" role="menubar" data-focuszone-id="FocusZone147">`

**How to fix:**
    Fix any of the following:
      Required ARIA children role not present: menuitem menuitemradio menuitemcheckbox

**MAS Impacted:** [MAS1.3.1](https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={54f28d1f-a2d1-4dcd-84e1-5c9b87e8aba4})

**Attachment:**
![MAS1 3 1_Ensures elements with an ARIA role](https://user-images.githubusercontent.com/60913131/74821883-4768be00-532a-11ea-9615-50c1ceffdd24.png)
![MAS1 3 1_Ensures elements with an ARIA role_Whole Filter dialog](https://user-images.githubusercontent.com/60913131/74821886-4899eb00-532a-11ea-8904-a29f43755ec7.png)

ashish315:
[...]

Issue #2054, opened 2020-02-23
[Accessibility][Info&Relationship-Filter Dialog]: Aria label and place holder both are defined same in the code on 'Filter Dialog' edit box.


ashish315:
**User Experience** 
Screen reader users will get confused if the same information is convey by the narrator twice.

**Test Environment**
Browser: Microsoft Edge Canary Version: 82.0.425.0 (Official build)
OS build: 2004 (19564.1005)
Screen Reader: Narrator
URL: [Bot Framework Composer](http://localhost:3000/dialogs/Main)
[**Prerequisite Here**](https://github.com/microsoft/BotFramework-Composer)

**Repro Steps:** 
1. Open the above mentioned URL in EDGE and set the zoom at 400%.
2. Select 'Design Flow' from left navigation.
3. Now, navigate to the filter dialog edit field.
4. Observe the narrator narration and find the issue.

**Actual:** 
Aria label and place holder both are defined same in the code on 'Filter Dialog' edit box due to which screen reader is conveying the same information twice.

**Expected:**
Aria label and place holder both should be defined differently so that same information will not be convey twice. In Aria Label it should be 'Search available filters'.

**Observation:**
Same issue observed with NVDA.

**MAS Impacted:** [MAS1.3.1](https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={54f28d1f-a2d1-4dcd-84e1-5c9b87e8aba4})

**Attachment:**
[MAS1.3.1_Aria lable and placeholder both are same.zip](https://github.com/microsoft/BotFramework-Composer/files/4240933/MAS1.3.1_Aria.lable.and.placeholder.both.are.same.zip)

beyackle:
Is this problem now fully fixed and we can move this into Waiting for Verification, or is there more?

ashish315:
As checked now the aria label and placeholder are different but due to "placeholder narrated twice" know issue it still announces placeholder twice already raised issue for this.
So, HCL is closing this issue.

PR #2637, opened 2020-04-14
a11y: Add role project tree

lei9444:
## Description
image

## Task Item
refs #2016


lei9444:
> #2016 looks good!
> #2054 behavior is not fixed yet.
> 
> Previously, the narrator would say 'Filter dialog. Search box. Filter dialog'
> Now, the narrator says 'Search box. Search box. Filter dialog'

Hi @corinagum I have removed the aria-label in searchbox, so this PR only fix #2016 now. I will create a new PR to fix #2054

Issue #2026, opened 2020-02-20
[Accessibility][Labels or Instructions - Save as]: Screen reader does not associate the table labels while reading the table contents.


ShellySoni:
**User Experience:**
Screen reader users will not get proper information and may get difficulty in performing the task efficiently if table Labels are not associated with there respective content.

**Test Environment:**
Environment: Chromium Edge (Anaheim) + Narrator
Build Version: 2004 (OS Build 19564.1005)​​
Microsoft Edge Version: 82.0.425.0 (Official build) canary (64-bit)​
Test URL: http://localhost:3000/home

**Repro Steps:**
1. Open URL: http://localhost:3000/home
2. Navigate on homepage with tab key up to 'Save as' button and press enter on it.
3. Observe the issue on table.

**Actual Result:**
While navigating within table using ctrl+Alt+arrow keys, Screen reader does not associate the table labels i.e. Name, data modified, while reading the table contents.

**Note:**
This issue is repro on every table throughout this application.

**Expected Result:**
While navigating within table using ctrl+Alt+arrow keys, Screen reader should associate the table labels i.e. Name, data modified, while reading the table contents.

**MAS Reference:**
https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={4b95a70d-258c-474a-820e-5c8b79d5c0bf}

**Attachment:**
[MAS3.3.2_Save as.zip](https://github.com/microsoft/BotFramework-Composer/files/4229356/MAS3.3.2_Save.as.zip)
[MAS3.3.2_Save as _tablecontents.zip](https://github.com/microsoft/BotFramework-Composer/files/4229393/MAS3.3.2_Save.as._tablecontents.zip)

corinagum:
PR above is a partial fix. @liweitian could you link the bug you filed with the fabric team on this issue for tracking?

liweitian:
> PR above is a partial fix. @liweitian could you link the bug you filed with the fabric team on this issue for tracking?

https://github.com/microsoft/fluentui/issues/12701. He gives a "solution" https://codepen.io/slimemold/pen/rNOxrjd?editors=0110. But we go to row first then go inside to cell. Is this behavior acceptable? You can try it. We are discussing issue and requirements.

corinagum:
@liweitian thanks for the link. Yes, I think this is a great solution. I don't see any problems with it. :)

ashish315:
[...]

PR #2640, opened 2020-04-14
a11y: make table content associated with table label

liweitian:
## Description
make table content associated with table label
## Task Item
Refs #2026


liweitian:
> With this change, the Narrator reads out each row at once, but does not allow navigation between cell via table or arrows.
> 
> Users via Assisted Technology should be able to examine each cell, with relevant information included.
> 
> Please check out this link for an example of expected behavior: https://www.w3.org/TR/wai-aria-practices/examples/grid/dataGrids.html When using Narrator in Edge, the user is able to navigate to each cell, learn what row and column they are focused on, and also hear the name of the column to indicate what data is reading out.
> 
> Here is some extra information: https://whatsock.com/training/matrices/
> (3.2.2: Grid Role)
> image
> 
> If you have any questions on this, please let me know and I am happy to help :)

I got what you mean. For now, we are using microsoft fabric UI to implement this grid. I did not find a way to access each cell. I left an issue on their github repository to see if they can fix this issue.

PR #2643, opened 2020-04-14
a11y: add announcement region to app and announce copy/deletion

### LINKED ISSUE #2050: [Accessibility][Status Message - Bot Responses-More option]: Screen reader doesn’t announce any confirmation on activating 'delete' menu item.
**User Experience:**
Screen reader users will get confused that whether the action is performed or not if user will not get any confirmation after executing that action.

**Test Environment:**
Environment: Chromium Edge (Anaheim) + Narrator
Build Version: 2004 (OS Build 19564.1005)​​
Microsoft Edge Version: 82.0.425.0 (Official build) canary (64-bit)​
Test URL: http://localhost:3000/home

**Repro Steps:**
1. Open URL: http://localhost:3000/home
2. Select 'Bot Responses' link from left navigation.
3. Navigate within the table using tab key.
4. Press right arrow key & navigate up to 'More option' button & activate it.
5. Select 'Delete' menu item and observe the issue.

**Actual Result:**
Screen reader doesn’t announce any confirmation on activating 'delete' menu item.

**Note:**
Screen reader read twice 'button button' information.

**Expected Result:**
Screen reader should announce the confirmation on activating 'delete' menu item. 

**MAS Reference:**
https://microsoft.sharepoint.com/:w:/r/sites/accessibility/CELA/Official_Microsoft_Accessibility_Standards/MAS%204.1.3%20%E2%80%93%20Status%20Messages.docx?d=wc154fbe5702f42fbb2bc87fbb3a23d91&csf=1&e=vba1jx

**Attachment:**
[MAS4.1.3_Bot responses_Delete.zip](https://github.com/microsoft/BotFramework-Composer/files/4239080/MAS4.1.3_Bot.responses_Delete.zip)

beyackle:
## Description

This creates an aria-live region that should be reading off announcements when an item in the Bot Responses table is duplicated or deleted. For some reason, though, it isn't.

## Task Item

closes #2050

## Screenshots

The new div, present in the page DOM:
image


beyackle:
I think I see the problem. The aria-live region, since it's inside a subpage that gets dynamically loaded, isn't being seen by the screen reader. That means it has to be somewhere outside the table, but the table has to be able to stick announcements into it.

Composer/packages/client/src/pages/language-generation/table-view.tsx
```diff
@@ -34,13 +35,23 @@ const TableView: React.FC<TableViewProps> = props => {
...
 
   const [focusedIndex, setFocusedIndex] = useState(0);
 
+  const _async = new Async();
+
+  const announce = (message: string) => {
+    setMessage(message);
+    _async.setTimeout(() => {
+      setMessage(undefined);
+    }, 2000);
```
yeze322, line 52: How this setTimeout works?

beyackle: This will reset the message to "undefined" after two seconds. Because the "aria-live" region is only read when it changes, we need to clear the message out after a pause. Otherwise doing the same action multiple times in a row (like deleting two rows from the table) will only be read once.

Composer/packages/client/src/App.tsx
```diff
@@ -123,13 +123,24 @@ const bottomLinks = [
...
 
   const openedDialogId = designPageLocation.dialogId || dialogs.find(({ isRoot }) => isRoot === true)?.id || 'Main';
   return (
     <Fragment>
+      <div
+        role="region"
+        aria-live="assertive"
+        style={{
+          position: 'absolute',
+          left: '-9999px',
```
corinagum, line 139: speaking from past experience, hiding the div way off to one side of the screen will cause scrollbar problems if/when RTL is brought into the mix. My suggestion would be to put it above the screen and/or change the div to 1x1 with no overflow to avoid the possibility.

beyackle: Thanks for the suggestion. I figured there'd be a better way to do that.

PR #2645, opened 2020-04-14
fix: add click behavior to notification table

### LINKED ISSUE #2635: should support mouse click but not double click for navigation errors
Double click can navigate. 
image

beyackle:
## Description

This draft is a compromise between the requested list behavior and the way DisplayList is meant to function. It enables a user to single-click on the icon at the head of a row and be taken to the page where the notification originates. Single-clicking anywhere else in the row will still "select" it by highlighting it, and this is necessary in order to stop breaking a11y functionality (where the keyboard arrows are used to highlight a row and then Enter will proceed to the notification's origin).

## Task Item

closes #2635


beyackle:
Superseded by link-clicking behavior

Issue #2091, opened 2020-02-26
[Accessibility][Usable - Bot Responses-Edit button]: Screen reader focus moves separately on the 'Edit' toggle button and to its text while navigating using up/down arrow keys in scan mode.


ShellySoni:
**User Experience:**
Screen reader users will get difficulties to interact the UI of the page as screen reader focus moves on sign and name of controls separately which will also create confusion. 

**Test Environment:**
Environment: Chromium Edge (Anaheim) + Narrator
Build Version: 2004 (OS Build 19569.1000)​​
Microsoft Edge Version: 82.0.436.0 (Official build) canary (64-bit)​
Test URL: http://localhost:3000/home

**Repro Steps:**
1. Open URL: http://localhost:3000/home
2. "Bot Framework Composer" homepage will gets opened.
3. Select any of the previously created bot from "Recent Bots" or create one from the available Bots under 'Examples' and activate it.
Note: Better to create new "Dialog Actions" bot from example.
4. Select 'Bot Responses' from left navigation.
5. Navigate throughout the page when "Edit toggle button is enabled".

**Actual Result:**
Screen reader focus moves separately on the 'Edit' toggle button and to its text while navigating using up/down arrow keys in scan mode.

**Expected Result:**
Screen reader should move on any control only once and it should cover the control as one entity and announce whole info in once. While pressing down arrow twice focus should move on the further control.

**MAG Guidelines:**
https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/Doc.aspx?sourcedoc=%7bF98F4235-B6BD-47D3-8DA4-F28B9B6FBFA8%7d&file=Guideline%20101%20%E2%80%93%20Recommended%20Assistive%20Technology%20Tools%20for%20MAS%20Testing%20of%20Microsoft%20Digital%20Properties.docx&action=default&mobileredirect=true

**Attachment:**
[Usable_Edit Toggle Button_Scan Foucs.zip](https://github.com/microsoft/BotFramework-Composer/files/4253512/Usable_Edit.Toggle.Button_Scan.Foucs.zip)

ashish315:
@cwhitten As checked, the issue is still repro.

We checked on the below environment and the latest 'Master' build downloaded on 23rd April.
**Test Environment**
Browser: Microsoft Edge Dev {Version 84.0.488.1 (Official build) dev (64-bit)}
OS build: 2004 (19562.1000)
Screen Reader: Narrator
[...]

PR #2646, opened 2020-04-14
a11y: update about page

beyackle:
## Description

This updates the links on the About page to be a little more a11y-friendly.

## Task Item

closes #2125

PR #2661, opened 2020-04-15
a11y: Accessibility 2091

liweitian:
## Description
We changed the UI. I think there are issues left only in about page. The others are solved.
## Task Item
Ref #2091


liweitian:
Hi @corinagum, you can review this pr now. Both https://github.com/microsoft/BotFramework-Composer/pull/2646 and this pr modify the about page. And the code is similar. The issues mentioned in #2091 only left in about page now. The others are solved.
微信图片_20200416161401

PR #2672, opened 2020-04-16
fix: cannot select nested actions in Visual Editor

### LINKED ISSUE #2667: Can't select certain action nodes
After pulling the latest commit in the master branch, all the prompt nodes, as well as condition nodes, have become unselectable.

yeze322:
## Description
fix #2667 by putting svg before actions

A trick about DOM event capturing:
- elements declared **before** `<svg />` are unreachable
- elements declared **after** `<svg />` are reachable

image


yeze322:
To fix this problem thoroughly, need refinement on zindex order since we have a nested action flow.
image

Issue #2520, opened 2020-04-06
Cross train cornet cases


boydc2014:
## Describe the bug

Current cross train design only cover the happy path that assume intent trigger and only intent trigger will followed by one or more begin dialog action.  

There are a few cases don't meet this assumption and thus not working as expected. 

1. beginDialog in OnBeginDialog trigger

Example:
 Main.dialog have 2 triggers
   OnBeginDialog, in this trigger, it directly started a dialog called DialogA
   OnIntent(Name=IntentB) trigger, in this trigger, started a dialog called DialogB

So Main.lu looks like
```
# IntentB
utterance for IntentB
```

The issue is that for dialogA.lu, we should cross train IntentB into dialogA.lu's _interruption intent.

which means in the config used, we should allow trigger key as "empty", such as
```jsonc
{
  "Main.lu":
   {
     "": ["dialogA.lu"],     // we previously assume key can't be empty
     "IntentB": ["dialogB.lu"],
   }
}
```

2. no beginDialog action in OnIntent trigger

Example:
 Main.dialog have 2 triggers
   OnIntent(Name=IntentA) in this trigger,  started a dialog called DialogA
   OnIntent(Name=IntentB) in this trigger, just an SendActivity action

Main.lu will be

```
#IntentA
utterance for IntentA

#IntentB
utterance for IntentB
```

The issue is, for dialogA.lu, i think IntentB should goes into _interruption after cross train. So the config we will support is

 
```jsonc
{
  "Main.lu":
   {
     "IntentA" : ["dialogA.lu"],    
     "IntentB": [],  // we previously assume key must have a dialog
   }
}
```

3. Local intent should not be cross-trained

Here by local intent, i mean those intents created with Inputs, I think today the library will bring local intents into _Interruption.

 

Example
 Main.dialog have 3 trigge
   OnBeginDialog, in this trigger, it has a TextInput which created a local intent (Inputs.TextInput.123)
   OnIntent(Name=IntentA) in this trigger, started a dialog called DialogA
   OnIntent(Name=IntentB) in this trigger, started a dialog called DialogB

In this case, I think today's implementation will bring Inputs.TextInput.123 into dialogA.lu's _intteruption intent, which i don't think it's right, because i assume what's goes into _interruption should only be trigger phrases, so that i can be correctly dispatched again when bubbling up. Otherwise, even we hit _interruption in children and buble up, it won't match.

[...]

PR #2673, opened 2020-04-16
fix: support the cross train corner case

lei9444:
## Description
1. for beginDialog in OnBeginDialog trigger
`main.lu: {dia1.lu: ""}`

2. for OnIntent(Name=IntentB) in this trigger, just an SendActivity action
config will be like `main.lu: {'':"IntentB"}`

3. ignore the local intent

But the config don't support multiple intent triggers -> 1 dialog now, I have create a issue https://github.com/microsoft/botframework-cli/issues/713

## Task Item
refs #2520


boydc2014:
Looks good to me, @feich-ms can you also help verify the behavior here?

feich-ms:
@lei9444 issue https://github.com/microsoft/botframework-cli/pull/714 is fixed here https://github.com/microsoft/botframework-cli/pull/714. Maybe you can wait for the fix in and update the bf-lu package then.

lei9444:
@boydc2014  @feich-ms I have updated the config and package now. And test it in composer, it works well now.

cwhitten:
@lei9444 please address conflicts

Issue #2312, opened 2020-03-20
Unify Adaptive elements' delete / create / copy behavior in Composer / Form / Visual


yeze322:
## Is your feature request related to a problem? Please describe.

In Visual Editor, delete / create / copy behavior are hooked with LU / LG resources heavily;
in Form Editor & composer, if we want a unified behavior, we have to implement the same thing again by copying the code.

Now that we are hooking dialog operations with more external resources such as LG / LU or even QnA, we really need a shared lib to handle the complexity of dialog element changes.

## Describe the solution you'd like

Have a shared dialog CRUD lib and hook the CRUD handlers with LG / LU resource management API.
Share the lib across Shell / Visual / Form

yeze322:
refs #2273 #2209

yeze322:
### Adaptive dialog CRUD operations checklist

#### Visual Editor
1. **DELETE** **_action_** via '...' flyout
2. **ADD** **_action_** via '+' flyout
3. **PASTE** **_action_** via '+' flyout

#### Form Editor
1. **UPDATE** **_action_**
2. **UPDATE** **_trigger_**
3. **UPDATE** **_dialog_**

#### Shell - Toolbar
1. **COPY** **_action_** selection
2. **CUT** **_action_** selection
3. **MOVE** **_action_** selection + **ADD** target **_trigger_** via modal
4. **DELETE** **_action_** selection

#### Shell - ProjectTree
1. **ADD** **_trigger_** via 'New Trigger...' modal
2. **ADD** **_dialog_** via 'New Dialog...' modal
3. **DELETE** **_trigger_** via '...' flyout
4. **DELETE** **_dialog_** via '...' flyout

### Resource types X Operations
1. **_Dialog_** x **Add** | **Update** | **Delete**
2. **_Trigger_** x **Add** | **Update** | **Delete**
3. **_Action_** x **Add** | **Update** | **Delete** | **Copy** | **Paste** | **Move**

@lei9444 @a-b-r-o-w-n @cwhitten please review, hope I didn't miss any operation

boydc2014:
@yeze322  the list looks good, one question is i think we were talking about if it's possible to be done entirely in Shell, totally transparent to Visual and Form.

Is that option still valid or not working so that we need a shared lib called by different component?

yeze322:
@boydc2014 This item is important but not necessarily to be included in R9.
I tagged it with 'R9' weeks ago, I also prefer to postpone it to R10

[...]

PR #2674, opened 2020-04-16
[not target master] feat: LU copy api

zhixzhan:
## Description

Part of Unify Adaptive elements' delete / create / copy behavior in Composer / Form / Visual #2312 

## Task Item

refs #2312


zhixzhan:
Hi @yeze322 please take a look, does it meet copy requirements.

PR #2676, opened 2020-04-16
fix: show bot url as tooltip

### LINKED ISSUE #2530: Expose bot port in Composer
## Describe the bug

In the skill doc, we have a step to use ngrok to configure the skillHostEndpoint, since our bot port is not restricted to 3979 any more. we need to have to way to let user know what is the port of the runing bot. 

## Browser

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

boydc2014:
image

fixes #2530


cwhitten:
Dong can you provide an icon that when clicked just copies the connection string to the clipboard?

cwhitten:
see this codepen https://codepen.io/cwhitten/pen/abvNwKR

boydc2014:
Thanks @cwhitten and @a-b-r-o-w-n, here is the updated version

image

Composer/packages/client/src/components/TestController/emulatorOpenButton.tsx
```diff
@@ -4,30 +4,43 @@
...
-      }}
-      onClick={onClick}
+    <TooltipHost
+      content={
+        <Fragment>
+          URL: {botEndpoint}{' '}
+          <IconButton iconProps={{ iconName: 'copy' }} onClick={() => navigator.clipboard.writeText(botEndpoint)} />
+        </Fragment>
+      }
+      closeDelay={50000}
```
cwhitten, line 34: you have configured 50 second closeDelay? why so high?

Composer/packages/client/src/components/TestController/emulatorOpenButton.tsx
```diff
@@ -4,30 +4,43 @@
...
   return (
-    <ActionButton
-      iconProps={{
-        iconName: 'OpenInNewTab',
-      }}
-      onClick={onClick}
+    <TooltipHost
+      content={
+        <Fragment>
+          URL: {botEndpoint}{' '}
```
cwhitten, line 30: Not a fan of this URL: prefix. Can we just remove it and show botEndpoint?

Composer/packages/client/src/components/TestController/emulatorOpenButton.tsx
boydc2014: accidentally checked in debugging configuration, corrected

Composer/packages/client/src/components/TestController/emulatorOpenButton.tsx
boydc2014: Thought it would be more like the Azure style, there is a key:value, but not sure, removing them is also good, done

PR #2423, opened 2020-03-30
feat: [draft] add TelemetryLoggerMiddleware to runtime

zidaneymar:
## Description

1. fixes #1639
2. support #2240 (not sure, just follow the documentation here: https://docs.microsoft.com/en-us/azure/bot-service/bot-service-channel-directline-extension-net-bot?view=azure-bot-service-4.0)
  
image

Issue #2410, opened 2020-03-29
Inspection Middleware disabled ?


InnoraG:
## Describe the bug

Emulator inspector not able to debug Composer generated bot

## Version

4.7.0-preview-191208-1

## Browser

- [x] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Composer -> Start Bot
2. Emulator -> Debug

## Expected behavior

Should see like "/INSPECT attach db2fbbba-da64-4d74-b4b8-3d4dc054f0a9"

luhan2017:
Yes, I disabled it because the trace of BotState is already by-default enabled in the Bot and emulator. 
Could I know what is the specific feature you want for InspectionMiddleware, can "BotState trace" meet your requirement? @InnoraG

InnoraG:
Thanks for clarity.
BotState can be ok when in local dev environment but inspection middleware is great for remote debugging of production bot, I am using inspection middleware to troubleshoot use case on real user on real channel for example.
Inspection middleware is the dependency of Bot Framework Emulator -> Debug -> Start Debugging so this feature doesn't currently work on Composer generated bots.
Anyway it could be activated easily.

luhan2017:
@InnoraG you can enabled it by runtime configuration now.

Issue #2240, opened 2020-03-11
Do composer bots support interaction through Direct Line App Service Extension?


jakeatmsft:
## Is your feature request related to a problem? Please describe.
Financial service customer wants the ease of use of the Composer tool, but the flexibility to deploy into an Internal App Service.  Can this be achieved in the preview?

## Describe the solution you'd like

Support for Direct Line App Service Extension

zidaneymar:
@luhan2017 @jakeatmsft 
I follow the instructions on this doc: https://docs.microsoft.com/en-us/azure/bot-service/bot-service-channel-directline-extension-net-bot?view=azure-bot-service-4.0

But it doesn't work as expect,

the 'ib' and 'ob' is false {"k":true,"ib":false,"ob":false,"initialized":true}

> Confirm Direct Line App Extension and the Bot are Initialized
In your browser, navigate to https://<your_app_service>.azurewebsites.net/.bot. If everything is correct, the page will return this JSON content: {"k":true,"ib":true,"ob":true,"initialized":true}. This is the information you obtain when everything works correctly, where

> k determines whether Direct Line App Service Extension (ASE) can read an App Service Extension Key from its configuration.
initialized determines whether Direct Line ASE can use the App Service Extension Key to download the bot metadata from Azure Bot Service
ib determines whether Direct Line ASE can establish an inbound connection with the bot.
ob determines whether Direct Line ASE can establish an outbound connection with the bot.

Need further investigation.

carlosscastro:
@zidaneymar: Myself and @DDEfromOR can help you or continue this experiment to confirm whether directline ASE just works with composer or if something is needed.

DDEfromOR:
I'm not familiar with which version of packages Composer configures bots to use, but assuming the bot builds* after modifying the Startup.cs code in accordance to the steps in the doc linked above there shouldn't be any additional changes needed to have the bot work with DL ASE. 

For general DL ASE troubleshooting, the JSON being returned by the .bot endpoint verifies DL ASE is up and running correctly, but unable to connect to the bot. Usually I'd ask the developer to try running their bot locally and confirm it's able to start up and talk with the Emulator over HTTP, just to rule out any errors preventing the bot from running at all. It can also sometimes be useful to manually run the bot from the Kudu console within the WebApp and watch for any errors during the startup sequence. 

[...]

PR #2679, opened 2020-04-16
feat: make runtime configurable

### LINKED ISSUE #2471: Revise the runtime code we provide to users for configurability
This is a draft ticket the requirements for this work are being gathered right now.

We need to make sure the runtime code we provide to users during the eject process (#2470) as well as when we deploy a bot in its unmodified form is well designed for configurability and extension.

Some major behaviors should be configuration driven:
* location of declarative assets
* state storage type
* TelementryloggerMiddleware (#2423)
* [Dong] ShowTypingMiddleware (#1781 )
* [Dong] InspectionMiddleware (#2410)
* set up to work from a sub-folder of a bot project
* TranscriptLoggerMiddleware 
* some channel related configures such as Direct Line App Service Extension(#2240 ), Direct Speech Channel(#1638)...

We will need to make sure that, when deployed, the runtime will still be able to find the assets, and those assets will be in a logical location.

MORE TBD

zidaneymar:
## Description

fixes #2471 

## Task Item

- [x]  location of declarative assets(already could set settings.bot to other location)
- [x]  state storage type (if settings.Feature.UseCosmosDB is true, will use CosmosDbPartitionedStorage for botstate storage)
- [x]  TelementryloggerMiddleware (#2423) (add settings.UseTelementryLoggerMiddleware)
- [x] [Dong] ShowTypingMiddleware (#1781) (add settings.UseShowTypingMiddleware)
- [x] [Dong] InspectionMiddleware (#2410) (add settings.UseInspectionMiddleware)
- [x] set up to work from a sub-folder of a bot project (resolved in another PR)
- [x]  TranscriptLoggerMiddleware (add settings.UseTranscriptLoggerMiddleware)
- [x] some channel related configures such as Direct Line App Service Extension(#2240 ), Direct Speech Channel(#1638)... (according to the [doc](https://docs.microsoft.com/en-us/azure/bot-service/bot-service-channel-directline-extension-net-bot?view=azure-bot-service-4.0), I add app.UseNamedPipes by default)


BotProject/Templates/CSharp/appsettings.json
```diff
@@ -1,4 +1,11 @@
 {
+  "feature": {
```
boydc2014, line 2: can we put this into "runtime" section, i mean runtime:feature:foo

benbrown: good idea

benbrown: actually on second thought, I'm not sure I like this.

There is a difference between a setting that affects Composer and should be controlled by Composer, and a something that should come out of the normal environment.  This would be the difference, in node, of having something in the settings.json file versus in the .env file.

benbrown: just to be clear, I'm not sure we want the Composer runtime behavior configuration (runtime.enableCustom, runtime.path, runtime.command) to be mingled with things about how the runtime actually works in prod (runtime.enableFeatureX)

BotProject/Templates/CSharp/appsettings.json
```diff
@@ -1,4 +1,11 @@
...
+    "UseTelementryLoggerMiddleware": false,
+    "UseTranscriptLoggerMiddleware": false,
```
luhan2017, line 4: If user set one of these settings to true, how can he know what else he want to configure? I suggest we put the other required settings in the same place with the switch, we can just make them "". Just want to give customer enough clue on this.

BotProject/Templates/CSharp/appsettings.json
luhan2017: feature is already a group, I don't like the idea of adding another group on top of that. We can think about a more appropriate group name if necessary.

zidaneymar: > just to be clear, I'm not sure we want the Composer runtime behavior configuration (runtime.enableCustom, runtime.path, runtime.command) to be mingled with things about how the runtime actually works in prod (runtime.enableFeatureX)

so those settings will only affect the prod environment, do we need to make them part of the deploy procedure, I mean, users will only need to configure those settings if they want to deploy the bot, rather than local debug the bot. Is this right?

cwhitten:
@garypretty can you please review the telemetry changes here?

garypretty:
Currently this PR is adding TelemetryLoggerMiddleware to the adapter, this is now outdated.

Please see this PR https://github.com/microsoft/BotFramework-Composer/pull/2700/files where I have wired up the telemetry in the template using the latest method and added additional settings to BotSettings.cs. The middleware that now needs to be configurable is TelemetryInitializerMiddleware.

cc @cwhitten

zidaneymar:
> Currently this PR is adding TelemetryLoggerMiddleware to the adapter, this is now outdated.
> 
> Please see this PR https://github.com/microsoft/BotFramework-Composer/pull/2700/files where I have wired up the telemetry in the template using the latest method and added additional settings to BotSettings.cs. The middleware that now needs to be configurable is TelemetryInitializerMiddleware.
> 
> cc @cwhitten
Now merged from Gary's branch, meet some test issues for some reason..

garypretty:
> > Currently this PR is adding TelemetryLoggerMiddleware to the adapter, this is now outdated.
> > Please see this PR https://github.com/microsoft/BotFramework-Composer/pull/2700/files where I have wired up the telemetry in the template using the latest method and added additional settings to BotSettings.cs. The middleware that now needs to be configurable is TelemetryInitializerMiddleware.
> > cc @cwhitten
> > Now merged from Gary's branch, meet some test issues for some reason..

Could this be due to the fact that this branch is now targeting the SDK nightly builds, from the merge in from my working branch?

@zidaneymar thanks for merging in my branch. I will close my PR now #2700

BotProject/Templates/CSharp/appsettings.json
boydc2014: @benbrown @luhan2017 i'm not sure i get the point, are you suggesting that we keep a top-level feature property, which is weird because there will appear at top level of settings object. which indeed is not a top-level concept like "luis", like "app", it's a configuration to a specific variation of runtime (CSharp). so here is the schema i like the settings to be

runtime {
    type: customized | built-in   // this value will decide the behavior of plugin to attach runtime or not
    version: string; 
    features: string[]; 
    path?: string; //only valid when type = customized
    command?: string; // only valid when type = customized
}

what do you guys think?
         
    
    path: ..
    command: ...
}

is this OK, or we want something more

boydc2014: actually, i think we should leave the discussion of overall settings property out this, so, in here, in your own project template, this shape is OK

PR #2681, opened 2020-04-16
fix: Auto-Suggestion for fields and values for structure LG

### LINKED ISSUE #2680: LG editing: card suggestion should suggest fields and possible values
when editing structure LG, the editor should be able to do auto-suggestions for fields of current struct and possible values.

cosmicshuai:
closes #2680 

Now when user entering structure LG. The editor is able to suggest all possible fields and a value for each field.  The effect can be 

Full LG mode:
full lg card
full lg suggest

Inline LG mode:
inline lg attachment
inline lg card


cosmicshuai:
@cwhitten 
I did some modification on your original proposal. 
If the editor do the suggestion for fields and values at one time, due to all the content can only be shown on a single line (newline character has no effect in suggestion window), users will see a messy suggestion.
So I changed the suggestions into seperated fileds way. Each filed and its correspongding value will become one suggestion item. And all the fields already existed in current card won't show up. 
After the last field is finished. The editor will do the trimming automatically.

cwhitten:
ok @cosmicshuai - is this still a draft or ready for review?

cosmicshuai:
> ok @cosmicshuai - is this still a draft or ready for review?

It's ready for review.

Composer/packages/tools/language-servers/language-generation/src/utils.ts
```diff
@@ -153,6 +153,20 @@ export const cardTypes = [
...
+  SuggestionActions: 'Text | ${Some_CardAction}',
+  subtitle: 'An Example Subtitle',
+  text: 'Some text',
+  image: 'https://example.com/demo.jpg',
+  buttons: 'Text | ${Some_CardAction}',
+  contenttype: 'adaptivecard',
+  content: '${json(fromFile("../../card.json"))}',
+  name: 'An Example Name',
+};
+
```
a-b-r-o-w-n, line 169: Can these use the vs code placeholder symantics?

For example: `text: '${1:Some text}'

See the example below of a snippet that does this.
Screen Recording 2020-04-16 at 09 19 AM

cosmicshuai: I tried using snippet string with placeholder during adding suggestion of builtin functions. However, the completionItem in LSP cannot do the placeholder directly. I think it needs some transformation in monaco client side. Furthermore, the snippet ${1: some text} looks confusing in the context of we can using expression as a value here.

a-b-r-o-w-n: I see. This aren't really "snippets" in the vs code sense.

a-b-r-o-w-n:
This was likely not introduced in this change, but I am seeing the suggestions being clipped:

image

cwhitten:
@a-b-r-o-w-n is the visual editor container doing the clipping?

a-b-r-o-w-n:
@cwhitten possibly. I thought it was the form, but didn't see any element with `overflow: hidden`. I didn't check the visual editor though.

Issue #2056, opened 2020-02-23
[Accessibility][Screen Readers-Filter Dialog] : Focus should move to the chart section after selecting any dialog.


ashish315:
**User Experience** 
Users will find it difficult to navigate through the application if focus order is incorrect. Here, focus should move to the flow chart section after selecting any of the dialog otherwise user would have to navigate through all the dialogs which will be time consuming.

**Test Environment**
Browser: Microsoft Edge Canary Version: 82.0.425.0 (Official build)
OS build: 2004 (19564.1005)
Screen Reader: Narrator
URL: [Bot Framework Composer](http://localhost:3000/dialogs/Main)
[**Prerequisite Here**](https://github.com/microsoft/BotFramework-Composer)

**Repro Steps:** 
1. Open the above mentioned URL in EDGE and set the zoom at 400%.
2. Select 'Design Flow' from left navigation.
3. Now, select any dialog available in the filter dialog pane.
4. Observe the issue.

**Actual:** 
When dialog gets selected, and user presses tab key, focus moves to the next dialog available in list item instead of the flow chart section.

**Expected:**
On selecting any of the dialog, the focus should move to the flow chart of the dialog so that the user knew about the opened chart.

**Note:**
1. Same issue found throughout the page, on activating any of the node property pane gets opened but the focus remains on the chart and when user press next tab new property pane gets opened. Due to this the keyboard user will not be able to reach property pane. I suggest the role should be assigned to the node or any shortcut should be defined to reach directly to the property pane.
[MAS2.4.3_Property pane.zip](https://github.com/microsoft/BotFramework-Composer/files/4241368/MAS2.4.3_Property.pane.zip)
 

**MAS Impacted:** [MAS2.4.3](https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={0de7fbe1-ad7e-48e5-bcbb-8d986691e2b9})

**Attachment:**
[MAS2.4.3_Focus should move to the first interactive element.zip](https://github.com/microsoft/BotFramework-Composer/files/4241053/MAS2.4.3_Focus.should.move.to.the.first.interactive.element.zip)

ashish315:
[...]

PR #2685, opened 2020-04-16
a11y: Focus move to visual panel after selecting dialog

lei9444:
## Description
1. change the focus component after selecting
2. update the key navigation in project tree

bug2

## Task Item
refs #2056


Composer/packages/client/src/pages/design/styles.ts
```diff
@@ -73,6 +73,9 @@ export const visualPanel = css`
   flex: 1;
   border-right: 1px solid #c4c4c4;
   position: relative;
+  &:focus {
```
corinagum, line 76: Focus is required to have styling so that a keyboard user can always see where they are on the page.

beyackle:
I agree with all three of @corinagum 's points. Those should be addressed before we merge this.

PR #2686, opened 2020-04-16
fix: update deprecated command to new command

nt-7:
## Description
"az group deployment" command is deprecated in the most recent version of the azure cli.
This will replace it with the new "az deployment group" command.

So, it should fix from deprecated "az group deployment" to  new "az deployment group" in create.ps1.

## Results (Screenshots)
- Before
image
> Creating resource group ...
> Validating Azure deployment ...
This command is implicitly **deprecated because command group 'group deployment' is deprecated** and will be removed in a future release. **Use 'deployment group' instead**.
> Deploying Azure services (this could take a while)...
This command is implicitly **deprecated because command group 'group deployment' is deprecated** and will be removed in a future release. **Use 'deployment group' instead**.

- After
image
> Creating resource group ...
> Validating Azure deployment ...
> Deploying Azure services (this could take a while)...

## Reference
fix: use deployment group instead of group deployment
https://github.com/microsoft/accessibility-insights-service/pull/662

refs #2687


a-b-r-o-w-n:
Thanks for the PR!

This command is not in the az cli in the version I have installed (`2.0.80`). I'm not sure the last time I updated, but it has me thinking that we should do 1 of 2 things:

1. Do a version check and invoke the appropriate command.
2. Prompt users to update to a newer version of the cli. 

@boydc2014 @luhan2017 do you have any thoughts?

nt-7:
@a-b-r-o-w-n thanks for your comment and posting the issue.
Just in case, my Azure CLI version is **2.3.1** using "az --version".

boydc2014:
> Thanks for the PR!
> 
> This command is not in the az cli in the version I have installed (`2.0.80`). I'm not sure the last time I updated, but it has me thinking that we should do 1 of 2 things:
> 
> 1. Do a version check and invoke the appropriate command.
> 2. Prompt users to update to a newer version of the cli.
> 
> @boydc2014 @luhan2017 do you have any thoughts?

I will vote for 2, supporting one approach which is the latest looks good enough to me, @zidaneymar  can you help check this?

BotProject/Templates/CSharp/Scripts/create.ps1
```diff
@@ -91,7 +91,7 @@ Write-Host "> Creating resource group ..."
 
 # Deploy Azure services
 Write-Host "> Validating Azure deployment ..."
-$validation = az group deployment validate `
+$validation = az deployment group validate `
```
boydc2014, line 94: Can you also add a version guard on top of the file? maybe pick the version az-cli switch to this new syntax? @zidaneymar any suggestions on the version here?

nt-7: @boydc2014 thanks for your reply!

As follows, it seems that az CLI version (az--version) cannot be parsed programmatically...
https://github.com/Azure/azure-cli/issues/11425

Meanwhile, we can get the version programmatically with the following new features, az version command. However, it is still a preview feature.
https://github.com/Azure/azure-cli/pull/11680

Do you have any good suggestions about this?

a-b-r-o-w-n:
@nt-7 we have deprecated the powershell script in favor of a cross-platform friendly node script. Thanks for the PR and for bringing this to our attention.

PR #2693, opened 2020-04-17
feat: various skills ux additions, export functionality

### LINKED ISSUE #2320: PVA: export to zip file
We need to introduce a control in the application to zip up declarative assets (.dialog, .lu, .lg, .manifest) and drop them on the user's filesystem.

cwhitten:
todo:

- [x] add types
- [x] extract the response object from localDiskStorage.zip
- [x] support windows/unix pathing

closes #2320

Screen Shot 2020-04-16 at 10 15 46 PM
Screen Shot 2020-04-16 at 10 15 53 PM
Screen Shot 2020-04-16 at 10 16 01 PM


Composer/packages/client/src/pages/skills/skill-list.tsx
```diff
@@ -84,7 +86,10 @@ const SkillList: React.FC<ISkillListProps> = props => {
   const { actions } = useContext(StoreContext);
 
   const { skills, projectId } = props;
+
   const [editIndex, setEditIndex] = useState<number | undefined>(undefined);
+  const [isModalOpen, setModalOpen] = useState<boolean>(false);
+  const [modalOpenIndex, setModalOpenIndex] = useState<number>(0);
```
a-b-r-o-w-n, line 92: This could be combined into a single state value since they always change together.

```ts
const onViewManifest = (skill) => {
  setSelectedSkill(skill)
}

const onHideManifest = () => {
  setSelectedSkill(null)
}

isOpen={!!selectedSkill}
```

Composer/packages/client/src/pages/skills/skill-settings.tsx
```diff
@@ -0,0 +1,76 @@
...
+  }, [props.botId, props.skillHostEndpoint]);
+
+  const handleFieldChange = event => {
+    const localSettings = {
+      ...skillSettings,
+      [event.target.id]: event.target.value,
+    };
+    setSkillSettings({ ...localSettings });
+    props.setSettings(props.projectId, props.botName, { ...props.settings, ...localSettings }, undefined);
+  };
```
a-b-r-o-w-n, line 42: Question: is it necessary to keep the skill settings in state? Does `props.setSettings` not update them through the props?

It's not clear to me why have the `useState` as well as a `useEffect` with the same props.

Composer/packages/client/src/store/action/export.ts
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { ActionCreator } from '../types';
+
+import httpClient from './../../utils/httpUtil';
+import { setError } from './error';
+
+export const exportToZip: ActionCreator = async (store, { projectId }) => {
```
a-b-r-o-w-n, line 9: It would probably be useful to add a comment here explaining that this is the only way to force all browsers to download the file instead of trying to open it.

Composer/packages/extensions/adaptive-form/src/components/AdaptiveForm/index.tsx
```diff
@@ -87,7 +87,7 @@ export const AdaptiveForm: React.FC<AdaptiveFormProps> = function AdaptiveForm(p
   }, [currentDialog, focusPath, localData]);
 
   if (!localData) {
-    return <>No Data</>;
+    return null;
```
a-b-r-o-w-n, line 90: Why this? Should I add a loading indicator?

Composer/packages/extensions/adaptive-form/src/utils/getOrderedProperties.ts
```diff
@@ -3,7 +3,7 @@
 import { UIOptions, JSONSchema7 } from '@bfc/extension';
 import cloneDeep from 'lodash/cloneDeep';
 
-const globalHiddenProperties = ['$kind', '$id', '$copy', '$designer', 'id'];
+const globalHiddenProperties = ['$kind', '$id', '$copy', '$designer', 'id', 'disabled'];
```
a-b-r-o-w-n, line 6: 💡 I wonder if there should be some api to set this in a plugin fashion?

Composer/packages/server/src/models/storage/localDiskStorage.ts
```diff
@@ -3,8 +3,10 @@
 
 import fs from 'fs';
 import { promisify } from 'util';
+import p from 'path';
```
a-b-r-o-w-n, line 6: This is a nit of mine. For built-in node modules, I don't like to rename the import. It makes it less clear what the module is.

Composer/packages/server/src/models/storage/localDiskStorage.ts
```diff
@@ -74,4 +76,35 @@ export class LocalDiskStorage implements IFileStorage {
...
+      archive.directory(directory, directory.split(source)[1]);
+    });
+
+    const files = await glob('*.dialog', { cwd: source, dot: true });
+    files.forEach(file => {
+      archive.file(p.format({ dir: `${source}/`, base: `${file}` }), { name: p.basename(file) });
+    });
+
+    archive.finalize();
+  }
```
a-b-r-o-w-n, line 109: A way to decouple this from express is to accept a callback function.

```ts
async zip(source: string, out: string, cb: (archiver: Archiver) => void {
  const archive = archiver('zip');

  cb(archiver)
}

// caller with a writeable stream in scope
zip('some-directory', (archiver) => {
  archiver.on('error', (err) => {
    res.status(500).send({ error: err.message });
  });

  archiver.pipe(res);
});
```

This decouples the archiving from what to do with the archive and would allow us to have an express archiver as well as a fs archiver.

Composer/packages/server/src/models/storage/localDiskStorage.ts
```diff
@@ -74,4 +76,35 @@ export class LocalDiskStorage implements IFileStorage {
   async rename(oldPath: string, newPath: string): Promise<void> {
     return await rename(oldPath, newPath);
   }
+
+  async zip(source: string, out: string, res) {
```
a-b-r-o-w-n, line 80: `out` can be removed.

Composer/packages/extensions/adaptive-form/src/components/AdaptiveForm/index.tsx
cwhitten: Yeah, let's either style this a bit or add a proper loading affordance

Composer/packages/server/src/models/storage/localDiskStorage.ts
cwhitten: will change

Composer/packages/client/src/pages/skills/skill-settings.tsx
cwhitten: I had this here for navigating in/out of the skills page or a hard refresh on the skills page. maybe it isn't necessary

Composer/packages/extensions/adaptive-form/src/components/AdaptiveForm/index.tsx
a-b-r-o-w-n: I'll cut a ticket

Composer/packages/server/src/models/storage/localDiskStorage.ts
cwhitten: donezo

Issue #2027, opened 2020-02-20
[Accessibility][Keyboard - Flow Chart]: Focus order at 'Design Flow' Page is not correct. 


ashish315:
**User Experience** 
Users who rely on AT will face difficulty and easily confused, if the navigate through the webpage is not logical.

**Test Environment**
Browser: Microsoft Edge Canary Version: 82.0.425.0 (Official build)
OS build: 2004 (19564.1005)
Screen Reader: Narrator
URL: [Bot Framework Composer](http://localhost:3000/dialogs/Main)
[**Prerequisite Here**](https://github.com/microsoft/BotFramework-Composer)

**Repro Steps:** 
1. Open the above mentioned URL in EDGE.
2. Select 'Design Flow' from left navigation.
3. Now, navigate the whole webpage with tab.
4. Observe the issue.

**Actual:** 
Focus order is not correct in the 'Design Flow' webpage. The focus navigates to the filter pane before the left navigation pane which is not logically correct.

**Expected:**
The focus order should be correct. In this case the focus should lands on the left navigation pane then reach to the filter dialog.

**Note:**
1. Same issue observed with the '+' i.e. add node available at the flow chart.
2. Same issue observed other at other pages i.e. Home, Design Flow, User Input, Bot Response, Notification, Settings, About.

**MAS Impacted:** [MAS2.4.3](https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={0de7fbe1-ad7e-48e5-bcbb-8d986691e2b9})

**Attachment:**
[Focus order is not correct.zip](https://github.com/microsoft/BotFramework-Composer/files/4229811/Focus.order.is.not.correct.zip)

ashish315:
As checked the issue is resolved. So, HCL is closing the issue.

PR #2698, opened 2020-04-17
a11y: accessibility 2027

liweitian:
## Description
the focus will go to left Navigate pane first now. 
## Task Item
Refs #2027


liweitian:
@corinagum In #2027 Note part. "Same issue observed with the '+' i.e. add node available at the flow chart" will wait for @alanlong9278  to fix. He is refactoring that part of code. We can probably open an new issue for it and merge this pr.

Issue #2022, opened 2020-02-19
[Accessibility][Reflow - Homepage]: Two way scrolling is provided on Homepage when zoom is set to 400%.


ShellySoni:
**User Experience:**
Users having low vision will get difficulties to interact with the whole page contents by scrolling in two ways I.e. horizontally also & vertically also.

**Test Environment:**
Environment: Chromium Edge (Anaheim) + Narrator
Build Version: 2004 (OS Build 19564.1005)​​
Microsoft Edge Version: 82.0.425.0 (Official build) canary (64-bit)​
Test URL: http://localhost:3000/home

**Repro Steps:**
1. Open URL: http://localhost:3000/home
2. Set page zoom at 400% and observe the issue.

**Actual Result:**
Two way scrolling is provided on Homepage when zoom is set to 400%.

**Expected Result:**
Vertical slider should only be present when zoom is set to 400% because it will create difficulties for low vision users to interact with the whole page contents by scrolling in two ways I.e. horizontally also & vertically also.

**Note:**
Main title of the page is also gets cropped when zoom is set to 400%.

**MAS Reference:**
https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/Doc.aspx?sourcedoc=%7BE12828DF-B35C-4F4D-9289-1F932E19F8CF%7D&file=MAS%201.4.10%20%E2%80%93%20Reflow.docx&action=default&mobileredirect=true&cid=2d02c9a8-ece2-4f80-ad8c-b88ed4018176

**Attachment:**
[MAS1.4.10_Homepage.zip](https://github.com/microsoft/BotFramework-Composer/files/4225044/MAS1.4.10_Homepage.zip)

corinagum:
![MAS1 4 10_Homepage](https://user-images.githubusercontent.com/14900841/75414173-7fe53a80-58dc-11ea-8c8f-05d52d6e0534.png)

Only .png provided, pasted above; no video in .zip :)

ashish315:
@cwhitten As checked and verified on below environment, the issue is not fixed. I downloaded below Master build file last updated on 18th April.

**Test Environment**
Browser: Microsoft Edge Dev {Version 83.0.467.0 (Official build) dev (64-bit)}
OS build: 2004 (19588.1000)
Screen Reader: Narrator
URL: [Bot Framework Composer](http://localhost:3000/dialogs/Main)
[...]

PR #2710, opened 2020-04-20
a11y: accessibility issue 2022 

liweitian:
## Description
remove x scroll bar when zoom-in

## Task Item
Refs #2022 

## Screenshots


beyackle:
Tested this out and it looks good.

PR #2716, opened 2020-04-20
fix: always regenerate id when bot creation 

### LINKED ISSUE #2401: new created bot will share the same runtime if bot name is same with the deleted bot
- create a bot named "EchoBot-1"
- start bot to create the bot runtime under plugin folder
- delete the bot asset under composer directory
- re-create the bot with same name  "EchoBot-1"
image

The bot id will be same, so all the new created bot assets will be saved in old bot runtime folder if I click start bot later. This will mixed new bot assets with old bot assets. 
image
image

boydc2014:
## Description

When a bot is deleted and then created at the same path, we are reusing the same id, which is not correct because certain impact on the plugin may not be cleaned. 

Fixes #2401


Composer/packages/server/src/services/project.ts
```diff
@@ -159,6 +159,16 @@ export class BotProjectService {
     return projectId.toString();
   };
 
+  // clean project registry based on path to avoid reuseing the same id
+  public static cleanProject = async (location: LocationRef): Promise<void> => {
+    for (const key in BotProjectService.projectLocationMap) {
+      if (BotProjectService.projectLocationMap[key] === location.path) {
+        delete BotProjectService.projectLocationMap[key];
+      }
+    }
```
a-b-r-o-w-n, line 168:
```suggestion
    delete BotProjectService.projectLocationMap[location.path];
```

boydc2014: not sure i understand this, the projectLocationMap is  id => path mapping, i need to remove by value, not by key, make sense?

a-b-r-o-w-n: That makes sense.

PR #2719, opened 2020-04-20
feat: add publish target configuration schema

### LINKED ISSUE #2644: Publish plugins should provide a form for the JSON configuration

As part of the plugin registration, the plugin should provide a basic form for the JSON configuration that can be used as a placeholder in the "add new profile" flow on the publish page.

As part of the payload to `addPublishMethod`, the plugin should profile a JSON shape.

This will then be included in the call to `/publish/types`, and made available to the front end client.

The front end client should use this JSON shape to pre-fill the JSON configuration field in the "add new profile" flow.

Related to #2355

a-b-r-o-w-n:
## Description

Enables UX in the monaco editor when editing the publishing config.

## Task Item

closes #2644


Composer/packages/client/src/pages/publish/index.tsx
```diff
@@ -310,7 +311,7 @@ const Publish: React.FC<PublishPageProps> = props => {
         await actions.publishToTarget(projectId, selectedTarget, { comment: comment }, sensitiveSettings);
 
         // update the target with a lastPublished date
-        const updatedPublishTargets = settings.publishTargets.map(profile => {
+        const updatedPublishTargets = settings.publishgargets.map(profile => {
```
benbrown, line 314: type

a-b-r-o-w-n: Nice catch. I'm going to update the settings type so this will be caught.

PR #2718, opened 2020-04-20
feat: LU copy & paste & delete with unified CRUD lib

yeze322:
## Description
1. Fix LU related copy / paste / delete behaviors ('Move' function requires #2742 be merged).
2. Migrate to a unified CRUD lib to bring unified *Action / LU / LG* resources' *Create / Read / Update / Delete / Copy / Paste / Move* experience.

close #2273 
close #2406 
close #2726 
refs #2699 
refs #2312 (only applied to Visual Editor)

Changes:
1. Implemented several shared hooks as unified CRUD lib
2. Copy / Paste behavior will always invoke action constructor (#2406 )
3. Broke 'Move' operation into several atomic CRUD action
4. make Visual Editor consumes the shared CRUD lib

#### External Resource API
- `useLgApi` -> operate lgTemplate
- `useLuApi` -> operate luIntent
#### Adaptive Dialog API
- `useActionApi` -> construct / copy / destruct actions
- ~~(not included) `useTriggerApi`~~
- `useDialogApi` -> operate dialog file
#### Editing API
- `useDialogEditApi` -> copySelection; cutSelection; pasteSelection; deleteSelection; moveSelection

PR #2724, opened 2020-04-21
fix: use ref to replace the stale dialogMap

lei9444:
## Task Item
refs #2718


yeze322:
validated with Leilei, it works!

PR #2729, opened 2020-04-21
feat: Update sdk schema

### LINKED ISSUE #2541: Output format not work for Date Prompt
## Describe the bug
When I use "Prompt for date", I want to set the output format to int, but I can not make it. (tried some other built-in functions, also not work.) Is there any introduction for this, which kind of format can we use here?
image

## Browser

- [x] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

luhan2017:
## Description

Tom has a commit in the sdk side to go through and update the schema. I am bring that update in composer. (the package update is included in #2679 )

Here are two issues I found for the schema update, others looks good to me. 

1. Enum is updated to one of enum, which caused all the enum type failed to render in composer. 
    EditArray->changeType, HttpRequest->responseType. Here is one example:
    
image

2. For value with different types, we should show string editor as a default instead of empty.
    Now: 
     
image

   Suggested Fix:
   
image

@cwhitten @a-b-r-o-w-n 

## Task Item

fixes #2541


a-b-r-o-w-n:
I'll make some PR's today to address those issues.

BotProject/Templates/CSharp/Schemas/update.cmd
```diff
@@ -0,0 +1,2 @@
+cd ..
+bf dialog:merge *.csproj -o schema/sdk.schema -v -b "R9"
```
tomlm, line 2: > "R9" [](start = 51, length = 5)

NO, this should be "" 
This switch is going away, and you should not be passing data to it, except an empty string.

a-b-r-o-w-n:
@luhan2017 update: I have been talking with @tomlm about make some more adjustments to the schema based on the usage by the property editor. He is in the process of applying those updates and I will push some commits into this branch. I have marked the PR as draft in the meantime.

luhan2017:
@a-b-r-o-w-n, 
1. Begindialog.dialog is missing "type" field, so we got such error in composer, I've created this PR is SDK side https://github.com/microsoft/botbuilder-dotnet/pull/3801
image

2. Not sure which version of the schema you are using. I see Tom make more changes on the schema and the nightly build didn't come out yet. we might need to update the schema to the latest version after it is ready.

a-b-r-o-w-n:
The schema was based off of microsoft/botbuilder-dotnet#3794. That has been merged to we should take the nightly build to generate the schema.

a-b-r-o-w-n:
@luhan2017 I have fixed 1 failing e2e test. The other is due to the issue you mentioned above. Ping me when you update the schema to the nightly build and I will do another look and then we can merge.

luhan2017:
@a-b-r-o-w-n , I've updated to the very latest schema and packages, and I did some sanity test, the samples looks good now. 

image

luhan2017:
I've found another issue about the validation, I've talke with @lei9444 on this, he will submit a PR to fix the validation logic. 

image

a-b-r-o-w-n:
@luhan2017 can you address the conflicts as well as investigate the e2e test failure?

Once those are addressed, go ahead and merge.

PR #2732, opened 2020-04-21
a11y: accessibility issue 2055

liweitian:
## Description
announce filtered dialogs.
## Task Item
Refs #2055


Composer/packages/client/src/components/ProjectTree/index.tsx
```diff
@@ -181,8 +184,13 @@ export const ProjectTree: React.FC<IProjectTreeProps> = props => {
           onChange={onFilter}
           iconProps={{ iconName: 'Filter' }}
         />
+        <div
+          aria-live={'polite'}
+          aria-label={`${res.groups.length} dialogs have been found
+          ${res.groups.length > 0 ? 'Press down arrow key to navigate the search results' : ''}`}
```
beyackle, line 190: Please use formatMessage so we can get translations of this. It would also be useful to take advantage of its built-in pluralization mechanism.

Composer/packages/client/src/components/ProjectTree/index.tsx
```diff
@@ -105,7 +106,7 @@ export const ProjectTree: React.FC<IProjectTreeProps> = props => {
   const groupRef: React.RefObject<IGroupedList> = useRef(null);
   const { dialogs, dialogId, selected, openNewTriggerModal, onSelect, onDeleteTrigger, onDeleteDialog, onAdd } = props;
   const [filter, setFilter] = useState('');
-
+  const delayedSetFilter = debounce(newValue => setFilter(newValue), 1000);
```
beyackle, line 109: 👍 Good use of debounce.

Composer/packages/client/src/components/ProjectTree/index.tsx
liweitian: done

Issue #2451, opened 2020-03-31
Write some boilerplate content to the bot project via storage at create time


benbrown:
## Describe the solution you'd like

In the past, during the "new bot" process, Composer would copy from a "templates" folder into the new project -- this was historically used to copy the entire bot project template.

We should bring back this feature, allowing us to copy some arbitrary boilerplate materials into every new bot project.

* In the future, this might also include things like a README file, deployment scripts, etc.
* Might include schema extensions
* This will be written to the storage system
* This will NOT include the entire bot code project. 

I propose we re-instate the process of copying files from the templates folder into the project when creating a bot project. For now this can include a README file only.

boydc2014:
We are talking about both SDK.schema and uischema, right? 

I'm not sure do we model UI schema in one single file now?  I see multiple places called uischema.ts, and seems the previous ui.schema in server/schemas folder is not removed. I thought we can just copy that into each bot. @a-b-r-o-w-n any ideas on how can we do this?

a-b-r-o-w-n:
The form is fully schema driven, so from it's point of view, it doesn't care where that schema came from. The only thing the shell needs to ensure is that a full schema is given to the form, including any user added customizations.

For the UI schema, it just needs to adhere to the interface defined in `Composer/packages/extensions/extension/src/types/uiSchema.ts` (minus any javascript). I am currently working on removing the iframe extension host which brings us a little closer to being able to send js into the extensions.

One quick note though: although technically a single runtime can operate on multiple bot's assets, what a "bot" is is a combination of content and runtime. The glue between those is the schema. So the problem of "what if a user changes the schema and the runtime that we manage doesn't support the change" is only a problem because we want to separate the runtime and the assets. So there may need to be some detection mechanism that can catch that case and have some UX for the user to correct it (by ejecting or something).

a-b-r-o-w-n:
Oh and one more thing, I don't think there should be a default ui schema for a bot. Maybe an empty file.

boydc2014:
> Oh and one more thing, I don't think there should be a default ui schema for a bot. Maybe an empty file.

yes, i agree, i don't think we need to let the schema files in user's bot folder work as a totally `replace`, it should work as "override and extend". 

[...]

PR #2733, opened 2020-04-21
feat: azure publish plugin

### LINKED ISSUE #2461: Feature: Make Azure Deploy Plugin

### LINKED ISSUE #2757: customized botruntime publish
Context:
- create a bot and click start bot
- bot will be published to local (built-in) as : C:\Repo\Composer\plugins\localPublish\hostedBots\25038.96675632076
- then user could specify the customized runtime as : 
image
- restart bot, bot will use the customized runtime but not the hostedBot runtime

And hostedBots runtime is not synced with customized runtime.
 if customized runtime path specified, azure publish still use built-in runtime, it will cause the issue.

benbrown:
## Task Item

close #2461 
close #2757 

## Screenshots

image


Composer/packages/server/src/controllers/publisher.ts
```diff
@@ -52,7 +52,7 @@ export const PublishController = {
     // append config from client(like sensitive settings)
     const configuration = {
       name: profile.name,
-      ...JSON.parse(profile.configuration),
+      customizeConfiguration: JSON.parse(profile.configuration),
```
benbrown, line 55: this will be confusing to developers who will expect the configuration to just be in the main object. why create this sub-object?

Composer/plugins/azurePublish/src/index.ts
```diff
@@ -0,0 +1,277 @@
...
+  subscriptionID: string;
+  luisAuthoringKey?: string;
+  luisAuthoringRegion?: string;
+  appPassword: string;
+  accessToken: string;
+  graphToken: string;
+}
+
+interface PublishConfig {
+  customizeConfiguration: CreateAndDeployResources;
```
benbrown, line 23: I don't love this. What is the reason to create this subobject?

Composer/packages/lib/bot-deploy/src/botProjectDeploy.ts
```diff
@@ -300,170 +357,186 @@ export class BotProjectDeploy {
...
+        // Update deploymentSettings with the luis config
+        const settings: any = await fs.readJson(this.deploymentSettingsPath);
+        settings.luis = luisConfig;
 
-      for (let k in luisAppIds) {
-        const luisAppId = luisAppIds[k];
-        this.logger({
-          status: BotProjectDeployLoggerType.DEPLOY_INFO,
-          message: `Assigning to luis app id: ${luisAppIds}`,
+        await fs.writeJson(this.deploymentSettingsPath, settings, {
```
benbrown, line 461: why only write to the build path?

Shouldn't these new LUIS configuration items be written to the main appsettings.json file as well?

benbrown:
@VanyLaw We need to simplify this in a few ways.

## do not provision in the plugin

The provisioning part (botProjectDeploy.create) should *not* be part of the Composer publish action. This should happen outside of the plugin, via a command line tool.  We need to figure out how to make this an easy to use script -- possibly bundle it with the ejected code.

Ideally, we would be able to do something like this:

```
provisionComposer --name <bot> --location westus --environment dev --appPassword <password> --luisAuthoringKey --<key> --token <token> --subscription <subid>
```

This would perform the provision, and after completion, output a configuration profile, like:
```
Your Azure hosting environment has been created! Copy paste the following configuration into a new profile in Composer's Publishing tab.

{
accessToken: <token>,
subscription: <subid>,
environment: <environment>,
luisAuthoringKey: <key>,
luisAuthoringRegion: <region>
}
```

## make sure the plugin does not expect the declarative assets to be local

Right now, the plugin is doing a simple recursive copy of the files from the bot project's source folder.  However, this will not work for all storage providers.

Instead, you should use the project.files array to access the file content and write them to the appropriate location.

## Consider alternative mechanisms for building/deploying

Right now, the plugin copies the declarative assets from the bot project along with the c sharp code project from the template into a single TEMP folder.  Inside this temp folder, it updates an appsettings.deployment.settings file after provisioning, and then uses that for the deploy.

There are a number of flaws in this approach, though a big one is that this requires provisioning to occur immediately before deploying and will only work for one project at a time, and only work once since provisioning status is held in the plugin's memory and not serialized anywhere.

* where should information about the provisioned resources live? in the appsettings.json file? In the publishing profile?
* How will this work with an ejected runtime?
* When using an ejected runtime, should the build occur in a temp folder or should it occur in the normal project folder?


## Error handling during provisioning

Right now, a lot of the errors just return and cause no error to be raised. These critical errors should probably throw an error and cause the promise to reject so that they can be handled properly.

Composer/packages/server/src/controllers/publisher.ts
VanyLaw: Because it's easy to have duplicated name between user-inputed configuration and name(profileName). Or we changed the name into profileName?

Composer/plugins/azurePublish/src/index.ts
VanyLaw: I will remove this and rename the name field into profileName.

VanyLaw:
@benbrown Can't agree more. I did some update today.

1.  To handle the token expire more user-friendly, I split the login into a js script which return a credential. user can simpliy run this before start composer. And no need to update the profile even the token expired. The credential will be read before composer start and refresh in memory.
2.  Instead of the a single TEMP folder, I used the hask key of profile config(name, location,env,````) as a folder name to save those declarative assets and runtime. and add the field ```create``` to defined provision or not. If user use the same config to publish different bots, it will only replace the    composerDialogs folder in it and use the same config and appsettings.Deployment.json to publish.

I will keep doing the rest and move the provision out of the plugin next.

VanyLaw:
@benbrown  Hi Ben, I fix the comments about use the project.files instead of dataDir. And I know we can simply implement the pluginLoader addRuntimeTemplate function to support eject, but I don't know where is the link we can click to eject the deploy runtime?  Is it the customize runtime page is only for local publish plugin?

Composer/packages/lib/bot-deploy/src/botProjectDeploy.ts
VanyLaw: done

benbrown:
> @benbrown Hi Ben, I fix the comments about use the project.files instead of dataDir. And I know we can simply implement the pluginLoader addRuntimeTemplate function to support eject, but I don't know where is the link we can click to eject the deploy runtime? Is it the customize runtime page is only for local publish plugin?

The runtime eject page is now in master - it is part of the settings.

Composer/packages/lib/bot-deploy/src/botProjectDeploy.ts
VanyLaw: I return the result in this script and save in provision script.

Composer/plugins/azurePublish/src/index.ts
VanyLaw: done

benbrown:
@VanyLaw I updated the provisionComposer.js file to format the JSON response and also include the accessToken as well as the full credential object.

I have several pieces of feedback on the current state of things:

* As discussed, the publish profile should accept the access token and use this to do the zip deploy.  I was unable to figure out how to do this - do you know? Is it possible?
* The current code uses the WebsiteManagementClient class to *change usernames and passwords* using a predictable pattern, which seems very insecure and probably a bad practice since that deploy user might be used for other purposes.  We have to find a way to do this that does not involve changing user credentials.  Hopefully, by using the access token. 
* After some consideration, we have decided to go with the Node versions of the provisionComposer script and login script.  We will copy these into the project at create time as part of the boilerplate material -- I still need to implement that in #2451 -- but for now it would be good to get the necessary class out of the @bfc/bot-deploy package and made available locally to that provisionComposer script.  That is, we should get that stand-alone javascript package ready.
* we should add to the schema all the fields that will be created by a successful provision instead of leaving it defined as a generic object

benbrown:
@VanyLaw small update - my latest change to this branch switches the zipdeploy to using the bearer token. It works!

However, it is still using the large `profile.credentials` blob to generate this instead of using an accesstoken in the profile instead.  Can you please make this change?  This will likely require a bit of refactoring to remove expectation that this.creds exists inside the botDeploy class. As I mentioned in the previous comment, we need to split the DEPLOY piece from the PROVISION piece anyways.

benbrown:
@VanyLaw this is looking pretty good.

A few things:

- [ ] When the token has expired, the first error it hits is a 400 error from LUIS. We should handle this and present a nice error message that tells the user to use `az account get-access-token`
- [ ] In provisionComposer: There is a particular error that happens when the provision fails because your subscription is full. It comes back as an InvalidTemplateDeployment.   Perhaps we can provide a better error message, or at least print out the `details` array which may have info in it.

Composer/packages/lib/bot-deploy/src/botProjectDeploy.ts
```diff
@@ -399,13 +399,22 @@ export class BotProjectDeploy {
...
+      try {
+        // Assign a LUIS key to the endpoint of each app
+        const getAccountUri = `${luisEndpoint}/luis/api/v2.0/azureaccounts`;
+        const options = {
+          headers: { Authorization: `Bearer ${this.accessToken}`, 'Ocp-Apim-Subscription-Key': luisAuthoringKey },
+        } as rp.RequestPromiseOptions;
+        const response = await rp.get(getAccountUri, options);
+        jsonRes = JSON.parse(response);
+      } catch (err) {
+        // handle the token invalid
```
benbrown, line 420: I think we should get a bit more sophisticated here, and look at the details of the error. 

IF the error is about an expired token, throw the token-specific error

Otherwise, pass back the error details.

Composer/packages/lib/bot-deploy/src/botProjectDeploy.ts
VanyLaw: outdated.

Composer/packages/lib/bot-deploy/src/botProjectDeploy.ts
VanyLaw: I check the message include token expiry or not.

BotProject/Templates/CSharp/Scripts/provisionComposer.js
```diff
@@ -19,6 +19,12 @@ const appPassword = argv.appPassword;
 const appId = argv.appId; // MicrosoftAppId
 const luisAuthoringKey = argv.luisAuthoringKey;
 const luisAuthoringRegion = argv.luisAuthoringRegion || 'westus';
+const createLuisResource = argv.createLuisAuthoringResource == 'true' ? true : false;
+const createLuisAuthoringResource = argv.createLuisAuthoringResource == 'true' ? true : false;
```
benbrown, line 24: all of these should default to true

benbrown:
@VanyLaw My progress report today:

* I was able to repeatedly provision and publish successfully! Hooray!
* I moved the location of the provision scripts into /assets/shared.  These get copied into every new bot project on create.  I also added a README file with instructions on using it, copying the profile, and refreshing the token.
* I made a change so that the azurePublish plugin will now clean up the publishBots folder after it completes.
* I made a change so that if the bot project has a customized runtime path, it will copy the runtime from this path instead of the default template (support ejected runtime)
* I turned off writing to publishHistory.txt by default. I think it is unsafe to write this history because it could contain sensitive information. The history is kept in memory for now.

A few things I could use your help on:

- [ ] If I do NOT have LUIS configured, the first error about the invalid token comes out of the zipDeploy step.  Please catch that error and make sure it responds properly. It currently responds with a VERY VERBOSE error with a lot of binary data in it that is impossible to read.
- [ ] In the provision script, we should default to provisioning ALL the resources. Currently these default to false, but they should default to true.
- [ ] The result of the provision includes the app password field TWICE.  This should only be present once.
- [ ] If I disable certain resources during the provision, the resulting profile should exclude those keys.  Currently it includes the keys with blank values.  For example, if I do not provision cosmosDB, there should be no cosmosDB key in the config.

BotProject/Templates/CSharp/Scripts/provisionComposer.js
VanyLaw: done

Composer/packages/lib/bot-deploy/src/botProjectDeploy.ts
```diff
@@ -4,7 +4,7 @@
 import * as path from 'path';
 import * as util from 'util';
 
-// import { WebSiteManagementClient } from '@azure/arm-appservice-profile-2019-03-01-hybrid';
+import { WebSiteManagementClient } from '@azure/arm-appservice-profile-2019-03-01-hybrid';
```
benbrown, line 7: Please back out this whole commit and go back to using the token.

As discussed in teams, use the error status on the zipdeploy -- if 403, this indicates a token failure.

VanyLaw: done

PR #2734, opened 2020-04-21
a11y: use useShell for global announcements

### LINKED ISSUE #2130: [Accessibility][Object Information - Property Pane_Connect to QnA Knowledgebase]: No info is provided for the 'Enter' button to perform action.
**User Experience** 
The screen reader user will get confused if the proper information is not announced by the narrator.

**Test Environment**
OS build: 2004 (19564.1005)
Screen Reader: Narrator

**Repro Steps:** 
1. Open the above mentioned URL in EDGE and turn on narrator.
2. Navigate to already created dialog tree item and activate it. Such as- Greetings, Actions, EndDialog etc.
3. Click on "+" and select 'Access External resources'. Then select 'Connect to QnA Knowledgebase' from the available options.
4. Navigate to edit fields under "Strict Filters' having functionality to enter.
5. Observer the issue in Property Pane.

**Actual:** 
The screen reader didn't announce any information about the functionality of 'Enter' to insert the data. Same issue found for the 'Value and Name' edit field.
Narrator only narrates the placeholder whereas the edit fields have functionality that if user press 'Enter' the data  gets inserted. 

**Expected:**
The narrator should announce the functionality of 'Enter'. It should announce as 'Press Enter to insert name or press tab to move the cursor on the next interactive element'. 
Similarly for value edit field 'Press Enter to insert value or press tab to move the cursor on the next interactive element'. 

**Note:**
Issue is observed for all the Property Pane which have similar control such as 'Property Pane_Ask a question (Other)'.

**MAS Impacted:** [MAS4.2.1](https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={a12f3779-bdf1-40c2-bbcc-0e314ea94858})

**Attachment:**
[No info is provided for the 'Enter' button to perform action.zip](https://github.com/microsoft/BotFramework-Composer/files/4273495/No.info.is.provided.for.the.Enter.button.to.perform.action.zip)
Narrator should announce the functionality of Enter Key

beyackle:
## Description

This uses the useShell mechanism to make the "announce" action easier to introduce various places in the code, including inside extensions like the adaptive form editor.

## Task Item

closes #2130


Composer/packages/client/src/pages/language-generation/table-view.tsx
```diff
@@ -17,42 +17,34 @@ import formatMessage from 'format-message';
...
+import { useShell } from '../../shell';
 
 interface TableViewProps extends RouteComponentProps<{}> {
   dialogId: string;
 }
 
 const TableView: React.FC<TableViewProps> = props => {
   const { state, actions } = useContext(StoreContext);
+  const { api } = useShell();
+  const { announce } = api;
```
a-b-r-o-w-n, line 34: This doesn't make sense to use here. This component should invoke the actions directly off of context.

Composer/packages/client/src/shell/useShell.ts
```diff
@@ -20,7 +20,7 @@ const FORM_EDITOR = 'PropertyEditor';
 
 type EventSource = 'VisualEditor' | 'PropertyEditor';
 
-export function useShell(source: EventSource): { api: ShellApi; data: ShellData } {
+export function useShell(source?: EventSource): { api: ShellApi; data: ShellData } {
```
a-b-r-o-w-n, line 23: See my comment above. The `useShell` hook is only intended to bind client state / actions to extension contexts. The `source` param should be required.

Composer/packages/client/src/store/action/error.ts
```diff
@@ -12,9 +14,18 @@ export const setError: ActionCreator = ({ dispatch }, error) => {
...
+
+  setTimeout(
+    () =>
+      dispatch({
+        type: ActionTypes.SET_MESSAGE,
+        payload: undefined,
+      }),
+    2000
+  );
+}, 500);
```
a-b-r-o-w-n, line 31: To type this correctly, the debounced func should be hoisted to be a module defined function and then the action should just invoke that function. 

This is because `type debounce = <T>(fn: T, opts) => T & Cancelable` which is not `ActionCreator`. Consumers of this action should not have knowledge of its implementation (that it can be cancelled for instance).

The reason the compiler likely doesn't catch this is due to a loose typing in `BoundActionCreators`.

Composer/packages/extensions/adaptive-form/src/components/fields/ObjectArrayField.tsx
```diff
@@ -20,6 +20,12 @@ import { objectArrayField } from './styles';
 import { ArrayFieldItem } from './ArrayFieldItem';
 import { UnsupportedField } from './UnsupportedField';
 
+const END_OF_ROW_LABEL = formatMessage('press Enter to add this item or Tab to move to the next interactive element');
+
+const INSIDE_ROW_LABEL = formatMessage(
+  'press Enter to add this name and advance to the next row, or press Tab to advance to the value field'
+);
```
a-b-r-o-w-n, line 27: These need to be defined inside the function or wrapped in functions to be invoked in the component. `formatMessage` needs to be able to translate just-in-time. This should produce a warning in the console.

PR #2740, opened 2020-04-22
a11y: accessibility issue 2026

liweitian:
## Description
Update code to make lg, lu, recent bot, notification and save as file selector tables' cell associate with table name. And the cell can only be focused through keyboard operation.

## Task Item
Refs #2026


liweitian:
> I approve this in general, but I'd really like some feedback about proper table behavior from someone who depends on using a screen-reader. I feel like a good SR should be handling this automatically. That's not a fault of this code, though - just a separate thought.
@beyackle 

You can check the info https://github.com/microsoft/BotFramework-Composer/pull/2640 here. @corinagum described a lot about the right behavior.

PR #2742, opened 2020-04-22
fix: concurrency control in persistence layer

lei9444:
## Description
Now in file persistence use notify to sync with server. Sometimes may run a series of async operations. 
For move will call following actions: create dialog(will create lg and lu) -> update lu -> update lg -> update dialog -> update lu -> update lg, all these run in parallel, So the update lg may be returned earlier than create lg and the data is overwritten.

update:
1. Add a queue to store the changed file info.
2. For persistence layer: add a tag to protect current concurrent events.
3. Merge the change for the same files in the queue.  and we can run multiple tasks concurrently for different files

## Task Item
refs #2718


Composer/packages/client/src/store/persistence/Semaphor.ts
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+class Semaphore {
```
boydc2014, line 4: So you used a Semaphaore to chain all async opertation right? 

what's the granularity of this? I mean, you make all actions synced, or just a set of actions in a certain transaction?

boydc2014:
Glad to see this change, please added a little bit more on the design. It's probably OK to be simply just make every change just sequential. 

The good news is this is now only a problem in persistence layer, whichever strategy we toke is a problem of persistence, which is great.

yeze322:
A js-styled solution
```ts
class FileWriter {
  private tasks = []
  private isRunning = false;

  public schedule(files) {
    this.tasks.push(files);
    this.runAll();
  }
  private async runAll() {
    if (this.isRunning) return;
    this.isRunning= true;
    while(tasks.length) {
      const files = this.tasks.shift();
      await save(files);
    }
    this.isRunning= false;
  }
}
```

yeze322:
A more advanced solution
```ts
class FilePersistence {
  private fileQueue = new FileQueue();
  private isRunning = false;

  public persist(file) {
    this.fileQueue.push(file);
    this.runAll();
  }

  private async runAll() {
    if (this.isRunning) return;
    this.isRunning= true;
    while(!this.fileQueue.isEmpty()) {
      const file = this.fileQueue.pop();
      await save(file);
    }
    this.isRunning= false;
  }
}
```

To provide a more advanced file update strategy, definea file queue.
```ts
class FileQueue {
  private tasks = [];
  private push(file) {
    const fileIndex = this.tasks.find(x => x.id === file.id);
    if (fileIndex > -1) {
      this.tasks[fileIndex] = file;
    }
    this.tasks.push(file);
  }
  private pop() {
    return this.tasks.shift();
  }
  private popList() {
    // pop all unique file ids
    return [...];
  }
}
```

yeze322:
validated on my local PC, can fix the behavior of 'Move'

Composer/packages/client/src/store/persistence/Semaphor.ts
lei9444: The file persistence will not influence the actions. This is used to make a serial order for file operate. I have update it and now can concurrently run several task now.

boydc2014:
Thanks @yeze322 , good advices, we need to make sure the granularity is not too small, which means we should allow concurrency in between dialogs, which is reasonable of this stage. 

But i'm not a fan of "FileQueue", a generic queue is good enough, there are plenty of implementation of threadpool is the old days using the model of queue+workers. 

It would be better for queue is just queue, only work as a storage as your message, but not aware of any internal data shape. 

Make sense?

boydc2014:
Perhaps let me be concrete about my suggestions @lei9444  @yeze322 , the princinple is decouple how to store the data with how you process the data. 

1. I will suggest explicit maintain a map { key => queue } , key is a "generic tasking sharding key" which in your case happens to be file.id, instead of creating new types for your case
2. I will suggest keep the original change diff intact without merging those, when you save to store.

Composer/packages/client/src/store/persistence/FilePersistence.ts
```diff
@@ -28,6 +29,14 @@ class FilePersistence {
   private _files: { [fileName: string]: FileOperation };
   private _projectId = '';
   private _handleError = name => error => {};
+  private _queue = new FileQueue();
+  private _lock = false;
```
boydc2014, line 33: this is not a `lock`, this is just a "isRunning" or "isSaveing"  flag if there is a saving function runnning.

boydc2014: use a more accurate name would be better. lock is confusing

Composer/packages/client/src/store/persistence/FileQueue.ts
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+import { ResourceInfo, FileChangeType, IFileTask } from './types';
+
+export default class FileQueue {
```
boydc2014, line 5: FileQueue looks like we are inventing new primitive types, usually something you don't do unless you have a strong reason

boydc2014: Looks like this more like a TaskPool, at least to be a better name.

But in general, you don't need this taskPool, a 
var _taskQueues: {[key: string] : Task[] } should be good enough.

If you have more features, like a customized saving behavior or you are planning to move the executor into taskPool which is also often seen, then probably you need a taskPool class

Composer/packages/client/src/store/persistence/FilePersistence.ts
```diff
@@ -71,61 +80,34 @@ class FilePersistence {
...
+    const { name, content } = file;
+    this.attach(name);
+    await this._files[name].createFile(name, content);
+  }
 
-      if (fileType === FileExtensions.Lu) {
-        const lu = luFiles.find(d => d.id === id);
-        if (!lu) return;
-        await this.doUpdate(`${id}.lu`, lu.content);
+  public async operate() {
```
boydc2014, line 100: call this flush() will be better

Composer/packages/client/src/store/persistence/FilePersistence.ts
```diff
@@ -28,6 +29,14 @@ class FilePersistence {
   private _files: { [fileName: string]: FileOperation };
```
boydc2014, line 29: fileOperators?

boydc2014: also, i don't think you should maintain a operator per file, operator should be stateless, ie, static functions.

Composer/packages/client/src/store/persistence/FilePersistence.ts
```diff
@@ -28,6 +29,14 @@ class FilePersistence {
...
   private _projectId = '';
   private _handleError = name => error => {};
+  private _queue = new FileQueue();
```
boydc2014, line 32: _fileChanges: {[key:string]: FileChange[]} = new {}

Composer/packages/client/src/store/persistence/FilePersistence.ts
```diff
@@ -28,6 +29,14 @@ class FilePersistence {
...
+  private _lock = false;
+
+  private _operator = {
+    [FileChangeType.CREATE]: this.create.bind(this),
```
boydc2014, line 36: you don't need bind once you make it `static`

Composer/packages/client/src/store/persistence/FilePersistence.ts
boydc2014: isFlushing is the best i can think

Composer/packages/client/src/store/persistence/FilePersistence.ts
```diff
@@ -1,165 +1,213 @@
...
-  // if create dialog, the lg and lu are created together
-  private async _doCreateForOtherFile(files: ResourceInfo[], extension: string, targetId: string) {
-    await Promise.all(
-      files
-        .filter(file => getBaseName(file.id) === targetId)
-        .map(async file => await this.doUpdate(`${file.id}${extension}`, file.content))
-    );
+  private _getFileChanges(changeInfo, id: string, previousState: State, currentState: State): IFileChange[] {
+    const { dialogs, luFiles, lgFiles } = currentState;
+    let files: IFileChange[] = [];
```
boydc2014, line 179: s/files/changes

Composer/packages/client/src/store/persistence/FilePersistence.ts
```diff
@@ -1,165 +1,213 @@
...
+      return { ...changes[lastIndex], type: ChangeType.CREATE };
     }
+    return changes[lastIndex];
   }
 
-  public async notify(previousState: State, currentState: State, action: ActionType) {
-    if (action.type === ActionTypes.GET_PROJECT_SUCCESS) {
-      this.init(action.payload);
-      return;
+  private _createResourceInfo(file: any, fileType: FileExtensions, changeType: ChangeType): IFileChange {
```
boydc2014, line 146: rename to contructChange?

Composer/packages/client/src/store/persistence/FilePersistence.ts
```diff
@@ -1,165 +1,213 @@
...
+        break;
+      }
+      case FileExtensions.Lu: {
+        const lu = luFiles.find(lu => lu.id === id);
+        files.push(this._createResourceInfo(lu, fileType, changeType));
+        break;
+      }
+      case FileExtensions.Lg: {
+        const lg = lgFiles.find(lg => lg.id === id);
+        files.push(this._createResourceInfo(lg, fileType, changeType));
```
boydc2014, line 206: think about if you renaming this files to changes, this _createResourceInfo to createChange or something else, it would be simply reads as

changes.push(this.makeChange(...))

saying clearly you are construct a change object to a changes array.

Composer/packages/client/src/store/persistence/FilePersistence.ts
boydc2014: and again, is this method has to refer by this.?

Composer/packages/client/src/store/persistence/types.ts
```diff
@@ -14,8 +14,14 @@ export enum FileExtensions {
...
 
 export interface ResourceInfo {
-  id: string;
+  name: string;
   content: string;
 }
 
 export type FileErrorHandler = (error) => void;
+
+export interface IFileChange {
```
boydc2014, line 23: if you have defined a IFileChange, why you need ResourceInfo?

Composer/packages/client/src/store/persistence/FilePersistence.ts
```diff
@@ -9,157 +8,211 @@ import { ActionTypes } from '../../constants';
...
+    if (action.type === ActionTypes.GET_PROJECT_SUCCESS) {
+      this.init(action.payload);
+      return;
     }
+
+    if (!this._projectId) return;
+
+    const fileChanges: IFileChange[] = this._getFileChanges(previousState, currentState, action);
+
+    if (!fileChanges.length) return;
```
boydc2014, line 57: this line is unncessary

lei9444:
> The issue with LG files being copied to new dialogs **unsuccessfully** is **not fixed** with this change, which makes me doubt that this change is effective in what it is attempting to do. @yeze322 @lei9444
> 
> Additionally, the Undo/Redo behavior is operating at too fine of a grain. When a user action spawns multiple actions, for example by using the "Move" feature, "Undo" should back out the last action that _user_ took, not the software. When I last discussed this with @yeze322 and @boydc2014 I was under the impression that this was being worked on. An identifier needs to be generated for a user action and attached to a series of contiguous actions beneath.

Thanks @cwhitten. Currently I use a table to record the action type and get the changed files from store by specific action types. I think I should use a more general ways to find all changed files. The file persistence can get previous and current state now, I can do a diff between them to find all changed files.

For the Undo/Redo, now I wrapped the action, and it's to heavy now to handle a series of actions. I will do a refactor in another PR to fix the Undo/Redo behavior. 

Maybe we can add just a action called 'Move', and do all  calculation in reducer.( this can reduce rendering times and calculation times in persistence)

lei9444:
> > The issue with LG files being copied to new dialogs **unsuccessfully** is **not fixed** with this change, which makes me doubt that this change is effective in what it is attempting to do. @yeze322 @lei9444
> > Additionally, the Undo/Redo behavior is operating at too fine of a grain. When a user action spawns multiple actions, for example by using the "Move" feature, "Undo" should back out the last action that _user_ took, not the software. When I last discussed this with @yeze322 and @boydc2014 I was under the impression that this was being worked on. An identifier needs to be generated for a user action and attached to a series of contiguous actions beneath.
> 
> Thanks @cwhitten. Currently I use a table to record the action type and get the changed files from store by specific action types. I think I should use a more general ways to find all changed files. The file persistence can get previous and current state now, I can do a diff between them to find all changed files.
> 
> For the Undo/Redo, now I wrapped the action, and it's to heavy now to handle a series of actions. I will do a refactor in another PR to fix the Undo/Redo behavior.
> 
> Maybe we can add just a action called 'Move', and do all calculation in reducer.( this can reduce re-rendering times and calculation times in persistence)

boydc2014:
> The issue with LG files being copied to new dialogs **unsuccessfully** is **not fixed** with this change, which makes me doubt that this change is effective in what it is attempting to do. @yeze322 @lei9444
> 
> Additionally, the Undo/Redo behavior is operating at too fine of a grain. When a user action spawns multiple actions, for example by using the "Move" feature, "Undo" should back out the last action that _user_ took, not the software. When I last discussed this with @yeze322 and @boydc2014 I was under the impression that this was being worked on. An identifier needs to be generated for a user action and attached to a series of contiguous actions beneath.

By "The issue with LG files being copied to new dialogs", which issues are you referring to, are you refering to `Move` then `Undo` and `Redo`? the intention of this PR is solve the race condition of multiple lg\lu operation in Move. The fix change from "everything operation is instantly sent to server, ie concurrent"  to  "operations on one file are sequential, and operations on different file are concurrent". The implementation is creating a queue for each file to store and accumulate changes there, before flushing to server.  The granularity of concurrency in this approach is less fine-grained than use a "transaction" id to control the concurrency in transaction level, but probabaly good enough in most cases. 

I think we tested Move here, but probably not Move then Undo, Redo. Let's tested on those cases @lei9444  @yeze322, and see we need a structural change here, or we can solve it on this level.

lei9444:
> > The issue with LG files being copied to new dialogs **unsuccessfully** is **not fixed** with this change, which makes me doubt that this change is effective in what it is attempting to do. @yeze322 @lei9444
> > Additionally, the Undo/Redo behavior is operating at too fine of a grain. When a user action spawns multiple actions, for example by using the "Move" feature, "Undo" should back out the last action that _user_ took, not the software. When I last discussed this with @yeze322 and @boydc2014 I was under the impression that this was being worked on. An identifier needs to be generated for a user action and attached to a series of contiguous actions beneath.
> 
> By "The issue with LG files being copied to new dialogs", which issues are you referring to, are you refering to `Move` then `Undo` and `Redo`? the intention of this PR is solve the race condition of multiple lg\lu operation in Move. The fix change from "everything operation is instantly sent to server, ie concurrent" to "operations on one file are sequential, and operations on different file are concurrent". The implementation is creating a queue for each file to store and accumulate changes there, before flushing to server. The granularity of concurrency in this approach is less fine-grained than use a "transaction" id to control the concurrency in transaction level, but probabaly good enough in most cases.
> 
> I think we tested Move here, but probably not Move then Undo, Redo. Let's tested on those cases @lei9444 @yeze322, and see we need a structural change here, or we can solve it on this level.

@cwhitten @boydc2014 I have test copy to new dialog and find a bug #2775 ,I have fixed it in #2776. Then I test the copy and it works.

boydc2014:
> > > The issue with LG files being copied to new dialogs **unsuccessfully** is **not fixed** with this change, which makes me doubt that this change is effective in what it is attempting to do. @yeze322 @lei9444
> > > Additionally, the Undo/Redo behavior is operating at too fine of a grain. When a user action spawns multiple actions, for example by using the "Move" feature, "Undo" should back out the last action that _user_ took, not the software. When I last discussed this with @yeze322 and @boydc2014 I was under the impression that this was being worked on. An identifier needs to be generated for a user action and attached to a series of contiguous actions beneath.
> > 
> > 
> > By "The issue with LG files being copied to new dialogs", which issues are you referring to, are you refering to `Move` then `Undo` and `Redo`? the intention of this PR is solve the race condition of multiple lg\lu operation in Move. The fix change from "everything operation is instantly sent to server, ie concurrent" to "operations on one file are sequential, and operations on different file are concurrent". The implementation is creating a queue for each file to store and accumulate changes there, before flushing to server. The granularity of concurrency in this approach is less fine-grained than use a "transaction" id to control the concurrency in transaction level, but probabaly good enough in most cases.
> > I think we tested Move here, but probably not Move then Undo, Redo. Let's tested on those cases @lei9444 @yeze322, and see we need a structural change here, or we can solve it on this level.
> 
> @cwhitten @boydc2014 I have test copy to new dialog and find a bug #2775 ,I have fixed it in #2776. Then I test the copy and it works.

OK, cool, a few more words so just to be clear about the layering, 

- persistence layer's responsibility is to make sure the changes in store been correctly applied into server and disk (with the proper order and frequency). So that actions never directly talked to server
- the case of redo\undo is a problem in one higher level, how do re-do\un-do an action or a set of actions, in this layer, undo is to make sure the store state is reverted a certain time ago, redo is to make sure the store state is brought back. 

is this a clear layering to you guys @cwhitten @yeze322  @lei9444? and in this PR is fix a race condition in persistence layer when persisting data.  Redo\Undo\Copy\Delete is handled separately, make sense?

lei9444:
> > > > The issue with LG files being copied to new dialogs **unsuccessfully** is **not fixed** with this change, which makes me doubt that this change is effective in what it is attempting to do. @yeze322 @lei9444
> > > > Additionally, the Undo/Redo behavior is operating at too fine of a grain. When a user action spawns multiple actions, for example by using the "Move" feature, "Undo" should back out the last action that _user_ took, not the software. When I last discussed this with @yeze322 and @boydc2014 I was under the impression that this was being worked on. An identifier needs to be generated for a user action and attached to a series of contiguous actions beneath.
> > > 
> > > 
> > > By "The issue with LG files being copied to new dialogs", which issues are you referring to, are you refering to `Move` then `Undo` and `Redo`? the intention of this PR is solve the race condition of multiple lg\lu operation in Move. The fix change from "everything operation is instantly sent to server, ie concurrent" to "operations on one file are sequential, and operations on different file are concurrent". The implementation is creating a queue for each file to store and accumulate changes there, before flushing to server. The granularity of concurrency in this approach is less fine-grained than use a "transaction" id to control the concurrency in transaction level, but probabaly good enough in most cases.
> > > I think we tested Move here, but probably not Move then Undo, Redo. Let's tested on those cases @lei9444 @yeze322, and see we need a structural change here, or we can solve it on this level.
> > 
> > 
> > @cwhitten @boydc2014 I have test copy to new dialog and find a bug #2775 ,I have fixed it in #2776. Then I test the copy and it works.
> 
> OK, cool, a few more words so just to be clear about the layering,
> 
> * persistence layer's responsibility is to make sure the changes in store been correctly applied into server and disk (with the proper order and frequency). So that actions never directly talked to server
> * the case of redo\undo is a problem in one higher level, how do re-do\un-do an action or a set of actions, in this layer, undo is to make sure the store state is reverted a certain time ago, redo is to make sure the store state is brought back.
> 
> is this a clear layering to you guys @cwhitten @yeze322 @lei9444? and in this PR is fix a race condition in persistence layer when persisting data. Redo\Undo\Copy\Delete is handled separately, make sense?

@cwhitten @boydc2014 I have a doc about the undo/redo refactor https://microsoft.sharepoint.com/:w:/t/ConversationalAI785/EWILGaqXP1hJssSYt6YFfaEBYw30KiDdpDoozvlgu2cjRw?e=U3Ff5X
The file persistence layer can not resolve the undo behavior. I will fix it in new PR

yeze322:
about Undo/Redo, I had a discussion with @lei9444 , we can rebuild it by 2 steps:
1. Force all dialog edit events fired throught the `useDialogEditApi`
2. Let the `useDialogEditApi` maintain a snapshot stack, undo / redo picks a snapshot to replace Shell store.

cwhitten:
Guys I'm referring to this comment here: https://github.com/microsoft/BotFramework-Composer/pull/2718#issuecomment-618772593

It is not addressed with this change like @yeze322 said it would be. That is what I'm inquiring about. I understand the Undo/Redo behavior is not a component of this work.

boydc2014:
> Guys I'm referring to this comment here: [#2718 (comment)](https://github.com/microsoft/BotFramework-Composer/pull/2718#issuecomment-618772593)
> 
> It is not addressed with this change like @yeze322 said it would be. That is what I'm inquiring about. I understand the Undo/Redo behavior is not a component of this work.

I expect  this PR is a dependency for Ze's PR on #2718. This is a issue reported by Ze and Leilei think this is a general issue, so this PR is against master. Next step is for @yeze322 to merge master on #2718 after we take this in master, and complete any fixes there, make sense? @cwhitten

yeze322:
This PR should make LU/LG copy paste move work, I've validated on several cases
- LG/LU copy paste -> paste to the same trigger, paste to another trigger / dialog
- LG/LU move -> move to a new dialog

Are there some senarios I didn't covered?  @cwhitten

lei9444:
Hi, @cwhitten @boydc2014 The throttle wait time has been set to 0, and I test move more then two actions works now. But we need to reopen the lg editor perf issue now. And I will try to fix the issue.

boydc2014:
> Hi, @cwhitten @boydc2014 The throttle wait time has been set to 0, and I test move more then two actions works now. But we need to reopen the lg editor perf issue now. And I will try to fix the issue.

Just remove the throttling is cleaner than set to 0. Rethink the perf issue with a follow up on parsing.

lei9444:
> > Hi, @cwhitten @boydc2014 The throttle wait time has been set to 0, and I test move more then two actions works now. But we need to reopen the lg editor perf issue now. And I will try to fix the issue.
> 
> Just remove the throttling is cleaner than setting to 0. Rethink the perf issue with a follow up on parsing.

Now set wait time to 0 can make it work. We need to remove it, but I think I should do some investigate about this. Some code related maybe need to be clean. I want to remove it in lg editor perf PR.

lei9444:
@cwhitten @boydc2014 #2823 will fix the throttle issue in move and copy, and I will have a new PR to fix the performance issue about lg and lu editors.

cwhitten:
sounds good @lei9444

Issue #2011, opened 2020-02-19
[Accessibility][Keyboard Navigation - Homepage]: Multiple controls on homepage including ‘Examples’ section is not accessible via Keyboard.


ShellySoni:
**User Experience:** 
For keyboard users it is difficult to complete the task is if any control is not accessible via keyboard.

**Test Environment:**
Environment: Chromium Edge (Anaheim) + Narrator
Build Version: 2004 (OS Build 19564.1005)​​
Microsoft Edge Version: 82.0.425.0 (Official build) canary (64-bit)​
Test URL: http://localhost:3000/home

**Repro Steps:**
1. Open URL: http://localhost:3000/home
2. Navigate on homepage with tab key and observe the issue.

**Actual Result:**
Multiple controls on homepage i.e. ‘New’, ‘Empty Bot-1’ buttons, Video tutorial links and ‘Examples’ section  is not accessible via Keyboard.

**Expected Result:**
Every interactive control on homepage should be accessible via Keyboard.

**MAS Reference:**
https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={8492c4eb-c179-40ae-8777-cd044ed725a2}

**Attachment:**
[MAS2.1.1_Bot Framework_Homepage.zip](https://github.com/microsoft/BotFramework-Composer/files/4223506/MAS2.1.1_Bot.Framework_Homepage.zip)

ashish315:
@beyackle  As checked and verified on below environment, the highlighted controls are still not accessible with keyboard. I downloaded below Master build file last updated on 18th April.
![image](https://user-images.githubusercontent.com/60913131/79884825-5398f680-8413-11ea-846d-bba706715476.png)

**Test Environment**
Browser: Microsoft Edge Dev {Version 83.0.467.0 (Official build) dev (64-bit)}
OS build: 2004 (19588.1000)
Screen Reader: Narrator
URL: [Bot Framework Composer](http://localhost:3000/dialogs/Main)
[**Prerequisite Here**](https://github.com/microsoft/BotFramework-Composer)
![image](https://user-images.githubusercontent.com/60913131/79755879-6c33de80-8337-11ea-90cd-8dd51ac56600.png)

ashish315:
@beyackle As well as now the focus lands on 'New' and "Recent Bot" card are accessible by keyboard but they are not opened when user press ENTER or Spacebar.
Please look into the attached video: 
[...]

PR #2750, opened 2020-04-23
a11y: fix accessibility issue 2011

liweitian:
## Description
And onkeydown to new and recent bot buttons

## Task Item
Refs #2011 

## Screenshots
aaaaa

### DIFF
--- a/Composer/packages/client/src/pages/home/ItemContainer.tsx
+++ b/Composer/packages/client/src/pages/home/ItemContainer.tsx
@@ -29,6 +29,15 @@ export const ItemContainer: React.FC<ItemContainerProps> = ({
   forwardedRef,
   ...rest
 }) => {
+  const handleKeyDown = async event => {
+    if (event.key.toLowerCase() === 'enter') {
+      event.preventDefault();
+      if (onClick) {
+        await onClick();
+      }
+    }
+  };
+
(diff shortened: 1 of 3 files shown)


Composer/packages/client/src/pages/home/ItemContainer.tsx
```diff
@@ -41,11 +50,12 @@ export const ItemContainer: React.FC<ItemContainerProps> = ({
       ref={forwardedRef}
       tabIndex={0}
       {...rest}
+      onKeyDown={handleKeyDown}
     >
```
beyackle, line 54: If we use a button here instead, then we'll get the onKeyDown behavior for free (because browsers using keyboard nav are smart enough to map "enter" onto a button click), and the screen-reader will read it as an interactive button. Right now it just calls it a "group" without saying it's interactable.

liweitian: done.

cwhitten:
I didn't have access to your fork @liweitian so I've created a new PR with some minor additions I've made to this change. In the interest of time, let's take that one and close this.

cwhitten:
https://github.com/microsoft/BotFramework-Composer/pull/2830

PR #2758, opened 2020-04-24
feat: Create a bot through bfcomposer protocol

### LINKED ISSUE #2322: PVA: deep link to create new bot/skill with a predefined template
From PVA, users will need to hit a well-known endpoint to create a new bot with a template selected. This functionality should allow any template to be selected and use the same functionality. For Composer-electron, this will be establishment of a file protocol similar to `bfemulator://` with support for parameterization.

srinaath:
## Description
This PR allows us to create a bot through the bfcomposer protocol
bfcomposer://create/template/ToDoBotWithLuisSample?schemaUrl=https%3A%2F%2Fraw.githubusercontent.com%2Fmicrosoft%2Fbotframework-sdk%2Fmaster%2Fschemas%2Fcomponent%2Fcomponent.schema%26name%3DFlightBot%26description%3DFlightBot

transforms to 

http://localhost:3000/home/create/template/ToDoBotWithLuisSample?schemaUrl=https://raw.githubusercontent.com/microsoft/botframework-sdk/master/schemas/component/component.schema&name=FlightBot&description=FlightBot

## Task Item
Fixes #2322 

## Screenshots

create


Composer/packages/client/src/components/CreationFlow/index.tsx
```diff
@@ -44,8 +44,10 @@ const CreationFlow: React.FC<CreationFlowProps> = () => {
   }, [storages]);
 
   const init = async () => {
-    fetchTemplates();
-    fetchStorages();
+    (async function() {
+      await fetchStorages();
+      await fetchTemplates();
+    })();
   };
```
beyackle: Is there a reason this can't just be:
```
const init = async () => {
      await fetchStorages();
      await fetchTemplates();
  };
```
?

Is wrapping it in an extra layer of async needed somewhere?

Composer/packages/client/src/components/CreationFlow/index.tsx
```diff
@@ -0,0 +1,141 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import Path from 'path';
```
a-b-r-o-w-n, line 4: Please add TODO to remove.

Composer/packages/client/src/components/CreationFlow/index.tsx
```diff
@@ -0,0 +1,141 @@
...
+    if (storages && storages.length) {
+      const storageId = storage.id;
+      const path = storage.path;
+      const formattedPath = Path.normalize(path);
+      fetchFolderItemsByPath(storageId, formattedPath);
+    }
+  }, [storages]);
+
+  const init = async () => {
+    (async function() {
```
a-b-r-o-w-n, line 47: This doesn't make sense to me. Why do we need to double wrap this init function?

Composer/packages/client/src/components/CreationFlow/index.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+import React, { useEffect, useContext, useRef, Fragment, Suspense } from 'react';
+import { RouteComponentProps, Router, navigate } from '@reach/router';
+
+import { LoadingSpinner } from '../../components/LoadingSpinner';
+import { CreationFlowStatus } from '../../constants';
+import { StoreContext } from '../../store';
+import Home from '../../pages/home';
+
+import { CreateOptions } from './CreateOptions/index';
```
a-b-r-o-w-n, line 14:
```suggestion
import { CreateOptions } from './CreateOptions';
```

Composer/packages/client/src/components/CreationFlow/index.tsx
```diff
@@ -0,0 +1,141 @@
...
+  };
+
+  const handleCreateNext = data => {
+    navigate(`./create/template/${data}`);
+  };
+
+  return (
+    <Fragment>
+      <Home />
+      <Suspense fallback={<LoadingSpinner />}>
```
a-b-r-o-w-n, line 108: None of these components look to be dynamically imported, so this suspense seems unnecessary.

Composer/packages/client/src/components/DialogWrapper/index.tsx
```diff
@@ -1,19 +1,34 @@
...
+  });
+
+  useEffect(() => {
+    if (overrideStyles) {
+      updateWrapperStyle({
+        modal: overrideStyles.modal,
+        dialog: overrideStyles.dialog,
+      });
+    }
+  }, [overrideStyles]);
```
a-b-r-o-w-n, line 31: Is this necessary? Can we just pass the prop directly to the components?

Composer/packages/client/src/router.tsx
```diff
@@ -6,22 +6,23 @@ import { jsx } from '@emotion/core';
...
 import { About } from './pages/about';
 import { data } from './styles';
 import { NotFound } from './components/NotFound';
 import { BASEPATH } from './constants';
 import { StoreContext } from './store';
-import { resolveToBasePath } from './utils/fileUtil';
 import { LoadingSpinner } from './components/LoadingSpinner';
 
-const Home = React.lazy(() => import('./pages/home'));
+// const Home = React.lazy(() => import('./pages/home'));
```
a-b-r-o-w-n, line 17: Remove

Composer/packages/client/src/router.tsx
```diff
@@ -6,22 +6,23 @@ import { jsx } from '@emotion/core';
...
 const DesignPage = React.lazy(() => import('./pages/design'));
 const LUPage = React.lazy(() => import('./pages/language-understanding'));
 const LGPage = React.lazy(() => import('./pages/language-generation'));
 const SettingPage = React.lazy(() => import('./pages/setting'));
 const Notifications = React.lazy(() => import('./pages/notifications'));
 const Publish = React.lazy(() => import('./pages/publish'));
 const Skills = React.lazy(() => import('./pages/skills'));
+const BotCreationFlowRouter = React.lazy(() => import('./components/CreationFlow/index'));
```
a-b-r-o-w-n, line 25:
```suggestion
const BotCreationFlowRouter = React.lazy(() => import('./components/CreationFlow));
```

Composer/packages/client/src/router.tsx
```diff
@@ -50,7 +52,7 @@ const Routes = props => {
             <Publish path="publish/:targetName" />
             <Skills path="skills/*" />
           </ProjectRouter>
-          <Home path="home" />
+          <BotCreationFlowRouter path="/home/*"></BotCreationFlowRouter>
```
a-b-r-o-w-n, line 55: Could this be something like:

```ts
<Home path="home" />
<BotCreationFlowRouter path="projects" />
```

And then in `BotCreationFlowRouter`:

```ts
          <CreateOptions path="create" />
          <DefineConversation path="create/:templateId" />
          <DefineConversation path=":projectId/save" />
          <OpenProject path="/open" />
```

Composer/packages/electron-server/package.json
```diff
@@ -59,6 +59,7 @@
     "debug": "4.1.1",
     "electron-updater": "4.2.5",
     "fix-path": "^3.0.0",
+    "jest": "^25.3.0",
```
a-b-r-o-w-n, line 62: This shouldn't be required if using `@bfc/test-utils`

Composer/packages/electron-server/src/utility/url.ts
```diff
@@ -3,20 +3,28 @@
 
 import lowerCase from 'lodash/lowerCase';
 
-export const parseDeepLinkUrl = deeplinkUrl => {
-  const convertedUrl = new URL(deeplinkUrl);
+import log from './logger';
+const error = log.extend('error');
```
a-b-r-o-w-n, line 7: There might be a better name for this. `url` or something.

Composer/packages/server/src/models/bot/botProject.ts
```diff
@@ -177,6 +178,21 @@ export class BotProject {
...
 
+  public async saveSchemaToProject(schemaUrl, pathToSave) {
+    try {
+      const response = await axios({
+        method: 'get',
+        url: schemaUrl,
+        responseType: 'stream',
+      });
+      const pathToSchema = `${pathToSave}/Schemas`;
+      fs.mkdirSync(pathToSchema);
```
a-b-r-o-w-n, line 189: Should use the async method here. Convert to promise with `utils.promisify`

```ts
import { promisify } from 'util';
import fs from 'fs';

const mkDir = promisify(fs.mkDir);

await mkDir(...);
```

Composer/packages/server/src/models/bot/botProject.ts
```diff
@@ -177,6 +178,21 @@ export class BotProject {
...
+      response.data.pipe(fs.createWriteStream(`${pathToSchema}/sdk.schema`));
```
a-b-r-o-w-n, line 190: I like this!

Composer/packages/client/src/components/DialogWrapper/index.tsx
srinaath: Yeah.. I was thinking of doing it but there were a few other components that were using the DialogWrapper and all of them followed the same style which meant it was using the styles.ts from the DialogWrapper folder itself. I figured it's safer to not touch them now rather override the style through props. But I agree importing the style in the individual components and passing it through to DialogWrapper seems better

Composer/packages/client/src/router.tsx
beyackle, line 17: This line can probably just go.

Composer/packages/client/src/router.tsx
srinaath: Just leaving a note here as a follow up of our convo on Teams
Since all the modals are overlayed on top of Home component. The above flow would mean rerendering of home component when we land on the home page and then an another reload when we go into the Create step wizard flow. We decided that 
<BotCreationFlowRouter path="projects/*" />
   <CreateOptions path="create" />
          <DefineConversation path="create/:templateId" />
          <DefineConversation path=":projectId/save" />
          <OpenProject path="/open" />
<BotCreationFlowRouter path="home" />
will prevent reloading of home component

Composer/packages/client/src/components/CreationFlow/index.tsx
srinaath: Yep these will be reverted. They were made just to check if an Integration test passed with this change.

Composer/packages/electron-server/src/utility/url.ts
srinaath: Changed it electron-deeplink-url

Composer/packages/client/src/components/CreationFlow/index.tsx
srinaath: Nice catch!!

Composer/packages/client/src/components/DialogWrapper/index.tsx
```diff
@@ -1,19 +1,27 @@
...
 export const DialogWrapper: React.FC<DialogWrapperProps> = props => {
-  const { isOpen, onDismiss, title, subText, children } = props;
+  const { isOpen, onDismiss, title, subText, children, dialogType } = props;
+  const [currentStyle, setStyle] = useState(styles[dialogType]);
+
+  useEffect(() => {
+    if (dialogType) {
+      setStyle(styles[dialogType]);
+    }
+  }, [dialogType]);
```
a-b-r-o-w-n, line 24: Just curious if this state is needed? What does it buy us? Could it just be `const currentStyle = styles[dialogType]`?

Composer/packages/client/src/router.tsx
```diff
@@ -42,6 +42,7 @@ const Routes = props => {
           />
           <Redirect from="/bot/:projectId/publish" to="/bot/:projectId/publish/all" noThrow />
           <Redirect from="/" to={resolveToBasePath(BASEPATH, 'home')} noThrow />
+          {/* <Redirect from="/bot/:projectId" to="/bot/:projectId/dialogs/Main" noThrow /> */}
```
a-b-r-o-w-n, line 45: can this be removed?

Composer/packages/client/src/pages/home/index.tsx
```diff
@@ -70,9 +69,10 @@ const Home: React.FC<RouteComponentProps> = () => {
     }
   };
 
-  const onClickTemplate = id => {
-    saveTemplateId(id);
+  const onClickTemplate = async (id: string) => {
+    await saveTemplateId(id);
```
a-b-r-o-w-n, line 73: I don't think `saveTemplateId` is awaitable.

Composer/packages/client/src/components/DialogWrapper/index.tsx
srinaath: Doesn't buy us anything now. This can be changed.

Composer/packages/client/src/components/CreationFlow/DefineConversation/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+      }
+
+      if (name) {
+        updatedFormData.name = name as string;
+      } else {
+        updatedFormData.name = getDefaultName();
+      }
+      setFormData(updatedFormData);
+    }
+  }, [templateId]);
```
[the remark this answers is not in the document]
srinaath: Template ID actually can change at this component. The reason being the component is exposed in a router path and can be hit directly. So I have to get the template id passed as props through the router and set it inside this component

Composer/cypress/support/commands.ts
```diff
@@ -24,6 +24,15 @@ Cypress.Commands.add('visitPage', page => {
...
 });
 
+Cypress.Commands.add('enterTextAndSubmit', (textElement: string, text: string, submitBtn?: string) => {
+  cy.findByTestId(textElement)
+    .clear()
+    .type(text);
+  if (submitBtn) {
+    cy.findByTestId(submitBtn).click();
+  }
+});
```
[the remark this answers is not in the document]
srinaath: I see. the reasoning behind it was actually get it to a review and see how u feel about it. Basically, it had its utility in a few places and I thought this could be a nice utility to type text and submit. But, u have a lot more expertise on this project and if you feel it does not add value i can remove it and let enter submit the components

Composer/packages/client/src/components/CreationFlow/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+        break;
+      case CreationFlowStatus.SAVEAS:
+        handleSaveAs(formData);
+        break;
+
+      default:
+        handleCreateNew(formData);
+    }
+    handleDismiss();
+  };
```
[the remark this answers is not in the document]
srinaath: I feel this can be refactored. But I see quite a few usages of these in other components like toolbar, home, reducers etc. Some of those components are not completely unit tested for me to confidently remove creationStatus from the reducer and all usages of it everywhere. For ex if the creation status is being used for something else. It seems like a refactor that can hold after R9. what do u think?

Composer/cypress/support/commands.ts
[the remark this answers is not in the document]
a-b-r-o-w-n: I don't think it abstracts away enough logic to warrant a new method. Also the name and its body are in conflict as it doesn't always submit, only if you pass in a test id to the submit button.

Composer/packages/client/src/components/CreationFlow/DefineConversation/index.tsx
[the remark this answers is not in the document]
a-b-r-o-w-n: In what scenario does `templateId` change across renders of this component? I just don't see it yet.

Composer/packages/client/src/components/CreationFlow/index.tsx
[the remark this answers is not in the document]
a-b-r-o-w-n: I'm not saying we need to refactor those completly away, but we don't need this switch statement in here. We can directly pass `handleCreateNew` or `handleSaveAs` to the components that require them.

PR #2761, opened 2020-04-24
fix: fix luis appid not found error

### LINKED ISSUE #2751: dotnet run - “” is not a valid LUIS application id.
## Describe the bug

If I create a bot with Composer (Todo with LUIS) and run in from Composer the testing in the emulator works fine. But if I run the bot from the CLI using dotnet run, I always get this message when it should detect the intent with LUIS:

“” is not a valid LUIS application id.

image

## Version

test

## Browser

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [x] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Create bot
2. Open cmd and run dotnet run

## Expected behavior

The bot should be using LUIS to detect the intent correctly.

zidaneymar:
## Task Item
fixes #2751 

1. After the runtime structure changes, need to update the settings path.
2. builder.UseLuisSettings() don't search the settings in parent folder so need to add "root": "../" to specify the location


BotProject/Templates/CSharp/Program.cs
```diff
@@ -26,8 +26,9 @@ public static void Main(string[] args)
             {
                 var env = hostingContext.HostingEnvironment;
 
-                builder.AddJsonFile($"ComposerDialogs/settings/appsettings.json", optional: true, reloadOnChange: true)
-                    .AddJsonFile("appsetting.json", optional: true, reloadOnChange: true)
+                var configFile = Path.GetFullPath(Path.Combine(@"../settings/appsettings.json"));
```
luhan2017, line 29: We should not do this, we need to get the bot path first, and then search in botfolder/settings/appsettings.json, otherwise it will not working for the deployment scenario.

luhan2017: And, we should consider the deployment scenario, if it is deployment, we need to get the bot and root location from appsettings.deployment.json.

zidaneymar: > We should not do this, we need to get the bot path first, and then search in botfolder/settings/appsettings.json, otherwise it will not working for the deployment scenario.

fixed

zidaneymar: > And, we should consider the deployment scenario, if it is deployment, we need to get the bot and root location from appsettings.deployment.json.

currently the deployment bot's path is hard-coded, do we need to make that configurable, I mean, move the 'bot': '../' to appsettings.development.json layer, and add 'bot': 'composerDialogs' to appsettings.deployment.json layer.

luhan2017: yes, we should make them configurable

PR #2772, opened 2020-04-24
fix: Server will now dynamically mount on an open port.

### LINKED ISSUE #2770: Allow Composer to dynamically find a port to mount the server to
Currently, Composer is set to run on port 5000, which can cause it to fail if the user's machine is using that port for anything else.

Composer should dynamically find an open port and mount the server using that port.

tonyanziano:
## Description

> Currently, Composer is set to run on port 5000, which can cause it to fail if the user's machine is using that port for anything else.

This PR enables Composer to dynamically find an open port before starting the server.

**NOTE: This has been changed to only apply to the production environment. The development environment will continue to mount on `process.env.PORT` or 5000.**

This works in both `server` and `electron-server`

## Task Item

closes #2770 

## Screenshots

**Dynamically finding a port**

image

**Starting Composer again after already having something running on 8000**

image


Composer/packages/electron-server/src/main.ts
```diff
@@ -90,6 +92,9 @@ async function loadServer() {
   log('Created app data directory.');
 
   log('Starting server...');
+  const availablePort = await getPortPromise();
```
a-b-r-o-w-n, line 95: Can this prefer 5000?

Composer/packages/server/src/init.ts
```diff
@@ -1,8 +1,14 @@
...
+import { getPortPromise } from 'portfinder';
+
 import { start } from './server';
 
-start().catch(err => {
+(async () => {
+  const availablePort = await getPortPromise();
+  process.env.FALLBACK_PORT = availablePort.toString();
+  await start();
+})().catch(err => {
```
a-b-r-o-w-n, line 12: Why does this need to be duplicated in electron-server?

tonyanziano: The way electron-server's startup script is structured right now requires it to know the port ahead of time before calling `/server/`'s `start()` method, but I just looked and think it can be rearranged to get rid of this duplication.

I'll ping you when I fix it.

Composer/packages/electron-server/src/main.ts
tonyanziano: yup!

tonyanziano:
Thanks for taking a look @beyackle !

@a-b-r-o-w-n pointed out that this breaks the webpack dev server proxy that points to the server, and so I'm currently looking into how to fix that.

This PR will remain in draft until then.

Issue #2775, opened 2020-04-26
Can't navigate to a new trigger under the new created dialog 


lei9444:
## Describe the bug
![bug](https://user-images.githubusercontent.com/39758135/80297330-cb20aa00-87b4-11ea-9dc1-865d31537a88.gif)

## Browser

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [x] Edge

## OS

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

## To Reproduce

1. Go to 'design page'
2. Click on 'new dialog'
3. click the trigger under the new dialog

PR #2776, opened 2020-04-26
fix: the action in new dialog should be an Array

lei9444:
## Description
The actions need to be set as an array

## Task Item
refs #2775


yeze322:
additional context:
1. `createDialog` started to receive the action seed after we shipped the 'Move' feature.
2. Semantically, I think `createDialog` requires a dialog seed (instead of action seed), so it would be easy to make a mistake passing `{}` to the handler.
3. About 'actions':
    - If the `actions` field of a dialog is unset, action API can generate correct data;
    - however, when the actions set to `{}` which is unexpected, the result will be wrong.

Actions I want to take:
1. make `createDialog` an action in `useDialogApi`
2. let toolbar use the action from `useDialogApi` hook as a follow of the 'unified CRUD lib' item

cwhitten:
Thank you @lei9444 - this was my mistake! Better typings would have caught this.

PR #2786, opened 2020-04-27
fix: unexpected file changes when editing

### LINKED ISSUE #2748: unexpected error when I set the property
## Describe the bug
When I edit the value in the set property, if I type in quickly, the result will be presented very slowly. And often throw an error.
Error

image

and some lg file will be deleted automatically.

image

## Version

"3a7dbce5c902d"

## Browser

- [x] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

### LINKED ISSUE #2793: When delete a dialog, server throw an error
## Browser

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

## To Reproduce

1. Go to 'Design page'
2. Click on 'Delete a dialog' -> delete a dialog
4. See  Server error

## Screenshots
image

image

zhixzhan:
## Description

Issue  #2748 is when editing, the concurrency bug in `reformat` methods
After some refactor by plugins PR #2023,  getProjectById is called at very high frequency. And every getProjectById() --> Project.index() , and inside it some `reformat` methods  executed would travels all files do reformat. When called in concurrency it produced a write bug.

By re-using an indexed project instead of call index() everytime we could avoid this type problem and improve performance.

Issue #2793 is when delete a dialog, the concurrency bug in `removeFile` methods.
`this.files` is not updated immediately.

## Task Item

close #2748 
close #2793


zhixzhan:
Nowadays, client handle most changes in local, so after first load, we also do not need server index anymore.

Composer/packages/server/__tests__/controllers/project.test.ts
```diff
@@ -159,15 +159,15 @@ describe('dialog operation', () => {
     const mockReq = {
       params: { projectId },
       query: {},
-      body: { name: 'test.dialog', content: '' },
```
zhixzhan, line 162: test is the bot name, don't allowed delete it's root dialog: test.dialog

Composer/packages/server/src/controllers/project.ts
```diff
@@ -50,7 +50,6 @@ async function createProject(req: Request, res: Response) {
     const currentProject = await BotProjectService.getProjectById(id, user);
     if (currentProject !== undefined) {
       await currentProject.updateBotInfo(name, description);
-      await currentProject.init();
```
zhixzhan, line 53: project.init() is already called on getProjectById() which is the single place of every project created. so here don't need call again,

Composer/packages/server/src/models/bot/botProject.ts
```diff
@@ -230,7 +230,7 @@ export class BotProject {
   };
 
   public deleteFile = async (name: string) => {
-    if (Path.resolve(name) === 'Main') {
+    if (Path.basename(name, '.dialog') === this.name) {
```
zhixzhan, line 233: now root dialog is not Main.dialog anymore

Composer/packages/server/src/models/bot/botProject.ts
```diff
@@ -406,10 +406,10 @@ export class BotProject {
     if (index === -1) {
       throw new Error(`no such file at ${relativePath}`);
     }
+    this.files.splice(index, 1);
 
     const absolutePath = `${this.dir}/${relativePath}`;
     await this.fileStorage.removeFile(absolutePath);
-    this.files.splice(index, 1);
```
zhixzhan, line 412: `this.files` can be different after `await` when concurrency requests come in. this is the bug code cause `unexpected file delete/not-delete`

Composer/packages/server/src/services/project.ts
```diff
@@ -88,9 +89,13 @@ export class BotProjectService {
       'turn.repeatedIds',
       'turn.activityProcessed',
     ];
+    const projectVariables = BotProjectService.getIndexedProjectById(projectId)
+      ?.files.filter(file => file.name.endsWith('.dialog'))
+      .map(({ content }) => extractMemoryPaths(content));
+
     const userDefined: string[] =
-      BotProjectService.currentBotProjects[projectId]?.dialogs.reduce((result: string[], dialog) => {
```
zhixzhan, line 92: `.dialogs ` is out of date, cause we don't index in server anymore.

Composer/packages/server/src/services/project.ts
```diff
@@ -180,9 +185,18 @@ export class BotProjectService {
...
+    // use indexed project
+    const indexedCurrentProject = BotProjectService.currentBotProjects.find(({ id }) => id === projectId);
+    if (indexedCurrentProject) return indexedCurrentProject;
+  }
+
+  public static getProjectById = async (projectId: string, user?: UserIdentity): Promise<BotProject> => {
     BotProjectService.initialize();
 
+    const cachedProject = BotProjectService.getIndexedProjectById(projectId);
+    if (cachedProject) return cachedProject;
```
zhixzhan, line 198: use indexed project

zhixzhan:
> Nowadays, client handle most changes in local, so after first load, we also do not need server index anymore.

#### Note for a special case.

If some changes are made in current opened bot , not via Composer, e.g 

delete a dialog from local file system, 
write file via my text editor, 

the Composer wouldn't  receive this changes. Cause this PR let Composer server use indexed project.

To make Composer sync with external file change, user need to **re-open** current project force trigger a server index.

Issue #2563, opened 2020-04-08
Manage schema and sample through runtime plugin


boydc2014:
## Is your feature request related to a problem? Please describe.

We decoupled runtime code from Composer core with the plugin architecture, that leaves us the questions of how should we manage schemas files for bot. 

Do we attach that at the beginning of creation bot or do we leave it later when ejecting the runtime?

## Describe the solution you'd like
Given schema is more close to runtime, and attaching schema at very beginning don't really support user to do customization without code ejected there. So it make sense to put them together when ejecting the bot. 

Then it also make sense to let the runtime plugin to provide schema since it manage the code, so we need a way to let runtime plugin to provide the schema. 

The design is 
1. Put a "version" field in bot settings
2. Plugin system allow the extension to register a lookup method to look up schema based on version
3. Server will call this method and get the schema and merged the schema found in bot folder and sent to client. 

the version can be just "latest" or "lastest_csharp"  before we version it very precisely. 

### Samples
A sample may not have to bind into one specific runtime version, so it make sense to support adding sample for multiple versions. 

For example, i would vision a sample extension API like this. 

```js
Composer.AddSample(sample: SampleDef)

Interface SampleDef
{
  id: string
  description: string
  path: string
  type: basic | advanced
  tags: []
  runtimes: string[]   // a list of versions of runtime this sample supports
}

```

benbrown:
What would the `runtimes` field be holding?

Issue #2470, opened 2020-04-01
Create Runtime Settings page


benbrown:
## Is your feature request related to a problem? Please describe.

By default, new projects created in Composer will not contain the code needed to run the project. This will be provided automatically by the localPublish plugin that manages the local run/test cycle.

However, for bots that have a customized runtime or for existing projects that use Composer to edit declarative dialogs, developers need a way to manage that behavior.

## Describe the solution you'd like

Composer needs a settings page where the location and behavior of the runtime can be configured.

# Runtime behavior setting

This setting will control how the localPublish plugin handles the runtime for local testing.

- [x] By default, new projects will ship with these options turned off, or in a default/disabled mode.
- [x] A developer should be able to enable the custom runtime. This should be a toggle switch that, when enabled, allows the user to modify 2 related settings: the path to the runtime, and the command used to start the runtime.
- [ ] The path should use the local filesystem browser widget to allow a user to browse for a folder and select it in the UI.
- [x] The start command should be a freeform text field.

If the custom runtime is enabled, the localPublish plugin will use these fields instead of using the "built in" runtime.  See #2450 for details.

# Eject runtime feature

As stated above, the default behavior is that new projects do not include the bot project code. However, we still need to provide this code to some developers at some point so they can extend or customize it.

The eject feature should:

- [x] Include a brief description of what will be provided
- [x] Allow user to pick from 1 or more runtime templates. (C#, Javascript, Python, etc)
- [x] Composer automatically "migrates" the current project to include the code AND the assets. the runtime will be written to a sub-folder of the bot project, and will be automatically configured to use the local bot assets.

boydc2014:
A few points

1. do we want allow user to configure **the path to the runtime**

I like that, but i just want to clarify the implication of this, because by saying that, we are saying we don't assume the bot runtime live with the bot assets.  

If that's true, when we are selecting a bot to open, we can't select the runtime+declarative folder a whole bot, right, because there is not way to judge the whole folder is a bot or not.

So for us, we either say
a. any bot opened by composer must open from the declarative folder, in this approach, Composer core is still a editor only cares about declartive bits, there wouldn't be two modes. Actually i like this way. 
b. if we want treat the runtime+declartive assets a whole bot to open\view in Composer,  like the old days, we should not allow user to specify "a location to runtime" and the path to declarative assets is no going to be configurable. 

I like a), because, supporting two modes (normal mode and ejected mode) sounds like we are looking for trouble.

2. eject feature
[...]

Issue #2760, opened 2020-04-24
load schema from the path of 'Use custom runtime'


yeze322:
## Describe the bug
As a follow up of #2572, after pointed the runtime to a customized position
![image](https://user-images.githubusercontent.com/8528761/80192740-3d28b000-864a-11ea-891a-93eb1a657dda.png)
Composer should also load the schema files under that runtime folder.

## Browser

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

cwhitten:
@yeze322  i'm closing this. this ticket is already tracked and represented.

PR #2787, opened 2020-04-27
fix: eject schema with runtime

boydc2014:
Pop out sdk.schema from runtime to bot/schemas/sdk.schema when ejection, this will complete the E2E flow of customized schema, in which user eject then modify runtime, and update schema. 

Also fix the file loading after ejection to not looking into runtime folder. 

refs #2470 
refs #2563


Composer/packages/server/src/models/bot/botProject.ts
```diff
@@ -434,7 +435,7 @@ export class BotProject {
       // load only from the data dir, otherwise may get "build" versions from
       // deployment process
       const root = this.dataDir;
-      const paths = await this.fileStorage.glob([pattern, '!(generated/**)'], root);
+      const paths = await this.fileStorage.glob([pattern, '!(generated/**)', '!(runtime/**)'], root);
```
benbrown, line 438: good catch

Composer/plugins/localPublish/src/index.ts
```diff
@@ -359,9 +359,12 @@ export default async (composer: any): Promise<void> => {
     eject: async (project: any, localDisk: IFileStorage) => {
       const sourcePath = path.resolve(__dirname, '../../../../BotProject/Templates/CSharp');
       const destPath = path.join(project.dir, 'runtime');
+      const schemaSrcPath = path.join(sourcePath, 'Schemas');
+      const schemaDstPath = path.join(project.dir, 'schemas');
       if (!(await project.fileStorage.exists(destPath))) {
         // used to read bot project template from source (bundled in plugin)
         await copyDir(sourcePath, localDisk, destPath, project.fileStorage);
+        await copyDir(schemaSrcPath, localDisk, schemaDstPath, project.fileStorage);
```
benbrown, line 367: If Schemas/ is already a subfolder of sourcePath, why do we have to copy it separately?

benbrown: oh I see it is moving it up to the root folder not the runtime folder.

got it.

boydc2014:
> I think that `_getSchemas` should also be updated with the lowercase `schemas` directory. Or maybe updated to load *.schema again.

updated, and just FYI, actually the folder cases seems to be not case-sensitive in Windows, i tested both lowercase and uppercase works in Windows

yeze322:
I suppose this PR can solve #2760 load schema from the path of 'Use custom runtime' ?

Issue #2794, opened 2020-04-28
cross train unexpected result


xhr0804:
## Describe the bug
When I use cross train, I got unexpected result in intent _Interruption in 2 scenarios.
1. In main dialog, I have two trigger A and B, A begins a dialog DA,  if B begins a dialog DB(with None recognizer), the _Interruption in A will not contains the utterances in trigger B. If B doesn't begin a dialog, _Interruption will not contains the utterances either.
2. In main dialog, I have a trigger intent BookMeetingRoom which will begins a dialog named BookMeetingRoomDialog. BookMeetingRoomDialog begins a dialog named SetTitle, and it also contains a trigger intent GetInput, the utterances in GetInput should be in the _Interruption of SetTitle, but it didn't.
![image](https://user-images.githubusercontent.com/44257569/80452131-2f2fa380-8958-11ea-9277-70e7fb327555.png)
![image](https://user-images.githubusercontent.com/44257569/80452160-3e165600-8958-11ea-8a73-be9cf293024a.png)
![image](https://user-images.githubusercontent.com/44257569/80452194-4cfd0880-8958-11ea-8c6d-bcde211bb560.png)

## Browser

- [x] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

lei9444:
Hi, @xhr0804 I have updated the config now and tested it. Could help to check if it can fix your issue?

cwhitten:
addressed via https://github.com/microsoft/BotFramework-Composer/pull/2796

PR #2796, opened 2020-04-28
fix: crosstrain for dialog without lu file

lei9444:
## Description
If a trigger start with OnIntent and there is no action or begin a dialog without lu
the config will be like   'tirgger_Intent':''

## Task Item
refs #2794


xhr0804:
Verified the change and the bug is fixed

PR #2800, opened 2020-04-28
fix: telemetry logger middleware

### LINKED ISSUE #2789: Runtime middleware configuration
Some feedback wasn't picked up before the PR that made the runtime configurable got merged.  https://github.com/microsoft/BotFramework-Composer/pull/2679#event-3276572882

The changes mentioned in the comments are...

 - **The TelemetryLoggerInitializer middleware should always be added to the adapter pipeline** (i.e. not have it added based on the UseTelementryLoggerMiddleware setting).  This setting should be removed. Then the telemetry settings (logActivities and logPersonalInformation) are used to control if further telemetry is logged, using the TelemetryLoggerMiddleware internally.  Telemetry capture is only ever enabled if a developer sets their App Insights instrumentation key and so all we do with this setting right now is mean they have to enable two things to turn on telemetry (it also can give the false impression that all telemetry is off if you have this set to false, which is not true, telemetry will just be incomplete).

 - **Right now the UseInspectionMiddleware setting is adding the TelemetryInitializerMiddleware, instead of adding the InspectionMiddleware**.  https://github.com/microsoft/BotFramework-Composer/blob/dff6dec2f5a7046200f0d0240129be85be2d7e40/BotProject/Templates/CSharp/Startup.cs#L57-L63
This needs to be fixed and then the TelemetryInitializerMiddleware needs to be added to the pipeline too (without a conditional setting as per above).

zidaneymar:
## Description

1. Remove UseTelemetryLoggerMiddleware section in appsettings.json, use telemetry section to control 
2. Add feature section to compose default settings page. 

## Task Item
fixes #2789

### DIFF
--- a/BotProject/Templates/CSharp/appsettings.json
+++ b/BotProject/Templates/CSharp/appsettings.json
@@ -1,6 +1,5 @@
 {
   "feature": {
-    "UseTelementryLoggerMiddleware": false,
     "UseTranscriptLoggerMiddleware": false,
     "UseShowTypingMiddleware": false,
     "UseInspectionMiddleware": false,
@@ -18,7 +17,7 @@
   },
   "telemetry": {
     "logPersonalInformation": false,
-    "logActivities": true
+    "logActivities": false
   },
   "blobStorage": {
     "connectionString": "",
--- a/BotProject/Templates/CSharp/appsettings.json
+++ b/BotProject/Templates/CSharp/appsettings.json
@@ -3,7 +3,7 @@
     "UseTranscriptLoggerMiddleware": false,
     "UseShowTypingMiddleware": false,
     "UseInspectionMiddleware": false,
-    "UseCosmosDb": false
+    "UseCosmosDbPersistentStorage": false
   },
   "microsoftAppId": "",
   "cosmosDb": {
@@ -17,7 +17,7 @@
   },
   "telemetry": {
     "logPersonalInformation": false,
-    "logActivities": false
+    "logActivities": true
   },
   "blobStorage": {
     "connectionString": "",
--- a/Composer/packages/server/src/models/settings/defaultSettingManager.ts
+++ b/Composer/packages/server/src/models/settings/defaultSettingManager.ts
@@ -19,6 +19,12 @@ export class DefaultSettingManager extends FileSettingManager {
 
   protected createDefaultSettings = (): any => {
     return {
+      feature: {
+        UseTranscriptLoggerMiddleware: false,
+        UseShowTypingMiddleware: false,
+        UseInspectionMiddleware: false,
+        UseCosmosDb: false,
+      },
       MicrosoftAppPassword: '',
       MicrosoftAppId: '',
       luis: {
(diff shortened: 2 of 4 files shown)


zidaneymar:
@luhan2017 Hi Lu, could you please review this pr about the settings part? Do we need to keep the UseTelemetryLoggerSettings or just use the 'telemetry' section to maintain.

BotProject/Templates/CSharp/appsettings.json
```diff
@@ -18,7 +17,7 @@
   },
   "telemetry": {
     "logPersonalInformation": false,
-    "logActivities": true
+    "logActivities": false
```
garypretty, line 20: I think we should set this to true.  This is the default in the SDK and activities will only be logged once someone populates their InstrumentationKey.  We already set LogPersonalInformation to false by default which I think is enough.

Composer/packages/server/src/models/settings/defaultSettingManager.ts
```diff
@@ -37,6 +43,7 @@ export class DefaultSettingManager extends FileSettingManager {
       },
       telemetry: {
         logPersonalInformation: false,
+        logActivities: false,
```
garypretty, line 46: As per comment above, this should be set to true by default.

Composer/packages/server/src/models/settings/defaultSettingManager.ts
```diff
@@ -19,6 +19,12 @@ export class DefaultSettingManager extends FileSettingManager {
 
   protected createDefaultSettings = (): any => {
     return {
+      feature: {
+        UseTranscriptLoggerMiddleware: false,
+        UseShowTypingMiddleware: false,
+        UseInspectionMiddleware: false,
+        UseCosmosDb: false,
```
garypretty, line 26: I think this setting name is ambiguous and confusing.  I had to check if this meant use CosmosDb instead of another persistent storage (e.g. Blob storage) or if it was simply enabling persistent storage in favor of using memory, where CosmosDb is the only option.  

In my opinion, we need to be more explicit here and change this to either **UsePersistentStorage** or **UseCosmosDbPersistentStorage**.

zidaneymar:
> I think we need to move all those settings to the bot\settings\appsettings.json page, and so that user can edit it in composer, otherwise user can't see this. in appsettings.json, we can only keep bot and root to specify the bot location.

yes actually I add a default settings in Composer to include the runtime settings, I'm not sure, is this config section reasonable? 


```
protected createDefaultSettings = (): any => {
    return {
      feature: {
        UseTranscriptLoggerMiddleware: false,
        UseShowTypingMiddleware: false,
        UseInspectionMiddleware: false,
        UseCosmosDb: false,
      },
      MicrosoftAppPassword: '',
      MicrosoftAppId: '',
      luis: {
```

Composer/packages/server/src/models/settings/defaultSettingManager.ts
zidaneymar: fixed

BotProject/Templates/CSharp/appsettings.json
zidaneymar: fixed

PR #2806, opened 2020-04-28
feat: support creating actions from ejected schema in Visual Editor

### LINKED ISSUE #2578: Need to support custom schema e2e
Scenarios listed as below.
User has a customized action eg. MSGraph. user would like to enable it in composer.
-[] Flyout format

yeze322:
## Description
fixes #2578 
Enable creating custom actions in the fly-out action menu.
(Custom actions is calculated automatically based on the builtin sdk.schema and user ejected sdk.schema).

#### Users can follow these steps to enable custom schemas:
1. Prepare the schema file (example: insert a new $kind to `sdk.schema` -> `definitions`)
  
image

2. Eject the runtime to composer by setting custom runtime path
  
image

3. Switch back to the Design page
4. Now custom actions will appear under the 'Custom Actions' menu.
5. If you want to custom the displayed label of an Action, just edit the `title` field in schema

image

#### About Editor schema injection (custom the visual node appearance)
Not covered in this PR. Visual & Form Editor render those custom actions with their default widgets.

We can bring up a plan and then implement it in R10
1. what's the format of an editor schema
2. where to put users' editor schemas
2. how it works with the pluggin system
@a-b-r-o-w-n 

## Task Item
#2578


Composer/packages/extensions/visual-designer/src/components/menus/EdgeMenu.tsx
```diff
@@ -5,13 +5,18 @@
 import { jsx, css } from '@emotion/core';
 import { useCallback, useContext, useState } from 'react';
 import classnames from 'classnames';
-import formatMessage from 'format-message';
-import { createStepMenu, DialogGroup, SDKKinds } from '@bfc/shared';
-import { IContextualMenu, ContextualMenuItemType } from 'office-ui-fabric-react/lib/ContextualMenu';
+import formatMessage, { custom } from 'format-message';
+import { createStepMenu, createStepSubmenu, DialogGroup, SDKKinds } from '@bfc/shared';
```
a-b-r-o-w-n, line 9: This would be a good time to remove `createStepMenu` & `createStepSubmenu` since they are only used in the visual editor.

Composer/packages/extensions/visual-designer/src/components/menus/EdgeMenu.tsx
```diff
@@ -26,12 +31,44 @@ interface EdgeMenuProps {
   addCoachMarkRef?: (ref: { [key: string]: HTMLDivElement }) => void;
 }
 
+const buildCustomActionsSubmenu = (
+  label: string,
+  customizedActionGroups: any[][],
```
a-b-r-o-w-n, line 36: Can this have a type?

Composer/packages/extensions/visual-designer/src/components/menus/EdgeMenu.tsx
```diff
@@ -26,12 +31,44 @@ interface EdgeMenuProps {
...
+  handleType,
+  factory
+): IContextualMenuItem => {
+  const itemGroups: IContextualMenuItem[][] = customizedActionGroups
+    .filter(actionGroup => Array.isArray(actionGroup) && actionGroup.length)
+    .map(actionGroup => {
+      return actionGroup.map(({ title: $kind }) => ({
+        key: $kind,
+        name: $kind,
+        $kind,
```
a-b-r-o-w-n, line 46: This should be wrapped in a `data` object to conform to the menu item type.

yeze322:
@a-b-r-o-w-n the discussion thread is under #2578

Composer/packages/extensions/visual-designer/src/components/menus/EdgeMenu.tsx
yeze322: @a-b-r-o-w-n I'd like to. I will move them out of shard package

Composer/packages/extensions/visual-designer/src/components/menus/EdgeMenu.tsx
yeze322: it's from the `oneOf` field of a schema file. I named it to `DefinitionSummary`

Composer/packages/extensions/visual-designer/src/components/menus/EdgeMenu.tsx
yeze322: After the refactoring work, the EdgeMenu only emits the key of the item.
Usually, the emitted key is set to the action `$kind`.

The motivation of doing so is to avoid coupling `DialogFactory` in menu components

PR #2811, opened 2020-04-28
fix: make error messages multiline

### LINKED ISSUE #2792: Validation issues - can we get some features back?
## Describe the bug
* Users are NOT able to expand the incomplete error message in the Language Generation inline editor. 
image

* Users are NOT able to expand the incomplete error message in the Language Understanding inline editor. 
image

* Users are able to expand the incomplete error message but need a link pointing to the [Adaptive Expressions](https://github.com/microsoft/BotBuilder-Samples/tree/master/experimental/common-expression-language) reference doc.
image

## Version

Latest master branch

## Browser
- [x] Chrome

## OS
- [x] Windows

## To Reproduce

1. Produce an error on purpose in the LG/LU inline editor
2. Click on the generated error message
3. See error

## Expected behavior
- Should allow users to expand the error message.
- Should provide corresponding reference docs for syntax. 

## Additional
What we used to have:
* **lg**
lg-inline-error

* **lu**
lu-inline-error

* **Expressions**
expand-expression-error

beyackle:
## Description

This makes our error messages multiline, which is needed since we removed the reveal arrows (for a11y reasons).

## Task Item

fixes #2792

## Screenshots

image

### DIFF
--- a/Composer/packages/lib/code-editor/src/BaseEditor.tsx
+++ b/Composer/packages/lib/code-editor/src/BaseEditor.tsx
@@ -223,7 +223,7 @@ const BaseEditor: React.FC<BaseEditorProps> = props => {
       {(hasError || hasWarning) && (
         <MessageBar
           messageBarType={hasError ? MessageBarType.error : hasWarning ? MessageBarType.warning : MessageBarType.info}
-          isMultiline={false}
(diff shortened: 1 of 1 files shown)


cwhitten:
Thank you Ben

PR #2814, opened 2020-04-28
fix: add a new help URL to the Expression editor

### LINKED ISSUE #2792: Validation issues - can we get some features back?

beyackle:
## Description

Adds in a help URL to the expression editor, plus some plumbing to make it easy to decorate other sorts of fields with whatever other syntax help URLs we need.

## Task Item

closes #2792

## Screenshots

image


Composer/packages/extensions/adaptive-form/src/components/FormRow.tsx
```diff
@@ -59,6 +59,9 @@ const FormRow: React.FC<FormRowProps> = props => {
 
   const { required = [] } = schema;
 
+  const HELP_URL =
+    'https://github.com/microsoft/BotBuilder-Samples/tree/master/experimental/common-expression-language';
+
```
a-b-r-o-w-n, line 64: This should probably exist in the expression plugin package.

beyackle:
Hang on - this is still not quite right.

Composer/packages/extensions/adaptive-form/src/components/SchemaField.tsx
```diff
@@ -34,9 +34,10 @@ const SchemaField: React.FC<FieldProps> = props => {
     ...rest
   } = props;
   const pluginConfig = usePluginConfig();
+
   const schema = resolveRef(baseSchema, definitions);
   const uiOptions = {
-    ...getUISchema(schema, pluginConfig.formSchema),
+    ...getUIOptions(schema, pluginConfig.formSchema, pluginConfig.roleSchema[schema.$role || '']),
```
a-b-r-o-w-n, line 40: `pluginConfig.roleSchema[schema.$role || '']` should happen inside of `getUIOptions` just like formSchema

Composer/packages/extensions/adaptive-form/src/utils/getUIOptions.ts
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+import { UISchema, UIOptions, JSONSchema7 } from '@bfc/extension';
+/**
+ * Merges overrides and plugins into default ui schema and returns the UIOptions
+ */
+export function getUIOptions(schema?: JSONSchema7, uiSchema?: UISchema, pluginOptions?: UIOptions): UIOptions {
```
a-b-r-o-w-n, line 7:
```suggestion
export function getUIOptions(schema?: JSONSchema7, uiSchema?: UISchema, roleSchema?: UIOptions): UIOptions {
```

Composer/packages/extensions/adaptive-form/src/utils/getUIOptions.ts
```diff
@@ -0,0 +1,15 @@
...
+  const { $kind } = schema?.properties || {};
+
+  const kind = $kind && typeof $kind === 'object' && ($kind.const as string);
+
+  const formOptions = uiSchema && kind && uiSchema[kind] ? uiSchema[kind] : {};
+
+  return { ...formOptions, ...pluginOptions };
```
a-b-r-o-w-n, line 14: Let's invert this so that formOptions override pluginOptions.

beyackle: Will do, although intuitively it feels to me like a plugin should be able to overwrite defaults.

beyackle:
I have no idea why fixing conflicts causes it to replay all of the commits back in this weird snowballing fashion.

Composer/packages/extensions/adaptive-form/src/utils/getUIOptions.ts
a-b-r-o-w-n: These are the merged result of all the plugins. Which have overwritten the defaults already.

Just in this case, I think that specifying something on a particular field should take precedence over an override for a role.

Composer/packages/extensions/adaptive-form/package.json
```diff
@@ -17,6 +17,7 @@
   },
   "license": "MIT",
   "peerDependencies": {
+    "@bfc/ui-plugin-expressions": "*",
```
a-b-r-o-w-n, line 20: This should be removed.

Composer/packages/client/src/pages/design/PropertyEditor.tsx
```diff
@@ -57,7 +57,7 @@ const PropertyEditor: React.FC = () => {
   }, []);
 
   const $uiSchema = useMemo(() => {
-    return getUISchema($schema, pluginConfig.formSchema);
+    return getUIOptions($schema, pluginConfig.formSchema);
```
a-b-r-o-w-n, line 60: This is missing passing `pluginConfig.roleSchema`

Composer/packages/extensions/adaptive-form/src/utils/getUIOptions.ts
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Merges overrides and plugins into default ui schema and returns the UIOptions
+ */
+export function getUIOptions(schema?: JSONSchema7, uiSchema?: UISchema, roleSchema?: RoleSchema): UIOptions {
+  const { $kind } = schema?.properties || {};
+
+  const kind = $kind && typeof $kind === 'object' && ($kind.const as string);
+
+  const formOptions = uiSchema && kind && uiSchema[kind] ? uiSchema[kind] : {};
+  const pluginOptions = roleSchema && kind && roleSchema[kind] ? roleSchema[kind] : {};
```
a-b-r-o-w-n, line 13:
```suggestion
  const roleOptions = roleSchema && kind && roleSchema[kind] ? roleSchema[kind] : {};
```

a-b-r-o-w-n: This will also need to get the $role  from the schema to index into the `roleSchema` instead of kind.

PR #2815, opened 2020-04-28
feat: Added skill manifest editor wizard

### LINKED ISSUE #2327: Skills: Manifest generation
In order for a Composer-built bot to be available as a Skill, a new file needs to be created (.manifest). This file is strongly typed using JSON-Schema and is versioned, similar to .dialog files.

We should leverage the strong typing of the file format to auto-generate form fields like we do the form editor in the Design page, as well as re-use UI controls across the forms and this manifest page.

tdurnford:
## Description
When the user clicks 'Export as skill' under the 'Export' toolbar menu, the skill manifest editor wizard renders in a modal. The user can create a new skill manifest by selecting which version of the manifest schema they want from the dropdown and then clicking the 'create' button. This will create a new file - `skill-manifest-${version}.manifest` - in the`skill-manifests` directory. If the user wants to edit an existing skill manifest instead, they can select the manifest they want to update from the list view. 

The next step uses the `AdaptiveForm` package to allow the user to enter metadata for the skill manifest. After this step in the designs, the user should be able to select which triggers from the root dialog they want to include so the wizard can generate the `dispatchModels` and `activities` properties; however, this is reliant on the user being able to create dialog schema files. Consequently, the generate manifest feature will be completed after the work pertaining to `*.dialog.schema` work is completed. At the end of the editor experience, the user can see entire skill manifest in the JsonCode editor and make any necessary changes to the manifest. 

## Task Item
Closes #2327 

## Screenshots

ef5a0e04-1b63-4e68-8c52-df4104018fc7


Composer/packages/extensions/adaptive-form/src/components/fields/ObjectField.tsx
```diff
@@ -7,7 +7,7 @@ import { getOrderedProperties } from '../../utils';
 import { FormRow } from '../FormRow';
 
 const ObjectField: React.FC<FieldProps<object>> = function ObjectField(props) {
-  const { schema, uiOptions, depth, value, ...rest } = props;
+  const { schema, uiOptions, depth, value, label, ...rest } = props;
```
a-b-r-o-w-n, line 10: Can you explain this? It doesn't look like it's being used which makes me think we just want to exclude it from `rest`. Is that right?

If so, this may be risky without doing a regression pass on the property editor.

Composer/packages/lib/code-editor/src/JsonEditor.tsx
```diff
@@ -10,10 +10,20 @@ interface JsonEditorProps extends Omit<BaseEditorProps, 'language' | 'value' | '
   onChange: (jsonData: any) => void;
   value?: object;
   schema?: any;
+  setError?: (error: string) => void;
```
a-b-r-o-w-n, line 13:
```suggestion
  onError?: (error: string) => void;
```

Composer/packages/lib/indexers/src/skillManifestIndexer.ts
```diff
@@ -0,0 +1,21 @@
...
+
+const index = (skillManifestFiles: FileInfo[]) => {
+  return skillManifestFiles.reduce((manifests: any[], { content, name, lastModified }) => {
+    try {
+      const jsonContent = JSON.parse(content);
+      return [...manifests, { content: jsonContent, id: getBaseName(name, '.manifest'), lastModified }];
+    } catch (error) {
+      return manifests;
+    }
+  }, [] as any[]);
```
a-b-r-o-w-n, line 16: Can this be typed?

Composer/packages/client/src/pages/design/exportSkillModal/content/SelectManifest.tsx
```diff
@@ -0,0 +1,182 @@
...
+import { calculateTimeDiff } from '../../../../utils';
+import { StoreContext } from '../../../../store';
+import { ContentProps } from '../constants';
+
+const VERSION_REGEX = /\d\.\d\.(\d+|preview-\d+)/i;
+
+const SCHEMA_URIS = [
+  'https://schemas.botframework.com/schemas/skills/skill-manifest-2.1.preview-1.json',
+  'https://schemas.botframework.com/schemas/skills/skill-manifest-2.0.0.json',
+];
```
a-b-r-o-w-n, line 30: I'm curious why you didn't put these in the constants file?

Composer/packages/client/src/pages/design/exportSkillModal/content/SelectManifest.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+  const options: IDropdownOption[] = SCHEMA_URIS.map((key, index) => {
+    const [version] = VERSION_REGEX.exec(key) || [];
+
+    return {
+      text: formatMessage('Version {version}', { version }),
+      key,
+      selected: !index,
+    };
+  });
```
a-b-r-o-w-n, line 62: Maybe memoize this?

Composer/packages/client/src/pages/design/index.tsx
```diff
@@ -30,8 +30,9 @@ import undoHistory from '../../store/middlewares/undo/history';
 import { navigateTo } from '../../utils';
 
 import { VisualEditorAPI } from './FrameAPI';
-import { CreateDialogModal } from './createDialogModal';
 import { AddSkillDialog } from './addSkillDialogModal';
+import { CreateDialogModal } from './createDialogModal';
+import { ExportSkillModal } from './exportSkillModal';
```
a-b-r-o-w-n, line 35: Can you dynamically import these since they are conditionally rendered?

Composer/packages/client/src/pages/design/exportSkillModal/content/SelectManifest.tsx
tdurnford: Because that would be a good place for them.

cwhitten: haha

Composer/packages/extensions/adaptive-form/src/components/fields/ObjectField.tsx
tdurnford: When you set the top-level `label` property to `false` in the form schema, that `label` value gets passed to every schema field in the object since the `label` value is included in `rest`. Subsequently, every field label will be hidden unintentionally. By including `label` when we deconstruct `props` it is no longer included in `rest`, and the labels appear as expected.

a-b-r-o-w-n: That makes sense. And we rely on the SchemaField to get a new label for the property anyways.

Composer/packages/client/src/pages/design/exportSkillModal/constants.tsx
```diff
@@ -0,0 +1,127 @@
...
+        primary: true,
+        text: formatMessage('Done'),
+        onClick: ({ onDismiss }) => onDismiss,
+      },
+    ],
+    content: ReviewManifest,
+    subText: formatMessage('The manifest can be edited and refined manually if and where needed.'),
+    title: formatMessage('Review and generate'),
+  },
+};
```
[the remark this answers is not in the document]
a-b-r-o-w-n: For future reference, I was suggesting wrapping this whole object in a function, but this works too!

Composer/packages/client/src/pages/design/index.tsx
```diff
@@ -425,28 +430,43 @@ function DesignPage(props) {
...
+      <Suspense fallback={<div />}>
+        {triggerModalVisible && (
+          <TriggerCreationModal
+            dialogId={dialogId}
+            isOpen={triggerModalVisible}
+            onDismiss={onTriggerCreationDismiss}
+            onSubmit={onTriggerCreationSubmit}
+          />
+        )}
+      </Suspense>
```
a-b-r-o-w-n, line 469: This just needs 1 Suspense wrapper. Also consider using the <LoadingSpinner /> component.

Issue #2818, opened 2020-04-29
Composer report error when select "false" in if-else step


DingmaomaoBJTU:
## Describe the bug
Composer report error when select "false" in if-else step
![image](https://user-images.githubusercontent.com/16643193/80559795-51392c80-8a11-11ea-8935-919e7277d5d7.png)

## Browser

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

## To Reproduce

1. create a if-else step
2. choose bool for condition, select "false"

## Expected behavior
Should not report error for this case

cwhitten:
addressed via https://github.com/microsoft/BotFramework-Composer/pull/2821

PR #2821, opened 2020-04-29
fix: Composer report error when select false in if-else step

lei9444:
## Description
when the value=false or 0, the if condition will be false

## Task Item
refs #2818


DingmaomaoBJTU:
verified and it works!

PR #2823, opened 2020-04-29
fix: only apply throttle on lu/lg `update` handler

yeze322:
## Description
close #2823
follows #2632 

Apply `throttle` only to `updateLgTemplate` and `updateLuIntent` to resolve concurrency issue.

(Many lines changed just because of the linr indent change)


yeze322:
@lei9444

lei9444:
> @lei9444

Verified with #2742 and the bug is fixed

PR #2826, opened 2020-04-29
fix: bot creation at an unexpected location

### LINKED ISSUE #2798: Bot gets created in the wrong folder
When creating a new bot and not changing the location folder (as it was ok from last time), the bot gets created in the folder C:\Users\<username>\Documents\Composer

## Version

SHA displays test

## Browser

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ X] Edge

## OS

- [ ] macOS
- [X ] Windows
- [ ] Ubuntu

## To Reproduce

1. Create new bot
2. Select Create from template, Echo Bot and click next
3. Leave location as is and click next
4. Check location of bot in file explorer

## Expected behavior

Storing in correct folder

liweitian:
## Description
Fix some bugs found during bot creation.
1. The location got reset during the process of decoding url. The location is set to '' and sent to server. So the server creates the new bot at default location.

2. The preset name is undefined when save as a bot because the templateId is not passed to that component. The corresponding url is also changed to 'projectId/templateId/save'

3. Add/Modify some Unit test.

## Task Item
closes #2798 

## Screenshots
aaaa

### DIFF
--- a/Composer/packages/client/src/pages/home/index.tsx
+++ b/Composer/packages/client/src/pages/home/index.tsx
@@ -121,7 +121,7 @@ const Home: React.FC<RouteComponentProps> = () => {
         },
         onClick: () => {
           setCreationFlowStatus(CreationFlowStatus.SAVEAS);
-          navigate(`projects/${state.projectId}/save`);
+          navigate(`projects/${state.projectId}/${state.templateId}/save`);
         },
       },
       align: 'left',
--- a/Composer/packages/client/__tests__/components/CreationFlow/DefineConversation/index.test.tsx
+++ b/Composer/packages/client/__tests__/components/CreationFlow/DefineConversation/index.test.tsx
@@ -5,6 +5,7 @@ import * as React from 'react';
 import { render, fireEvent } from '@bfc/test-utils';
 
 import { StoreContext } from '../../../../src/store';
+import { StorageFolder } from '../../../../src/store/types';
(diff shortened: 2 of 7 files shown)


liweitian:
Hi @srinaath, can you help to verify whether this pr breaks https://github.com/microsoft/BotFramework-Composer/pull/2758. And I changed the url in Save as bot operation. Is this OK?

srinaath:
Checking it now @liweitian !

srinaath:
@liweitian  PR looks good. I checked out the branch and it doesn't break the bfcomposer protocol stuff.

Composer/packages/client/src/pages/home/index.tsx
```diff
@@ -121,7 +121,7 @@ const Home: React.FC<RouteComponentProps> = () => {
         },
         onClick: () => {
           setCreationFlowStatus(CreationFlowStatus.SAVEAS);
-          navigate(`projects/${state.projectId}/save`);
+          navigate(`projects/${state.projectId}/${state.templateId}/save`);
```
srinaath, line 124: I wonder if `projects/${state.projectId}/save/${state.templateId}/` makes more meaning than `projects/${state.projectId}/${state.templateId}/save/`

Basically  `http://localhost:3000/projects/8012.804143209351/save/EchoBot/` instead of `http://localhost:3000/projects/8012.804143209351/EchoBot/save`

Since we are saving the project with a different template. But, I'm fine with the current url too if no one else has opinions

PR #2837, opened 2020-04-29
feat: Copy boilerplate material into new bot projects

### LINKED ISSUE #2451: Write some boilerplate content to the bot project via storage at create time

benbrown:
## Description
This introduces a step in the bot project creation process that copies a set of files from the `assets/shared` folder into the project.

This will be used to copy the new provisioning script into the projects so they can be provisioned and deployed without first requiring a full eject.

Currently the shared assets include just a README file with a link to the Composer repo. However, this file will soon include instructions on using the provisioning tool and the publishing tool inside Composer to provision and deploy the bot to Azure.

I wanted to do this as a clean PR, but this is needed to finish #2733 

## Task Item
Closes #2451

### DIFF
--- a/Composer/packages/server/assets/shared/README.md
+++ b/Composer/packages/server/assets/shared/README.md
@@ -1,0 +1,6 @@
+This folder contains a Bot Project created with Bot Framework Composer.
+
+The full documentation for Composer lives here:
+https://github.com/microsoft/botframework-composer
+
+To test this bot locally, open this folder in Composer, then click "Start Bot"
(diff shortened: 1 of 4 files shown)


cwhitten:
LGTM!

PR #2838, opened 2020-04-29
feat: Change banner color

### LINKED ISSUE #2755: Visual: Banner color update from black to blue
In the visual alignment review of the FUSE products and the deltas with the Fluent style guide Composer did very well. There were a few areas where we're unnecessarily different:

1. Banner color (currently black) and 
2. Icon in the upper left : we have our custom icon, Fluent apps have the 9 grid which on click brings a pane with 1 click access to related products.

Number 2 does not seem a reasonable one to introduce at this time.
Could we do 1?

@DesignPolice - could you include a pointer to the design?

cwhitten:
closes #2755


DesignPolice:
Thanks Chris!

PR #2842, opened 2020-04-30
fix: do not over fetch the project

a-b-r-o-w-n:
## Description

This reduces the number of fetches made by the client by only doing so when the route project id does not match what is in state.

Also fixes an issue with the form editor by making sure schema definitions are required.

## Task Item

refs #2023


Composer/packages/client/src/styles.ts
```diff
@@ -53,6 +53,8 @@ export const content = css`
   flex: 1;
   display: flex;
   flex-direction: column;
+
+  label: Content;
```
beyackle, line 57: I don't see "label" as a standard CSS property. What's this do?

Composer/packages/extensions/adaptive-form/src/utils/resolveBaseSchema.ts
```diff
@@ -6,7 +6,8 @@ import { MicrosoftAdaptiveDialog } from '@bfc/shared';
 import { resolveRef } from './resolveRef';
 
 export function resolveBaseSchema(schema: JSONSchema7, data: MicrosoftAdaptiveDialog): JSONSchema7 | undefined {
-  if (schema.definitions && schema.definitions?.[data.$kind] && schema.definitions?.[data.$kind] !== true) {
-    return resolveRef(schema.definitions?.[data.$kind], schema.definitions);
+  const defSchema = schema.definitions?.[data.$kind];
+  if (defSchema && typeof defSchema === 'object') {
+    return resolveRef(defSchema, schema.definitions);
```
beyackle, line 11: Nice - much cleaner.

Composer/packages/client/src/styles.ts
a-b-r-o-w-n: This is a special property for emotion. It just appends this to its generated class name.

This isn't required when using the babel transform (it gets inferred), but since we are using the typescript compiler we need to add label for something human recognizable.

It's not required, but is helpful when debugging instead of looking at class names like `css-sdkj21`.

Composer/packages/extensions/adaptive-form/src/utils/resolveBaseSchema.ts
a-b-r-o-w-n: I updated this because typescript started complaining. I'm not really sure why though.

Issue #2852, opened 2020-04-30
[DCR] Interchangeable Runtimes: definition, requirements, commands and code impact


carlosscastro:
Currently composer supports a single runtime for Azure deployment, local ejection and local execution. The only choice of runtime is in C# and using Asp.Net Core web apps. Conceivably, Composer should be runtime agnostic, and as should be Composer code with the exception of the plugins. At the code level, this means evaluating lots of assumptions such as removing direct calls to dotnet build / run. At the extensibility and artifact level, the implications of interchangeable runtimes are that each runtime needs a different set of commands and actions. Thus, runtimes need a common mechanism to define those actions, since Composer needs to answer questions such as 'How do I build this runtime?' 'How do deploy this runtime to a non-azure environment?'.

The scope of this DCR and proposal is:

- **Definition of a runtime manifest along with its axiomatic actions:** What are the set of commands that runtimes should provide for appropriate flexibility? Right now init command, build command and package/publish command could be a good start to ignite discussion
- **Definition of interface defining a physical instance of the runtime in composer:** At the code level, a well defined Runtime interface that includes access to the manifest and potentially could be used as an insertion point for custom code
- **Cleanup of runtime specific code** from the Composer codebase

Given that the Javascript runtime is coming close to ready and that we are preparing for multiple hosting environments such as Azure WebApps and Azure functions, this is a great timing for this DCR.

The foundational work being done as part of DCR https://github.com/microsoft/BotFramework-Composer/issues/2840 also lays a solid foundation for multi-runtime support from a structural point of view.

carlosscastro:
@cwhitten @benbrown FYI this is Post R9 but wanted to run it by you early.

cwhitten:
@carlosscastro closing as we changed direction a bit

PR #2855, opened 2020-05-01
feat: Runtime refactor, new directory structure and Azure Function introduction phase 1

### LINKED ISSUE #2840: [DCR] Bot runtime directory structure changes
Suggesting to change the concept of BotProject to bot runtime conceptually, since it seems like a more appropriate name.

As part of an effort to get Azure Functions integrated into composer I'm shuffling things around, and it marks a good time to re-think the folder structure for the bot runtime.

Proposed folder structure:

- **runtime/dotnet/** Dotnet root
- **runtime/dotnet/core/** Dotnet core shared libraries. Most of the dotnet code should be here.
- **runtime/dotnet/azurefunctions/**
- **runtime/dotnet/azurewebapp/**
- **runtime/node/core/**
- **runtime/node/azurefunctions/**
- **runtime/node/azurewebapp/**
- **runtime/node/someothertechnology/**

Note that we use `dotnet` in order to match with the botbuilder-* repos. We keep `node` since it really will host node bots for the first iteration. This also leaves the open slot for a js folder if / when we support browser js bots.

In a more generic way, the folder structure will be:

- **runtime/{language}/core/**
- **runtime/{language}/{hostingTechnology1}/**
- **runtime/{language}/{hostingTechnology2}/**

One neat thing about this, is that the eject or publish commands will onlyu require to move the core library plus the selected integration library.

carlosscastro:
Conceptual change of BotProject to bot runtime, since it makes it more universally clear and easily recognizable outside of the team. Improvements in c# runtime, refactors and laying the foundation for azure functions support, including alpha version of the azure functions runtime.

# Folder Structure

As part of an effort to get Azure Functions integrated into composer I'm shuffling things around, and it marks a good time to re-think the folder structure for the bot runtime.

Proposed folder structure:

- **runtime/dotnet/** Dotnet root
- **runtime/dotnet/core/** Dotnet core shared libraries. Most of the dotnet code should be here.
- **runtime/dotnet/azurefunctions/**
- **runtime/dotnet/azurewebapp/**
- **runtime/node/core/**
- **runtime/node/azurefunctions/**
- **runtime/node/azurewebapp/**
- **runtime/node/someothertechnology/**

Note that we use `dotnet` in order to match with the botbuilder-* repos. We keep `node` since it really will host node bots for the first iteration. This also leaves the open slot for a js folder if / when we support browser js bots.

In a more generic way, the folder structure will be:

- **runtime/{language}/core/**
- **runtime/{language}/{hostingTechnology1}/**
- **runtime/{language}/{hostingTechnology2}/**

One neat thing about this, is that the eject or publish commands will onlyu require to move the core library plus the selected integration library.

Directory structure proposal here: https://github.com/microsoft/BotFramework-Composer/issues/2840

the build system forced me to add the line below as well :)

fixes #2840

# Runtime Refactor

Separating the BotProject runtime into the following structure

- **Microsoft.BotFramework.Composer.Core:** Core logic to be used generally by an assortment of runtimes. All code here is runtime agnostic.
- **Microsoft.BotFrameowrk.Composer.AzureWebApp:** Thin head on top of the core library to support Azure WebApps. Only controllers and startup, calls the core library right away and no business logic here
- **Microsoft.BotFrameowrk.Composer.AzureFunctions:** Thin head on top of the core library to support Azure Functions. Only HttpTriggers and FunctionsStartup, calls the core library right away and no business logic here. 
- **Microsoft.BotFrameowrk.Composer.Tests:** Test project for all c# code.

# Composer server update

Updating the composer source to work with the new folder structure.

# Next steps and phase 2

This change is not final in that 1) new functionality needs to be added and 2) the current functionality needs to be improved.

In the very short term (i.e. single digit days), the next steps are as follows

- Test end to end for regressions on azure web app and add fixes as needed
- Test Functions runtime with skills
- Remove all hardcoded occurrences of `azurewebapp` and move to a variable model where each time we use `azurewebapp`, we have in scope an object with runtime metadata for the selected runtime, including the name (`azurewebapp` or `azurefunctions`, init command, run command and package command). This is a solid first step towards both progress and gaining understanding for the DCR https://github.com/microsoft/BotFramework-Composer/issues/2852. This work to support functions will be a basic solution and not the final resolution of the DCR. 

In the medium term, we could more refinements to this after proper planning and prioritization, such as

- Javascript runtime + Full support for DCR https://github.com/microsoft/BotFramework-Composer/issues/2852
- Integrated azure login


Composer/plugins/localPublish/src/index.ts
```diff
@@ -200,7 +201,7 @@ class LocalPublisher {
     const commandAndArgs =
       settings.runtime && settings.runtime.customRuntime === true
         ? settings.runtime.command.split(/\s+/)
-        : ['dotnet', 'run'];
+        : ['dotnet', 'run', '--project', 'azurewebapp']; //TODO: ccastro should pick up the bot start command here. After, remove azurewebapp arg
```
benbrown, line 204: is this a todo that can be done now?

carlosscastro: I think it should be in scope, yes. I'll try that later!

runtime/dotnet/azurefunctions/SkillsTrigger.cs
```diff
@@ -0,0 +1,57 @@
...
+
+            var body = await req.ReadAsStringAsync();
+            var activity = JsonConvert.DeserializeObject<Activity>(body, ActivitySerializationSettings.Default);
+            var result = await _skillHandler.HandleReplyToActivityAsync(req.Headers["Authorization"], conversationId, activityId, activity);
+
+            return new JsonResult(result, ActivitySerializationSettings.Default);
+        }
+
+
+        //[FunctionName("skills")]
```
carlosscastro, line 43: This will come along in a second PR that adds function support.

runtime/dotnet/azurefunctions/Startup.cs
```diff
@@ -0,0 +1,208 @@
+using Microsoft.ApplicationInsights.Extensibility;
```
carlosscastro, line 1: copyright header

Composer/package.json
```diff
@@ -40,7 +40,7 @@
     "start:client": "yarn workspace @bfc/client start",
     "start:server": "yarn workspace @bfc/server start",
     "start:server:dev": "yarn workspace @bfc/server start:dev",
-    "runtime": "cd ../BotProject/CSharp/ && dotnet build && dotnet run",
+    "runtime": "cd ../runtime/dotnet/azurewebapp && dotnet build && dotnet run",
```
cwhitten, line 43: good start here. i imagine we'll expand these script targets to configure the hosting platform.. 

`runtime:azurewebapp`, `runtime:azurefunction`

or accept a cli parameter, ex. "language", "host"
"runtime": "cd ../runtime/${language}/${host} .."`

but even then i'm not sure we'll be happy with it given that dotnet, node, python etc have disparate api's and it will require reworking this quite a bit.

Composer/packages/lib/bot-deploy/src/botProjectDeployConfig.ts
```diff
@@ -35,7 +35,7 @@ export interface BotProjectDeployConfig {
   // The ARM template file path, default is 'DeploymentTemplates/template-with-preexisting-rg.json'
   templatePath?: string;
 
-  // Dotnet project path, default is 'BotProject.csproj'
+  // Dotnet project path, default is 'Microsoft.BotFramework.Composer.WebApp.csproj'
```
carlosscastro, line 38: All of these hardcoded paths are going away when we do the interchangeable runtimes DCR later on https://github.com/microsoft/BotFramework-Composer/issues/2852

Composer/plugins/azurePublish/src/index.ts
```diff
@@ -258,7 +262,9 @@ class AzurePublisher {
       };
 
       // append provision resource into file
-      const resourcePath = path.resolve(this.getProjectFolder(resourcekey), 'appsettings.deployment.json');
+      // TODO: here is where we configure the template for the runtime, and should be parameterized when we
```
cwhitten, line 265: yes exactly, we'll need to hoist this "appsettings" stuff out

Composer/plugins/localPublish/src/index.ts
```diff
@@ -210,7 +211,7 @@ class LocalPublisher {
 
       // take the 0th item off the array, leaving just the args
       const startCommand = commandAndArgs.shift();
-
+
```
cwhitten, line 214: kill this?

Composer/package.json
carlosscastro: Yes, this PR is just having what worked before, to work with the new structure. No new functionality. Having this be configurable is more of an R10 thing.

Composer/plugins/azurePublish/src/index.ts
carlosscastro: https://github.com/microsoft/BotFramework-Composer/issues/2852 In r10 we should get all these harcodings out

carlosscastro:
Having some branch problems, fixing in a few mins :)

carlosscastro:
@cwhitten calling it node seems reasonable. Specially if in the future we might have js, node and typescript (even if we don't implement it, somebody might, right?). I can add that and update DCRs after I sort out some branch / fork issues that I'm fixing.

Composer/plugins/localPublish/src/index.ts
carlosscastro: Ha, I did this on purpose because I love spaces in code. It's the C++ code in me. I'll let the js flow through my veins, removing the space now :)

Composer/package.json
carlosscastro: For composer to support interchangeable runtimes, composer needs to make no assumptions about the runtimes so we agree on that one. There will be significant reworking proportionally, but in practice there are 4 classes in composer that do mos tof the runtime interaction: plugins (azure + local ), assetManager and BotDeployer. It's reasonable to rework those plus command line tools, considering that in exchange we gain interchangeable runtimes if we engineer it right.

carlosscastro:
> love this.
> 
> @carlosscastro what do we think about being more specific for the incoming JS stuff and call it "node" instead of "js"?

Just changed folder name, readme files and DCRs to use the `node` name. Added an explanation in the DCR https://github.com/microsoft/BotFramework-Composer/issues/2840

> Note that we use dotnet in order to match with the botbuilder-* repos. We keep node since it really will host node bots for the first iteration. This also leaves the open slot for a js folder if / when we support browser js bots.

PR #2860, opened 2020-05-01
fix: Fixed unset electron production flag.

### LINKED ISSUE #2770: Allow Composer to dynamically find a port to mount the server to

tonyanziano:
## Description

The `NODE_ENV` env variable was expected to be set to `production` when executing the packaged Composer Electron app, but it was not being set. This was breaking some production functionality such as dynamically finding an open port to mount the server on.

## Task Item

fixes #2770

### DIFF
--- a/Composer/packages/electron-server/src/main.ts
+++ b/Composer/packages/electron-server/src/main.ts
@@ -32,6 +32,12 @@ const getBaseUrl = () => {
   return `http://localhost:${serverPort}/`;
 };
 
+// set production flag
+if (app.isPackaged) {
+  process.env.NODE_ENV = 'production';
+}
+log(`${process.env.NODE_ENV} environment detected.`);
+
 function processArgsForWindows(args: string[]): string {
   const deepLinkUrl = args.find(arg => arg.startsWith(composerProtocol));
   if (deepLinkUrl) {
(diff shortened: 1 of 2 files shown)


Composer/packages/electron-server/src/main.ts
```diff
@@ -32,6 +32,12 @@ const getBaseUrl = () => {
   return `http://localhost:${serverPort}/`;
 };
 
+// set production flag
+if (app.isPackaged) {
```
srinaath, line 36: Is this the first place where app.isPackaged is set to true @tonyanziano . Can we move it outside directly inside main.ts not nested in any function?

srinaath: Or what if we use cross-env NODE_ENV=production inside the command yarn dist:full?

tonyanziano: I'm not sure I understand the wording of your question. Where would you like to move it to?

tonyanziano: Setting the variable at the script level doesn't work after testing. This is because `process.env` is scoped to the process at runtime, so attaching it to the `yarn dist:full` script would simply set it for the duration of building the production version, but at runtime it would be unset.

PR #2861, opened 2020-05-01
fix: bot content not updating after restart

### LINKED ISSUE #1747: Nothing can be reused in C# bot projects?
So nothing  from Composer that can be reused in C# bot project?

a-b-r-o-w-n:
closes #1747 

Description

Calling `child_process.kill()` will kill that process but does not kill all processes spawned by that process. We need to run the bot in a detached process which gets its own process group, then call `process.kill(-child.pid)` to kill all porcesses in that group.


boydc2014:
This is not compatible in windows, see https://github.com/nodejs/node/issues/3617.

And i believe this PR also touched something in promise resolving, causing the publish request, now unable to resolve in Windows.

PR #2863, opened 2020-05-01
fix: use lefthand nav icon selector in NotificationPage.spec.ts

### LINKED ISSUE #2866: fragile e2e test - Notifications view
## Describe the bug

Our e2e tests are a little fragile when it comes to finding the Notifications view; they're looking for the little "2 ⨂" icon in the top bar and often not finding it. Searching for the Notifications icon in the left-hand Nav bar should be less prone to failure, since it's not likely to have been scrolled off to the right somewhere.

## Expected behavior

E2E tests should pass consistently

beyackle:
## Description

Uses a different selector to bring up the Notifications view in our e2e tests

## Task Item

closes #2866


Composer/cypress/integration/NotificationPage.spec.ts
```diff
@@ -71,7 +71,7 @@ context('Notification Page', () => {
         .wait(1000);
     });
 
-    cy.get('[data-testid="notifications-info-button"]').click();
+    cy.get('[data-testid="LeftNav-CommandBarButtonNotifications"]').click();
```
a-b-r-o-w-n, line 74: Can I get you to update these to the testing library method?

`cy.findByTestId(id)`

beyackle: Will do. Thanks for reminding me of that.

PR #2904, opened 2020-05-06
fix: improved lu/lg editor performance

lei9444:
## Description
The main cause of the performance issue is that client is a single-threaded and generally runs in the UI thread. But the lu/lg parse functions don't run in a background thread, and some big lu/lg files will take a long time to parse. So the UI will freeze until the parse are done.

The web worker was added to do the time-consuming calculations and it won't block the main thread.

Now all editors are uncontrolled, so the files info in store will not effect the content showed in the editor. But the error message is from parse result, so if the file is big, the error may be delay display.

## Task Item
refs #2627

## Screenshots
before:
before

after:
after


Composer/packages/client/src/store/parsers/baseWorker.ts
```diff
@@ -0,0 +1,49 @@
...
+  private rejects = {};
+  private globalMsgId = 0;
+
+  constructor(worker: Worker) {
+    this.worker = worker;
+    this.worker.onmessage = this.handleMsg;
+  }
+
+  sendMsg = <T>(type: string, payload: T) => {
+    const msgId = this.globalMsgId++;
```
a-b-r-o-w-n, line 17: Let's use unique ids instead of incrementing.

Composer/packages/client/src/store/parsers/baseWorker.ts
```diff
@@ -0,0 +1,49 @@
...
+      // error condition
+      const reject = this.rejects[id];
+      if (reject) {
+        if (error) {
+          reject(error);
+        } else {
+          reject('Got nothing');
+        }
+      }
+    }
```
a-b-r-o-w-n, line 43: Maybe this should be inverted.

```ts
if (error) {
  // reject
} else {
  // resolve
}
```

Composer/packages/client/src/store/parsers/baseWorker.ts
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// Wrapper class
+export class BaseWorker {
```
a-b-r-o-w-n, line 5: Let's make this accept a generic type for the action types.

```ts
class BaseWorker<DispatchType> {
  sendMsg = <Payload>(type: DispatchType, payload: Payload) 
}
```

a-b-r-o-w-n:
@lei9444 I am unable to make any edits to this PR. Please make sure to check the "Allow edits from maintainers" when creating PRs.

a-b-r-o-w-n:
I've also found a couple of issues.

1. The error in the inline editor always says 'undefined'
2. The console is getting spammed with error messages.

Screen Recording 2020-05-06 at 03 58 PM

lei9444:
> @lei9444 I am unable to make any edits to this PR. Please make sure to check the "Allow edits from maintainers" when creating PRs.

Hi @a-b-r-o-w-n I will check these today.

lei9444:
> I've also found a couple of issues.
> 
> 1. The error in the inline editor always says 'undefined'
> 2. The console is getting spammed with error messages.
> 
> Screen Recording 2020-05-06 at 03 58 PM

The issue will be fixed in https://github.com/microsoft/botframework-cli/pull/763. when the PR is merged, I will update the bf-lu in composer

lei9444:
> I've also found a couple of issues.
> 
> 1. The error in the inline editor always says 'undefined'
> 2. The console is getting spammed with error messages.
> 
> Screen Recording 2020-05-06 at 03 58 PM

This issue will be fixed in https://github.com/microsoft/botframework-cli/pull/763. when the PR is merged, I will update the bf-lu in composer

hibrenda:
Tested in calendar skill and todo skill with multi lu/lg files and big lu/lg files.
Editing performance is hugely improved. great job! 
The only issue I observed is the auto-suggestion may delay for lu entity if this is a big lu file (eg. about 250 lines.)

Composer/packages/client/src/store/parsers/luWorker.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export enum LuActionType {
+  Parse = 'parse',
+  AddIntent = 'add-intent',
+  UpdateIntent = 'update-intent',
+  RemoveIntent = 'remove-intent',
+}
+
+// Wrapper class
+class LuWorker extends BaseWorker<string> {
```
a-b-r-o-w-n, line 27:
```suggestion
class LuWorker extends BaseWorker<LuActionType> {
```

Composer/packages/client/src/store/parsers/baseWorker.ts
```diff
@@ -0,0 +1,43 @@
...
+  private worker: Worker;
+  private resolves = {};
+  private rejects = {};
+
+  constructor(worker: Worker) {
+    this.worker = worker;
+    this.worker.onmessage = this.handleMsg;
+  }
+
+  sendMsg = <Payload>(type: DispatchType, payload: Payload) => {
```
a-b-r-o-w-n, line 17: So type is only used in the case of the lu worker. It makes more sense to make it a part of the payload instead of adding an argument to this method. 

Therefore we can remove the `DispatchType` generic because it will be part of the payload.

a-b-r-o-w-n:
Thanks @lei9444 !

PR #2913, opened 2020-05-06
fix: can not capture events in visual editor

### LINKED ISSUE #2907: Cannot use multi-selection when Visual Editor container is focused
## Describe the bug
Observed by Brenda after #2898 merged.
bug-focus-beh

## Browser

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

alanlong9278:
## Description

Fixed: can not use event capture in visual editor.
- Before
bug-focus-beh

- After
capture

After fixed it, the container focused state will be removed when children doms' event triggered.

## Why need  change the style?
This is because children doms' mouse event will be blocked when container dom has focus::after styles, like below
```
 &::after {
      content: '';
      position: absolute;
      top: 0;
      bottom: 0;
      left: 0;
      right: 0;
      border: black solid 1px;
    }
```
So I removed the after styles.

## Task Item

fix #2907

### DIFF
--- a/Composer/packages/client/src/pages/design/styles.ts
+++ b/Composer/packages/client/src/pages/design/styles.ts
@@ -83,18 +83,8 @@ export const visualPanel = css`
   position: relative;
 
   &:focus {
-    border-right: none;
+    border: black solid 1px;
     outline: none;
-
-    &::after {
-      content: '';
-      position: absolute;
-      top: 0;
-      bottom: 0;
-      left: 0;
-      right: 0;
-      border: black solid 1px;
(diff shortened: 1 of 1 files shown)


boydc2014:
So, is drawing a border on the whole visual canvas is expected?  the issue is just the border is draw in a sub-optimal way? 

@cwhitten  @a-b-r-o-w-n did you see a black box on the whole visual on Mac like this ?

image

Issue #2854, opened 2020-05-01
[Accessibility][Screen Readers-Filter Dialog] : Search results of filter dialogs is not accessible with down arrow key.


ashish315:
**User Experience** 
If a screen reader will not provide any incorrect way to reach to the search result, the screen reader dependent user will not be able to access the seach result.

**Test Environment**
Browser: Microsoft Edge Dev Version: 84.0.495.2 (Official build)
OS build: 2004 (19562.1000)
Screen Reader: Narrator
URL: [Bot Framework Composer](http://localhost:3000/dialogs/Main)
[**Prerequisite Here**](https://github.com/microsoft/BotFramework-Composer)

**Repro Steps:** 
1. Open the above mentioned URL in EDGE and set the zoom at 400%.
2. Select 'Design Flow' from left navigation.
3. Now, navigate to the filter dialog edit field.
4. Type anything in the search edit field of 'Filter Dialogs'.
5. Observe the narrator narration and try to navigate with down arrow key.
6. Observe the issue.

**Actual:** 
The narrator inform about the search result as "2 results found press down arrow key to navigate" whereas the search results are accessible with tab key instead of down arrow.

**Expected:**
As per narrator narration, the search results should be accessible by down arrow key.

**Note:**
Same issue observed with NVDA.

**MAS Impacted:** [MAS2.1.1](https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={8492c4eb-c179-40ae-8777-cd044ed725a2})

**Attachment:**

[Search results of filter dialogs is not accessible with down arrow key..zip](https://github.com/microsoft/BotFramework-Composer/files/4562672/Search.results.of.filter.dialogs.is.not.accessible.with.down.arrow.key.zip)

PR #2914, opened 2020-05-06
a11y: accessibility2854

liweitian:
## Description
The dialog tree is controller through down arrow keys now
Note: once a dialog/trigger is selected, the visual editor will get focused. I think this is an expected behavior added recently. 
## Task Item
Refs #2854 
## Screenshots
aaaa


Composer/packages/client/src/components/ProjectTree/index.tsx
```diff
@@ -163,18 +164,19 @@ export const ProjectTree: React.FC<IProjectTreeProps> = props => {
...
-          styles={searchBox}
-          onChange={onFilter}
-          iconProps={{ iconName: 'Filter' }}
-          underlined
-        />
-        <div
-          aria-live={'polite'}
-          aria-label={formatMessage(
-            `{
+        <FocusZone direction={FocusZoneDirection.vertical} isCircularNavigation={true}>
```
beyackle, line 167: This is exactly what I was trying to make work, and I think that "direction" prop is what I was missing before. Good job finding that.

Composer/packages/client/src/components/ProjectTree/treeItem.tsx
```diff
@@ -94,6 +107,7 @@ export const TreeItem: React.FC<ITreeItemProps> = props => {
         data-testid={`DialogTreeItem${link.id}`}
         onRenderItem={onRenderItem}
         onRenderOverflowButton={onRenderOverflowButton(link.isRoot, isActive)}
+        doNotContainWithinFocusZone={true}
```
beyackle, line 110: We should make a note (here in a comment, maybe) to remove this prop once we upgrade to Fluent 8.0 (whenever that happens); this behavior is going to be the default from then on.

boydc2014:
@liweitian this one is good to go, the only thing left is to merge with master.

PR #2916, opened 2020-05-06
fix: Register IBotTelemetryClient and add telemetry initializer middleware…

### LINKED ISSUE #2915: Telemetry configuration broken in runtime
Some changes to the runtime have broken the telemetry configuration.

- IBotTelemetryClient is no longer registered for DI
- The TelemetryInitializerMiddleware is never added to the adapter pipeline

I have a fix ready to PR for this - will raise the PR once this issue is opened.

garypretty:
Fixes #2915 - registers the IBotTelemetryClient and adds the TelemetryInitializerMiddleware to the pipeline of the adapter.


runtime/dotnet/azurewebapp/Startup.cs
```diff
@@ -131,17 +132,13 @@ public void ConfigureServices(IServiceCollection services)
...
-            });
-            services.AddSingleton<TelemetryInitializerMiddleware>(sp =>
-            {
-                var httpContextAccessor = sp.GetService<IHttpContextAccessor>();
-                var telemetryLoggerMiddleware = sp.GetService<TelemetryLoggerMiddleware>();
-                return new TelemetryInitializerMiddleware(httpContextAccessor, telemetryLoggerMiddleware, settings.Telemetry.LogActivities);
-            });
+            services.AddSingleton<IBotTelemetryClient, BotTelemetryClient>();
+
+            var sp = services.BuildServiceProvider();
```
carlosscastro, line 137: The github actions is issuing some good advice. This is expensive to run, please remove it. Instead, in line 169 you can do s.getservice there which is a factory method and will be called after the service provider is built.

If you need to use the telemetry somehwere else, just do as we did with the bot and use the factory method instead of the instance method to do deferred resolution. Ping me if you want to chat more about this.

runtime/dotnet/azurewebapp/Startup.cs
```diff
@@ -80,15 +80,16 @@ public IStorage ConfigureStorage(BotSettings settings)
             return storage;
         }
 
-        public BotFrameworkHttpAdapter GetBotAdapter(IStorage storage, BotSettings settings, UserState userState, ConversationState conversationState, IServiceProvider s)
+        public BotFrameworkHttpAdapter GetBotAdapter(IStorage storage, BotSettings settings, UserState userState, ConversationState conversationState, IServiceProvider s, TelemetryInitializerMiddleware telemetryInitializerMiddleware)
```
carlosscastro, line 83: Please port these to the `runtime/dotnet/azurefunctions` folder too. I can help you verify if you need.

runtime/dotnet/azurewebapp/Startup.cs
garypretty: Sorry, I had done this for passing in TelemetryInitializerMiddleware on line 160, but missed the fact that I had access to the service provider here (which I didn't think I did). Thanks for the catch. I'll amend now,

runtime/dotnet/azurewebapp/Startup.cs
garypretty: Added the changes to functions app as requested.

PR #2922, opened 2020-05-06
feat: Goto Begin Dialog after clicking dialog

### LINKED ISSUE #2696: goto first trigger(i.e. BeginDialog) after click a dialog

Currently we just open the default view

Or at least when clicking from begin dialog action

tdurnford:
## Description
Navigates to the `BeginDialog` trigger when the user selects the dialog if it has a `BeginDialog` trigger. If the dialog does not have a `BeginDialog` trigger, it reverts to the existing default behavior. 

## Task Item
Closes #2696


Composer/packages/client/src/pages/design/index.tsx
```diff
@@ -118,6 +118,11 @@ const DesignPage: React.FC<RouteComponentProps<{ dialogId: string; projectId: st
     }
   }, [dialogId, dialogs, location]);
 
+  useEffect(() => {
+    const index = currentDialog.triggers.findIndex(({ type }) => type === SDKKinds.OnBeginDialog);
+    ~index && selectTo(createSelectedPath(index));
```
beyackle, line 123: The `~index &&` is a cute trick, but I'd rather see something more legible like `if (index >= 0)`.

PR #2923, opened 2020-05-06
feat: Runtime + Deployment: Azure function support (preview version)

### LINKED ISSUE #2924: Runtime: Azure Functions support
Preview version of azure functions support. Runtime changes done as part of https://github.com/microsoft/BotFramework-Composer/issues/2471 allowed for code reuse and integration with azure functions.

This issue is to allow a first, simple integration based on allowing deployment to Functions of the current assets.

carlosscastro:
[Note that this is ccastro/azure-functions -> benbrown/azurefunctions, no master involved here]

fixes #2924

Preview version of Azure functions deploy support.

- Runtime: Tweak and improve AzureFunctions project
- Plugin: Minimal changes for function support
- Provision script: Now supports WebApp, functions and custom ARM templates. I used the custom one to create an azure function with a fancy premium plan for example
- ARM template: New arm template for functions equivalent to our web app arm template

Some limitations to the scope and considerations:

- AzureWebApp plugn and AzureFunction plugin can have more code reuse. We can do that refactor as part of the interchangeable runtime work when we bring in the js runtime.
- Functions can be skills, but ~1 day of work is required for azure functions to be able to be skill hosts. 
- This should be a preview feature, to gather data about how it is used. If we support multiple runtimes, we might want to be in a position where we don't have to support the deployment mechanisms of the underlying runtimes. More design should come into play around how the user experience is for deployment. How about for more mature projects that have source control?

Experience:

image

image


carlosscastro:
The devops failure seems to be luis related: 

```
0 passing (17s)
  1 failing

  1) Luis Deploy
       can deploy luis success:
     TestingLibraryElementError: Timed out retrying: Unable to find an element with the text: Start Bot. This could be because the text is broken up by multiple elements. In this case, you can provide a function for your text matcher to make your matcher more flexible.
      at Object.getElementError (http://localhost:3000/__cypress/tests?p=cypress/support/index.ts-557:749:19)
      at http://localhost:3000/__cypress/tests?p=cypress/support/index.ts-557:2743:38
      at http://localhost:3000/__cypress/tests?p=cypress/support/index.ts-557:2726:17
      at baseCommandImpl (http://localhost:3000/__cypress/tests?p=cypress/support/index.ts-557:558:16)
      at commandImpl (http://localhost:3000/__cypress/tests?p=cypress/support/index.ts-557:561:40)
      at getValue (http://localhost:3000/__cypress/tests?p=cypress/support/index.ts-557:585:23)
      at resolveValue (http://localhost:3000/__cypress/tests?p=cypress/support/index.ts-557:625:35)

```

runtime/dotnet/azurefunctions/appsettings.json
```diff
@@ -1,9 +1,10 @@
 {
   "feature": {
+    "UseTelementryLoggerMiddleware": false,
     "UseTranscriptLoggerMiddleware": false,
     "UseShowTypingMiddleware": false,
     "UseInspectionMiddleware": false,
-    "UseCosmosDbPersistentStorage": false
+    "UseCosmosDb": false
```
benbrown, line 7: why this change?
also we are hoping to deprecate this flag

runtime/dotnet/azurefunctions/Startup.cs
```diff
@@ -37,20 +37,14 @@ private IConfigurationRoot BuildConfiguration(string rootDirectory)
 
             config
                 .SetBasePath(rootDirectory)
-                .AddJsonFile("ComposerDialogs/settings/appsettings.json", optional: true, reloadOnChange: true)
+                .AddJsonFile("appsettings.json", optional: true, reloadOnChange: true)
```
benbrown, line 40: are you writing the settings differently?
in teh web app version, we load appsettings, then settings/appsettings, then finally appsettings.deployment.

the `settings/appsettings` version contains the settings managed inside Composer.  the root appsettings.json is only modified by hand by the user.

carlosscastro: Great catch! Pushed a fix for this. Still testing a bit more.

runtime/dotnet/azurefunctions/appsettings.json
carlosscastro: This is the new name in web app as well, updated for parity. See: https://github.com/microsoft/BotFramework-Composer/blob/master/runtime/dotnet/azurewebapp/appsettings.json#L7

PR #2930, opened 2020-05-06
fix: Flyout labels, property editor title/subtitles for action nodes updated

### LINKED ISSUE #2845: Consolidate labeling in the property & visual editors
Screen Shot 2020-04-29 at 12 52 57 PM

Currently, we are displaying the default name of an action in the secondary label. As a result, when you first created an action node you get duplicate names in the main label and the secondary label. Let's change the secondary label content to the raw control name. i.e. Microsoft.IfCondition in the screenshot above.

sangwoohaan:
## Description
Updated following labels

- Action node flyout menu
- Property editor main titles for action nodes
- Property editor subtitles for action nodes

## Task Item

closes #2845 

- Matched the flyout menu labels with the editor main titles
- Updated subtitles.


Composer/packages/ui-plugins/prompts/src/formSchema.ts
```diff
@@ -47,7 +47,7 @@ const formSchema: UISchema = {
     helpLink: 'https://aka.ms/bfc-ask-for-user-input',
     properties: {
       prompt: {
-        label: () => formatMessage('Prompt with multi-choice'),
+        label: () => formatMessage('Choice Input formschema'),
```
a-b-r-o-w-n, line 50: Is this correct?

sangwoohaan: No, that was an accident.

PR #2937, opened 2020-05-07
fix: copy skill manifests into azurePublish plugin 

### LINKED ISSUE #2918: skill manifest url
In order to call a remote skill, we need to provide the web access for manifest file to the caller bot.
This url should support both localhost and azure host.

http://localhost:3979/manifests/EchoBot-13-2-1-preview-1-manifest.json
https://echobot-13-0506-dev.azurewebsites.net/manifests/EchoBot-13-2-1-preview-1-manifest.json

With the current bot runtime structure, both locahost and azure deploy can not get the right file. 

For local host, this need to be hosted under "localPublish\hostedBots\27159.75275676705\runtime\azurewebapp" 
For azure publish, we need to include the manifest file as well.

VanyLaw:
## Description

Fix copy the manifest in wwwroot folder under azure plugin.

## Task Item

close #2918 

## Screenshots
after deploy, can get as static file

image


Composer/plugins/azurePublish/src/index.ts
```diff
@@ -69,9 +74,17 @@ class AzurePublisher {
     if (!botExist) {
       mkdirSync(botFolder, { recursive: true });
     }
-    // save bot files
+    // save bot files and manifest files
     for (const file of botFiles) {
```
benbrown, line 78: this could be done with botfiles.filter

PR #2950, opened 2020-05-08
feat: Custom Action Sample

### LINKED ISSUE #2824: Add a customized runtime(include action and schema) sample
## Describe the bug

Synced with Brenda and Dong, here is our proposal.

We can create a sample(bot template) folder with the customization part of the runtime.  Here is the bot looks like:
```
bot
  /bot.dialog
  /language-generation
  /language-understanding
  /dialogs 
     /custmozedDialogs
  /runtime
     /customizedCode
  /schemas
     customized.schema
```

With all the customization part in the template, when user create a bot from that template, we will help user copy all the assets above, and user can configure use custom schema, we will help user merge the default bot runtime with the customized code. then we can directly run it. We need to explain all the details in docs. 

Here are the actions items:
- [ ] create a sample for customized action and schema. -Lu
- [ ] update the plugin to enable the merge of default runtime with customized part. -Wenyi
- [ ] design the name, description and icon of the new template -Marc?
- [ ] docs -Yan

@cwhitten and @benbrown, please help to review the above proposal. 

## Browser

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

### LINKED ISSUE #2896: Updates to ejected /schemas folder
1.  update readme.md to explain custom actions and define prerequisites better
2. add a .sh file with same behavior as the .cmd file
3. update the dialog:merge command to work with the current /runtime folder (does not work today)

luhan2017:
## Description

In this PR, I added a CustomAction solution in the C# runtime folder which includes the custom code and schema.

Here are the steps to enable CustomAction:
1. Create a echo bot. 
2. Enable ejected runtime, the runtime will be placed in the bot folder.
3. In the bot/runtime folder add a project reference in azurewebapp/Microsoft.BotFramework.Composer.WebApp.csproj, 
```
<ProjectReference Include="..\customaction\Microsoft.BotFramework.Composer.CustomAction.csproj" />
```
4. Then remove line 28, uncomment line 29 and 139 in azurewebapp/Startup.cs file
```
using Microsoft.BotFramework.Composer.CustomAction;
// This is for custom action component registration.
ComponentRegistration.Add(new CustomActionComponentRegistration());
```
5. Run dotnet build to verifty if it passes build.
```
EchoBot-9\runtime\azurewebapp>dotnet build
```
6. Go to azurewebapp/schemas folder, run update.cmd, the sdk.schema will be updated.
```
EchoBot-9\runtime\azurewebapp\schemas>update.cmd
```
7. Copy the sdk.schema to the bot/schemas
8. Go back to composer, reopen the bot and you can add a custom action.  

## Task Item

Create your bot based on custom actions and try out in emulator. 

fixes #2824 
fixes #2896 

## Screenshots
image

image

image

image

image

image


Composer/plugins/samples/assets/projects/CustomAction/customaction.dialog
```diff
@@ -0,0 +1,100 @@
+{
```
boydc2014, line 1: I thought since we modified the built-in CSharp template (code + schema), we don't actually need this sample,  every sample can be used as customized action sample.

boydc2014: see https://github.com/microsoft/BotFramework-Composer/issues/2824#issuecomment-621603697

Composer/plugins/samples/assets/projects/CustomAction/schemas/sdk.schema
```diff
@@ -0,0 +1,8901 @@
+{
```
boydc2014, line 1: this should be an update to the default sdk.schema, instead of creating a new one.

Composer/plugins/samples/assets/projects/CustomAction/schemas/MultiplyDialog.schema
```diff
@@ -0,0 +1,25 @@
+﻿{
```
boydc2014, line 1: This is unnecessary, and doing the schema here and doing the code in another place actually make the CSharp Bot template not "self-contained".

luhan2017:
If we want to keep the custom schema "custom", we need keep two versions of schema in composer:
1. composer/packages/server/schemas/sdk.schema, this is our default schema and should not include custom schema. 
2. runtime/azurewebapp/schemas/sdk.schema. this is default+custom schema.

which means, every time we update the sdk package+schema, we need to keep in mind to do twice of the schema merge for the above two, which doesn't look good for me.

So my hacky way is to put custom schema in the sample folder and when ejection you can customize you sdk.schema by rerun the dialog:merge.

I think both way are not perfect, how about we just explain how to use it in documents. I am not sure providing a single click experience on custom schema is a good direction which will make user confused.  ("I don't need to do anything to enable custom schema"?)

@srinaath is working on a readme to explain how to enable this. we can decide whether that is enough for user. if adding this sample just introduces more confusion, let's close this. 

@cwhitten @benbrown  thoughts?

hibrenda:
I prefer to let user know how to merge the sdk.schema instead of just give them the customized sdk.schema as if user really use this feature later, he need to know how to do the schema merge.
So until we have the UX trade off to implement this step later, we just provide the readme file(manual step) to user. 

For custom schema, there are 2 steps. 
1. custom runtime (self-contained now)
- The customized runtime can be set up through the ejection 
- User can learn from this sample for how to write a customized code and also how to register the customized code into startup.cs . User can easily extend it based on this sample code.

2. customized schema (manual for R9)
- The customized schema (sample) is located in the sample folder  
- User need to follow readme to install bf-dialog tool and manually execute "bf dialog:merge" to get the updated sdk.schema file.
* Just like we did the manual step for azure provision, if user don't follow the step, they can not make it.

srinaath:
@luhan2017  I used this PR to actually build my README file for adding custom actions. I think there is a lot of value in including this sample and I'm mentioning about this sample in the README too. Granted the PR does a little magic of showing the merged schema directly. But the README coupled with this sample should give a user a good idea to add custom actions.

runtime/dotnet/azurewebapp/Startup.cs
```diff
@@ -168,6 +169,9 @@ public void ConfigureServices(IServiceCollection services)
             var resourceExplorer = new ResourceExplorer().AddFolder(botDir);
             var rootDialog = GetRootDialog(botDir);
 
+            // register customized types, please note this is only needed for customized actions. 
+            resourceExplorer.RegisterType<MultiplyDialog>("MultiplyDialog");
```
srinaath, line 173: @luhan2017 so currently the default runtime that is ejected out is going to include this registration for all samples correct? Do you think if we could include this registration only if they eject out CustomActions sample?

benbrown: yes we should include it for everyone so they have a clear example of how it works.

carlosscastro:
@luhan2017 @benbrown should this be added to the functions runtime as well? If yes, we could place the dialog in the core project (/runtime/core) and then jsut reference it from the web app startup.cs and from the functions startup.cs.

benbrown:
IMO - yes!

luhan2017:
To who evolved in the PR, @benbrown @benbrown @hibrenda @cwhitten @srinaath 

I have removed an action sample and added a custom action project in the runtime folder. which user need to refer to it explicitly to enable it. See the details at the beginning. 

@carlosscastro , if user want to enable it for azure function, he/she will need to add a project reference for azure function and add CustomActionComponentRegistration in Azurefunction/startup.cs. which I didn't explicitly add right now.

srinaath:
Awesome @luhan2017 .. Adding my README commits to this PR too. Once added I think we can request a rereview and merge it

srinaath:
README updated and update.sh script to run bfdialog tool on Mac/Win added. PR ready for rereview

PR #2960, opened 2020-05-08
fix: Added missing flag to prevent concurrent downloads of the same update.

### LINKED ISSUE #2226: Implement Auto-Update UX for Composer Electron
Work with @DesignPolice to implement in-app UX for automatic application updates.

tonyanziano:
## Description

This PR adds a line that flips a flag to prevent multiple downloads of the same update.

(Must have missed this in the original auto update PR 😮)

[This check was already there](https://github.com/microsoft/BotFramework-Composer/blob/master/Composer/packages/electron-server/src/appUpdater.ts#L49), but nothing was ever flipping `downloadingUpdate` to `true`.

## Task Item

fixes #2226

### DIFF
--- a/Composer/packages/electron-server/src/appUpdater.ts
+++ b/Composer/packages/electron-server/src/appUpdater.ts
@@ -57,6 +57,7 @@ export class AppUpdater extends EventEmitter {
 
   public downloadUpdate() {
     if (!this.downloadingUpdate) {
+      this.downloadingUpdate = true;
       autoUpdater.downloadUpdate();
     }
   }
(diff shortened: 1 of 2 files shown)


Composer/packages/electron-server/src/appUpdater.ts
```diff
@@ -57,6 +57,7 @@ export class AppUpdater extends EventEmitter {
 
   public downloadUpdate() {
     if (!this.downloadingUpdate) {
+      this.downloadingUpdate = true;
```
cwhitten, line 60: do we need to set this to false when it's finished?

tonyanziano: It already is being set to false in [`resetToIdle()`](https://github.com/microsoft/BotFramework-Composer/blob/master/Composer/packages/electron-server/src/appUpdater.ts#L120) which is called on error, finished downloading, or update unavailable.

PR #2964, opened 2020-05-09
fix: remove explicit `UseCosmosDbPersistentStorage` `UseTranscriptLoggerMiddleware` settings and use API to get tenant ID

### LINKED ISSUE #2858: Bot with Luis throws 500 error once deployed on Azure
A bot that does not have Luis enabled runs fine on Azure once it has been deployed. However when switching on Luis as the recognizer type and deploying the bot to Azure, it fails with a status code 500:
_**D:\home\site\wwwroot\ComposerDialogs\mybot.dialog error: Failed to find resource named mybot.lu.dialog or mybot.lu.**_

(Hunch) It seems it is missing the en-us in the filename, as that one does exist.

## Version

d7499777

## To Reproduce

1. Create a new bot and enable Luis
2. Create the Azure Resources
3. Deploy the bot
4. Fill in the setting for Luis
4. Enable applications insights for the webapp in Azure
5. Run the bot in the Azure Bot channels registration
6. Check the 500 error in application insights

### LINKED ISSUE #2927: Do we need explicit `UseCosmosDbPersistentStorage` setting
Currently, to enable the cosmos db features in the runtime, you must BOTH specify a cosmosDB config section.

Can we use the presence or absence of the cosmosDB config section to control whether this feature is used? Having 2 settings required seems easy to miss.

zidaneymar:
## Description

1. Remove explicit `UseCosmosDbPersistentStorage` `UseTranscriptLoggerMiddleware` settings
2. Add more error details to provision script
3. Add guid identifier to resources (referring to[ OA Solution](https://github.com/microsoft/botframework-solutions/blob/6d40fa8ae05f96b0c5e0464e01361a9e1deb696c/samples/csharp/assistants/enterprise-assistant/VirtualAssistantSample/Deployment/Resources/template.json))
4. Add tenant id retrieving in provision script.

## Task Item

fixes #2927 
fixes #2858 
## Screenshots

Have more details about the failed provision
image


runtime/dotnet/azurefunctions/Startup.cs
```diff
@@ -173,12 +173,9 @@ public override void Configure(IFunctionsHostBuilder builder)
 
         public void ConfigureTranscriptLoggerMiddleware(BotFrameworkHttpAdapter adapter, BotSettings settings)
         {
-            if (settings.Feature.UseTranscriptLoggerMiddleware)
+            if (!string.IsNullOrEmpty(settings.BlobStorage.ConnectionString) && !string.IsNullOrEmpty(settings.BlobStorage.Container))
```
carlosscastro, line 176: Thanks a lot for updating the functions one as well :)

carlosscastro:
This is great! One small comment: Since you deprecated a few of the settings, it would be great to remove them from everywhere. They are still referenced in the appsettings.json file in both azure functions and azure web app runtimes. Example [here](https://github.com/microsoft/BotFramework-Composer/blob/master/runtime/dotnet/azurewebapp/appsettings.json#L7). I'll set myself as waiting just to track this pending item. If there is a reason to keep them that I don't know of, all good -- just let me know :)

Composer/packages/lib/bot-deploy/src/botProjectDeploy.ts
```diff
@@ -89,6 +90,22 @@ export class BotProjectDeploy {
...
+    try {
+      const tenantUrl = `https://management.azure.com/tenants?api-version=2020-01-01`;
+      const options = {
+        headers: { Authorization: `Bearer ${this.accessToken}` },
+      } as rp.RequestPromiseOptions;
+      const response = await rp.get(tenantUrl, options);
+      const jsonRes = JSON.parse(response);
+      return jsonRes.value[0].tenantId;
+    } catch (err) {
+      throw new Error(`Get Tenant Id Failed`);
```
carlosscastro, line 106: I think we should output the error, so users can give us that information if they run into this, since there are many things that could go wrong here, from network connection, to unauthorized access token, etc.

Composer/packages/lib/bot-deploy/src/botProjectDeploy.ts
```diff
@@ -89,6 +90,22 @@ export class BotProjectDeploy {
       value: scope,
     };
   }
+  private async getTenantId() {
+    if (!this.accessToken) {
+      throw new Error('Invalid Access Token');
```
carlosscastro, line 95: Should we say something more precise, such as

> Error: Missing access token. Please provide a non-expired Azure access token. Tokens can be obtained by running `az account get-access-token`

Doesn't have to be like that, but should be actionable for the user. The only reason the access token is null is because thye didn't provide it.

Composer/packages/lib/bot-deploy/src/botProjectDeploy.ts
```diff
@@ -89,6 +90,22 @@ export class BotProjectDeploy {
...
+    }
+    try {
+      const tenantUrl = `https://management.azure.com/tenants?api-version=2020-01-01`;
+      const options = {
+        headers: { Authorization: `Bearer ${this.accessToken}` },
+      } as rp.RequestPromiseOptions;
+      const response = await rp.get(tenantUrl, options);
```
carlosscastro, line 102: Optional: It would be cool to have a comment with a link to the documentation of the api you are calling so future maintainers of this code can quickly see the shape and input / output of the api.

Composer/plugins/samples/assets/shared/scripts/DeploymentTemplates/template-with-preexisting-rg.json
```diff
@@ -145,6 +145,19 @@
     "LuisAuthoringAccountName": "[concat(parameters('luisServiceName'), '-Authoring')]"
   },
   "resources": [
+    {
```
carlosscastro, line 148: I'm curious, what is this for?

Composer/plugins/samples/assets/shared/scripts/provisionComposer.js
```diff
@@ -170,6 +171,23 @@ const unpackObject = output => {
...
+  try {
+    const tenantUrl = `https://management.azure.com/tenants?api-version=2020-01-01`;
+    const options = {
+      headers: { Authorization: `Bearer ${accessToken}` },
+    };
+    const response = await rp.get(tenantUrl, options);
+    const jsonRes = JSON.parse(response);
+    return jsonRes.value[0].tenantId;
+  } catch (err) {
+    throw new Error(`Get Tenant Id Failed`);
```
carlosscastro, line 187: Same comment as before, this error should not be swallowed, `err` has valuable information for the user and for support engineers helping users troubleshoot. b

Composer/plugins/samples/assets/shared/scripts/provisionComposer.js
```diff
@@ -170,6 +171,23 @@ const unpackObject = output => {
   return unpacked;
 };
 
+const getTenantId = async (accessToken) => {
+  if (!accessToken) {
```
carlosscastro, line 175: Same error as in the other file

Composer/plugins/samples/assets/shared/scripts/provisionComposer.js
```diff
@@ -170,6 +171,23 @@ const unpackObject = output => {
   return unpacked;
 };
 
+const getTenantId = async (accessToken) => {
```
carlosscastro, line 174: Can we reuse this function somehow? I'm not familiar whether we have a common codebase if we do, it would be great to avoid duplication.

Composer/plugins/samples/assets/shared/scripts/provisionComposer.js
```diff
@@ -270,6 +288,30 @@ const provisionFailed = msg => {
   });
 };
 
+const getErrorMesssage = (err) => {
+  if (err.body) {
+    if (err.body.error) {
+      if (err.body.error.details) {
+        const details = err.body.error.details;
+        let errMsg = "";
```
carlosscastro, line 296: single quotes?

Composer/plugins/samples/assets/shared/scripts/provisionComposer.js
```diff
@@ -270,6 +288,30 @@ const provisionFailed = msg => {
   });
 };
 
+const getErrorMesssage = (err) => {
```
carlosscastro, line 291: Nice, extracting this info is very useful, thanks!

runtime/dotnet/azurefunctions/Startup.cs
```diff
@@ -211,5 +208,10 @@ private string GetRootDialog(string folderPath)
 
             throw new Exception($"Can't locate root dialog in {dir.FullName}");
         }
+
+        private bool ConfigSectionValid(string val)
```
carlosscastro, line 212: This method doesn't access member state, so it can be made static

```suggestion
        private static bool ConfigSectionValid(string val)
```

runtime/dotnet/azurewebapp/Startup.cs
```diff
@@ -212,5 +209,10 @@ private string GetRootDialog(string folderPath)
 
             throw new Exception($"Can't locate root dialog in {dir.FullName}");
         }
+
+        private bool ConfigSectionValid(string val)
```
carlosscastro, line 213: This could be made static

```suggestion
        private static bool ConfigSectionValid(string val)
```

runtime/dotnet/azurefunctions/appsettings.json
```diff
@@ -1,9 +1,7 @@
 {
   "feature": {
-    "UseTranscriptLoggerMiddleware": false,
     "UseShowTypingMiddleware": false,
-    "UseInspectionMiddleware": false,
-    "UseCosmosDbPersistentStorage": false
+    "UseInspectionMiddleware": false
```
carlosscastro, line 4: (Y) Thanks for updating this!

Composer/plugins/samples/assets/shared/scripts/provisionComposer.js
carlosscastro: Optional, but we tend to use single quotes for what I saw in this codebase.

carlosscastro:
Thanks for the update, it's looking great. Most of my comments are minor, the only one that I feel is very relevant is that we need to somehow allow users to reach the error information when retrieving the tenant, otherwise this will be extremely hard to troubleshoot when things go wrong.

luhan2017:
Thanks Qi, with this change: "Add guid identifier to resources (referring to OA Solution)" I can query the provision instance number for composer runtime in azure data explorer now:

image

Composer/packages/lib/bot-deploy/src/botProjectDeploy.ts
```diff
@@ -84,12 +85,55 @@ export class BotProjectDeploy {
...
+      );
+    }
+    try {
+      const tenantUrl = `https://management.azure.com/tenants?api-version=2020-01-01`;
+      const options = {
+        headers: { Authorization: `Bearer ${this.accessToken}` },
+      } as rp.RequestPromiseOptions;
+      const response = await rp.get(tenantUrl, options);
+      const jsonRes = JSON.parse(response);
+      return jsonRes.value[0].tenantId;
```
benbrown, line 131: is it guaranteed to be at least 1 item in this array?

seems like something we should potentially error check

Composer/packages/lib/bot-deploy/src/botProjectDeploy.ts
```diff
@@ -84,12 +85,55 @@ export class BotProjectDeploy {
...
+    } catch (err) {
+      throw new Error(`Get Tenant Id Failed, details: ${this.getErrorMesssage(err)}`);
```
carlosscastro, line 133: Thanks, this will 100% save support costs and improve the user experience, glad you fixed it!

Composer/packages/lib/bot-deploy/src/botProjectDeploy.ts
carlosscastro: The documentation for the api is not super clear what happens when there is no tenants. However this api lists all tenants for the account. Two things stem from this:

- The case where there are zero values in this array is when your account has no tenants. I'm not sure if this is possible, but we can easily just print an error message if this happens and there are zero elements in the json response.
- Accounts with multiple tenants (which is a thing that is allowed) will be forced to use the first tenant in the list. This is a problem that we should fix at some point. We might want to file an issue to track this. Once we allow for azure sign in, the tenant issue will exacerbate, since the tenant selection is important for resource creation and AAD app creation.

Composer/packages/lib/bot-deploy/src/botProjectDeploy.ts
```diff
@@ -128,7 +128,11 @@ export class BotProjectDeploy {
       } as rp.RequestPromiseOptions;
       const response = await rp.get(tenantUrl, options);
       const jsonRes = JSON.parse(response);
-      return jsonRes.value[0].tenantId;
+      const tenantId = jsonRes.value[0].tenantId;
```
benbrown, line 131: this will throw an error if value has no elements. i know this is a low chance of happening, but the error correction should be about does jsonRes exist, does jsonRes.value exist, is jsonRes.value.length > 0 before accessing the inner property tenantId

zidaneymar: > this will throw an error if value has no elements. i know this is a low chance of happening, but the error correction should be about does jsonRes exist, does jsonRes.value exist, is jsonRes.value.length > 0 before accessing the inner property tenantId

fixed

zidaneymar:
@benbrown  Hi Ben, the issues about the json array empty error handling has been fixed. Could you please help review again, thanks!

Composer/packages/lib/bot-deploy/src/botProjectDeploy.ts
zidaneymar: Has been fixed, add more error handling .

Composer/plugins/samples/assets/shared/scripts/provisionComposer.js
zidaneymar: > Can we reuse this function somehow? I'm not familiar whether we have a common codebase if we do, it would be great to avoid duplication.

This function is used only in this provision file. And currently the provisionComposer.js is a single file to maintain all provision details.

Composer/packages/lib/bot-deploy/src/botProjectDeploy.ts
```diff
@@ -84,12 +85,62 @@ export class BotProjectDeploy {
...
+        headers: { Authorization: `Bearer ${this.accessToken}` },
+      } as rp.RequestPromiseOptions;
+      const response = await rp.get(tenantUrl, options);
+      const jsonRes = JSON.parse(response);
+      if (
+        jsonRes.value === undefined ||
+        (jsonRes.value && jsonRes.value.length === 0) ||
+        (jsonRes.value && jsonRes.value.length > 0 && jsonRes.value[0].tenantId === undefined)
+      ) {
+        throw new Error(`Get empty tenantId from tenants API`);
```
carlosscastro, line 136: Small text improvement suggestion:

```suggestion
        throw new Error(`No tenants found in the account.`);
```

Composer/plugins/samples/assets/shared/scripts/provisionComposer.js
```diff
@@ -175,6 +176,30 @@ const unpackObject = output => {
...
+  }
+  try {
+    const tenantUrl = `https://management.azure.com/tenants?api-version=2020-01-01`;
+    const options = {
+      headers: { Authorization: `Bearer ${accessToken}` },
+    };
+    const response = await rp.get(tenantUrl, options);
+    const jsonRes = JSON.parse(response);
+    if (jsonRes.value === undefined || (jsonRes.value && jsonRes.value.length === 0) || (jsonRes.value && jsonRes.value.length > 0 && jsonRes.value[0].tenantId === undefined)) {
+      throw new Error(`Get empty tenantId from tenants API`);
```
carlosscastro, line 195: Small text improvement suggestion:

```suggestion
      throw new Error(`No tenants found in the account.`);
```

cxbuilders:
The getTenantId function is not correct. It just gets the first tenant out of the list and uses that. This then fails as I have multiple tenants. The token from the call creds.getToken on line 346 already contains the tenantId that teh user actually authenticated against. So I suppose that should be used?

For me this now fails as it is the wrong tenant it uses.

PR #2966, opened 2020-05-09
[WIP] Add suggestions for options and global template functions

cosmicshuai:
Add suggestions for LG evaluation options. 

Todo: 
Depending on the SDK side, https://github.com/microsoft/botbuilder-js/pull/2188 is not included in js sdk released version.
After it is ready, I will add suggestions for global template functions
@boydc2014


cosmicshuai:
will rework on suggestion

Issue #2983, opened 2020-05-11
incorrect generated predictionOptions


xieofxie:
## Describe the bug

The generated predictionOptions have typo and can't be loaded.
![MicrosoftTeams-image](https://user-images.githubusercontent.com/2876650/81535209-f8b44880-939b-11ea-8a62-3eeba8a351d6.png)

It turns out that in SDK, these options are bool not expressions.
![MicrosoftTeams-image (1)](https://user-images.githubusercontent.com/2876650/81535214-fb16a280-939b-11ea-9a34-ed71784589f5.png)

## Version

Current master.

## Browser

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

## To Reproduce

1. Create a new project with Luis
2. Run

## Additional context

It seems updating cli will fix this.

boydc2014:
Looks like this is caused by bf-lu writing out options that core SDK runtime haven't supported yet. bf-cli RC6 fixed this by revert this feature. Update to RC6 will fix this.

cwhitten:
https://github.com/microsoft/BotFramework-Composer/pull/2984

PR #2984, opened 2020-05-11
fix: update bf-lu version to 4.9.0-RC6

lei9444:
## Description
update lu related packages to RC6

## Task Item
#refs #2983


boydc2014:
Looks like this is caused by bf-lu writing out options that core SDK runtime haven't supported yet. bf-cli RC6 fixed this by revert this feature. Update to RC6 will fix this.

PR #2990, opened 2020-05-11
fix: fix issue with addWebRoute() api extension method

benbrown:
## Description

This fixes a regression introduced a while ago but discovered today by a user that broke the addWebRoute method.

The method needed to be bound back to the Express object to function correctly.

closes #2990


Composer/packages/extensions/plugin-loader/src/composerPluginRegistration.ts
```diff
@@ -121,7 +121,7 @@ export class ComposerPluginRegistration {
       const method = this.loader.webserver[type.toLowerCase()];
 
       if (typeof method === 'function') {
-        method(url, ...handlers);
+        method.bind(this.loader.webserver)(url, ...handlers);
```
a-b-r-o-w-n, line 124: I think it's better to use `call` in this case.

```suggestion
        method.call(this.loader.webserver, url, ...handlers);
```

benbrown:
It was throwing an exception because Express wasn't bound to the method anymore, so it was unable to find certain internal methods.

Without this, using the addWebAPI method throws this error:

`TypeError: Cannot read property 'lazyrouter' of undefined`

This turns out to be the source problem:

PR #2991, opened 2020-05-11
fix: remove github secrets from config

benbrown:
config inside the github sample plugin included (old) secrets


benbrown:
These secrets have been completely invalid for a while... I *think* that is sufficient, but I'm not an expert!

srinaath:
Got it @benbrown .. I also agree with you. Should be sufficient.

Issue #2882, opened 2020-05-04
Focus indication in Electron is not consistent


cwhitten:
At least on Mac. See attached for a couple examples.

![Screen Shot 2020-05-04 at 9 38 07 AM](https://user-images.githubusercontent.com/1156704/80990406-24a75b00-8deb-11ea-9a16-f7d9207c7b02.png)

![Screen Shot 2020-05-04 at 9 38 18 AM](https://user-images.githubusercontent.com/1156704/80990380-16f1d580-8deb-11ea-9c80-bb23852bc003.png)

PR #2997, opened 2020-05-11
fix: remove outer focus container from visual editor

a-b-r-o-w-n:
## Description

Removes the outer level focus container from the visual editor and instead sends focus to the keyboard zone when navigating to a new trigger.

## Task Item

refs #2882

## Screenshots

Screen Recording 2020-05-11 at 02 28 PM


a-b-r-o-w-n:
@beyackle fixed.

PR #3002, opened 2020-05-12
fix: lu/lg editor update strategy

### LINKED ISSUE #2982: LUIS trigger phrases fail to show in the inline editor after trigger creation
Describe the bug
I selected LUIS as a recognizer type and created an Intent Recognized trigger. 
I enter the intent name and trigger phrases in the **Create a trigger** window, but after submit, the data is now shown on the Properties panel. 
Note: the data can be published to LUIS. It's just that they are not shown in the inline lu editor as they are supposed to. 

## Version
Latest master branch.

## Browser
- [x] Chrome

## OS
- [x] Windows

## To Reproduce

1. Select LUIS as a recognizer type
2. Select **+ Add** then **+ Add new trigger** in the toolbar to create an Intent Recognized trigger. 
3. Enter the intent name and trigger phrases, select **Submit**.
4. After creation, in the **Properties** panel on the right side, you cannot see the trigger phrases you created. The lu editor is empty. 

## Expected behavior
Should be able to show the trigger phrases in inline lu editor after creating the intent recognized trigger.

## Screenshots
* Create Intent Recognzied trigger with trigger phrases. 
image

* After creation, the inline lu editor was empty. 
image

* Was able to publish the training data to LUIS. 
image

zhixzhan:
## Description

Removed form-editor previous maintained localValue( for performance consideration ), let `BaseEditor` handle it, it would check `id` to see need update from external value or not. 

Because of no `localValue` anymore, we also don't need `sync` method.

## Task Item

close #2982 
refs #2627 
refs #2904 

## Screenshots
Untitled1

### DIFF
--- a/Composer/packages/ui-plugins/luis/src/LuisIntentEditor.tsx
+++ b/Composer/packages/ui-plugins/luis/src/LuisIntentEditor.tsx
@@ -3,7 +3,7 @@
 
 /** @jsx jsx */
 import { jsx } from '@emotion/core';
-import React, { useState } from 'react';
+import React from 'react';
 import { LuEditor, inlineModePlaceholder } from '@bfc/code-editor';
 import { FieldProps, useShellApi } from '@bfc/extension';
 import { filterSectionDiagnostics } from '@bfc/indexers';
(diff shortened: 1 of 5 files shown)


Composer/packages/ui-plugins/luis/src/LuisIntentEditor.tsx
```diff
@@ -20,13 +20,12 @@ const LuisIntentEditor: React.FC<FieldProps<string>> = props => {
...
     (luFile && luFile.intents.find(intent => intent.Name === intentName)) ||
-      ({
-        Name: intentName,
-        Body: '',
-      } as LuIntentSection)
-  );
+    ({
+      Name: intentName,
+      Body: '',
+    } as LuIntentSection);
```
a-b-r-o-w-n, line 28: How does this perform in a large bot with many intents?

PR #3003, opened 2020-05-12
chore: update resultProperty descption of http request step

### LINKED ISSUE #974: [Hackathon] HTTP Request variable definition/usage should be made less error prone.
When using the Composer component to make an HTTP request (Integrations-->Http Request), the variable that is definted will not receive the body of the response but instead will receive an object that has a content property (among others) which can be used to retrieve the response from the webservice. Discoverability of "you have to add .content" to the variable name would help.

- Ideally when I use variables I should get intelisense.
- Additionally when defining the variable that receive the result of the HTTP Request, it could show me right under the name of the variable that
<variableName>.content is what I should use to access the answer.
- Finally, if I try to use <variableName> directly without a .<Something>, highlight the fact that this is likely not what I want.

There may be better solutions.

**Additional context**

alanlong9278:
## Description

update resultProperty descption of http request step

## Task Item

fix #974 

## Screenshots

update resultProperty descption of http request step
image

### DIFF
--- a/Composer/packages/server/schemas/sdk.schema
+++ b/Composer/packages/server/schemas/sdk.schema
@@ -3315,7 +3315,7 @@
 				"resultProperty": {
 					"$ref": "#/definitions/stringExpression",
 					"title": "Result property",
-					"description": "Property to store the result of this action. The result includes 4 properties from the http response: statusCode, reasonPhrase, content and headers. If the content is json it will be a deserialized object.",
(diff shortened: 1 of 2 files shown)


alanlong9278:
> @alanlong9278 This looks fine, but we need to create a PR upstream in the dotnet repo to update the schema.

@luhan2017 will help to add it in the dotnet repo

PR #3011, opened 2020-05-12
fix: add links to .NET Core SDK to publish error

### LINKED ISSUE #2926: Some users report problems with start bot: "dotnet build" failed
## Describe the bug

Some users have reported a problem with the start bot button reporting "dotnet build" failing.

image001

The dotnet build step happens in the `localPublish` plugin as part of the management of the local runtime: https://github.com/microsoft/BotFramework-Composer/blob/master/Composer/plugins/localPublish/src/index.ts#L173

So far, people who have reported this are building locally and using the browser version. Both are using Windows.

beyackle:
## Description

This will detect if a publishing error relates to `dotnet`, and if so will direct a user to the .NET Core SDK with a download link.

## Task Item

closes #2926 

## Screenshots

image


Composer/packages/server/src/controllers/publisher.ts
```diff
@@ -83,6 +83,7 @@ export const PublishController = {
         // set status and return value as json
         res.status(results.status).json(response);
       } catch (err) {
+        console.error('Publishing error: ' + err.message);
```
benbrown, line 86: this is not exactly what I meant, since this is just going to say 
"failed to run dotnet build"

we can attach to the spawned child process and output the errors it generates to get more detail.

benbrown: the actual `exec`  happens inside the localPublish plugin.

PR #3012, opened 2020-05-12
fix: dereference schema definitions after loading project

### LINKED ISSUE #2998: expression validation no longer working
## Describe the bug

The new schema has move all expressions into definitions so the validation no longer works because the schema does not have `$role: 'expression'`.

a-b-r-o-w-n:
## Description

fixes an issue where expressions would not be validated

## Task Item

closes #2998

## Screenshots

image


Composer/packages/lib/shared/src/schemaUtils/dereference.ts
```diff
@@ -0,0 +1,92 @@
...
+): S extends JSONSchema7[] ? JSONSchema7[] : S extends JSONSchema7 ? JSONSchema7 : JSONSchema7Definition {
+  // properties, oneOf, allOf, anyOf, items, additionalProperties
+  if (Array.isArray(schema)) {
+    const arraySchema: JSONSchema7[] = [];
+
+    (schema as JSONSchema7[]).forEach(s => {
+      arraySchema.push(dereference(s, definitions));
+    });
+
+    return arraySchema as never;
```
beyackle, line 53: I'm confused by this casting things to never. Why do we need it when the return type of this function already has logic to say what return types should be for the given realizations of S?

Issue #3023, opened 2020-05-13
Maximum call stack size exceeded when open a bot


lei9444:
## Describe the bug
![image](https://user-images.githubusercontent.com/39758135/81790833-8b89ea00-9538-11ea-8e7b-5d759e37136d.png)

## Version
master

## Browser

- [x] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Go to 'home'
2. Click on 'open a new bot'
3. See error

yeze322:
@a-b-r-o-w-n Andy do you know what's happening here?
The first time I open a bot, it could be loaded correctly; if I go back to home page and reopen the bot again, whole page crashed.

PR #3025, opened 2020-05-13
fix: Maximum call stack size exceeded when open a bot

lei9444:
## Description
I have add a debug information with defName, invoked times and recursion depth  in the getRef function 

image

this three types
```
Microsoft.BeginDialog
Microsoft.QnAMakerDialog
Microsoft.AdaptiveDialog
```
are looping infinitely

The first time we open the bot, the definitionCache is empty. After doing dereference. The cache has stored the definition.

Then open a new bot again, The Microsoft.BeginDialog is stored in cache already. If we use the value and do dereference continue. There will be a infinite loop. The Microsoft.BeginDialog has a type Microsoft.IDialog, and IDialog has a  Microsoft.BeginDialog type.

The schema about Microsoft.IDialog
```		
"Microsoft.IDialog": {
			"title": "Microsoft Dialogs",
			"description": "Components which derive from Dialog",
			"$role": "interface",
			"oneOf": [
				{
					"type": "string"
				},
				{
					"$ref": "#/definitions/Microsoft.QnAMakerDialog"
				},
				{
					"$ref": "#/definitions/Microsoft.AdaptiveDialog"
				},
				{
					"$ref": "#/definitions/Microsoft.BeginDialog"
				},
				{
					"$ref": "#/definitions/Microsoft.BreakLoop"
				},
```

 and part of Microsoft.BeginDialog are 

```				"dialog": {
					"oneOf": [
						{
							"$kind": "Microsoft.IDialog",
							"pattern": "^(?!(=)).*",
							"title": "Dialog",
							"$ref": "#/definitions/Microsoft.IDialog"
						},
						{
							"$ref": "#/definitions/equalsExpression",
							"examples": [
								"=settings.dialogId"
							]
						}
					],
					"title": "Dialog name",
					"description": "Name of the dialog to call."
				},
```

## Task Item
refs #3023


Composer/packages/lib/shared/src/schemaUtils/dereference.ts
```diff
@@ -80,6 +80,8 @@ export function dereference<S extends JSONSchema7 | JSONSchema7[] | JSONSchema7D
 export function dereferenceDefinitions(definitions: SchemaDefinitions): SchemaDefinitions {
   const resolvedDefs: SchemaDefinitions = {};
 
+  definitionCache.clear();
```
boydc2014, line 83: So, based on your investigation, re-init this definitionCache will falls into an infinite loop, right? but do it once is ok, so you clear it, make sure every time deferencing start from scratch.  

that perhaps work as a hot fix, FYI @a-b-r-o-w-n, may know this better

lei9444: I think we can clear the cache when init a bot or if a type is detected in cache, we return the value directly.

a-b-r-o-w-n: It's not very clear to me why this prevents the stackoverflow. I don't disagree with doing this though -- its reasonable that a schema definition could be different if the schema has been modified.

boydc2014:
> I'm going to look into this a bit further this morning and add some unit tests. This change seems fine to me, but I'd like to understand why.

The issue is the dereference code will hit a circle once the defCache has value for certain key (which only happens the second time, we try to dereference on on exiting cache), this fix is just to clear the cache, so everytime it's a new cache and won't hit cirle.

boydc2014:
> I'm going to look into this a bit further this morning and add some unit tests. This change seems fine to me, but I'd like to understand why.

Seems like leilei added a few logs in above, it's pretty much the IDialog causing the circle.

Issue #2286, opened 2020-03-17
Support disabled nodes in visual editor 


hibrenda:
https://www.figma.com/file/04DymSq4xElhOBdAyutvvTbq/Composer-SH-06.19-?node-id=4586%3A0

- User can disable/enable nodes (single/multi selected) nodes
- copy/paste/move does not change nodes state
- condition nodes will be taken as "one block" to be disabled 

If nodes have been disabled, corresponding LU intent will not be published as well.

yeze322:
TODOs:
- [ ] disabled effect: if an action's `disabled` field set to true, make this node and its child nodes gray (leveragin the css `filter`)
- [ ] In Shell toolbar, provide 2 buttons - `Disable` | `Enable`, make them clickable when selected nodes are detected to be able to toggle the state
    - If all selected nodes are `disabled`, show **| Disable (unclickable) | Enable (clickable) |**
    - If all selected nodes are `NOT disabled`, show **| Disable (clickable) | Enable (unclickable) |**
    - If selected nodes are half `disabled` and half `NOT disabled`, show **| Disable (clickable) | Enable (clickable) |**
    - If no selected node, show **| Disable (unclickable) | Enable (unclickable) |**

yeze322:
LU publish part (@hibrenda who is responsible for the LU publish?)
- [ ] If nodes have been disabled, corresponding LU intent will not be published as well

yeze322:
@sangwoohaan do you provide what's the icon names of 'Disable' and 'Enable' button?
![image](https://user-images.githubusercontent.com/8528761/81818988-4cbc5a00-9561-11ea-836c-de1ae7858498.png)

hibrenda:
> LU publish part (@hibrenda who is responsible for the LU publish?)
> 
> * [ ]  If nodes have been disabled, corresponding LU intent will not be published as well

@lei9444 be aware of the lu content for luis publish.

PR #3027, opened 2020-05-13
draft: 'Disable' selected nodes via toolbar

yeze322:
## Description
User can disable/enable selected actions via toolbar buttons.
If a node is disabled, it won't be executed by runtime.

#### Works:
- [ ] 'disabled' effect - only apply to Node, won't apply to edge.
  image

- [ ] Shell toolbar - provides 2 buttons
  image

- [ ] dialog editing API - implement an API to disable selected nodes (Form supports disable a single node)
    - provide 2 APIs named `disableSelectedNodes` and `enableSelectedNodes`
    - disabled/enabled state should be applied to all child nodes recursively consider IfCondition/SwitchCondition/Foreach

## Task Item
#2286

### DIFF
--- a/Composer/packages/extensions/extension/src/hooks/useDialogEditApi.ts
+++ b/Composer/packages/extensions/extension/src/hooks/useDialogEditApi.ts
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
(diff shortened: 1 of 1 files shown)


yeze322:
Closed it because of found some obstacles during this PR, working on a refactoring PR, then I will open a new one to implement that feature

PR #3044, opened 2020-05-14
fix: Update schema

### LINKED ISSUE #3042: Adding an action to connect to skill errors: SkillDialog not registered in factory
## Describe the bug
When I add an action to connect to a skill, my bot throws a 500 error for:
```
An unhandled exception was thrown by the application.
System.Exception: virtualassistant.dialog error: Type Microsoft.SkillDialog not registered in factory
```

This leaves me unable to send anything to the bot.

## Version

https://github.com/microsoft/BotFramework-Composer/commit/652868ab95e779eb9b9fa8c399626dd35f4ebfc0
Runtime SDK v4.9.0

## Browser
- [x] Edge

## OS
- [x] Windows

## To Reproduce
1. Add a Skill manifest.
2. Add a new dialog with an action to connect to said skill.
3. Start bot and new conversation, it returns 500 error immediately.

## Expected behavior
User can add skill manifest and related dialog action without crashing the bot.

## Screenshots
image

## Additional context
This worked as of whatever the latest RC build was on May 6.

This may be unreleated but using the show code viewer it highlights an issue on a property,
image

luhan2017:
## Description

Update schema, which brings in the SkillDialog->BeginSkill renaming.
We've verified that 4.9.1 didn't fix the dialog id issue, so I reverted the 4.9.1 update. 

## Task Item
fixes #3042

### DIFF
--- a/Composer/packages/extensions/adaptive-form/src/defaultUiSchema.ts
+++ b/Composer/packages/extensions/adaptive-form/src/defaultUiSchema.ts
@@ -295,7 +295,7 @@ const DefaultUISchema: UISchema = {
     subtitle: () => formatMessage('Set Properties'),
     helpLink: 'https://aka.ms/bfc-using-memory',
   },
-  [SDKKinds.SkillDialog]: {
+  [SDKKinds.BeginSkill]: {
     label: () => formatMessage('Connect to a skill'),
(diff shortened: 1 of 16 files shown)


hibrenda:
Verified the skill can be called local to local with the schema updates. no 500 error.

PR #3050, opened 2020-05-14
fix: can not open a bot if manifest url is invalid

### LINKED ISSUE #3047: can not open a bot if manifest url is invalid
- create a consumer bot
- create a skill connection and skillbot hosted in local as following
image
- shut down composer
- restart composer 
- open consumer bot will hit following error
- bot can not be opened

image

Workaround: start the skillbot and make sure in port 3979.

zhixzhan:
## Description

manifest url is extracted at bot open, fixed by passing this message as a notification to client.

## Task Item
close #3047 

## Screenshots
Untitled1
image


Composer/packages/server/src/models/bot/botProject.ts
```diff
@@ -92,7 +94,12 @@ export class BotProject {
       // when re-index opened bot, file write may error
     }
     this.settings = await this.getEnvSettings('', false);
-    this.skills = await extractSkillManifestUrl(this.settings?.skill || []);
+    try {
+      this.skills = await extractSkillManifestUrl(this.settings?.skill || []);
+    } catch (_e) {
+      const notify = new Diagnostic('Accessing skill manifest url error', 'settings', DiagnosticSeverity.Error);
```
boydc2014, line 100: Can we encode the url also here, to show exact which url?

boydc2014:
Also, like you said, once this is an error, there really no good timing to trigger this re-check again, unless reopen the bot, which is weird. 

What i would expect is, even the url is invalid, we will still show the url in the table, and at the end the row, there is a warnning sign and a button to refresh, which seems fair to me. 

@cwhitten @mareekuh any ideas?

Composer/packages/server/src/models/bot/botProject.ts
zhixzhan: Sure

PR #3052, opened 2020-05-14
feat: delete bot

### LINKED ISSUE #2974: I can't delete a bot.
## Describe the bug

I have a EmptyBot-01 turd in my documents folder (see #2973 ) and I can't delete it. I don't know where it put it (see #2973 ).  And now there is no way for me to delete it from composer.

## Version

0.0.1 (2 days ago)

## Browser
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [x] Edge (chrome)

## OS
- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce
1. open compose
2. try to get rid of EmptyBot-01

## Expected behavior

1. Right Click gives menu to "Delete Bot"
2. Opening the bot gives me menu to "delete the bot"

liweitian:
## Description
1. add delete button in setting page (Will update after the UI is finalized)
2. delete bot folder
3. stop runtime process and delete runtime data BotFramework\Composer\Composer\plugins\localPublish\hostedBots\{projectId}
4. delete DialogSetting and LuFileStatus
5. delete data in data.json
6. reset many store states

## Task Item
Closes #2974 

## Screenshots
Already updated code that make the checkbox unchecked at the beginning and the modal wider
aaa


Composer/packages/server/src/controllers/publisher.ts
```diff
@@ -240,4 +240,50 @@ export const PublishController = {
       message: `${method} is not a valid publishing target type. There may be a missing plugin.`,
     });
   },
+  removeLocalRuntimeData: async (req, res) => {
```
benbrown, line 243: I don't think this functionality should be exposed as an external API. This is very specific to one type of publishing scenario -- the local runtime.

I would propose that the the "delete bot" feature simply call the stopbot and removeLocalRuntimeData directly.

Composer/packages/server/src/controllers/project.ts
```diff
@@ -98,6 +98,29 @@ async function getProjectById(req: Request, res: Response) {
...
+  if (!projectId) {
+    res.status(400).json({
+      message: 'parameters not provided, require storage id and path',
+    });
+    return;
+  }
+  try {
+    const currentProject = await BotProjectService.getProjectById(projectId);
+    if (currentProject !== undefined) {
+      await currentProject.deleteAllFiles();
```
benbrown, line 112: I would propose we call stopBot and removeLocalRuntme stuff here.

Composer/plugins/localPublish/src/index.ts
```diff
@@ -378,6 +378,19 @@ class LocalPublisher implements PublishPlugin<PublishConfig> {
     delete LocalPublisher.runningBots[botId];
   };
 
+  private removeRuntimeData = async (botId: string) => {
```
benbrown, line 381: if this is private how does it get called from outside?

Composer/packages/client/src/components/Modal/Confirm.jsx
```diff
@@ -41,9 +53,12 @@ const ConfirmDialog = props => {
         styles: dialogModal,
       }}
     >
-      {onRenderContent(subTitle, BuiltInStyles[style])}
+      <div css={[confirmationContainer, styles.content]}>
+        {onRenderContent(subTitle, BuiltInStyles[style])}
```
a-b-r-o-w-n, line 57: So what if the caller could control whether or not this is disabled?

I'm envisioning something like:

```tsx
{onRenderContent({ subTitle, styles: BuiltInStyles[style], setDisabled })}
```
and in the call site:
```tsx
const setting = {
  disabled: true,
  onRenderContent: (props) => <DeleteConfirm {...props} />
};

function DeleteConfirm(props) {
  return <checkbox onChange={(checked) => props.setDisabled(!checked)} />
}
```

This way the component give's its callers a way to manage this state on their own. For instance, what if we have another modal that requires the user to type in the name of the bot to delete? Or if there is a dropdown the user must choose before continuing? This affords that flexibility.

Composer/packages/client/src/components/ToolBar/index.jsx
```diff
@@ -87,6 +88,26 @@ export function ToolBar(props) {
             />
           </div>
         )}
+        {window.location.href.indexOf('/settings/') !== -1 && (
```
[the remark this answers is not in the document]
a-b-r-o-w-n: I have an item for R10 to rework the tool bar so this won't be necessary.

Composer/packages/server/src/models/bot/botProject.ts
```diff
@@ -313,6 +314,33 @@ export class BotProject {
...
+      await this.fileStorage.rmDir(this.dir);
+      await BotProjectService.cleanProject({ storageId: 'default', path: this.dir });
+      await BotProjectService.deleteRecentProject(this.dir);
+    } catch (e) {
+      throw new Error(e);
+    }
+    return true;
+  }
+
+  private async deleteFilesFromBottomToUp(path) {
```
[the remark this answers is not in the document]
liweitian: For localStorage, rmDir can only remove empty folders at this time. Do I need to change it to rm -rf? Maybe it is not safe to use rm -rf ？

beyackle:
Continued in #3092

liweitian:
continued in another PR. So close this one.

PR #3079, opened 2020-05-15
fix: Goto Begin Dialog after clicking dialog

### LINKED ISSUE #3070: Cannot edit dialog properties, always jump to 'dialog started' trigger
## Describe the bug
bug-cannot-focusdialog

## Browser

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

tdurnford:
## Description
The `designPageLocation` object being included in the `usEffect` dependency array causes the selected trigger to be set to the `BeginDialog` trigger after the user selects the dialog because `designPageLocation.selected` gets set to null when a trigger isn't selected which causes the hook to execute unnecessarily. Removing `designPageLocation` from the `usEffect` dependency array resolves the issue.

## Task Item
Closes #3070

## Screenshots
recording

### DIFF
--- a/Composer/packages/client/src/pages/design/index.tsx
+++ b/Composer/packages/client/src/pages/design/index.tsx
@@ -130,7 +130,6 @@ const DesignPage: React.FC<RouteComponentProps<{ dialogId: string; projectId: st
   }, [dialogId, dialogs, location]);
 
   useEffect(() => {
-    console.log(designPageLocation);
     const index = currentDialog.triggers.findIndex(({ type }) => type === SDKKinds.OnBeginDialog);
     if (index >= 0 && !designPageLocation.selected) {
       selectTo(createSelectedPath(index));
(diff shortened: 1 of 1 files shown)


Composer/packages/client/src/pages/design/index.tsx
```diff
@@ -130,11 +130,12 @@ const DesignPage: React.FC<RouteComponentProps<{ dialogId: string; projectId: st
   }, [dialogId, dialogs, location]);
 
   useEffect(() => {
+    console.log(designPageLocation);
```
cwhitten, line 133: can you remove this?

tdurnford: I shouldn't open PR's before I've had coffee.

PR #3082, opened 2020-05-15
feat: Added custom windows / mac installer graphics.

### LINKED ISSUE #3083: feat: Add custom graphics to Electron installers
The current Windows and Mac (.dmg) installers use the default graphics. These graphics should be updated with Composer branding.

tonyanziano:
## Description

This PR adds custom images to the Windows and Mac (.dmg) installers.

## Task Item

Fixes #3083 

### **Before - Windows**

installer-before-1
installer-before-2

### **After - Windows**

installer-after-1
installer-after-2

### **Before - Mac**

installer-mac-before-1

### **After - Mac**

image


Composer/packages/electron-server/electron-builder-config.json
```diff
@@ -56,6 +57,8 @@
     "runAfterFinish": true,
     "installerIcon": "resources/composerIcon.ico",
     "uninstallerIcon": "resources/composerIcon.ico",
```
srinaath, line 59: can we remove `resources/` now since we used it above as a directory?

srinaath:
This looks neat!!

benbrown:
I have never been part of a product launch that had an installer that looks as nice as this one!!

Composer/packages/electron-server/electron-builder-config.json
tonyanziano: good catch! I think we can

cwhitten:
whohoo!

PR #3084, opened 2020-05-15
fix: allow tenantId to be specified in provisioning script

### LINKED ISSUE #3075: provisionComposer script fails
## Describe the bug

provisionComposer fails as it picks the wrong TenantId in the script. This is introduced in #2964

The getTenantId function is not correct. It just gets the first tenant out of the list and uses that, whilst I need my default one. This then fails as I have multiple tenants. The token from the call creds.getToken on line 346 already contains the correct tenantId (I guess that's the default one). So I suppose that could be used instead.

For me this now fails as the wrong tenant is used.

## Version

6cd92ee

benbrown:
## Description

In provision script, use the tenantId provided by the login instead of choosing the first in the list of available tenants.

Also, allow user to specify a tenantId as a cli parameter.

## Task Item

fixes #3075


Composer/plugins/samples/assets/shared/scripts/provisionComposer.js
```diff
@@ -85,15 +85,16 @@ const appPassword = argv.appPassword;
...
+const luisAuthoringKey = argv.luisAuthoringKey; // not currently used
+const luisAuthoringRegion = argv.luisAuthoringRegion || 'westus'; // not currently used
 
 // Get option flags
 const createLuisResource = argv.createLuisResource == 'false' ? false : true;
 const createLuisAuthoringResource = argv.createLuisAuthoringResource == 'false' ? false : true;
 const createCosmosDb = argv.createCosmosDb == 'false' ? false : true;
 const createStorage = argv.createStorage == 'false' ? false : true;
 const createAppInsignts = argv.createAppInsignts == 'false' ? false : true;
+var tenantId = argv.tenantId ? argv.tenantId : '';
```
srinaath, line 97: const here and maybe we can do the || instead of ternary

Composer/plugins/samples/assets/shared/scripts/provisionComposer.js
```diff
@@ -345,7 +366,18 @@ const create = async (
   if (!tenantId) {
     const token = await creds.getToken();
     const accessToken = token.accessToken;
-    tenantId = await getTenantId(accessToken);
+    // the returned access token will almost surely have a tenantId. 
+    // use this as the default if one isn't specified.
+    // otherwise, fetch a list and use the first, but print available options.
+    if (token.tenantId) {
```
srinaath, line 372: Maybe token && token.tenantId?

cwhitten:
taking this now to kick a build off (0.0.4)

PR #3098, opened 2020-05-16
fix: Fix opening links in Electron

cwhitten:
closes #3098


beyackle:
> closes #3098

Is there a bug this closes, or does pointing a PR to itself short-circuit the need to link it to a bug?

Composer/packages/client/src/pages/home/ItemContainer.tsx
```diff
@@ -16,8 +18,12 @@ import {
...
 } from './styles';
 
-interface ItemContainerProps extends Omit<IButtonProps, 'onChange' | 'styles' | 'title'> {
-  onClick?: () => void | Promise<void>;
+const { openExternal } = window;
+
+type onClickFunctionType = () => void | Promise<void>;
+
+interface ItemContainerProps {
+  onClick?: onClickFunctionType | boolean;
```
beyackle, line 26: Instead of abusing booleans for this, what if we defined a single-valued type in here and then imported that type/value in `index.tsx` for use in the onClick prop? Or used a Symbol?

cwhitten: we cleaned this up.

PR #3118, opened 2020-05-18
fix: update recent bots list after deleting a bot project

### LINKED ISSUE #3110: deleting a bot should clean up client state
## Describe the bug

After deleting a bot, the client state still holds a reference to the project so when on the home screen, the deleted project still appears in the recent projects list as well as impacts bot creation (i.e. appending a number to the bot name)

liweitian:
## Description
You can repro this bug easily by deleting an Empty bot.

1. The root reason is a race condition. Pls check the code I have commented there.

2. Set a global error, if we fail to delete a bot.

## Task Item

Closes #3110 

## Screenshots

aaaa

### DIFF
--- a/Composer/packages/client/src/pages/setting/index.tsx
+++ b/Composer/packages/client/src/pages/setting/index.tsx
@@ -104,7 +104,7 @@ const SettingPage: React.FC<RouteComponentProps<{ '*': string }>> = () => {
     };
     const res = await OpenConfirmModal(title, null, settings);
     if (res) {
-      actions.deleteBotProject(projectId);
+      await actions.deleteBotProject(projectId);
       navigateTo('home');
     }
   };
--- a/Composer/packages/client/src/router.tsx
+++ b/Composer/packages/client/src/router.tsx
@@ -77,7 +77,7 @@ const ProjectRouter: React.FC<RouteComponentProps<{ projectId: string }>> = prop
     if (state.projectId !== props.projectId && props.projectId) {
       actions.fetchProjectById(props.projectId);
     }
-  }, [props.projectId, state.projectId]);
+  }, [props.projectId]);
 
   if (props.projectId !== state.projectId) {
     return <LoadingSpinner />;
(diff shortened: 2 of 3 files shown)


Composer/packages/client/src/pages/setting/index.tsx
```diff
@@ -104,7 +104,7 @@ const SettingPage: React.FC<RouteComponentProps<{ '*': string }>> = () => {
     };
     const res = await OpenConfirmModal(title, null, settings);
     if (res) {
-      actions.deleteBotProject(projectId);
+      await actions.deleteBotProject(projectId);
       navigateTo('home');
```
liweitian, line 108: When navigating to home page, we will fetch the recent bot lists. But the recent bot list is probably not updated because we did not wait on the http request. Adding await will solve the problem.

Composer/packages/client/src/router.tsx
```diff
@@ -77,7 +77,7 @@ const ProjectRouter: React.FC<RouteComponentProps<{ projectId: string }>> = prop
     if (state.projectId !== props.projectId && props.projectId) {
       actions.fetchProjectById(props.projectId);
     }
-  }, [props.projectId, state.projectId]);
+  }, [props.projectId]);
```
liweitian, line 80: After deleting the bot, the state.projectId will set to ''. It changed so  actions.fetchProjectById(props.projectId). But the bot {props.projectId} is already deleted. 

I think the logic is, if the url changed, props.projectId will change, so we fetch the project. The we successfully fetched the project, then change store projectId.

beyackle: This isn't a criticism on this change, but a general question for people who understand more about this design here than I do. If the intended logic for the useEffect doesn't involve `state.projectId`, can we simplify line 77 to just `if (props.projectId !== '')`? Why do we care about the state inside that useEffect call?

Composer/packages/client/src/store/action/project.ts
```diff
@@ -87,7 +87,13 @@ export const deleteBotProject: ActionCreator = async (store, projectId) => {
...
     });
   } catch (e) {
-    console.log(e);
+    store.dispatch({
+      type: ActionTypes.SET_ERROR,
+      payload: {
+        message: e.message,
+        summary: 'Delete Bot Error',
+      },
+    });
```
beyackle, line 96: Good catch here.

PR #3119, opened 2020-05-18
fix: change local storage key

### LINKED ISSUE #3120: Bot name should not be used as the localstorage key
BotName is not unique. We can use project Id instead,

liweitian:
## Description
Use projectId rather than botName as the localstorage key. BotName is not unique.
## Task Item
closes #3120


cwhitten:
@liweitian do we need to consider any migrations in areas where we have serialized botName anywhere? I'm thinking the info we put in local storage of the browser..

liweitian:
> @liweitian do we need to consider any migrations in areas where we have serialized botName anywhere? I'm thinking the info we put in local storage of the browser..

@cwhitten I talked with @DingmaomaoBJTU on this. She had some bots with the same name. So we probably cannot migrate. Even if we do not migrate, the users only need to fill some forms again right?  So how about cleaning those data with a bot Name as the key.

cwhitten:
@liweitian can we add a test that exercises _both_ mechanisms given that we are not migrating old bots. Can you add this?

liweitian:
> @liweitian can we add a test that exercises _both_ mechanisms given that we are not migrating old bots. Can you add this?

@cwhitten If two bots have the same name, the data stored about the second bot will override the first one. We can probably migrate dialogSettings data from botName to projectid. Because the sensitive properties remain the same for a user. For the luFileStatusSetting data, we can probably remove it, which means the users need to re-publish. What do you think?

cwhitten:
Sorry @liweitian, can you address conflicts on this please?

PR #3125, opened 2020-05-19
build: update typescript, eslint, prettier and more

### LINKED ISSUE #2570: [chore] update typescript, eslint, prettier

a-b-r-o-w-n:
* [Typescript 3.9](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html)
* [Prettier 2.0](https://prettier.io/blog/2020/03/21/2.0.0.html)
* [Eslint 7](https://eslint.org/blog/2020/05/eslint-v7.0.0-released)

closes #2570


beyackle:
This is a huge set of file diffs to look through - am I correct that the biggest changes here are to put parentheses around function arguments (e.g. `(x) => x+2`) and to alphabetize props in components everywhere? I support the former, but I don't know if I agree with the latter. It does make props easier to find in a long list, but it means they can't be grouped functionally - e.g. if I wanted to put all of the `onEvent` props at the end of a component as a convention.

a-b-r-o-w-n:
@beyackle You are correct. You are also in luck, the alphabetized props already groups event handlers `on*` together at the end!

a-b-r-o-w-n:
@beyackle I'm open to discuss new / changes to eslint rules. With respect to sorting props, I like that you don't have to think about it -- every prop has its place!

beyackle:
Giving it some more thought, I think I agree with the alphabetized props. It's one fewer way code can be changed without changing its meaning. If we decide it's too annoying, we can revert that part.

Composer/packages/client/src/store/persistence/FileOperation.ts
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) Microsoft Corporation.
```
beyackle, line 1: Make sure we're not re-including files that got deleted/moved.

Composer/packages/client/src/pages/setting/publisher/index.jsx
```diff
@@ -0,0 +1,225 @@
+// Copyright (c) Microsoft Corporation.
```
beyackle, line 1: Make sure we're not re-including files that got deleted/moved.

PR #3129, opened 2020-05-19
fix: show all error messages in log

### LINKED ISSUE #3087: Publishing flow inconsistently provides detailed error log
## Describe the bug
I've had a series of publishing attempts without enough error output to understand what needs to be fixed. The log provided does not show enough of what is going on, it only shows messages regarding DEPLOY_INFO. After investigating in source it turns out that I hit a LUIS publishing issue, however this exception is not bubbled up in the logs available in UI,

### Various publish attempts
image

### Click on See Log
image

### Check output from process
image

## Version
Latest

## Browser
- [x] Edge

## OS
- [x] Windows

## Expected behavior
Error/exception should be provided to understand what needs to be resolved

VanyLaw:
## Description

push all the error into log. 

## Task Item

close #3087


a-b-r-o-w-n:
What is an example of an object error? Can we do better than simply stringifying it?

Composer/plugins/azurePublish/src/index.ts
```diff
@@ -195,11 +195,16 @@ class AzurePublisher {
...
+        this.logMessages.push(JSON.stringify(error));
+      } else {
+        this.logMessages.push(error);
+      }
       // update status and history
       const status = this.getLoadingStatus(botId, profileName, jobId);
       if (status) {
         status.status = 500;
-        status.result.message = error ? error.message : 'publish error';
+        status.result.message = error && error.message ? error.message : 'publish error';
```
beyackle, line 207: We can use the new null coalescing to write this more cleanly:
`status.result.message = error?.message ?? 'publish error'`

Composer/plugins/azurePublish/src/index.ts
```diff
@@ -195,11 +195,16 @@ class AzurePublisher {
       }
     } catch (error) {
       console.log(error);
+      if (typeof error === 'object') {
```
beyackle, line 198: I'd be happier if we could guarantee we'd see actual `Error` objects whenever an exception is thrown (because those have a known shape instead of just being generic objects, and then we could just test `error.message`), but I don't know how we can get that assurance. MDN says that literally anything can be used with `throw`, so it looks like we do need a check like this.

However, I think we should first check if `error` is an instance of `Error`, because then if someone is nice enough to follow the conventions, we can then log `error.message` instead of spitting out raw JSON.

VanyLaw:
> What is an example of an object error? Can we do better than simply stringifying it?

It can be a Errror or other object type exception. If error is a simple object type exception, we need to stringify it before we put into log.
image

Composer/plugins/azurePublish/src/index.ts
VanyLaw: done

Composer/plugins/azurePublish/src/index.ts
VanyLaw: done

Composer/plugins/azurePublish/src/index.ts
```diff
@@ -195,11 +195,16 @@ class AzurePublisher {
       }
     } catch (error) {
       console.log(error);
+      if (typeof error === 'object' && !(error instanceof Error)) {
```
beyackle, line 198: I think these `if` statements should have 3 branches:
- `if (error instanceof Error)` (where we log `error.message`),
- `else if (typeof error === 'object')` where we stringify it, and
- `else`, where we just push it.

The last two are the same as you have them here.

Composer/plugins/azurePublish/src/index.ts
```diff
@@ -319,7 +324,11 @@ class AzurePublisher {
       return response;
     } catch (err) {
       console.log(err);
-      this.logMessages.push(err.message);
+      if (typeof err === 'object' && !(err instanceof Error)) {
```
beyackle, line 327: See previous comment - the same should happen here.

Composer/plugins/azurePublish/src/index.ts
a-b-r-o-w-n: To add to this, this logic should be extracted into a single method.

Composer/plugins/azurePublish/src/index.ts
VanyLaw: done

Composer/plugins/azurePublish/src/index.ts
VanyLaw: done

PR #3131, opened 2020-05-19
ci: add base path for coveralls action

a-b-r-o-w-n:
## Description

Fixes coveralls reporting. Report [here](https://coveralls.io/builds/30892746)

## Task Item

refs #2570

## Screenshots

image


Composer/package.json
```diff
@@ -86,7 +86,7 @@
     "@typescript-eslint/parser": "2.10.0",
     "chalk": "^4.0.0",
     "concurrently": "^4.1.0",
-    "coveralls": "^3.0.7",
+    "coveralls": "^3.1.0",
```
a-b-r-o-w-n, line 89: This wasn't required to get the fix. I think the problem was that coveralls wasn't recognizing the file paths correctly since they were relative to the Composer directory, but the action is run in the root of the project. So something like `./packages/client/src/foo.ts` wasn't found in the project root, only from the `Composer` directory.

PR #3137, opened 2020-05-20
fix: Use IStorage for skill's id factory

### LINKED ISSUE #3133: Runtime should use IStorage for SkillConversationIdFactory
I looked at the runtime code for [SkillConversationIdFactory.cs](https://github.com/microsoft/BotFramework-Composer/blob/master/runtime/dotnet/core/SkillConversationIdFactory.cs#L19) and noticed that is currently using a ConcurrentDictionary to store skill conversation references.

This is OK for unit tests but for real bot we should use a version that uses IStorage (there is an example [here](https://github.com/microsoft/BotBuilder-Samples/blob/master/samples/csharp_dotnetcore/81.skills-skilldialog/DialogRootBot/SkillConversationIdFactory.cs).

In dev mode IStorage should be MemoryStorage but when deployed it should use CosmosDB to ensure the state is shared if there are several web apps running the same bot.

zidaneymar:
## Description

use IStorage rather than concurrent dictionary for id factory in Composer
## Task Item
fixes #3133


runtime/dotnet/core/SkillConversationIdFactory.cs
```diff
@@ -1,45 +1,78 @@
...
+        public override async Task<string> CreateSkillConversationIdAsync(SkillConversationIdFactoryOptions options, CancellationToken cancellationToken)
+        {
+            if (options == null)
+            {
+                throw new ArgumentNullException(nameof(options));
+            }
+
+            // Create the storage key based on the SkillConversationIdFactoryOptions.
+            var conversationReference = options.Activity.GetConversationReference();
+            var skillConversationId = $"{conversationReference.Conversation.Id}-{options.BotFrameworkSkill.Id}-{conversationReference.ChannelId}-skillconvo";
```
gabog, line 37: You can keep using the same key you had before if you want:

```
var key = $"{options.FromBotId}-{options.BotFrameworkSkill.AppId}-{skillConversationReference.ConversationReference.Conversation.Id}-{skillConversationReference.ConversationReference.ChannelId}-skillconvo";
```

zidaneymar:
> Check my comment and see if it makes sense.
> Also, you may want to write a unit test for this. You can use MemoryStorage for it.

I have added the unit tests, referring to the SkillHandlerTests in botbuilder-dotnet

runtime/dotnet/core/SkillConversationIdFactory.cs
```diff
@@ -1,45 +1,78 @@
...
+        public override async Task<string> CreateSkillConversationIdAsync(SkillConversationIdFactoryOptions options, CancellationToken cancellationToken)
+        {
+            if (options == null)
+            {
+                throw new ArgumentNullException(nameof(options));
+            }
+
+            // Create the storage key based on the SkillConversationIdFactoryOptions.
+            var conversationReference = options.Activity.GetConversationReference();
+            var skillConversationId = $"{options.FromBotId}-{options.BotFrameworkSkill.AppId}-{conversationReference.Conversation.Id}-{conversationReference.ChannelId}-skillconvo";
```
luhan2017, line 37: what does this "skillconvo" mean?

gabog: It is just an arbitrary string, so you know that is a skill conversation ID when debugging, it not required at all, is up to the dev what the ID looks like.

Issue #3134, opened 2020-05-20
showcode will mess up the dialog content across different dialog


hibrenda:
- create a empty bot
- create a new dialog A
- click show code
- switch between  Main (greeting triggers) and A (begin dialog) dialog, you will see the everything becomes into begin dialog 
- but sounds like both a dialog and the wrong begin dialog action are not serialized into disk

https://recordit.co/NSlcuxcReb
![image](https://user-images.githubusercontent.com/43408889/82393044-f2972980-9a77-11ea-8b0b-059a380f9e60.png)
![image](https://user-images.githubusercontent.com/43408889/82393072-03479f80-9a78-11ea-8557-bf909290dac8.png)
![image](https://user-images.githubusercontent.com/43408889/82393091-1195bb80-9a78-11ea-9582-d1729b36128d.png)

hibrenda:
By the way sounds like I hit the thread blocking, so after create the dialog A, this is not written to the disk. so that's why content has not been updated . I create another bot and did not hit thread blocking, so after show code messed the dialog content, all the errors are serialized into disk.

lei9444:
> By the way sounds like I hit the thread blocking, so after create the dialog A, this is not written to the disk. so that's why content has not been updated . I create another bot and did not hit thread blocking, so after show code messed the dialog content, all the errors are serialized into disk.

lei9444:
> By the way sounds like I hit the thread blocking, so after create the dialog A, this is not written to the disk. so that's why content has not been updated . I create another bot and did not hit thread blocking, so after show code messed the dialog content, all the errors are serialized into disk.

fix in #3138

xieofxie:
a real sad story

cwhitten:
https://github.com/microsoft/BotFramework-Composer/pull/3138

PR #3138, opened 2020-05-20
fix: show code mess up the dialog content and the persistence layer block issue

lei9444:
## Description

1. If the server return error when handling file update. The flushing flag was not set to false, then the task queue will never do flush.

2. When click show code button, the editor will get initial value. If we switch the dialog, the value will update to the editor, and the editor will call the onChange function to update dialog. The update function in design page will use a stale dialog id to update, so the previous dialog was covered by the new dialog.

In this PR I just re-render the the editor if the id changed. 

It works If we use a reference attached the dialog id to update the dialog too. But the ctrl-z will undo the update and the value will be covered again.

## Task Item
refs #3134


boydc2014:
Can you clarify more about why this will fix https://github.com/microsoft/BotFramework-Composer/issues/3134? Why the server will fails in that case?

boydc2014:
LGTM cc @cwhitten

PR #3139, opened 2020-05-20
feat: add RemoveRecipientMention config to runtime

### LINKED ISSUE #3104: feature: add middleware to remove @mentions for MSTeams

Messages from MS Teams include an @mention of the bots name at the beginning of every message.  This generally needs to be stripped off using `ActivityExtensions.RemoveRecipientMention()` otherwise `TextInput` actions will include it as part of the users response. I ejected my host and customized my `OnTurnAsync()` implementation to include the block below but we should really formalize this into a piece of middleware that can be turned on & off as a feature like the inspection middleware. 

```CS
       public override async Task OnTurnAsync(ITurnContext turnContext, CancellationToken cancellationToken = default(CancellationToken))
        {
            if (turnContext.Activity.Type == "message") {
                turnContext.Activity.RemoveRecipientMention();
            }
```

zidaneymar:
## Description

Add RemoveRecipientMention config to runtime, only works in Teams channel

## Task Item
Closes #3104


runtime/dotnet/core/Settings/BotFeatureSettings.cs
```diff
@@ -15,5 +15,8 @@ public class BotFeatureSettings
 
         // Use InspectionMiddleware
         public bool UseInspectionMiddleware { get; set; }
+
+        // Use RemoveRecipientMentionExtensions, only works in Teams channel
```
luhan2017, line 19: I don't think it only works in Teams channel, is it?

zidaneymar: > I don't think it only works in Teams channel, is it?

Yes, according to the code, it doesn't depend on the channel type. Let me fix the comment.
```
        public static string RemoveMentionText(this IMessageActivity activity, string id)
        {
            foreach (var mention in activity.GetMentions().Where(mention => mention.Mentioned.Id == id))
            {
                if (mention.Text == null)
                {
                    activity.Text = Regex.Replace(activity.Text, "<at>" + Regex.Escape(mention.Mentioned.Name) + "</at>", string.Empty, RegexOptions.IgnoreCase).Trim();
                }
                else
                {
                   activity.Text = Regex.Replace(activity.Text, Regex.Escape(mention.Text), string.Empty, RegexOptions.IgnoreCase).Trim();
                }
            }

            return activity.Text;
        }
```

PR #3141, opened 2020-05-20
fix: lg file mess up

### LINKED ISSUE #3136: LG parse may mess up the content in lg file
- create a dialog
- add sendactivity with "${$ xxx}"
- add more sendactivity steps with "hello", "welcome" ...
- back to the first sendactivity inline editing , quickly type "backspaces" 
- all the following  sendactivity content will be deleted. (eg. "hello", "welcome" will be deleted)

This will mess up the whole lg file content .

zhixzhan:
## Task Item

close #3136


Composer/packages/server/src/models/bot/botProject.ts
```diff
@@ -778,16 +778,6 @@ export class BotProject {
 
     for (const file of files) {
       const { realFilePath, content } = file;
-      await this.fileStorage.removeFile(realFilePath);
```
boydc2014, line 781: Remove this dangerous delete and write back is ok, but i think we can take the chance to totally disable this whole "tranformer" work for now,  which is done somewhat very ad-hoc. 

One top thing in R10 is about version, we should take the transformer work with that to have a better transformer. 

What do you think @cwhitten

zhixzhan: I agree remove  those two "transformer" methods, they already exist about 2 month on master, user's bot should already be migrated.

zhixzhan:
> Please also update the lg package to fix the parsing issue.

Sure, there are some upgrade issue haven't resolve, added `WIP` tag to hold this PR.

boydc2014:
> > Please also update the lg package to fix the parsing issue.
> 
> Sure, there are some upgrade issue haven't resolve, added `WIP` tag to hold this PR.

FYI, the js fix is in master, you should be unblocked now

zhixzhan:
> > > Please also update the lg package to fix the parsing issue.
> > 
> > 
> > Sure, there are some upgrade issue haven't resolve, added `WIP` tag to hold this PR.
> 
> FYI, the js fix is in master, you should be unblocked now

updated.

PR #3151, opened 2020-05-20
fix: Updated the skill manifest editor to not reorder properties

### LINKED ISSUE #3140: Generated Skill Manifest not correct
When exporting (strange action name for generating a manifest) as a skill the generated manifest is not correct. The $schema is below the name;
image

It should be:
image

When manually changing this it will automatically put it back in the wrong order again.
Power Virtual Agents won't accept the manifest with the wrong order, however this takes ages to figure out as it won't tell you anything about what was wrong. After eliminating step by step I found out the $schema was placed on the wrong line.

tdurnford:
## Description
Updated the skill manifest editor to order the manifest properties correctly and to allow the user to reorder the properties.

## Task Item
Closes #3140


Composer/packages/client/src/pages/design/exportSkillModal/content/Description.tsx
```diff
@@ -80,7 +81,9 @@ export const Description: React.FC<ContentProps> = ({ errors, value, schema, onC
   );
 
   useEffect(() => {
-    onChange({ $id: botName, name: botName, ...value });
+    if (!value.$id) {
+      onChange({ $schema, $id: botName, name: botName, ...rest });
```
srinaath, line 85: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify. So looks like stringify cannot guarantee order when run on structures that dont maintain order. Wonder if u want to sort the keys to make sure schema always comes up first or we let PVA do it.

cwhitten: @srinaath PVA is taking a bug to not enforce this constraint. I think we're OK here.

PR #3155, opened 2020-05-21
refactor: clean some code

liweitian:
## Description
use rimraf to replace deleteFilesFrombottomToUp

## Task Item

#minor


Composer/packages/server/src/models/storage/localDiskStorage.ts
```diff
@@ -61,8 +63,12 @@ export class LocalDiskStorage implements IFileStorage {
...
 
-  async rmDir(path: string): Promise<void> {
-    await rmDir(path);
+  async rmDir(path: string, force?: boolean): Promise<void> {
+    if (force) {
+      await forceRmDir(path);
+    } else {
+      await rmDir(path);
+    }
   }
```
beyackle, line 72: `rmdir` and `rm -rf` are different operations. `rm -rf` is not "force rmdir to happen", it's "remove a directory and everything inside it", and it ought to be its own method here. I'm fine with using `rimraf` as a way to implement it, but the boolean switch to choose between the two is not the way I'd go. `rmdir` should stay as pure `rmdir`. As a side benefit, this means you won't have to change the type of `rmDir` here, so those extra `force` arguments you added elsewhere can go away.

Composer/packages/server/src/models/bot/botProject.ts
```diff
@@ -325,8 +325,7 @@ export class BotProject {
 
   public async deleteAllFiles(): Promise<boolean> {
     try {
-      await this.deleteFilesFromBottomToUp(this.dir);
-      await this.fileStorage.rmDir(this.dir);
+      await this.fileStorage.rmDir(this.dir, true);
```
beyackle, line 328: This is where you'd call the new method (which can just be an invocation of `rimraf`; see the other comment).

Composer/packages/server/src/models/bot/botProject.ts
```diff
@@ -363,31 +362,16 @@ export class BotProject {
...
-      const childStat = await this.fileStorage.stat(curPath);
-      if (childStat.isDir && curPath.startsWith(this.dir)) {
-        await this.deleteFilesFromBottomToUp(curPath);
-        await this.fileStorage.rmDir(curPath);
-      } else {
-        await this.fileStorage.removeFile(curPath);
-      }
-    }
-  }
-
```
beyackle, line 380: 👍 to deleting this.

Composer/packages/server/src/models/storage/localDiskStorage.ts
srinaath: Do we need rmDir though @beyackle as it deletes just an empty directory. What if we change the function to removeDirectory and it accepts an options object like 
```
{
  recurse: true,
  force: true
}
```
and implement stuff based on the options object

boydc2014:
Nice to see the cleanup. Can you take the chance to add more UT in the code you touched here if possible.

Composer/packages/server/src/models/bot/botProject.ts
boydc2014: +1, also can you also just delete AzureBLobStorage class which is not used and maintained anymore

Composer/packages/server/src/models/storage/localDiskStorage.ts
liweitian: will only use rimraf here to implement rmDir. I am adding some test cases to this pr.

boydc2014: @liweitian i think what Ben is suggesting here is to explicitly add a interface for rimraf

liweitian:
> Nice to see the cleanup. Can you take the chance to add more UT in the code you touched here if possible.

added UT to test delete all files function

Composer/packages/server/src/models/bot/botProject.ts
liweitian: deleted

Composer/packages/server/src/models/storage/localDiskStorage.ts
liweitian: add one more interface rmrfDir

PR #3159, opened 2020-05-21
refactor: add useForm hook

a-b-r-o-w-n:
## Description

We have various forms through out the client app which all handle validation in similar ways. I have been thinking we need a form hook for a while so I decided to whip up something as a POC.

## Task Item

#minor


boydc2014:
Looks neat, @liweitian please check this out

Composer/packages/client/src/components/CreationFlow/DefineConversation/index.tsx
```diff
@@ -146,22 +127,24 @@ const DefineConversation: React.FC<DefineConversationProps> = (props) => {
...
-      return;
-    }
+  const handleSubmit = useCallback(
+    (e) => {
+      e.preventDefault();
+      if (hasErrors) {
+        return;
+      }
+
+      onSubmit({
```
srinaath, line 141: We are memoizing this function if formData or hasErrors changes. But, it still is going to call the onSubmit even if there is no change to formData here though right. should we check for changes in formData or do u think its trivial change and it can continue to submit even if no changes were present

Composer/packages/client/__tests__/hooks/useForm.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+      expect(result.current.formErrors).toMatchObject({
+        customValidationField: 'custom',
+        asyncValidationField: 'async',
+      });
+    });
+
+    it('validates required fields', async () => {
+      const { result, waitForNextUpdate } = hooks.renderHook(() => useForm(fields));
+
+      await hooks.act(async () => {
```
srinaath, line 93: Hook testing is sweet!!

Composer/packages/client/src/pages/design/createDialogModal.tsx
```diff
@@ -32,47 +33,40 @@ export const CreateDialogModal: React.FC<CreateDialogModalProps> = (props) => {
...
   const { dialogs } = state;
   const { onSubmit, onDismiss, isOpen } = props;
-  const initialFormData: DialogFormData = { name: '', description: '' };
-  const [formData, setFormData] = useState(initialFormData);
-  const [formDataErrors, setFormDataErrors] = useState<{ name?: string }>({});
+  const formConfig: FieldConfig<DialogFormData> = {
+    name: {
+      required: true,
+      validate: (value) => {
+        const nameRegex = /^[a-zA-Z0-9-_.]+$/;
```
srinaath, line 40: can we move this constant outside at the module level scope rather than function level

Composer/packages/client/src/components/CreationFlow/DefineConversation/index.tsx
a-b-r-o-w-n: I think in this case it's fine. We just validate that the `formData` is valid before submitting. So when `hasErrors` changes, we need a new function that reads that value.

Composer/packages/client/src/pages/design/createDialogModal.tsx
a-b-r-o-w-n: I think since this is the only place that `nameRegex` is used, there is no need to promote to a module constant. I actually find that to be more of a cognitive load (needing to scan the file to see where the const is defined, etc). If this were used in multiple places in this module, moving it would be appropriate.

Composer/packages/test-utils/index.js
```diff
@@ -3,10 +3,12 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
 
 const testingLib = require('@testing-library/react');
+const hooksLib = require('@testing-library/react-hooks');
 
 const createConfig = require('./createConfig');
 
 module.exports = testingLib;
+module.exports.hooks = hooksLib;
```
a-b-r-o-w-n, line 11: I would like to have done something like 

```ts
import { act } from '@bfc/test-utils/hooks'`
```

But I couldn't figure out how to expose that in the typings correctly. If anyone has some TS wizardry, I would appreciate.

a-b-r-o-w-n:
Going to merge this to get the hooks testing utils.

PR #3165, opened 2020-05-22
fix: LUIS authoringKey used for endpoint calls instead of endpointKey

### LINKED ISSUE #3145: LUIS authoringKey used for endpoint calls instead of endpointKey
## Describe the bug

When testing a Bot through Composer locally, you provide an authoringKey enabling Composer to manage your LUIS resources. All good. However when testing, it should use an endpointKey if provided especially as using the authoringKey for an endPointKey has a limited quota which when expired would leave a customer stuck no 

## Version

1.0

## Browser

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [x] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

- In Bot Settings within Composer, provide an endpointKey in addition to the authoringKey
- Observe in Ibiza Portal->LUIS Key->Metrics that successful calls don't increase as you perform testing.

## Expected behavior

EndpointKey to be used instead of authoringKey and design work to ensure that developers understand that authoringKey should be used for authoring only, endpointKey should be provided for testing/prod

VanyLaw:
## Description

pass endpointKey if exist to runtime.

**Notice**:
The endpointKey must bound with your luis application. Can bound in Luis portal like:
image

Otherwise, it will cause the 403 error.
image

## Task Item

close #3145 

## Screenshots

endpointKey Resource will be call:
image


Composer/plugins/localPublish/src/index.ts
```diff
@@ -289,14 +289,8 @@ class LocalPublisher implements PublishPlugin<PublishConfig> {
       configList.push(config.MicrosoftAppPassword);
     }
     if (config.luis) {
-      if (config.luis.authoringKey) {
-        configList.push('--luis:endpointKey');
-        configList.push(config.luis.authoringKey);
-      }
-      if (config.luis.authoringRegion) {
```
cwhitten, line 296: I see this removed and it suggests that the `--luis:endpoint` is not necessary when the user supplies an authoringKey instead of endpointKey. Is this correct?

VanyLaw: endpoint is the URL base on luis region, it's not necessary because c# runtime will make it itself in `builder.UseLuisConfigAdapter().UseLuisSettings()` in Program.cs file (It's a feature of C# SDK).

Issue #3164, opened 2020-05-22
master branch can't build in windows now


lei9444:
## Describe the bug
![image](https://user-images.githubusercontent.com/39758135/82650789-c8976000-9c4d-11ea-9b97-f98209802881.png)

## Browser

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

cxbuilders:
had same issue, reverted back to a commit from a day ago to keep working

lei9444:
in Windows we use CRLF
![image](https://user-images.githubusercontent.com/39758135/82651016-29269d00-9c4e-11ea-909d-049be84dc5dd.png)

if we use lint:fix will commit a lot of files.

I not sure use 
`"endOfLine": "auto"`
in .prettierrc is a good Solution #3162

zxyanliu:
Same here. I was unable to build Composer using the master branch in Windows.

a-b-r-o-w-n:
@lei9444 I think we should enforce the files in git to be LF (can be done using a git attribute). This is the most cross platform friendly solution, but it may require windows users to reclone (according to https://prettier.io/docs/en/options.html#end-of-line).

Can you experiment by adding the following to a `.gitattributes` file in the repo?

```
* text=auto eol=lf
``` 

Since I'm on mac, this is already happening, but I am curious what Windows users will need to do if this setting is present. 

What about using the "auto" option? Did that not work? I see your PR #3162 is closed but without comment.

lei9444:
> @lei9444 I think we should enforce the files in git to be LF (can be done using a git attribute). This is the most cross platform friendly solution, but it may require windows users to reclone (according to https://prettier.io/docs/en/options.html#end-of-line).
> 
> Can you experiment by adding the following to a `.gitattributes` file in the repo?
> 
> ```
> * text=auto eol=lf
> ```
> 
> Since I'm on mac, this is already happening, but I am curious what Windows users will need to do if this setting is present.
> 
> What about using the "auto" option? Did that not work? I see your PR #3162 is closed but without comment.

Sorry, I have tested the "auto" option. It works, but this will make windows and mac users commit different line-endinds. I'm not sure it is good way to fix this issue, so I closed it.

timoconnellaus:
In lieu of this being addressed I was able to build the project on windows by running "npm run lint -- --fix" this command before running "yarn build"

[...]

PR #3178, opened 2020-05-25
fix: use git to replace crlf with lf

lei9444:
## Description
Add * text=auto eol=lf to the repo's .gitattributes file. But Windows users need to re-clone the repo after this change to ensure git has not converted LF to CRLF on checkout.

## Task Item
refs #3164

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf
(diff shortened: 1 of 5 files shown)


boydc2014:
I'm closing this one in favor of #3162

PR #3179, opened 2020-05-25
fix: add settings file ignore

### LINKED ISSUE #3105: Projects appsettings.json file checked in by default
## Describe the bug

There's no `.gitignore` file added to the root of new composer projects. This results in the projects `appsettings.json` being checked in by default.  That file can contain access tokens and authoring keys which makes checking it in a security risk. We should add a default `.gitignore` file to the root of the project that prevents this file from being checked in.
 
## Version
1.0.0

VanyLaw:
## Description

add appsettings.json ignore in root project. In order to prevent file be mis-uploaded.

## Task Item

close #3105


VanyLaw:
Hi @cwhitten I add gitignore file into our sample bot folder, so when user create a new bot, it will include the gitignore file. But it can only take effect in new bot. Do you think is it necessary to check the gitignore file existed or not in code when bot opened?

boydc2014:
> Hi @cwhitten I add gitignore file into our sample bot folder, so when user create a new bot, it will include the gitignore file. But it can only take effect in new bot. Do you think is it necessary to check the gitignore file existed or not in code when bot opened?

@VanyLaw , please use the feature of BoilerPlate, you just need to add one .gitignore in `plugins/samples/assets/shared`, it will be copied to generated bot

VanyLaw:
> > Hi @cwhitten I add gitignore file into our sample bot folder, so when user create a new bot, it will include the gitignore file. But it can only take effect in new bot. Do you think is it necessary to check the gitignore file existed or not in code when bot opened?
> 
> @VanyLaw , please use the feature of BoilerPlate, you just need to add one .gitignore in `plugins/samples/assets/shared`, it will be copied to generated bot

Got it, Thank you for reminding me.

Composer/plugins/samples/assets/projects/ToDoBotWithLuisSample/.gitignore
```diff
@@ -1,9 +0,0 @@
-# NuGet Packages
-*.nupkg
-# The packages folder can be ignored because of Package Restore
-**/[Pp]ackages/*
-# except build/, which is used as an MSBuild target.
-!**/[Pp]ackages/build/
-
-!Properties/launchSettings.json
-!packages/packages.json
```
a-b-r-o-w-n, line 9: Are we sure these should not also get ignored? The new `.gitignore` only ignores `appsettings.json`.

VanyLaw: I think those folder were stale, those code added in 7 month ago. I remember we had the C# runtime copy with sample bot previously, and it's relevant with those structure only.

PR #3188, opened 2020-05-26
fix: Fixed expression array styling

### LINKED ISSUE #3187: [PropertyEditor] Prompt validator field styling
##  Describe the bug
The items in the 'Validation Rules' array do not fill the full width of the Property Editor, and errors are rendered next to the field instead of below it.

## Version
master

## To Reproduce

1. Add a prompt
2. Go to the 'Other' tab
3. Add a validation with an error

## Expected behavior
The field should fill the full width of the editor, and errors should render below the field.

## Screenshots
fccde5ab-b55f-440b-bdad-17b11bfc6d8e

a16d8aa6-d96d-4d1f-a35b-8788168663ee

tdurnford:
## Description
Fixed the 'Validation Rules' array styling so the expression expands to use the full width of the property editor and errors render below the text field instead of next to it.

## Task Item
Closes #3187

## Screenshots
ca0007b2-7a78-4ff7-92e8-2f1f56ebf260

### DIFF
--- a/Composer/packages/extensions/adaptive-form/src/components/fields/ArrayField.tsx
+++ b/Composer/packages/extensions/adaptive-form/src/components/fields/ArrayField.tsx
@@ -68,6 +68,7 @@ const ArrayField: React.FC<FieldProps<any[]>> = (props) => {
             label={false}
             rawErrors={rawErrors[idx]}
             schema={itemSchema}
+            stackArrayItems={true}
             uiOptions={uiOptions}
             value={element.value}
             {...getArrayItemProps(arrayItems, idx, handleChange)}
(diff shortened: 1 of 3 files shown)


tdurnford:
@beyackle Yes, I agree - the `label` prop is not very intuitive and has room to be improved. Here's a quick overview of the prop's current behavior.

- `undefined`: If the `label` is `undefined`, it uses the default label from the SDK Schema
- `string`: Overrides the default label from the SDK Schema
- `false`: If the `label` is `false`, the label is hidden in the Property Editor. This is typically the case in arrays, so we don't have repeated labels.

PR #3193, opened 2020-05-27
refactor: clean up the 'slot' abstraction for abs-h

lei9444:
## Description
clean the code

## Task Item
#minor


Composer/packages/server/src/models/settings/interface.ts
```diff
@@ -4,7 +4,7 @@
 export const OBFUSCATED_VALUE = '*****';
 
 export interface ISettingManager {
-  get(slot?: string, obfuscate?: boolean): Promise<any | null>;
-  set(slot: string, settings: any): Promise<void>;
+  get(obfuscate?: boolean): Promise<any | null>;
```
boydc2014, line 7: Should we also take the chance of remove this `obfuscate` feature, @cwhitten what do you think? I think we chatted about this

cwhitten: Let's leave it in for now ad revisit once some customers have successfully deployed Composer and we get some additional feedback.

PR #3198, opened 2020-05-27
feat: enable js runtime (preview)

### LINKED ISSUE #1682: Output code in NodeJs
[Feature request]

We're working on NodeJS platform and have implemented, custom logging, custom dialogs etc.

The bot composer is great but I'm unable to use my existing code to integrate in it as the output code is in c#.

It'd be great if we have option to select output code language in bot composer as NodeJs allowing us to integrate existing codebase easily.

### LINKED ISSUE #3272: Enable Js runtime as azurewebapp

VanyLaw:
## Description
**Current Solution**
- currently, I use source code as runtime template.
    - in local publilsh, we copy runtime template and bot folder into temp folder. Do the npm install and build and start in temp folder.
    - in azure publish, we copy source code and bot folder into temp folder, build it and zip it and upload to azure. Do the npm install and start in azure. Azure web app only install node_modules in the first time deploy.
    - in eject, we copy source code, then doing npm install and npm build.

**Solution for switch different type of runtime (c#, JS) Maybe we need some UI design later**
- Currently, I add a **key** field in settings.runtime to clarify which runtime we are using. If we set this into **csharp-azurewebapp** we use c# runtime as template, set **node-azurewebapp** to use JS.

Enable js runtime:
- [x]  support eject js runtime and switch c# to js.
- [x]  support local publish
- [x]  support deploy to azure
- [x]  js runtime support both azure webapp and azure functions

## Task Item

close #3272
closes #1682 
merge #3453


Composer/plugins/localPublish/src/index.ts
```diff
@@ -201,21 +223,23 @@ class LocalPublisher implements PublishPlugin<PublishConfig> {
     await this.zipBot(dstPath, srcDir);
   };
 
-  private saveSkillManifests = async (dstPath: string, srcDir: string) => {
-    const manifestSrcDir = this.getManifestSrcDir(srcDir);
-    const manifestDstDir = this.getManifestDstDir(dstPath);
+  private saveSkillManifests = async (dstPath: string, srcDir: string, runtimeType: string) => {
+    if (runtimeType === 'C#') {
```
benbrown, line 227: why would this only happen for C# bots?

runtime/node/azurewebapp/src/index.ts
```diff
@@ -0,0 +1,109 @@
...
+    appPassword: process.env.microsoftAppPassword,// || '4ve=H:PYD?.NXn9jHORhW8e9mr9SSnm_',
+});
+adapter.use(new LanguageGeneratorMiddleWare(resourceExplorer));
+
+const bot = new DialogManager();
+bot.userState = new UserState(new MemoryStorage());
+bot.conversationState = new ConversationState(new MemoryStorage());
+bot.rootDialog = resourceExplorer.loadType(mainDialog) as AdaptiveDialog;
+
+// Find settings json file
```
benbrown, line 60: all this settings stuff should be moved into a function/class and out of this file. including it inline here makes the file much harder to understand.

runtime/node/azurewebapp/src/index.ts
```diff
@@ -0,0 +1,109 @@
...
+// Create adapter.
+// See https://aka.ms/about-bot-adapter to learn more about .bot file its use and bot configuration.
+const adapter = new BotFrameworkAdapter({
+    appId: process.env.microsoftAppID,// || '97f8a0c9-93d0-4bad-a3e3-30b1aaf79aa0',
+    appPassword: process.env.microsoftAppPassword,// || '4ve=H:PYD?.NXn9jHORhW8e9mr9SSnm_',
+});
+adapter.use(new LanguageGeneratorMiddleWare(resourceExplorer));
+
+const bot = new DialogManager();
+bot.userState = new UserState(new MemoryStorage());
```
benbrown, line 56: we will need to apply all the same configurable options to this bot as we did for C# - that is, ability to swap out storage providers, turn on middlewares, etc.

carlosscastro:
Great work, thanks!!! The general feedback is that we should never be asking `if is C#`. Instead, this should be segregated into separate plug ins. The result will be that 

- Platform specific code will be isolated in its own plugin
- Core code will be platform and technology agnostic
- Core code can be tested more easily
- Platform code can be tested in isolation
- Good pattern that scales when we have N runtimes rather than 2
- Sets the foundation for future external runtime plugin implementations.

carlosscastro:
**Folder organization and reuse comment:** The same way we have a core directory in dotnet, it would be ideal to have the following directory structure:

- `runtime/dotnet/azurewebapp`
- `runtime/dotnet/core`
- `runtime/dotnet/azurefunctions`
- `runtime/node/core` : here is pretty much all the code you have in azure web app right now
- `runtime/node/azurewebapp` : here is only a piece of code that references `runtime/node/core`, and only starts a server

Then, in the future we can add azure functions or other hosting options and reuse the code in core. This would be amazingly useful and seems to be a low hanging fruit.

Composer/packages/server/src/models/settings/defaultSettingManager.ts
```diff
@@ -47,6 +47,7 @@ export class DefaultSettingManager extends FileSettingManager {
       },
       runtime: {
         customRuntime: false,
+        name: pluginLoader.extensions.runtimeTemplates[0]?.name,
```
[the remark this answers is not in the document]
VanyLaw: Yes, agree. But do we need to set default as C# runtime? I mean I just want a value here.

Composer/plugins/localPublish/src/index.ts
VanyLaw: Yes, currently JS runtime and reference SDK not support manifest feature. I will implement it once the SDK support.

VanyLaw:
Hi @carlosscastro, @benbrown , Thanks for your comments. I will follow the C# runtime to modify it. 
But I have a few questions about switching different type of runtime to host in local. Do we need to support this? Or we just default to c# runtime, and eject other runtime whenever user want to use other runtimes? Currently, I just add a name field in settings and match it with name in runtimeTemplate in localpublish plugin, I know it's not a good way to handle, we need to figure out a better solution for this.

Besides, I don't know much about azure function. Could we host c# runtime as azure function locally? Because we just default by webapp in local publish.

Composer/packages/server/src/models/settings/defaultSettingManager.ts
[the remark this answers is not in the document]
benbrown: yes, c# should be the default for now.

benbrown:
> But I have a few questions about switching different type of runtime to host in local. Do we need to support this? Or we just default to c# runtime, and eject other runtime whenever user want to use other runtimes? 

Since the behavior of the runtime should not really be different, we should default to c# runtime for "managed" scenarios. Choice of platform should only matter in ejected scenario. Am I wrong?

benbrown:
I want to second Carlos' comment above - we need to do the effort to move the platform specific pieces into the plugins, so that platform specific details do not leak into core. Early next week I can propose some mechanisms to do this!

VanyLaw:
Hi @benbrown @carlosscastro, In order to remove `if is C#` in localPublish and azurePublish plugins, Maybe we need to add a subFolder under those two plugins, and extract share code.
maybe the folder structure like:
`azurePublish/shared`
`azurePublish/azurewebapp/shared`
`azurePublish/azurewebapp/dotnet`
`azurePublish/azurewebapp/node`

`azurePublish/azureFunctions/dotnet`
`azurePublish/azureFunctions/node`
`azurePublish/azureFunctions/shared`

And currently bot-deploy package was under `composer/package/lib` folder, which is not in the same workspace with plugins. Then it will cause some synchronized issue like https://github.com/microsoft/BotFramework-Composer/pull/3322. Maybe we can move it out of `/lib` folder and add it into `azurePublish/shared`. Does that make sense to you?

chon219:
> Hi @benbrown @carlosscastro, In order to remove `if is C#` in localPublish and azurePublish plugins, Maybe we need to add a subFolder under those two plugins, and extract share code.
> maybe the folder structure like:
> `azurePublish/shared`
> `azurePublish/azurewebapp/shared`
> `azurePublish/azurewebapp/dotnet`
> `azurePublish/azurewebapp/node`
> 
> `azurePublish/azureFunctions/dotnet`
> `azurePublish/azureFunctions/node`
> `azurePublish/azureFunctions/shared`
> 
> And currently bot-deploy package was under `composer/package/lib` folder, which is not in the same workspace with plugins. Then it will cause some synchronized issue like #3322. Maybe we can move it out of `/lib` folder and add it into `azurePublish/shared`. Does that make sense to you?

`bot-deploy` package is actually a bunch of scripts for manipulating azure resources and bot files, which is not used by any composer components. Thus `bot-deploy` should not be a part of composer's packages, it should be a shared library/script among plugins azurePublish/azureFunctionsPublish.
However, we didn't find a practical way to create a _shared _library__ among plugin packages, any suggestions? Thanks.

benbrown:
I want to make sure we are thinking about the compartmentalization of this functionality correctly:

* All of the functionality required for platform specific actions (build, run, deploy) should be part of the addRuntimeTemplate API (in a manner similar to how the eject feature is currently implemented)
* We will need to track which version of the runtime has been ejected
* We will need to introduce a way for Composer to understand which runtime to use as the default (currently C#)

I think we should take an opportunity to extract the call to `.addRuntimeTemplate` currently in the localPublish plugin into its own plugin called `dotnetRuntime`

Then we should create a parallel one called `jsRuntime` which contains the definition of that plugin,its eject steps, necessary build / run / deploy actions.

Composer/plugins/runtimes/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "main": "lib/index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "build": "tsc"
+  },
+  "extendsComposer": true,
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "adm-zip": "^0.4.14",
```
cwhitten, line 14: why are `adm-zip` and `archiver` both required dependencies?

Composer/plugins/runtimes/src/index.ts
```diff
@@ -0,0 +1,197 @@
...
+  composer.addRuntimeTemplate({
+    key: 'csharp-azurewebapp',
+    name: 'C#',
+    startCommand: 'dotnet run --project azurewebapp',
+    path: path.resolve(__dirname, '../../../../runtime/dotnet'),
+    build: async (runtimePath: string, _project: any) => {
+      // TODO: copy source into temporary folder
+      // copyDir(path.resolve(__dirname, '../../../../../runtime/dotnet'), runtimePath);
+
+      // do stuff
```
cwhitten, line 27: this is not a helpful comment

Composer/plugins/runtimes/src/index.ts
```diff
@@ -0,0 +1,197 @@
...
+export default async (composer: any): Promise<void> => {
+  // register the bundled c# runtime used by the local publisher with the eject feature
+  composer.addRuntimeTemplate({
+    key: 'csharp-azurewebapp',
+    name: 'C#',
+    startCommand: 'dotnet run --project azurewebapp',
+    path: path.resolve(__dirname, '../../../../runtime/dotnet'),
+    build: async (runtimePath: string, _project: any) => {
+      // TODO: copy source into temporary folder
+      // copyDir(path.resolve(__dirname, '../../../../../runtime/dotnet'), runtimePath);
```
cwhitten, line 25: let's not forget to do this

Composer/plugins/runtimes/src/index.ts
```diff
@@ -0,0 +1,197 @@
...
+      }
+      composer.log('Run dotnet build...');
+      const { buildOut, buildErr } = await exec('dotnet build', { cwd: runtimePath });
+      if (buildErr) {
+        throw new Error(buildErr);
+      }
+      composer.log('FINISHED BUILDING!');
+    },
+    run: async (project: any, localDisk: IFileStorage) => {
+      // do stuff
```
cwhitten, line 45: this is not a helpful comment

Composer/plugins/runtimes/src/index.ts
```diff
@@ -0,0 +1,197 @@
...
+        stdio: 'pipe',
+      });
+      const { install2Out, install2Err } = await exec('npm build', {
+        cwd: runtimePath,
+        stdio: 'pipe',
+      });
+      composer.log('BUILD COMPLETE');
+    },
+    run: async (project: any, localDisk: IFileStorage) => {
+      // do stuff
```
cwhitten, line 152: this function implementation is empty?

Composer/plugins/runtimes/tsconfig.json
```diff
@@ -5,9 +5,9 @@
     "declaration": true,
     "sourceMap": true,
     "esModuleInterop": true,
-    "skipLibCheck": true,
     "outDir": "./lib",
     "rootDir": "./src",
+    "skipLibCheck": true,
```
cwhitten, line 10: why was this change necessary?

runtime/node/lib/shared/composerBot.d.ts
```diff
@@ -0,0 +1,9 @@
+import { ResourceExplorer } from "botbuilder-dialogs-declarative";
+export declare class ComposerBot {
+    private dialogManager;
+    private readonly resourceExplorer;
+    private readonly rootDialogPath;
+    constructor(resourceExplorer: ResourceExplorer, rootDialog: string, settings: any);
+    private loadRootDialog;
+    onTurn: (context: any) => Promise<void>;
```
cwhitten, line 8: can we get the typed turnContext from the botbotbuilder-schema package?

runtime/node/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "node-",
```
cwhitten, line 2: this doesn't seem like a good name

runtime/node/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "version": "1.0.0",
+  "description": "",
+  "main": "./lib/webapp.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
```
cwhitten, line 7: what is the plan for test execution?

Composer/packages/client/src/pages/setting/runtime-settings/ejectModal.tsx
```diff
@@ -14,13 +14,14 @@ import { StoreContext } from '../../../store';
...
 export interface EjectModalProps {
-  ejectRuntime: (templateKey: string) => void;
+  ejectRuntime: (templateKey: string) => Promise<void>;
   hidden: boolean;
   closeModal: () => void;
 }
 
 export const EjectModal: React.FC<EjectModalProps> = (props) => {
   const [selectedTemplate, setSelectedTemplate] = useState<string | undefined>();
+  const [copying, setCopy] = useState(false);
```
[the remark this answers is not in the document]
VanyLaw: done

Composer/packages/client/src/store/action/publisher.ts
```diff
@@ -5,9 +5,38 @@ import formatMessage from 'format-message';
...
+  }
+  dispatch({
+    type: ActionTypes.RUNTIME_POLLING_UPDATE,
+    payload: null,
+  });
+};
+
+export const startPollingRuntime: ActionCreator = (store) => {
+  const botStatusInterval = store.getState().botStatusInterval;
+  const projectId = store.getState().projectId;
```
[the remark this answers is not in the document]
VanyLaw: done

Composer/packages/extensions/plugin-loader/src/pluginLoader.ts
```diff
@@ -109,6 +111,30 @@ export class PluginLoader {
...
   }
 
+  // return a reference to the plugin used by the app
+  public getRuntimeByProject(project): RuntimeTemplate {
+    const type = project.settings.runtime?.key || DEFAULT_RUNTIME;
+    const templates = this.extensions.runtimeTemplates.filter((t) => t.key === type);
+    if (templates.length) {
+      return templates[0];
+    } else {
+      throw new Error(`Support for runtime with name ${type} not available`);
```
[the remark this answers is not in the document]
VanyLaw: done

Composer/packages/extensions/plugin-loader/src/pluginLoader.ts
```diff
@@ -109,6 +111,30 @@ export class PluginLoader {
...
+  // return a reference to the plugin used by the app
+  public getRuntime(type: string | undefined): RuntimeTemplate {
+    if (!type) {
+      type = DEFAULT_RUNTIME;
+    }
+    const templates = this.extensions.runtimeTemplates.filter((t) => t.key === type);
+    if (templates.length) {
+      return templates[0];
+    } else {
+      throw new Error(`Support for runtime type ${type} not available`);
```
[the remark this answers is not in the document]
VanyLaw: done

runtime/node/package.json
VanyLaw: sorry about that, fixed it.

Composer/plugins/azurePublish/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "azurePublish",
+  "name": "plugin-azure-publish",
```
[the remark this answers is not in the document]
VanyLaw: I am not sure which is better, it's modified by Ben. But every time we change this name, the old deploy profile would throw error, because it's the key to find the plugin.

Composer/packages/client/src/store/reducer/index.ts
```diff
@@ -539,9 +552,16 @@ const publishFailure: (title: string) => ReducerFunc = (title) => (state, { erro
 const getPublishStatus: ReducerFunc = (state, payload) => {
   // the action below only applies to when a bot is being started using the "start bot" button
   // a check should be added to this that ensures this ONLY applies to the "default" profile.
-  if (payload.target.name === 'default' && payload.endpointURL) {
-    state.botStatus = BotStatus.connected;
-    state.botEndpoints[state.projectId] = `${payload.endpointURL}/api/messages`;
+  if (payload.target.name === 'default') {
```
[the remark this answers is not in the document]
VanyLaw: it's the default publish target name, which means localpublish. I add a const in shared folder to fix it.

Composer/plugins/azurePublish/src/schema.ts
```diff
@@ -125,9 +125,9 @@ const schema: JSONSchema7 = {
       },
       luis: {
         authoringKey: '<authoring key>',
-        authoringEndpoint: '<luis authoring endpoint>',
+        authoringEndpoint: '',
```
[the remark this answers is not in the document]
VanyLaw: because when user don't input a value, the placeholder will be set as a value. And in the server it will not use the default endpoint, and it will trigger a absolute path error.

Composer/plugins/runtimes/package.json
VanyLaw: remove the unused package.

Composer/plugins/runtimes/src/index.ts
VanyLaw: removed

Composer/plugins/runtimes/src/index.ts
VanyLaw: do it in publish plugin.

Composer/plugins/runtimes/src/index.ts
VanyLaw: removed

Composer/plugins/runtimes/src/index.ts
benbrown: I still need to move the implementation of the run into this plugin, but I wanted to get it into the code while I was updating the form of the API.  I will execute on this ASAP.

Composer/plugins/azurePublish/package.json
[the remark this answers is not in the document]
benbrown: this got changed when the files were moving back and forth - i think it should be "azurePublish"

Composer/plugins/azurePublish/src/luis.ts
```diff
@@ -0,0 +1,247 @@
...
+      const luisAppIds: any = {};
+
+      // Read in all the luis app id mappings
+      for (const luisConfigFile of luisConfigFiles) {
+        const luisSettings = await fs.readJson(luisConfigFile);
+        Object.assign(luisAppIds, luisSettings.luis);
+      }
+
+      // Create the base LUIS config object
+      // const luisConfig: any = {
```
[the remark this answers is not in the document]
benbrown: I will go in and clean this out - we no longer need to construct the configuration like this here.

Composer/plugins/azurePublish/package.json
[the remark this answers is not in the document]
VanyLaw: done

Composer/plugins/runtimes/tsconfig.json
VanyLaw: because the packages in node_modules would be checked without this, and may cause type check error.

runtime/node/lib/shared/composerBot.d.ts
VanyLaw: use TurnContext instead of any, and get it from botbuilder-core package.

runtime/node/package.json
VanyLaw: add a test project like dotnet runtime.

benbrown:
* The code in this runtime needs substantially more inline code documentation to explain what is happening.
* The main file is also quite long and contains lots of stuff about manipulating settings. We should think carefully about what code the developer sees when they open this in VS
* This need to also include a readme file of some sort that explains what it contains and how to configure it.

Composer/plugins/azurePublish/src/index.ts
```diff
@@ -34,281 +35,366 @@ interface PublishConfig {
...
+      } = config;
 
+      // get the bot id from the project
+      const botId = project.id;
+
+      // generate an id to track this deploy
+      const jobId = uuid();
+
+      // resource key to map to one provision resource
+      const resourcekey = md5([project.name, name, environment, settings?.MicrosoftAppPassword].join());
```
benbrown, line 393: it is probably not a great idea to use the app password this way even though we are hashing it.  should get a unique enough name from the project name, name, environment combo

Composer/plugins/localPublish/.eslintrc.js
```diff
@@ -1,10 +1,10 @@
...
-    rules: {
-        'security/detect-non-literal-fs-filename': 'off',
-    },
+  extends: ['../../.eslintrc.js'],
+  parserOptions: {
+    project: './tsconfig.json',
+    tsconfigRootDir: __dirname,
+  },
+  rules: {
+    // 'security/detect-non-literal-fs-filename': 'off',
```
benbrown, line 8: why did this get removed?

Composer/plugins/localPublish/src/index.ts
```diff
@@ -317,13 +388,15 @@ class LocalPublisher implements PublishPlugin<PublishConfig> {
 
     child.on('exit', (code) => {
       if (code !== 0) {
-        reject(erroutput);
+        // this.setBotStatus(botId, { status: 500, result: { message: erroutput } });
```
benbrown, line 391: Are there scenarios where the non-zero code does not indicate some sort of error?

How can we properly handle this? I don't like just silently eating this

runtime/dotnet/azurefunctions/Startup.cs
```diff
@@ -43,19 +43,16 @@ private IConfigurationRoot BuildConfiguration(string rootDirectory)
             var config = new ConfigurationBuilder();
 
             // Config precedence 1: root app.settings
-            config
-                .SetBasePath(rootDirectory)
-                .AddJsonFile("appsettings.json", optional: true, reloadOnChange: true)
-                .UseLuisConfigAdapter()
-                .UseLuisSettings();
+            config.SetBasePath(rootDirectory);
+            config.AddJsonFile("appsettings.json", optional: true, reloadOnChange: true);
```
benbrown, line 47: What is this additional appsettings.json file for?

Why do we need it as well as the one in settings/appsettings.json?

runtime/dotnet/azurewebapp/Program.cs
```diff
@@ -27,31 +22,20 @@ public static void Main(string[] args)
...
-                if (env.IsDevelopment())
-                {
-                    // Local Debug
-                    builder.AddJsonFile("appsettings.development.json", optional: true, reloadOnChange: true);
-                }
+                // Use Composer bot path adapter
+                builder.UseBotPathConverter(env.IsDevelopment());
 
                 var configuration = builder.Build();
+                builder.AddJsonFile("appsettings.json", optional: true, reloadOnChange: true);
```
benbrown, line 29: why 2 settings files?

runtime/node/src/webapp.ts
```diff
@@ -0,0 +1,192 @@
...
+import { AdaptiveDialogComponentRegistration } from "botbuilder-dialogs-adaptive";
+import { ResourceExplorer } from "botbuilder-dialogs-declarative";
+import { AuthenticationConfiguration, Claim, JwtTokenValidation, SimpleCredentialProvider, SkillValidation } from 'botframework-connector';
+import { ComposerBot } from "./shared/composerBot";
+import { BotSettings } from "./shared/settings";
+import { SkillConversationIdFactory } from './shared/skillConversationIdFactory';
+import { SkillsConfiguration } from './shared/skillsConfiguration';
+
+
+export const getProjectRoot = (): string => {
```
benbrown, line 17: all these helper methods should be moved into a different file and imported

runtime/node/src/webapp.ts
```diff
@@ -0,0 +1,192 @@
...
+  // get the root folder according to environment
+  if (process.env.node_environment === "production") {
+    return path.join(__dirname, "../azurewebapp/ComposerDialogs");
+  } else {
+    return path.join(__dirname, "../../");
+  }
+};
+
+export const getRootDialog = (folderPath: string): string => {
```
benbrown, line 26: move to helpers

runtime/node/src/webapp.ts
```diff
@@ -0,0 +1,192 @@
...
+  for (let file of files) {
+    if (file.endsWith(".dialog")) {
+      rootDialog = file;
+      break;
+    }
+  }
+  return rootDialog;
+};
+
+export const getBotAdapter = (settings: BotSettings, userState: UserState, conversationState: ConversationState): BotFrameworkAdapter => {
```
benbrown, line 39: move to helpers

runtime/node/src/webapp.ts
```diff
@@ -0,0 +1,192 @@
...
+    } catch (err) {
+      console.error(`\n [onTurnError] Exception caught in sendErrorMessage: ${ err }`);
+    }
+    await conversationState.clear(turnContext);
+    await conversationState.saveChanges(turnContext);
+  };
+  return adapter;
+}
+
+export const getSettings = (projectRoot: string): BotSettings => {
```
benbrown, line 70: move to helpers

runtime/node/src/webapp.ts
```diff
@@ -0,0 +1,192 @@
...
+    for (let file of generatedFiles) {
+      if (file.endsWith(".json")) {
+        const items = require(path.join(generatedPath, file));
+        settings.luis = Object.assign(settings.luis, items.luis); // merge luis settings
+      }
+    }
+  }
+
+  // load settings from arguments
+  for (let key in argv) {
```
benbrown, line 93: move to helpers or do this in more off the shelf way

runtime/node/src/webapp.ts
```diff
@@ -0,0 +1,192 @@
...
+    }
+  }
+  settings.MicrosoftAppId = settings.MicrosoftAppId || process.env.MicrosoftAppId;
+  settings.MicrosoftAppPassword = settings.MicrosoftAppPassword || process.env.MicrosoftAppPassword;
+  return settings;
+};
+
+// Create HTTP server.
+const server = restify.createServer();
+const argv = require("minimist")(process.argv.slice(2));
```
benbrown, line 116: require this at the top like normal libraries

runtime/node/src/webapp.ts
```diff
@@ -0,0 +1,192 @@
...
+// prefer the argv port --port=XXXX over process.env because the parent Composer app uses that.
+const port = argv.port || process.env.port || process.env.PORT || 3979;
+
+server.listen(port, (): void => {
```
benbrown, line 120: normally this is towards the bottom of the code file

runtime/node/src/webapp.ts
```diff
@@ -0,0 +1,192 @@
...
+  console.log(
+    `\nGet Bot Framework Emulator: https://aka.ms/botframework-emulator`
+  );
+  console.log(
+    `\nTo talk to your bot, open http://localhost:${ port }/api/messages in the Emulator.`
+  );
+});
+
+const memoryStorage = new MemoryStorage();
```
benbrown, line 129: all of this needs inline code documentation

runtime/node/src/webapp.ts
```diff
@@ -0,0 +1,192 @@
...
+const projectRoot = getProjectRoot();
```
benbrown, line 130: As much of this as possible should be moved inside the ComposerBot class so the constructor can have less parameters.

runtime/node/src/webapp.ts
```diff
@@ -0,0 +1,192 @@
...
+const resourceExplorer = new ResourceExplorer();
+resourceExplorer.addFolders(projectRoot, ["runtime"], false);
+resourceExplorer.addComponent(new AdaptiveDialogComponentRegistration(resourceExplorer));
+
+const conversationIdFactory = new SkillConversationIdFactory();
+const credentialProvider = new SimpleCredentialProvider(settings.microsoftAppId, settings.microsoftAppPassword);
+const skillClient = new SkillHttpClient(credentialProvider, conversationIdFactory);
+const defaultLocale = settings.defaultLanguage || 'en-us';
+
+const bot = new ComposerBot(
```
benbrown, line 149: so many parameters! why do these have to be defined outside and passed in? Are they used directly anywhere else?

runtime/node/src/webapp.ts
```diff
@@ -0,0 +1,192 @@
...
+  skillClient,
+  conversationIdFactory,
+  undefined,
+  rootDialog,
+  defaultLocale,
+  settings.feature && settings.feature.removeRecipientMention || false
+);
+
+// Create and initialize the skill classes
+const skillsConfig = new SkillsConfiguration(settings);
```
benbrown, line 162: all of this skills stuff should be in its own file

runtime/node/src/shared/composerBot.ts
```diff
@@ -0,0 +1,78 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { ActivityHandler, ActivityTypes, BotFrameworkClient, BotTelemetryClient, ConversationState, SkillConversationIdFactoryBase, StatePropertyAccessor, TurnContext, UserState } from "botbuilder";
```
benbrown, line 4: this file needs inline documentation

Composer/packages/client/src/pages/setting/runtime-settings/ejectModal.tsx
```diff
@@ -22,6 +22,8 @@ export interface EjectModalProps {
 
 export const EjectModal: React.FC<EjectModalProps> = (props) => {
   const [selectedTemplate, setSelectedTemplate] = useState<string | undefined>();
+  // const [copying, setCopying] = useState(false);
```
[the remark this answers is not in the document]
VanyLaw: because currently main branch split 2 dialogs to render the eject and ejecting. So the copying state is staled.

Composer/packages/client/src/recoilModel/persistence/FilePersistence.ts
```diff
@@ -76,7 +76,7 @@ class FilePersistence {
               clearInterval(timer);
               resolve(true);
             }
-          }, 100);
+          }, 500);
```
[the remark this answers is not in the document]
VanyLaw: revert.

Composer/plugins/localPublish/.eslintrc.js
VanyLaw: because Chris commented on this before, and seems like it's not necessary, so I remove this which default is 'on'.

Composer/plugins/azurePublish/src/index.ts
VanyLaw: done

runtime/dotnet/azurewebapp/Program.cs
VanyLaw: removed

runtime/dotnet/azurefunctions/Startup.cs
VanyLaw: remove the appsettings.json

Composer/plugins/runtimes/src/index.ts
```diff
@@ -95,7 +90,7 @@ export default async (composer: any): Promise<void> => {
       // write settings to disk in the appropriate location
       const settingsPath = path.join(publishFolder, 'ComposerDialogs', 'settings', 'appsettings.json');
       // Set the bot and root fields to `ComposerDialogs` - this points the runtime to the appropriate deployed location
-      // todo: are both necessary?
+      // root parameter is read by botbuilder SDK, bot parameter is read by dotnet runtime.
```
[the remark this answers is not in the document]
VanyLaw: remove those parameter,  get the root folder base on environment in runtime.

Composer/plugins/localPublish/src/index.ts
VanyLaw: add this back,  and remove event listener before killing process to solve the non-zero exit code.

boydc2014:
>  Notice Because we use daily build botbuilder-js, so please set your registry into myget before try.
>  FYI like npm config set registry https://botbuilder.myget.org/F/botbuilder-v4-js-daily/npm/

Is this a must have or can we put this in some config file?

Composer/plugins/azurePublish/src/luis.ts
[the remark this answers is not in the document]
VanyLaw: > > Notice Because we use daily build botbuilder-js, so please set your registry into myget before try.
> > FYI like npm config set registry https://botbuilder.myget.org/F/botbuilder-v4-js-daily/npm/
> 
> Is this a must have or can we put this in some config file?

No, because I set the registry as myget in .npmrc file in runtime/node folder. It also make sure the azure webapp service can download those packages after we deployed.

VanyLaw:
> > Notice Because we use daily build botbuilder-js, so please set your registry into myget before try.
> > FYI like npm config set registry https://botbuilder.myget.org/F/botbuilder-v4-js-daily/npm/
> 
> Is this a must have or can we put this in some config file?

No, because I set the registry as myget in .npmrc file in runtime/node folder. It also make sure the azure webapp service can download those packages after we deployed.

runtime/node/.npmrc
```diff
@@ -0,0 +1,2 @@
+registry=https://botbuilder.myget.org/F/botbuilder-v4-js-daily/npm
+always-auth=true
```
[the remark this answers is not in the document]
VanyLaw: Ok.

runtime/node/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "start:function": "node ./lib/function.js",
+    "build": "tsc"
+  },
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "botbuilder": "4.10.0-preview-151936",
+    "botbuilder-dialogs": "4.10.0-preview-151936",
+    "botbuilder-dialogs-adaptive": "4.10.0-preview-151936",
+    "botbuilder-dialogs-declarative": "4.10.0-preview-151936",
```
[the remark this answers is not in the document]
VanyLaw: Use the  **botbuilder-dialogs-adaptive@4.10.0-rc0-preview** in npmjs instead.

runtime/node/src/shared/helpers.ts
```diff
@@ -0,0 +1,194 @@
...
+    const port = argv.port || process.env.port || process.env.PORT || 3979;
+    return port;
+};
+
+/**
+ * Get path of bot project.
+ */
+export const getProjectRoot = (): string => {
+    // get the root folder according to environment
+    if (process.env.node_environment === "production") {
```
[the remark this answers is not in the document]
VanyLaw: the NODE_ENV will be set to production when yarn start, but we need to getProjectRoot specialize by local run or deploy. So we use this new environment instead.

runtime/node/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "node-runtime",
```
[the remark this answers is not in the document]
VanyLaw: Do you means we should update the version, like increase 1 automatically ?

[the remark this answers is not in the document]
a-b-r-o-w-n: No. We should just change the name of this package. Like `botname-node-runtime` or something. Not really necessary, just an idea.

VanyLaw:
> First round of review. I will continue tomorrow. Here are some asks though:
> 
> 1. The start command for the bot should be `npm start`
> 2. There is no schemas directory when ejecting
1. if we use npm start instead, it may cause issue when restart. Because the process about js runtime is more than one, and could not be kill completely to reuse the port.

2. Because the JS SDK have not supported schema yet. We can add this support as soon as the SDK updated.

Composer/plugins/runtimes/src/index.ts
```diff
@@ -0,0 +1,220 @@
...
+    path: path.resolve(__dirname, '../../../../runtime/dotnet'),
+    build: async (runtimePath: string, _project: any) => {
+      composer.log(`BUILD THIS C# PROJECT! at ${runtimePath}...`);
+      composer.log('Run dotnet user-secrets init...');
+      // TODO: capture output of this and store it somewhere useful
+      const { initOut, initErr } = await exec('dotnet user-secrets init --project azurewebapp', {
+        cwd: runtimePath,
+      });
+      if (initErr) {
+        throw new Error(initErr);
```
a-b-r-o-w-n, line 31: Is `initErr` already an Error instance? If so, we can just `throw initErr`

Composer/plugins/runtimes/src/index.ts
```diff
@@ -0,0 +1,220 @@
...
+  composer.addRuntimeTemplate({
+    key: 'csharp-azurewebapp',
+    name: 'C#',
+    startCommand: 'dotnet run --project azurewebapp',
+    path: path.resolve(__dirname, '../../../../runtime/dotnet'),
+    build: async (runtimePath: string, _project: any) => {
+      composer.log(`BUILD THIS C# PROJECT! at ${runtimePath}...`);
+      composer.log('Run dotnet user-secrets init...');
+      // TODO: capture output of this and store it somewhere useful
+      const { initOut, initErr } = await exec('dotnet user-secrets init --project azurewebapp', {
```
a-b-r-o-w-n, line 27: How does this work? The docs say that the return value of exec is `{ stdout, stderr }`.

Composer/plugins/runtimes/src/index.ts
```diff
@@ -0,0 +1,220 @@
...
+        cwd: runtimePath,
+      });
+      if (initErr) {
+        throw new Error(initErr);
+      }
+      composer.log('Run dotnet build...');
+      const { buildOut, buildErr } = await exec('dotnet build', { cwd: runtimePath });
+      if (buildErr) {
+        throw new Error(buildErr);
```
a-b-r-o-w-n, line 36: Same thing about the error here as well as the output from `exec`

Composer/plugins/runtimes/src/index.ts
```diff
@@ -0,0 +1,220 @@
...
+      }
+      composer.log('FINISHED BUILDING!');
+    },
+    run: async (project: any, localDisk: IFileStorage) => {
+      composer.log('RUN THIS C# PROJECT!');
```
a-b-r-o-w-n, line 41: Does this need to be implemented?

Composer/plugins/runtimes/src/index.ts
```diff
@@ -0,0 +1,220 @@
...
+      try {
+        const { stdout, stderr } = await exec(
+          `dotnet publish "${dotnetProjectPath}" -c release -o "${publishFolder}" -v q`,
+          {
+            cwd: runtimePath,
+          }
+        );
+        composer.log('OUTPUT FROM BUILD', stdout);
+        if (stderr) {
+          console.error('ERR FROM BUILD: ', stderr);
```
a-b-r-o-w-n, line 75: Use the composer logger here.

Composer/plugins/runtimes/src/index.ts
```diff
@@ -0,0 +1,220 @@
...
+        }
+      } catch (err) {
+        console.error('Error doing dotnet publish', err);
```
a-b-r-o-w-n, line 78: Composer logger.

Composer/plugins/runtimes/src/index.ts
```diff
@@ -0,0 +1,220 @@
...
+        overwrite: true,
+        recursive: true,
+      });
+
+      // write settings to disk in the appropriate location
+      const settingsPath = path.join(publishFolder, 'ComposerDialogs', 'settings', 'appsettings.json');
+      if (!(await fs.pathExists(path.dirname(settingsPath)))) {
+        fs.mkdirSync(path.dirname(settingsPath), { recursive: true });
+      }
+      fs.writeFileSync(settingsPath, JSON.stringify(settings, null, 2));
```
a-b-r-o-w-n, line 95: We use a weird mix of sync and async methods. Can we just use all async?

Composer/plugins/runtimes/src/index.ts
```diff
@@ -0,0 +1,220 @@
...
+      // update manifst into runtime wwwroot
+      if (mode === 'azurewebapp') {
+        const manifestDstDir = path.resolve(dstRuntimePath, 'azurewebapp', 'wwwroot', 'manifests');
+
+        if (await fs.pathExists(manifestDstDir)) {
+          await removeDirAndFiles(manifestDstDir);
+        }
+
+        if (await fs.pathExists(srcManifestDir)) {
+          copyDir(srcManifestDir, srcStorage, manifestDstDir, dstStorage);
```
a-b-r-o-w-n, line 139: Should this be `await`ed?

Composer/plugins/runtimes/src/index.ts
```diff
@@ -0,0 +1,220 @@
...
+        }
+      }
+    },
+  });
+
+  composer.addRuntimeTemplate({
+    key: 'node-azurewebapp',
+    name: 'JS (preview)',
+    startCommand: 'node ./lib/webapp.js',
```
a-b-r-o-w-n, line 148: Update start command to `npm start`

Composer/plugins/runtimes/src/index.ts
```diff
@@ -0,0 +1,220 @@
...
+    path: path.resolve(__dirname, '../../../../runtime/node'),
+    build: async (runtimePath: string, _project: any) => {
+      // do stuff
+      composer.log('BUILD THIS JS PROJECT');
+      const { installOut, installErr } = await exec('npm install --dev', {
+        cwd: runtimePath,
+        stdio: 'pipe',
+      });
+      const { install2Out, install2Err } = await exec('npm run build', {
```
a-b-r-o-w-n, line 157: If we are not using these values, then just wait the exec.

Composer/plugins/runtimes/src/index.ts
```diff
@@ -0,0 +1,220 @@
...
+      // do stuff
+      composer.log('BUILD THIS JS PROJECT');
+      const { installOut, installErr } = await exec('npm install', {
+        cwd: path.resolve(runtimePath, '../'),
+        stdio: 'pipe',
+      });
+      const { install2Out, install2Err } = await exec('npm run build', {
+        cwd: path.resolve(runtimePath, '../'),
+        stdio: 'pipe',
+      });
```
a-b-r-o-w-n, line 176: Same with these. Also, refer to my previous comment about the return value of `exec`

Composer/plugins/runtimes/src/index.ts
```diff
@@ -0,0 +1,220 @@
...
+      fs.writeFileSync(settingsPath, JSON.stringify(settings, null, 2));
+
+      composer.log('BUILD COMPLETE');
+      return path.resolve(runtimePath, '../');
+    },
+    eject: async (project: any, localDisk: IFileStorage) => {
+      const sourcePath = path.resolve(__dirname, '../../../../runtime/node');
+      const destPath = path.join(project.dir, 'runtime');
+      // const schemaSrcPath = path.join(sourcePath, 'azurewebapp/Schemas');
+      // const schemaDstPath = path.join(project.dir, 'schemas');
```
a-b-r-o-w-n, line 192: remove

Composer/plugins/runtimes/src/index.ts
```diff
@@ -0,0 +1,220 @@
...
+      if (!(await project.fileStorage.exists(destPath))) {
+        // used to read bot project template from source (bundled in plugin)
+        const excludeFolder = new Set<string>().add(path.resolve(sourcePath, 'node_modules'));
+        await copyDir(sourcePath, localDisk, destPath, project.fileStorage, excludeFolder);
+        // await copyDir(schemaSrcPath, localDisk, schemaDstPath, project.fileStorage);
```
a-b-r-o-w-n, line 197: remove

runtime/node/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "start": "node ./lib/webapp.js",
+    "start:function": "node ./lib/function.js",
+    "build": "tsc",
+    "lint": "eslint --quiet --ext .js,.ts,.tsx ./src",
+    "lint:fix": "npm run lint --fix"
+  },
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "@types/lodash": "^4.14.159",
```
a-b-r-o-w-n, line 18: dev dependencies.

runtime/node/src/shared/helpers.ts
```diff
@@ -0,0 +1,204 @@
...
+  const port = argv.port || process.env.port || process.env.PORT || 3979;
+  return port;
+};
+
+/**
+ * Get path of bot project.
+ */
+export const getProjectRoot = (): string => {
+  // get the root folder according to environment
+  if (process.env.node_environment === 'production') {
```
a-b-r-o-w-n, line 41: use `NODE_ENV` instead.

runtime/node/src/shared/helpers.ts
```diff
@@ -0,0 +1,204 @@
...
+  }
+
+  // load generated settings
+  const generatedPath = path.join(projectRoot, 'generated');
+  if (fs.existsSync(generatedPath)) {
+    const generatedFiles = fs.readdirSync(generatedPath);
+    for (const file of generatedFiles) {
+      if (file.endsWith('.json')) {
+        const items = JSON.parse(fs.readFileSync(path.join(generatedPath, file), 'utf8'));
+        settings.luis = merge(settings.luis, items.luis); // merge luis settings
```
a-b-r-o-w-n, line 68:
```suggestion
        settings.luis = merge({}, settings.luis, items.luis); // merge luis settings
```

`merge` mutates the source object.

runtime/node/src/webapp.ts
```diff
@@ -0,0 +1,49 @@
...
+// Configure skill endpoint.
+configureSkillEndpoint(server, adapter, bot, skillConversationIdFactory);
+
+// Configure manifests endpoint.
+configureManifestsEndpoint(server);
+
+// Get port and listen.
+const port = getServerPort();
+server.listen(port, (): void => {
+  console.log(`\nGet Bot Framework Emulator: https://aka.ms/botframework-emulator`);
```
a-b-r-o-w-n, line 47: These lines should only be printed in the dev environment.

Composer/packages/client/src/recoilModel/dispatchers/publisher.ts
```diff
@@ -87,6 +112,37 @@ export const publisherDispatcher = () => {
...
+  });
+
+  const stopPollingRuntime = useRecoilCallback((callbackHelpers: CallbackInterface) => async () => {
+    const { set, snapshot } = callbackHelpers;
+    const botStatusInterval = await snapshot.getPromise(botStatusIntervalState);
+    if (botStatusInterval) {
+      clearInterval(botStatusInterval);
+      set(botStatusIntervalState, undefined);
+    }
+  });
```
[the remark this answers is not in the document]
VanyLaw: remove in recoil and add it in TestController component.

Composer/packages/extensions/plugin-loader/src/composerPluginRegistration.ts
```diff
@@ -56,10 +56,19 @@ export class ComposerPluginRegistration {
   /**************************************************************************************
    * Publish related features
    *************************************************************************************/
-  public async addPublishMethod(plugin: PublishPlugin, schema?: JSONSchema7, instructions?: string) {
+  public async addPublishMethod(
+    plugin: PublishPlugin,
+    schema?: JSONSchema7,
+    instructions?: string,
+    customName?: string,
+    customDescription?: string
```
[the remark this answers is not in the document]
VanyLaw: move all into PublishPlugin interface.

Composer/packages/extensions/plugin-loader/src/composerPluginRegistration.ts
```diff
@@ -89,6 +98,16 @@ export class ComposerPluginRegistration {
     this.loader.extensions.runtimeTemplates.push(plugin);
   }
 
+  // return a reference to the plugin used by the app
```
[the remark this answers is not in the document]
VanyLaw: done

Composer/plugins/azurePublish/src/mergeDeep.ts
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) Microsoft Corporation.
```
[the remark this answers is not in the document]
VanyLaw: because it's deep merge two object, it can't be replace simply by lodash merge. I use lodash mergeWith re-implement this function instead.

Composer/plugins/runtimes/src/index.ts
VanyLaw: No, it's string.

Composer/plugins/runtimes/src/index.ts
VanyLaw: used { stderr: initErr } to map it now.

runtime/node/src/shared/helpers.ts
VanyLaw: the NODE_ENV will be set to production when yarn start, but we need to getProjectRoot specialize by local run or deploy. So we use this new environment instead.

Composer/plugins/runtimes/src/index.ts
VanyLaw: currently it's just a placeholder, we want to move the spawn process which to host the local runtime into this function in the next step.

Composer/plugins/runtimes/src/index.ts
VanyLaw: if we use npm start instead, it may cause issue when restart. Because the process about js runtime is more than one, and could not be kill completely to reuse the port.

Composer/plugins/azurePublish/src/deploy.ts
```diff
@@ -0,0 +1,184 @@
...
+            environment,
+            this.accessToken,
+            language,
+            luisEndpoint,
+            luisAuthoringEndpoint,
+            luisEndpointKey,
+            luisAuthoringKey,
+            luisAuthoringRegion,
+            luisResource
+          );
```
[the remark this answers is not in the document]
VanyLaw: will change it to object

Composer/plugins/azurePublish/src/deploy.ts
```diff
@@ -0,0 +1,184 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import * as path from 'path';
+
+import * as fs from 'fs-extra';
+import * as rp from 'request-promise';
```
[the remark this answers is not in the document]
VanyLaw: I tried to use axios instead, but fail to upload the zip file with it, I try to wrap it in formData but it didn't work.

cwhitten:
Additionally, if you already have an ejected runtime and want to move to another runtime technology (swapping between JS & C#), we get an error.

Screen Shot 2020-08-14 at 8 44 40 AM

Can we not error here, but inform the user that they already have a runtime at this location, and ask them to confirm if they really do want to replace the runtime directory?

luhan2017:
The csharp_azurewebapp not available is fixed. I met such error in my environment, and fixed that by installing the "Visual Studio C++ core features", not sure why this is needed, we need to figure out the prerequisite to eject the js runtime. 

image:

I still met this error to switch between C# and JS
```
If this problem persists, please file an issue onGitHub
Details
EBUSY: resource busy or locked, rmdir 'C:\Users\luhan\Documents\Composer\ToDoBotWithLuisSample-3\runtime'
```

VanyLaw:
> I'm seeing the following when creating new bots (browser/macOS):
> 
> Starting a newly created bot:
> Screen Shot 2020-08-14 at 8 22 28 AM
> Screen Shot 2020-08-14 at 8 23 15 AM
> 
> ejecting the runtime to select C# or JS:
> Screen Shot 2020-08-14 at 8 27 03 AM

@cwhitten Sorry about that, I check this and find it's an issue because package import error in runtime plugin. It's fixed now.

VanyLaw:
> The csharp_azurewebapp not available is fixed. I met such error in my environment, and fixed that by installing the "Visual Studio C++ core features", not sure why this is needed, we need to figure out the prerequisite to eject the js runtime.
> 
> image:
> 
> I still met this error to switch between C# and JS
> 
> ```
> If this problem persists, please file an issue onGitHub
> Details
> EBUSY: resource busy or locked, rmdir 'C:\Users\luhan\Documents\Composer\ToDoBotWithLuisSample-3\runtime'
> ```
The first issue is about npm install, I will check if it is because npm version.

The second issue is because we need to stop runtime before switch it, currently composer can only switch before runtime running. I will add this part later.

VanyLaw:
Hi @luhan2017, I add the stop bot before switch eject runtime, it should fix the `resource busy or locked`, could you have a try?

cwhitten:
Seeing the following when I try and move an ejected runtime from C# to JS
Screen Shot 2020-08-17 at 1 34 16 PM

VanyLaw:
Hi @cwhitten, Do you use nvm? Could you go into the runtime/node folder and run the `npm install` and `npm run build`? Seems like it's an issue because the node version is diff from the one in the PATH variable.

zhixzhan:
I tried out a work around  in MacOS Catalina 
1. Install full Xcode
2. Open Xcode and the CLT will install
3. Then run `sudo xcode-select -s /Applications/Xcode.app/Contents/Developer`

Next you we met another error 
```
--scripts-prepend-node-path` option to include the path for the node binary npm was executed with
```

To solve this

 4. Add config in `composer/.npmrc`
```
scripts-prepend-node-path=true
```

Finally I successfully added a `js runtime`
image

luhan2017:
@cwhitten , We had investigated on this and found that all the failures we met are related to gyp. Wenyi checked in a readme file which explain the prerequisite to run the js runtime. 
For windows, users need to install windows-build-tools, and for linux, users need to install xcode-select.

Currently the ejection process will do the npm install for user and it will fail with gyp without installing the prerequisite.
How about we just copy the runtime, and tell user to follow the instructions in readme to install and build.
Then come back to composer to start the runtime? 

I think the main purpose for ejection is to do the custom runtime, so it is nature for user to do the npm install & run build by their own.

VanyLaw:
Thanks for the info @zhixzhan, I add scripts-prepend-node-path=true in .npmrc to fix about the npm path issue. But for the `node-gyp` error, could only install xcode-select to fix? Like run `xcode-select --install`.

runtime/node/src/shared/helpers.ts
VanyLaw: changed

benbrown: NODE_ENV is set to production for what?

I am with @a-b-r-o-w-n here. NODE_ENV should not just be set to production by default. If this is an environment inherited from the Composer app, we should find a way around this but I don't think introducing a new name is the right fix.

zxyanliu:
I tested on my Window machine. I did not see any errors. 

Steps I followed: 
- set registry into myget: `npm config set registry https://botbuilder.myget.org/F/botbuilder-v4-js-daily/npm/`
- `git clone https://github.com/microsoft/BotFramework-Composer.git`
- `yarn`, `yarn build`, `yarn start`
- create an echo bot with Composer and tested in Emulator work
- Eject runtime 
image
- Then see
Picture1

I did not see any errors throughout the process. What are the next steps?

zhixzhan:
> Thanks for the info @zhixzhan, I add scripts-prepend-node-path=true in .npmrc to fix about the npm path issue. But for the `node-gyp` error, could only install xcode-select to fix? Like run `xcode-select --install`.

@VanyLaw Before that I have tried `Xcode-select -- install` or `softwareupdate --install --all` but both failed.

VanyLaw:
@zxyanliu Thanks you for the try. Because we use all the packages from npm, you can skip the "set registry step". After you eject the runtime, the next step is clicking the start bot button to start it, then you can test in your emulator.

zxyanliu:
> @zxyanliu Thanks you for the try. Because we use all the packages from npm, you can skip the "set registry step". After you eject the runtime, the next step is clicking the start bot button to start it, then you can test in your emulator.

After ejecting the JS runtime, I tested the echo bot.  

image

cwhitten:
End to end tests are failing the SaveAs spec. I'm seeing unexpected behavior when trying to Save As. The bot being copied is _not_ the currently loaded bot. 

Screenshot shows EmptyBot being loaded up to copy, but I currently have the TodoLuis bot open.
Screen Shot 2020-08-19 at 10 29 34 AM

zxyanliu:
> End to end tests are failing the SaveAs spec. I'm seeing unexpected behavior when trying to Save As. The bot being copied is _not_ the currently loaded bot.
> 

This behavior happens to: _ToDoWithLuis_,  _SimpleToDo_ and _Interruptions_ samples. Every time after you create an Empty Bot first, then create the three samples mentioned will load the Empty Bot template. 

This behavior happens in the main branch as well.

cwhitten:
> @cwhitten
> 
> > End to end tests are failing the SaveAs spec. I'm seeing unexpected behavior when trying to Save As. The bot being copied is _not_ the currently loaded bot.
> 
> This behavior happens to: _ToDoWithLuis_, _SimpleToDo_ and _Interruptions_ samples. Every time after you create an _Empty Bot_ first, then creating the three samples mentioned will load the _Empty Bot_ template.
> 
> This behavior happens in the main branch as well.

Got it. Thanks @zxyanliu 

Let's address this issue outside this PR

PR #3210, opened 2020-05-28
test: create option test cases

liweitian:
## Description
add test cases
## Task Item
#minor


Composer/packages/client/__tests__/components/CreationFlow/CreateOptions/index.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+    return render(
+      <StoreContext.Provider value={storeContext}>
+        <CreateOptions
+          path="create"
+          templates={templates}
+          onDismiss={handleDismissMock}
+          onNext={handleCreateNextMock}
+        />
+      </StoreContext.Provider>
+    );
```
a-b-r-o-w-n, line 46: You can use the `renderWithStore` utility function here.

Composer/packages/client/__tests__/components/CreationFlow/CreateOptions/index.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+        saveTemplateId: saveTemplateMock,
+      },
+      state: {
+        templateId: '',
+      },
+    };
+  });
+
+  it('should save empty bot template id', async () => {
+    component = renderComponent();
```
a-b-r-o-w-n, line 61: Since component is meant to be isolated in each test, make this a `const` instead of assigning.

PR #3218, opened 2020-05-29
fix: remove unnecessary lg parsing

zhixzhan:
## Description

current diagnostics can be get from shell in real-time, local parsing for diagnostics are no longer need, and even cause page bad performance.

## Task Item

#minor


zhixzhan:
> Can this happen in the lu editor as well?

Other place usage already be cleaned, this is the only place left.

Issue #3224, opened 2020-05-29
"Learn more" link in settings page is a 404


Andrea-Orimoto:
![image](https://user-images.githubusercontent.com/17152540/83288305-faea1580-a197-11ea-81af-959aa26aacd0.png)

zxyanliu:
Hey @Andrea-Orimoto Thanks for reporting the broken doc link. I have already fixed it in [PR3238](https://github.com/microsoft/BotFramework-Composer/pull/3238).

Issue #3222, opened 2020-05-29
HTTP Request properties "learn more" tooltips link botbuilder-samples


sgellock:
## Describe the bug
For the HTTP Request property sheet, the "Learn more" links for the "?" help icons all navigate to BotBuilder-Samples/experimental/common-expression-language/README.md file on GitHub

## Version
1.0.0

## Browser

Electron distribution
## OS
- [x] macOS
- [ ] Windows
- [ ] Ubuntu

## To Reproduce

1. Open a bot project
2. On the design canvas, add an HTTP Request by [+] -> Access external resources -> Send an HTTP Request 
3. From the property sheet for the HTTP Request invoke the tooltips for any of the following properties:  Url, body, result property, content type, or response type.
4. Click on the "Learn more" link found in the tooltip.

We launch a browser window that displays the boatbuilder-samples/experimental/common-expression-language README.md on GitHub.
 
## Expected behavior
Expect to see a browser window launched that takes me to appropriate documentation

zxyanliu:
Hey @sgellock Thanks for reporting the broken doc link. I have already fixed it in [PR3238](https://github.com/microsoft/BotFramework-Composer/pull/3238).

PR #3238, opened 2020-05-30
fix: update expression and publish doc links

zxyanliu:
## Description
Addressing issue #3222
Addressing issue #3224  
#minor

## Additional 
Tested by local build and both doc links reported in the issues now point to the right places.


cwhitten:
Thank you @zxyanliu !

PR #3266, opened 2020-06-01
refactor: remove tildes

beyackle:
## Description

This replaces tildes across the codebase with easier-to-read equivalents (e.g. `~array.indexOf(x)` becomes `array.includes(x)`).

## Task Item

refs #3307


Composer/packages/server/src/models/bot/luPublisher.ts
```diff
@@ -208,7 +208,7 @@ export class LuPublisher {
 
     //add the lu file that are not in interuption folder.
     files.forEach((file) => {
-      if (!~paths.indexOf(file.name)) {
+      if (!paths.includes(file.name)) {
```
srinaath, line 211: This becomes a O(n^2) with the lookup inside for loop. Can paths be a set?

Composer/packages/lib/indexers/src/dialogUtils/validation.ts
```diff
@@ -29,8 +29,8 @@ const isExpression = (value: string | boolean | number, types: string[]): boolea
 //TODO: returnType is number, schem type is string, need map or unify
 const checkReturnType = (returnType: ReturnType, types: string[]): string => {
   return returnType === ReturnType.Object ||
-    ~types.indexOf(ExpressionTypeMapString[returnType]) ||
-    (returnType === ReturnType.Number && ~types.indexOf(ExpressionType.integer))
+    types.includes(ExpressionTypeMapString[returnType]) ||
+    (returnType === ReturnType.Number && types.includes(ExpressionType.integer))
```
srinaath, line 33: cant this just be types[ExpressionType.integer] instead of doing includes everytime?

Composer/packages/server/src/models/bot/luPublisher.ts
srinaath: Figured most of this code was already there. If u have time for a little cleanup will be great

a-b-r-o-w-n:
@beyackle consider adding the [no-bitwise](https://eslint.org/docs/rules/no-bitwise) rule to our eslint config as well.

beyackle:
> @beyackle consider adding the [no-bitwise](https://eslint.org/docs/rules/no-bitwise) rule to our eslint config as well.

That's an excellent idea. I didn't know that existed, but I'll add it in.

Composer/packages/server/src/models/bot/luPublisher.ts
beyackle: Will do.

Composer/packages/lib/indexers/src/dialogUtils/validation.ts
beyackle: 👍

beyackle: Looking at this again, I'm not sure we can do anything about it. Types is an array but ExpressionType.integer is a string; we want to check if the string is a value in the array, so `includes` is appropriate here. The more correct thing here would be to make types a `Set<string>`, but that ought to be its own cleanup ticket later - it's a non-trivial amount of work.

a-b-r-o-w-n:
@srinaath can you revisit this PR today?

srinaath:
PR is good on my end. Once @a-b-r-o-w-n  approves i think we are good to merge

PR #3270, opened 2020-06-02
fix: suppress websocket error after connection lost

### LINKED ISSUE #3167: Something went wrong dialog displays in various contexts
## Describe the bug
If I leave version 1.0.0 of Composer running on my Mac overnight, in the morning a "Something went wrong!" dialog is displayed.  Clicking Ok closes the dialog, and best I can tell, the product has no adverse behavior that I can detect.

## Version
1.0.0

## Browser
For this repro, I'm running the Electron distribution on macOS

## OS

- [x ] macOS
- [ ] Windows
- [ ] Ubuntu

## To Reproduce

1. Launch Composer
2. Perform authoring activities. (unclear if this has any impact on the issue or not)
3. Leave Composer running overnight.  May also require more than 24 hours to reproduce.  0.0.3 reproduced it overnight.  1.0.0 seems a little more resilient and requires more idle time to repro.

## Screenshots
Screen Shot 2020-05-22 at 9 38 27 AM

## Additional context
I've gotten this regardless of which authoring activity page I'm on.  I've gotten it when I'm on the main authoring canvas.  I've also gotten it when I was on the Publish page.

zhixzhan:
## Description

LG/LU editor use WebSocket connect to LSP server, this may failed when page is idle or network is poor, and at this moment inline editor's inside suggestion/check would not work. 

we have auto re-connection, so mute the failed alert sounds reasonable.

## Task Item
close #3167

#### Connection lost & resume after editor open.
Untitled2

#### Connection lost & resume before editor open.
Untitled3

### DIFF
--- a/Composer/packages/lib/code-editor/src/LgEditor.tsx
+++ b/Composer/packages/lib/code-editor/src/LgEditor.tsx
@@ -47,8 +47,13 @@ declare global {
 async function initializeDocuments(lgOption: LGOption | undefined, uri: string) {
   const languageClient = window.monacoLGEditorInstance;
   if (languageClient) {
-    await languageClient.onReady();
-    languageClient.sendRequest('initializeDocuments', { uri, lgOption });
+    try {
+      await languageClient.onReady();
+      languageClient.sendRequest('initializeDocuments', { uri, lgOption });
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error('LSP WebSocket connection lost', error);
+    }
   }
 }
 
--- a/Composer/packages/lib/code-editor/src/LgEditor.tsx
+++ b/Composer/packages/lib/code-editor/src/LgEditor.tsx
@@ -1,26 +1,21 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import { listen, MessageConnection } from 'vscode-ws-jsonrpc';
 import get from 'lodash/get';
-import { MonacoServices, MonacoLanguageClient } from 'monaco-languageclient';
+import { MonacoServices } from 'monaco-languageclient';
 import { EditorDidMount } from '@monaco-editor/react';
 
 import { registerLGLanguage } from './languages';
-import { createUrl, createWebSocket, createLanguageClient } from './utils/lspUtil';
+import { createUrl, createWebSocket, createLanguageClient, SendRequestWithRetry } from './utils/lspUtil';
 import { BaseEditor, BaseEditorProps, OnInit } from './BaseEditor';
+import { LGOption } from './utils';
 
 const LG_HELP =
   'https://github.com/microsoft/BotBuilder-Samples/blob/master/experimental/language-generation/docs/lg-file-format.md';
 const placeholder = `> To learn more about the LG file format, read the documentation at
 > ${LG_HELP}`;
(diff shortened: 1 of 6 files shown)


Composer/packages/lib/code-editor/src/LgEditor.tsx
```diff
@@ -47,8 +47,13 @@ declare global {
...
   if (languageClient) {
-    await languageClient.onReady();
-    languageClient.sendRequest('initializeDocuments', { uri, lgOption });
+    try {
+      await languageClient.onReady();
+      languageClient.sendRequest('initializeDocuments', { uri, lgOption });
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error('LSP WebSocket connection lost', error);
+    }
```
a-b-r-o-w-n, line 56: This seems suspect to me to just swallow all errors here.... Are we sure there are no other error types here?

zhixzhan:
> I think I'd like to see something that clearly shows that these errors are, in fact, connection errors.
> 
> In the description you mention we already have retry logic, so it's a little confusing to me that this code would throw. I'm not super familiar with the language servers / monaco though so I take my thoughts lightly. At minimum I would expect something like:
> 
> ```ts
> try {
>   await languageClient.onReady();
>   languageClient.sendRequest('initializeDocuments', { uri, luOption });
> } catch (err) {
>   if (err instanceof ConnectionError) {
>     await languageClient.reconnect();
>   }
> }
> ```
> 
> This code could already be happening somewhere, but the error manifestation seems odd to me if so.

The retry logic happens after `initializeDocuments` executed, like
1. open a LSP editor, initializeDocument success.
2. connection lost
3. retry connecting
4. connection resume

Yes, It didn't cover initialize document phase-1 error, and this phase error caused an alert, and if phase-1 failed, retry connecting would not exec `initializeDocuments` again.

I should let retry cover initialize document, please HOLD.

zhixzhan:
@a-b-r-o-w-n updated, the screen gif shows two connection lost scenarios.

Composer/packages/lib/code-editor/src/LuEditor.tsx
```diff
@@ -91,6 +78,49 @@ const LuEditor: React.FC<LULSPEditorProps> = (props) => {
...
+
+        const m = monacoRef.current;
+        if (m) {
+          editor.addCommand(m.KeyMod.Shift | m.KeyCode.Enter, function () {
+            const position = editor.getPosition();
+            SendRequestWithRetry(languageClient, 'labelingExperienceRequest', { uri, position });
+          });
+        }
+
+        SendRequestWithRetry(languageClient, 'initializeDocuments', { luOption, uri });
```
a-b-r-o-w-n, line 109: Shouldn't this be awaited to ensure the document has been initialized?

a-b-r-o-w-n: @zhixzhan I am going to go ahead and take this but please respond here when you get back online and create a follow-up PR if this needs to be fixed.

zhixzhan: I checked code below `SendRequestWithRetry(languageClient, 'initializeDocuments', {`, nothing relied on it.
One effect is before `initializeDocuments `, user interact in LSP editor interface, all information needs come from LSP server ( diagnostics/suggustions...) would be empty, looks like writing in a plain text-editor.

Issue #3130, opened 2020-05-19
Support customize built-in expressions for LG and AdaptiveExpression


luhan2017:
## Describe the solution you'd like

https://github.com/microsoft/BotBuilder-Samples/blob/master/samples/csharp_dotnetcore/language-generation/20.extending-with-custom-functions/Bots/CustomFunctionBot.cs
In this sample, we support user to add a custom expression function in code. But if we add this in composer runtime, we can't use it in composer UX, since the function is not registered in the LSP.

Here is the piece of code I add in composer runtime:
```
           const string mySqrtFnName = "contoso.sqrt";
           // Add custom sqrt function
            Expression.Functions.Add(mySqrtFnName, (args) =>
            {
                object retValue = null;
                if (args[0] != null)
                {
                    double dblValue;
                    if (double.TryParse(args[0], out dblValue))
                    {
                        retValue = System.Math.Sqrt(dblValue);
                    }
                }
                return retValue;
            });
```

But I will see this error in the UX:

![image](https://user-images.githubusercontent.com/32191031/82315979-bc699380-99fe-11ea-8549-3c02095e265d.png)
![image](https://user-images.githubusercontent.com/32191031/82316028-cab7af80-99fe-11ea-84a4-359436d24d69.png)

The easy way might be we add those custom functions as a white list in composer?

didaskein:
Hi, 
Also a question linked to the custom Function, Could it be possible to have natively access to the locale of the user inside the code of the function ? 

I would like to avoid to add a parameter in the function (not very user friendly ;))

![image](https://user-images.githubusercontent.com/13362276/83766429-b91c0c00-a67c-11ea-8d38-81cdf7db771d.png)

Thank you.

didaskein:
Also 2 features evolutions request linked to custom function : 
  - Be able to Call Async code inside a custom function
  - Be able to return Activity (Speak,Event...) not only text

Thanks.

cwhitten:
Closing this workitem as complete - see https://github.com/microsoft/BotFramework-Composer/issues/3656 for continued investment in this area.

PR #3273, opened 2020-06-02
feat: support validate lg custom functions

lei9444:
## Description
1. If the expression contain custom functions, the validation will fail. 
We can collect all custom functions from lg files. 
Now use string template to filter the diagnostics and ignore it.

  - If the lg file content is like 
   image
  we can use '=foo.bar()' in dialog and will not show error

  - if namespace doesn't exist use file name as namespace, '=test.en-us.bar()'

2. Separate the dialog validate from parser. 

## Task Item
#minor


luhan2017:
@lei9444 , will this PR fix this issue? #3130 , please let me know if you need my help to verify this.

PR #3274, opened 2020-06-02
feat: support endpoint for luis publish

### LINKED ISSUE #2986: Need to support different endpoint for luis publish
User would like to publish the luis to mooncake. 
eg. chinaeast2 region is targeting https://chinaeast2.api.cognitive.azure.cn instead of https://${region}.api.cognitive.microsoft.com.  

So need to support different endpoint for luispublish
https://github.com/microsoft/botframework-cli/issues/778

### LINKED ISSUE #3228: Bad error message when trying to start a bot
## Describe the bug
I have cloned the https://github.com/microsoft/meeting_assistant repo. I opened up the Timer project in composer. I enter in my LUIS Key, AppID, App Password. It states.
bug_composer

## Version
v1.0.0

## Browser

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ x] Edge

## OS

- [ ] macOS
- [ X] Windows
- [ ] Ubuntu

## To Reproduce

1. Go to https://github.com/microsoft/meeting_assistant clone the repo.
2. Open in composer the MSMeeting-Timer bot which is under skils
3. Change setting to add in AppId, AppPassword, Luis Key
4. Start Bot.

lei9444:
## Description
lubuild use endpoint to publish now. This is a breaking change and make luis publish fail.

now add endpoint to setting.
if endpoint is empty use `https://${config.region}.api.cognitive.microsoft.com`
or use endpoint to publish

## Task Item
closes #2986
closes #3228
refs microsoft/botbuilder-dotnet#4011


Composer/packages/server/src/models/bot/luPublisher.ts
```diff
@@ -218,7 +222,7 @@ export class LuPublisher {
   private _loadLuConatents = async (paths: string[]) => {
     return await this.builder.loadContents(
       paths,
-      this.config?.defaultLanguage || '',
+      'en-us',
```
a-b-r-o-w-n, line 225: Why has this change been made? Seems unrelated to the PR.

cwhitten: @lei9444 we'll take this change but please follow-up here as it seems to hardcode 'en-us'

lei9444: Sorry, @a-b-r-o-w-n @cwhitten this is another issue @luhan2017 ever mentioned. And it is not a good solution to use hardcode here. I commit another PR to fix this #3281

PR #3293, opened 2020-06-03
fix: polish dotnet error message shown in client

### LINKED ISSUE #3182: Composer needs .NET Core SDK
## Describe the bug

Can't Start Bot

## Version

1.0.0

## Browser

image

Desktop Application

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Go to 'new'
2. Go to 'create from template'
3. Select a template and click on next
4. Click on Start Bot
5. See error

Composer needs .NET Core SDK

To run this bot, Composer needs .NET Core SDK. Learn more.

Install Microsoft .NET Core SDK

## Expected behavior

Bot should be hosted

## Screenshots

image

## Additional context

.net core sdk is already installed
`dotnet --version`
`3.1.300`

Reinstalling Bot Framework Composer doesn't solve the issue.

VanyLaw:
## Description

Previously our error messages in running dotnet commands will be easily replaced in client.
Solution: pipe the error message to throw to client, and filter those in client more accurrate.

## Task Item

close #3182


VanyLaw:
Hi @beyackle, I did the changes in this PR because I think If the dotnet core version is not larger than 3.0, it will broken in calling dotnet user-secrets command. And if not install dotnet core, it will cause error about install .net core sdk.
What's more, I change the inherit into pipe is because we not only console the detail error message, but also pass it to client, which make filter become possible.

PR #3302, opened 2020-06-03
fix: allow inputting path

### LINKED ISSUE #3009: File Open doesn't allow me to paste a path in
## Version

0.0.1

## Browser

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [x] Edge (chrome)

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Go to OPEN BOT
2. I have a bot in a known location.  I can't type in a path, I can't paste a path. 

## Expected behavior

To be able to PASTE in a path or TYPE A path

liweitian:
## Description
We can input a path now.

## Task Item
Closes #3009 
## Screenshots
aaaa


liweitian:
@a-b-r-o-w-n I got it. It looks like this is the same as inputting a path in the address bar of 'This PC'. 
@sangwoohaan let me know if you have any UX concerns on creating a new folder, at least we need a new button?

sangwoohaan:
I don't think you should be able to directly type to create a new folder, since there can be so many mistakes happening. I'd rather keep the direct typing only for navigating existing folders.
Here's UX flow I put together for creating a new folder
https://www.figma.com/file/LlTlh5vXwq91zjGnFtrUUR6l/Composer-master-design-spec-UI-library?node-id=2309%3A20258

Let me know if you have any feedback.

liweitian:
Hi @sangwoohaan,  do we need to enter the folder after creating it? Because in the design, I see the path also changed (under the label "location", it says it gets updated automatically ). But it is weird. I think user should click the folder then enter it just like the normal behavior when we create a folder on Windows.

sangwoohaan:
@liweitian Yes. So when you clicked 'create a new folder', a text field to enter a new folder name would appear in the file navigator part. As you type in the text field, 'location' path will automatically get updated according to what you typed in the file navigator. Hope this would make it clear.

cwhitten:
ping @liweitian

liweitian:
> ping @liweitian

Conflicts is solved. this pr is ready for review.

sangwoohaan:
Just checked the functionality, everything seems to be working as intended. Can we also add one more behavior that when you hit the escape key when you are typing the new folder name, it just cancels the new folder creation(instead of closing the entire pop-up dialog).

liweitian:
> Just checked the functionality, everything seems to be working as intended. Can we also add one more behavior that when you hit the escape key when you are typing the new folder name, it just cancels the new folder creation(instead of closing the entire pop-up dialog).

it is added now.

Composer/packages/client/src/components/CreationFlow/LocationBrowser/FileSelector.tsx
```diff
@@ -46,52 +62,147 @@ type SortState = {
...
+
+  const cancelEditOperation = () => {
+    setEditMode(EditMode.NONE);
+    setNameError('');
+    setCurrentPath(initialPath);
+    setFolderName('');
+  };
+
+  const handleKeydown = (e, index) => {
+    if (e.key === 'Enter' && index) {
```
a-b-r-o-w-n, line 150: What if `index` is 0?

Composer/packages/server/src/services/storage.ts
```diff
@@ -110,14 +111,47 @@ class StorageService {
...
   };
 
+  public createFolder = (path: string) => {
+    try {
+      if (!fs.existsSync(path)) {
+        fs.mkdirSync(path);
+      }
+    } catch (e) {
+      throw new Error(e);
+    }
```
a-b-r-o-w-n, line 138: No need to wrap in a try/catch if you are just going to re-throw the error.

Composer/packages/server/src/services/storage.ts
```diff
@@ -110,14 +111,47 @@ class StorageService {
...
+    const newPath = Path.join(path, newName);
+    try {
+      if (fs.existsSync(currentPath)) {
+        fs.renameSync(currentPath, newPath);
+      } else {
+        throw new Error(`The folder ${currentPath} does not exist`);
+      }
+    } catch (e) {
+      throw new Error(e);
+    }
```
a-b-r-o-w-n, line 152: Same thing here.

Composer/packages/client/src/components/CreationFlow/LocationBrowser/FileSelector.tsx
liweitian: This is on purpose. The folder[0] is the go to parent folder button. The handleKeydown function should never be triggered on index 0.

a-b-r-o-w-n: Let's make that more explicit in the code. `index > 0` with a comment explaining that.

liweitian: done

a-b-r-o-w-n:
@liweitian please address conflicts.

PR #3303, opened 2020-06-03
test: add test for useLgApi and useLuApi hooks

lei9444:
## Description
add unit tests

## Task Item
#minor

### DIFF
--- a/Composer/packages/client/__tests__/shell/lgApi.test.tsx
+++ b/Composer/packages/client/__tests__/shell/lgApi.test.tsx
@@ -1,0 +1,96 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { renderHook } from '@testing-library/react-hooks';
+import React from 'react';
+
+import { useLgApi } from '../../src/shell/lgApi';
+import { StoreContextProvider } from '../testUtils';
+
+describe('use lgApi hooks', () => {
+  const updateLgTemplate = jest.fn((value) => value.template.body);
+  const copyLgTemplate = jest.fn((value) => value.toTemplateName);
+  const removeLgTemplate = jest.fn((value) => value.templateName);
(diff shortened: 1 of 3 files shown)


Composer/packages/client/__tests__/shell/lgApi.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+    expect(updateLgTemplate).toHaveReturnedWith('update');
+  });
+
+  test('should call copy lg template action', () => {
+    const { result } = renderHook(() => useLgApi(), { wrapper });
+
+    result.current.copyLgTemplate('test.en-us', 'from', 'to');
+
+    expect(copyLgTemplate).toBeCalledTimes(1);
+    expect(copyLgTemplate).toHaveReturnedWith('to');
```
a-b-r-o-w-n, line 76: Instead of asserting what the mock returns, you should assert what the function under test returns.

```ts
const output = result.current.copyLgTemplate(...);

expect(output).toEqual('to');
```

Its also a good idea to assert what the mock was called with


```ts
expect(copyLgTemplate).toBeCalledWith({
  // ...
});
```

Composer/packages/client/__tests__/testUtils/renderWithStore.tsx
```diff
@@ -24,8 +24,17 @@ export function renderWithStore(subject, state: Partial<State> = {}, actions = {
       lgImportresolver: jest.fn(),
       lgFileResolver: jest.fn(),
       luFileResolver: jest.fn(),
+      ...resolvers,
     },
   };
 
-  return render(<StoreContext.Provider value={store}>{subject}</StoreContext.Provider>);
+  return <StoreContext.Provider value={store}>{children}</StoreContext.Provider>;
+};
```
a-b-r-o-w-n, line 32: This can be removed with my suggestion above.

Issue #3307, opened 2020-06-03
Refactor for code style and cleanliness


beyackle:
The current codebase has a lot of style and typing issues that we should be refactoring out in the name of code cleanliness. This issue is just here so we have a unified place to track PRs related to that instead of just relegating them all to "minor".

PR #3308, opened 2020-06-03
refactor: remove double negations and other misc cleanup

beyackle:
## Description

This replaces instances of `!!` with more appropriate logic and also cleans up some data structures that weren't being used efficiently.
## Task Item

refs #3307


Composer/packages/client/src/components/ProjectTree/index.tsx
```diff
@@ -107,17 +107,18 @@ export const ProjectTree: React.FC<IProjectTreeProps> = (props) => {
...
     return (
-      <span ref={props.group && props.group.data.isRoot && addMainDialogRef} role="grid">
+      <span ref={props.group?.data.isRoot && addMainDialogRef} role="grid">
         <TreeItem
           depth={0}
-          isActive={!props.group!.isCollapsed}
-          isSubItemActive={!!selected}
-          link={props.group!.data}
+          isActive={!props.group?.isCollapsed}
+          isSubItemActive={selected != null}
```
srinaath, line 120: Is use of null warranted here? Should these variable be assigned null in the first place. The decision could be part of our style guide decision as well. Like can we get through almost all scenarios with just using undefined

beyackle: The null check using != will check for either null or undefined (using !== is the one for specifically null values).

Composer/packages/client/src/pages/language-understanding/code-editor.tsx
```diff
@@ -39,9 +39,9 @@ const CodeEditor: React.FC<CodeEditorProps> = (props) => {
 
   const hash = props.location?.hash ?? '';
   const hashLine = querystring.parse(hash).L;
-  const line = Array.isArray(hashLine) ? +hashLine[0] : typeof hashLine === 'string' ? +hashLine : 0;
+  const line = Array.isArray(hashLine) ? parseInt(hashLine[0]) : typeof hashLine === 'string' ? parseInt(hashLine) : 0;
 
-  const inlineMode = !!intent;
+  const inlineMode = intent != null;
```
a-b-r-o-w-n, line 44: Are you sure that this is the same intended behavior as before? `''` will now set `inlineMode` to true.

Composer/packages/client/src/pages/publish/createPublishTarget.tsx
```diff
@@ -49,7 +49,7 @@ const CreatePublishTarget: React.FC<CreatePublishTargetProps> = (props) => {
     if (!newName || newName.trim() === '') {
       setErrorMsg(formatMessage('Must have a name'));
     } else {
-      const exists = !!props.targets?.find((t) => t.name.toLowerCase() === newName?.toLowerCase);
+      const exists = props.targets?.some((t) => t.name.toLowerCase() === newName?.toLowerCase);
```
a-b-r-o-w-n, line 52: `t.name.toLowerCase() === newName?.toLowerCase` pretty sure this is a bug. Should be `newName?.toLowerCase()`

Composer/packages/client/src/store/reducer/index.ts
```diff
@@ -466,8 +466,8 @@ const setUserToken: ReducerFunc<UserTokenPayload> = (state, user = {}) => {
   return state;
 };
 
-const setUserSessionExpired: ReducerFunc = (state, { expired } = {}) => {
-  state.currentUser.sessionExpired = !!expired;
+const setUserSessionExpired: ReducerFunc = (state, { expired }: { expired: boolean }) => {
```
a-b-r-o-w-n, line 469:
```suggestion
const setUserSessionExpired: ReducerFunc<{ expired: boolean }> = (state, { expired }) => {
```

a-b-r-o-w-n: Also, are you sure its safe to remove the default parameter value?

Composer/packages/client/src/utils/electronUtil.ts
```diff
@@ -5,5 +5,5 @@
  * Returns true if the client is embedded in the Composer Electron environment.
  */
 export function isElectron(): boolean {
-  return !!window.__IS_ELECTRON__;
+  return window.__IS_ELECTRON__ ?? false;
```
a-b-r-o-w-n, line 8:
```suggestion
  return Boolean(window.__IS_ELECTRON__);
```

It could be possible for `window.__IS_ELECTRON__` to be something other than a boolean.

Composer/packages/client/src/utils/fileUtil.ts
```diff
@@ -6,19 +6,19 @@ import moment from 'moment';
 import { FileTypes, SupportedFileTypes } from '../constants';
 import { File } from '../store/types';
 
-export function getExtension(filename?: string): string | any {
-  if (typeof filename !== 'string') return filename;
-  return filename.substring(filename.lastIndexOf('.') + 1, filename.length) || filename;
+export function getExtension(filename: any): string {
```
a-b-r-o-w-n, line 9: I don't understand why these changes were made. If we want better type safety we can write this as

```ts
function getExtension<T = string>(filename?: T): T | undefined
```

Composer/packages/extensions/extension/src/hooks/useLgApi.ts
```diff
@@ -11,11 +11,8 @@ export const useLgApi = (shellApi: ShellApi) => {
 
   const deleteLgTemplates = (lgFileId: string, lgTemplates: string[]) => {
     const normalizedLgTemplates = lgTemplates
-      .map((x) => {
-        const lgTemplateRef = LgTemplateRef.parse(x);
-        return lgTemplateRef ? lgTemplateRef.name : '';
-      })
-      .filter((x) => !!x);
+      .map((x) => LgTemplateRef.parse(x)?.name)
+      .filter((x) => x != null) as string[];
```
a-b-r-o-w-n, line 15:
```suggestion
      .filter(Boolean) as string[];
```

Composer/packages/extensions/visual-designer/src/utils/mergePluginConfig.ts
```diff
@@ -7,8 +7,8 @@ import { defaultFlowSchema } from '../schema/defaultFlowSchema';
 import { defaultFlowWidgets } from '../schema/defaultFlowWidgets';
 
 export const mergePluginConfig = (...plugins: PluginConfig[]): Required<FlowEditorConfig> => {
-  const externalWidgets = plugins.map((x) => x.visualSchema?.widgets).filter((x) => !!x);
-  const externalSchema = plugins.map((x) => x.visualSchema?.schema).filter((x) => !!x);
+  const externalWidgets = plugins.map((config) => config.visualSchema?.widgets).filter((x) => x != null);
```
a-b-r-o-w-n, line 10:
```suggestion
  const externalWidgets = plugins.map((config) => config.visualSchema?.widgets).filter(Boolean);
```

Composer/packages/extensions/visual-designer/src/utils/mergePluginConfig.ts
```diff
@@ -7,8 +7,8 @@ import { defaultFlowSchema } from '../schema/defaultFlowSchema';
...
+  const externalSchema = plugins.map((config) => config.visualSchema?.schema).filter((x) => x != null);
```
a-b-r-o-w-n, line 11:
```suggestion
  const externalSchema = plugins.map((config) => config.visualSchema?.schema).filter(Boolean);
```

Composer/packages/client/src/components/ProjectTree/index.tsx
srinaath: I agree with @a-b-r-o-w-n . !selected would solve what !=null solves right. We can use either === {value} or !value.

srinaath: Also do we see any use of null in our codebase. Should variables even be instantiated to null?

Composer/packages/client/src/utils/fileUtil.ts
```diff
@@ -6,19 +6,19 @@ import moment from 'moment';
...
+export function getBaseName(filename: any): string {
+  if (typeof filename === 'string') return filename.substring(0, filename.lastIndexOf('.'));
+  return filename;
 }
 
-export function upperCaseName(filename?: string): string | any {
-  if (typeof filename !== 'string') return filename;
-  return filename.charAt(0).toUpperCase() + filename.slice(1);
+export function upperCaseName(filename: any): string {
+  if (typeof filename === 'string') return filename.charAt(0).toUpperCase() + filename.slice(1);
```
srinaath, line 20: I personally like the syntax isString(filename). We leverage the entire lodash library. We could use it to have a cleaner check

Composer/packages/client/src/pages/design/exportSkillModal/index.tsx
```diff
@@ -113,7 +113,8 @@ const ExportSkillModal: React.FC<ExportSkillModalProps> = ({ onSubmit, onDismiss
               {buttons.map(({ disabled, primary, text, onClick }, index) => {
                 const Button = primary ? PrimaryButton : DefaultButton;
                 const buttonText = text();
-                const isDisabled = typeof disabled === 'function' ? disabled({ manifest: skillManifest }) : !!disabled;
+                const isDisabled =
+                  typeof disabled === 'function' ? disabled({ manifest: skillManifest }) : disabled ?? false;
```
srinaath, line 117: lodash's isFunction(disabled)?

Composer/packages/client/src/utils/fileUtil.ts
a-b-r-o-w-n: I actually prefer using primitives where it makes sense. No need to hide some pretty basic syntax behind a utility function.

That being said, I am still a fan of lodash for other, more complex, operations (map, reduce, etc) because they often work on 'collections` as opposed to just arrays.

srinaath: Yah. It was just a personal preference for me. I was writing apps that required support for older browsers. By just practice I was using all lodash functions if available to avoid things breaking. But we could go either way on typeof checks as they are primitive

Composer/packages/client/src/components/ProjectTree/index.tsx
a-b-r-o-w-n: What is the type of `let foo;`? This breaks type safety. Variable declarations must be assigned a value in the typescript world.

beyackle:
Closing this as outdated; this will be revisited later once we've gotten a more solid set of guidelines to follow.

PR #3326, opened 2020-06-04
fix: Cleaned up handling of external links in Electron.

tonyanziano:
## Description

Improves upon and reverts #3098

Instead of adding custom electron logic to the client side to handle opening external links, this PR listens for the `new-window` event on the Electron browser window's web contents to open external links in the user's default browser.

## Task Item

closes #3326


cwhitten:
Whoa! Nice.

cwhitten:
Here is what failed Component Governance:
[INFO] |Security Alerts
[INFO] |CVE-2020-7660 |serialize-javascript 2.1.2 | High

https://cxsecurity.com/cveshow/CVE-2020-7660

Composer/packages/electron-server/src/electronWindow.ts
```diff
@@ -63,4 +64,19 @@ export default class ElectronWindow {
...
+  private isExternalLink(url: string): boolean {
+    return /^http(s)?:\/\//.test(url);
+  }
+
+  /** Intercepts any requests to open a new browser window (window.open or <a target="_blank">) */
+  private onOpenNewWindow(event: Event, url: string) {
+    if (this.isExternalLink(url)) {
+      // do not open a new Electron browser window, and instead open in the user's default browser
+      event.preventDefault();
+      shell.openExternal(url, { activate: true });
```
beyackle, line 78: 🆒

Composer/packages/client/src/pages/home/ItemContainer.tsx
```diff
@@ -76,12 +74,7 @@ export const ItemContainer: React.FC<ItemContainerProps> = ({
     <DefaultButton
       css={[itemContainerWrapper(disabled), styles.container]}
       onClick={async (e) => {
-        // todo: clean this up
-        const { href } = rest as Partial<{ href: string }>;
-        if (openExternal) {
-          e.preventDefault();
-          return openExternalLink(href, { activate: true });
-        } else if (onClick && !href) {
+        if (onClick) {
```
beyackle, line 77: nitpick: `if (onclick != null)`

tonyanziano:
@cwhitten should the security vulnerability be fixed in a separate PR? This PR doesn't touch the `yarn.lock` or any dependencies

cwhitten:
> @cwhitten should the security vulnerability be fixed in a separate PR? This PR doesn't touch the `yarn.lock` or any dependencies

yeah, that's fine. @srinaath is looking at addressing some of these right now. He can include this in his work.

srinaath:
It's weid

> > @cwhitten should the security vulnerability be fixed in a separate PR? This PR doesn't touch the `yarn.lock` or any dependencies
> 
> yeah, that's fine. @srinaath is looking at addressing some of these right now. He can include this in his work.

Hmm the only reason might be they categorized it as vulnerable this morning? @tonyanziano

Composer/packages/electron-server/src/electronWindow.ts
srinaath: Super clean fix!

Composer/packages/electron-server/src/electronWindow.ts
```diff
@@ -46,6 +46,7 @@ export default class ElectronWindow {
     }
     this._currentBrowserWindow = new BrowserWindow(browserWindowOptions);
     this._currentBrowserWindow.on('page-title-updated', (ev) => ev.preventDefault()); // preserve explicit window title
+    this._currentBrowserWindow.webContents.on('new-window', this.onOpenNewWindow.bind(this));
```
srinaath, line 49: Do we want to bind the function outside and pass the bound function to this event?

tonyanziano: Would there be any benefit to doing that?

After reading [the docs for `on(event, listener)`](https://nodejs.org/api/events.html#events_emitter_on_eventname_listener) it sounds like it just adds the listener to the array of listeners for that event.
> Adds the listener function to the end of the listeners array for the event named eventName.

I don't think the bound function gets instantiated every time the event is fired.

srinaath: I was hoping that would be the case. We are good with it then. In that case there wouldn't be any benefit.

PR #3327, opened 2020-06-04
refactor: code cleanup phase 1

beyackle:
## Description

Phase one of some miscellaneous cleanup on our codebase. I'm starting with the `client/src/` directory and going through there alphabetically, which is why this contains what it does. The point here is just refactoring with no functional changes.

## Task Item

refs #3307


Composer/packages/client/src/components/CreationFlow/DefineConversation/index.tsx
```diff
@@ -45,7 +44,7 @@ interface DefineConversationProps
 
 const DefineConversation: React.FC<DefineConversationProps> = (props) => {
   const { onSubmit, onDismiss, onCurrentPathUpdate, saveTemplateId, templateId, focusedStorageFolder } = props;
-  const files = get(focusedStorageFolder, 'children', []);
+  const files = focusedStorageFolder.children ?? [];
```
a-b-r-o-w-n, line 47:
```suggestion
  const files = focusedStorageFolder?.children ?? [];
```

Composer/packages/client/src/components/CreationFlow/DefineConversation/index.tsx
```diff
@@ -72,17 +70,15 @@ const DefineConversation: React.FC<DefineConversationProps> = (props) => {
         }
 
         const newBotPath =
-          focusedStorageFolder && Object.keys(focusedStorageFolder as Record<string, any>).length
+          focusedStorageFolder != null && Object.keys(focusedStorageFolder as Record<string, any>).length
```
a-b-r-o-w-n, line 73:
```suggestion
          focusedStorageFolder !== null && Object.keys(focusedStorageFolder as Record<string, any>).length
```

Composer/packages/client/src/components/CreationFlow/DefineConversation/index.tsx
beyackle: Will do.

Composer/packages/client/src/components/CreationFlow/DefineConversation/index.tsx
beyackle: Good catch. Will fix.

PR #3329, opened 2020-06-04
fix: bump version of serialize-javascript to help tests pass

beyackle:
## Description

Our builds are failing because of a security warning. This change will bring in a newer version of the package that's throwing the warning that doesn't have the same issue.

## Task Item

#minor


Composer/package.json
```diff
@@ -4,7 +4,8 @@
   "private": true,
   "resolutions": {
     "@types/react": "16.9.23",
-    "mkdirp": "^0.5.2"
+    "mkdirp": "^0.5.2",
+    "serialize-javascript": "^3.1.0"
```
srinaath, line 8: Why is this library added here. Isn't this a transitive dependacy fixed by bumping webpack?

beyackle:
Closing this in favor of a more widespread upgrade.

beyackle:
Looks like we might need this after all.

PR #3362, opened 2020-06-09
a11y: reapply tooltip additions

### LINKED ISSUE #2006: [Accessibility][ Sensory Characteristics - Flow Chart]: Tooltips are not defined for ‘+’ and other controls.
**User Experience** 
Users having hearing impairment will get difficulties to know the use of the given controls as no text Sensory Characteristics is given for the controls present in calendar.

**Test Environment**
OS build: 2004 (19564.1005)
Screen Reader: Narrator

**Actual Result**
Tooltip of ‘+’ control is not defined.

**Expected:**
Tooltip’s should be defined for the controls which doesn’t have text associated with them.

**Repro Steps:**
3. Now, navigate to '+' button with tab in the flow chart.
4. Observe the issue.

**Note:**
Same issue observed throughout the website, tooltips are not defined for any of the controls.

**MAS Impacted:** [MAS1.3.3]( https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={1823da5a-e8fb-4e63-b881-7eccb312c358})

**Attachment:**
[tooltips are not defined for the + button.zip](https://github.com/microsoft/BotFramework-Composer/files/4225289/tooltips.are.not.defined.for.the.%2B.button.zip)

beyackle:
## Description

This is a reapplication of the changes from @corinagum 's work on #2691, after that old PR and the master branch diverged enough that trying to rebase just leading to a larger and larger mess over time.

## Task Item

refs #2006 

## Screenshots

See screenshots in #2691


Composer/cypress/integration/LUPage.spec.ts
```diff
@@ -8,7 +8,7 @@ context('LU Page', () => {
   });
 
   it('can open language understanding page', () => {
-    cy.findByTestId('LeftNav-CommandBarButtonUser Input').click();
+    cy.findByTestId('LeftNav-CommandBarButtonUser Input').click({ force: true });
```
a-b-r-o-w-n, line 11: Let's make this work without using `force`.

Composer/cypress/integration/NotificationPage.spec.ts
```diff
@@ -59,7 +59,7 @@ context('Notification Page', () => {
       cy.findByTestId('FieldErrorMessage').should('exist');
     });
 
-    cy.findByTestId('LeftNav-CommandBarButtonNotifications').click();
+    cy.findByTestId('LeftNav-CommandBarButtonNotifications').click({ force: true });
```
a-b-r-o-w-n, line 62: Ditto

Composer/packages/client/src/App.tsx
```diff
@@ -67,17 +71,19 @@ export const App: React.FC = () => {
...
+            <TooltipHost content={globalNavButtonText} directionalHint={DirectionalHint.rightCenter}>
+              <IconButton
+                ariaLabel={globalNavButtonText}
+                css={globalNav}
+                data-testid={'LeftNavButton'}
+                iconProps={{
+                  iconName: 'GlobalNavButton',
+                }}
+                onClick={() => {
+                  setSideBarExpand(!sideBarExpand);
```
a-b-r-o-w-n, line 83:
```suggestion
                  setSideBarExpand((current) => !current);
```

When using the current state to get to a new state, use the updater function instead.

beyackle: Will do.

Composer/cypress/integration/LUPage.spec.ts
beyackle: I'm not sure we can. Without `force`, the e2e tests were failing because, while Cypress could see the element it wanted to click, it complained that the element was being covered by something, and some investigation showed that that something else was the tooltip's active region. I could see if making Cypress click on the tooltip region works, but that doesn't seem like a natural way to test the behavior we actually want.

beyackle: On further inspection, the e2e tests are still failing even with `force`, so I'll figure out another way to do this.

a-b-r-o-w-n: Thanks for looking at this.

beyackle: It looks like those failures were transient. They were definitely failing without `force`, though, so I think we still might need it.

a-b-r-o-w-n: We should figure out why we need it. To me, this lowers my confidence in these tests because it could lead to false positives.

Composer/cypress/integration/NotificationPage.spec.ts
```diff
@@ -41,7 +41,7 @@ context('Notification Page', () => {
     cy.findAllByText('__TestToDoBotWithLuisSample').should('exist');
   });
 
-  it('can show dialog expression error ', () => {
+  it.only('can show dialog expression error ', () => {
```
a-b-r-o-w-n, line 44: whoops.

Composer/cypress/integration/LUPage.spec.ts
```diff
@@ -15,12 +15,12 @@ context('LU Page', () => {
...
     cy.get('.toggleEditMode button').should('not.exist');
 
-    // by default is table view
+    // by default it goes to table view
     cy.findByTestId('LUPage').findByTestId('table-view').should('exist');
 
     // nav to ToDoBotWithLuisSample.main dialog
     cy.findByTestId('ProjectTree').within(() => {
-      cy.findByText('__TestToDoBotWithLuisSample').click();
+      cy.findByText('__TestToDoBotWithLuisSample').click(64, 0);
```
a-b-r-o-w-n, line 23: Since this is not a typical usage of `click`, can you add a comment why this is necessary?

Composer/cypress/integration/NotificationPage.spec.ts
beyackle: Oops. Yeah, I'll fix that.

Composer/cypress/integration/LUPage.spec.ts
beyackle: Will do.

beyackle: Found another way to fix this.

PR #3374, opened 2020-06-10
refactor: add typing to toolbar items

beyackle:
## Description

Adds some type cleanliness to the ToolBar component and the things that use it. This lets us take advantage of having a global store of actions instead of having to wire them through props everywhere.

## Task Item

refs #3307


Composer/packages/client/src/components/ToolBar/index.tsx
```diff
@@ -30,29 +54,25 @@ function itemList(action, index) {
...
-  if (toolbarItems && toolbarItems.length > 0) {
-    left = toolbarItems.filter((item) => {
-      return item.align === 'left';
-    });
-    right = toolbarItems.filter((item) => {
-      return item.align === 'right';
-    });
+    actions: { openNewTriggerModal, onboardingAddCoachMarkRef, showSkillManifestModal, createDialogBegin, exportToZip },
+    state: { projectId },
+  } = useContext(StoreContext);
```
a-b-r-o-w-n, line 62: Check out the `useStoreContext` helper hook. This makes stubbing the store context much easier in test.

beyackle:
E2E tests are failing and not for a transient reason. So this is going to take some time to fix before it's ready to review again.

Composer/packages/client/src/components/ToolBar/index.tsx
beyackle: 👍

PR #3375, opened 2020-06-10
docs: Update download links to match latest version

bartbilliet:
## Description

I'm proposing to change the download reference to the latest releases path on Github (URL: BotFramework-Composer/**releases/latest**). 

Currently, the individual download links for Windows, Mac and Linux use an aka.ms reference which is still pointing to version 1.0.0. It would require continuous updating of the aka.ms links to always refer to the latest releases. 
I believe pointing directly to the latest releases on Github would be more manageable in the long run, but feel free to let me know if updating the aka.ms links is still preferred for other reasons.

#minor


cwhitten:
hey @bartbilliet can we keep the direct links and add a link to `releases/latest`? we will be better about updating those aka.ms links.

bartbilliet:
@cwhitten Just making sure I understand well, your proposal is to keep the different aka.ms links, but change their target URL so that they refer to a dynamic `releases/latest` path for every OS? I thought about that, but the distributable filenames include the version number, so I think the target path of the aka.ms link would somehow still need to get updated at every release?

image

Or did you try to say something else? 

(e.g. Keep the aka.ms links as they are in the current docs, but *add* a line of text indicating something along the lines "Click here to learn about our most recent releases of Composer"..?)

cwhitten:
@bartbilliet I'm saying we should keep the links as is (we'll be better updating them with as we release new stable version) and _add_ a link to `releases/latest` in the docs

bartbilliet:
@cwhitten I committed an update based on your feedback. Feel free to make adjustments, or close the PR if you believe the link to `releases/latest` is better added on different docs.microsoft.com pages (The repo *composer-docs-pr* is private).

PR #3383, opened 2020-06-11
refactor: update more components

beyackle:
## Description

This finishes up the Components directory, mostly fixing typos and a small validation bug, and converting Tree to TSX.

## Task Item

refs #3307


Composer/packages/client/src/Onboarding/index.tsx
```diff
@@ -35,9 +35,9 @@ const Onboarding: React.FC = () => {
       .map((stepSet) => ({
         ...stepSet,
         steps: stepSet.steps.filter(({ targetId }) => {
-          if (!dialogs.length) {
+          if (dialogs.length > 0) {
             return !(targetId === 'mainDialog' || targetId === 'newTrigger' || targetId === 'action');
-          } else if (!dialogs[0].triggers.length) {
+          } else if (dialogs[0]?.triggers.length > 0) {
```
a-b-r-o-w-n, line 40: Shouldn't this be `=== 0` instead?

Composer/packages/client/src/components/Pagination/index.tsx
```diff
@@ -16,52 +17,52 @@ export interface IPaginationProps {
...
-  return options;
+  return range(pageCount).map((_, i) => ({
+    key: `page ${i}`,
+    text: `${i}`,
+  }));
 };
 
 export const Pagination: React.FC<IPaginationProps> = (props) => {
-  const [index, setIndex] = useState(0);
+  const [index, setIndex] = useState<number>(0);
```
a-b-r-o-w-n, line 27: Is number not inferred by the initial argument?

Composer/packages/client/src/components/Pagination/index.tsx
```diff
@@ -16,52 +17,52 @@ export interface IPaginationProps {
...
 
+  const changePage = (page: number) => {
+    setIndex(page);
+    onChange(page + 1);
+  };
+
   const handlePageSelected = (event: React.FormEvent<HTMLDivElement>, item?: IDropdownOption, index?: number) => {
-    setIndex(index || 0);
-    onChange(index ? index + 1 : 1);
+    if (index == null) return;
```
a-b-r-o-w-n, line 40: Double `==` here. Use `===` instead.

Composer/packages/client/src/components/Pagination/index.tsx
```diff
@@ -16,52 +17,52 @@ export interface IPaginationProps {
...
+  const handlePreviousClick = () => {
+    changePage(index - 1);
   };
 
-  const hangleNextClick = () => {
-    const current = index + 1;
-    setIndex(current);
-    onChange(current + 1);
+  const handleNextClick = () => {
+    changePage(index + 1);
```
a-b-r-o-w-n, line 49: Consider using `useCallback` for these two handlers as they have a dependency on `index`.

Composer/packages/client/src/components/SkillForm/CreateSkillModal/index.tsx
```diff
@@ -182,7 +181,7 @@ const CreateSkillModal: React.FC<ICreateSkillModalProps> = (props) => {
               onChange={updateForm('manifestUrl')}
             />
             {isValidating && (
-              <Spinner css={SpinnerLabel} label="validating..." labelPosition="right" size={SpinnerSize.medium} />
+              <Spinner css={SpinnerLabel} label="Validating..." labelPosition="right" size={SpinnerSize.medium} />
```
a-b-r-o-w-n, line 184:
```suggestion
              <Spinner css={SpinnerLabel} label={formatMessage('Validating...')} labelPosition="right" size={SpinnerSize.medium} />
```

Composer/packages/client/src/components/SkillForm/CreateSkillModal/index.tsx
```diff
@@ -62,12 +61,12 @@ const CreateSkillModal: React.FC<ICreateSkillModalProps> = (props) => {
   const validateForm = (newData: ISkillFormData): ISkillFormDataErrors => {
     const errors: ISkillFormDataErrors = { ...formDataErrors };
 
-    if (has(newData, 'manifestUrl')) {
+    if (newData.manifestUrl != null) {
```
a-b-r-o-w-n, line 64: triple equals here and below.

Composer/packages/client/src/components/SkillForm/CreateSkillModal/validateManifestUrl.ts
```diff
@@ -3,15 +3,15 @@
 
 import httpClient from '../../../utils/httpUtil';
 
-export async function validateManifestUrl(projectId: string, manifestUrl: any): Promise<string | undefined> {
-  // skip validation if has local other errors.
-  if (!manifestUrl) {
+export async function validateManifestUrl(projectId: string, manifestUrl: string): Promise<string | undefined> {
+  // skip validation if there are other local errors.
+  if (manifestUrl == null || manifestUrl === '') {
```
a-b-r-o-w-n, line 8: triple equals.

Composer/packages/client/src/components/Tree/index.tsx
```diff
@@ -3,17 +3,16 @@
 
 /** @jsx jsx */
 import { jsx } from '@emotion/core';
-import { PropTypes } from 'prop-types';
 
 import { container } from './styles';
 
-export const Tree = (props) => (
+type ITreeProps = {
```
a-b-r-o-w-n, line 9: Prefer interface over type.

Composer/packages/client/src/components/Tree/index.tsx
```diff
@@ -3,17 +3,16 @@
...
+  variant: string;
+  children: Element;
+};
+
+export const Tree = (props: ITreeProps) => (
```
a-b-r-o-w-n, line 14: In order to preserve component names in the bundled code, export components like:

```ts
const ComponentName: React.FC<ComponentPropsInterface> = (props) => { ... }
export { ComponentName };
```

Composer/packages/client/src/Onboarding/index.tsx
beyackle: Oops. Thanks for catching that.

Composer/packages/client/src/components/Pagination/index.tsx
beyackle: Yeah, it is. I don't know why I felt compelled to change this either.

Composer/packages/client/src/components/Pagination/index.tsx
beyackle: I think this is one place I feel comfortable carving out an exception. `== null` is currently the cleanest way there is to say "is `null` or `undefined`", and in TypeScript that's a common thing to say because of how it narrows a `?` type into a realized one.

Composer/packages/client/src/components/Pagination/index.tsx
beyackle: I don't think that would get us much. These handlers only fire if the index changes, so `useCallback` would just always go off. There's no way to take an action on the paginator that leaves the page unchanged except for the edge case of selecting it from the dropdown (and even them I'm not sure if the component fires its `onChange` if the same item gets selected as is already selected).

Composer/packages/client/src/components/SkillForm/CreateSkillModal/index.tsx
beyackle: Oops. Good catch.

Composer/packages/client/src/components/Tree/index.tsx
beyackle: Okay.

Composer/packages/client/src/Onboarding/index.tsx
```diff
@@ -35,9 +35,9 @@ const Onboarding: React.FC = () => {
       .map((stepSet) => ({
         ...stepSet,
         steps: stepSet.steps.filter(({ targetId }) => {
-          if (!dialogs.length) {
+          if (dialogs.length > 0) {
```
a-b-r-o-w-n, line 38: This needs to be `dialogs.length === 0`

beyackle: Fixed.

PR #3397, opened 2020-06-12
fix: add untracked settings to composer default settings

### LINKED ISSUE #3286: Question: Conversation logging
Is there any documentation for logging conversations with Composer?

We are porting a bot written in C# and logging is a hard requirement for us for auditing and repudiation reasons.

I see #2339 has raised this, the response appears to address storing bot state in Cosmos DB and not logging. I have set this up locally and all I see in my local Cosmos instance is bot state - no logs.

Thanks.

### LINKED ISSUE #3328: Need visibility into the publishing logs and errors
## Describe the bug

When publishing fails, there's very little information about the error available to the user, which makes problems very hard to debug. E.g.:
 
1. Create a bot with an ejected runtime and a custom action
2. In the custom action reference and external dll outside of the runtime folder (e.g. ..\..\mydll.dll)
3. Try publishing 

Result: Publish will fail but the only error will be that the dotnet build command failed. It fails because during the publish the runtime folder is copied to a temporary location, so the dll reference fails. However, this is not at all visible to the user so its really hard to diagnose.

I was able to find the error by running Composer from the source (yarn startall), which gave me visibility into the logs where I could clearly see the problem.

## Version

4.9

## Browser
Chrome

## OS

- [ ] macOS
- [X ] Windows
- [ ] Ubuntu

zidaneymar:
## Description
1. The default settings should include the cosmosdb settings and blobstorage settings.
2. Add missing logger for botproject deploy error

## Task Item

fixes #3328 
fixes #3286 (previously the blobstorage, application insights, cosmosdb are incomplete in settings)

## Screenshots

logging error in dotnet publish procedure
image

### DIFF
--- a/Composer/packages/lib/bot-deploy/src/botProjectDeploy.ts
+++ b/Composer/packages/lib/bot-deploy/src/botProjectDeploy.ts
@@ -492,7 +492,7 @@ export class BotProjectDeploy {
         const luisAppId = luisAppIds[k];
         this.logger({
           status: BotProjectDeployLoggerType.DEPLOY_INFO,
-          message: `Assigning to luis app id: ${luisAppIds}`,
+          message: `Assigning to luis app id: ${luisAppId}`,
         });
 
         const luisAssignEndpoint = `${luisEndpoint}/luis/api/v2.0/apps/${luisAppId}/azureaccounts`;
@@ -595,7 +595,10 @@ export class BotProjectDeploy {
         message: 'Publish To Azure Success!',
       });
     } catch (error) {
-      console.log(error);
+      this.logger({
+        status: BotProjectDeployLoggerType.DEPLOY_ERROR,
+        message: JSON.stringify(error, Object.getOwnPropertyNames(error)),
+      });
       throw error;
     }
   }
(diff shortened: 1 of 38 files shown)


Composer/packages/lib/bot-deploy/src/botProjectDeploy.ts
```diff
@@ -595,7 +595,10 @@ export class BotProjectDeploy {
         message: 'Publish To Azure Success!',
       });
     } catch (error) {
-      console.log(error);
+      this.logger({
+        status: BotProjectDeployLoggerType.DEPLOY_ERROR,
+        message: JSON.stringify(error, Object.getOwnPropertyNames(error)),
```
a-b-r-o-w-n, line 600: What is the intention of the second argument here? If I recall the api, it should be a replacer function, but `Object.getOwnPropertyNames` returns an array.

zidaneymar: > What is the intention of the second argument here? If I recall the api, it should be a replacer function, but `Object.getOwnPropertyNames` returns an array.

I'm trying to log down the native javascript exception error to json string, and referring to this answer on stackoverflow: https://stackoverflow.com/a/26199752

a-b-r-o-w-n: Ah ok. I learned something new about `JSON.stringify`!

PR #3417, opened 2020-06-15
doc: Added assessment of recoil

GeoffCoxMSFT:
## Description

Creating a PR of this document so that team can review and comment using the PR process.

## Task Item

No task.

## Screenshots

None.

#minor

### DIFF
--- a/docs/internal/GeoffsRecoilAssessment.md
+++ b/docs/internal/GeoffsRecoilAssessment.md
@@ -1,0 +1,176 @@
+This document describes my assessment of moving to [Recoil](http://recoiljs.org) for application state management based on converting an application to Recoil.  It provides the pros and cons I discovered, what I think influences the cost of adoption/conversion, my best practices, and an example of implementing the Dispatcher pattern.
+
+# Background : Assessment Process
+I have a React application (a text labeler) initially written using MobX. The application state followed the typical approach for MobX - a hierarchy of models which use @observable and @computed. Update to model state was done with @action methods in the models (although following a dispatch/handler pattern would have been an easy alternative). The React components for the app were a mixture of container components, presentation components, and pure components. The container components were wrapped with the MobX observer() method to respond to state changes.  The presentation components relied on prop changes and React.useEffect.  The pure components relied only on prop changes.
+
+I rewrote the application without MobX and using only using React's context API and useEffect to respond to state changes. The application state remained in a hierarchy, but without any observer/observable pattern.  The state was passed down the component hierarchy via context.  Container components detected state changes via additional useEffect calls. In order to coordinate larger reset of the application state (such as clearing most data), an application pub/sub mechanism was used to communicate between components to orchestrate the sequence of clearing state. This was because the useEffect order was difficult to guarantee.
+
+Finally, I rewrote the application to use Recoil. I flattened the application state into atoms. Because relationships between the data create a logical hierarchy, I wrote utility functions to make cross referencing between data islands easy. I then wrote selectors to shape the data for my React components. The container components leveraged useRecoilValue and useRecoilState to respond to state changes. Because the same application state change be updated from different React components, I implemented a dispatcher pattern based on useRecoilCallback.  I had several missteps as I learned how Recoil works, and discovered how my application was affected by flattening the state.
+
+# Recoil API: Pros, Cons, Unknowns, and Hazards
+
+## Overall
+- PRO: Built for hooks from the start and follows React concepts
+- PRO: Atom and selectors allow for composing applications from small pieces without introducing unnecessary dependencies.
+- PRO: Significantly reduces the complexity and code volume in components.
+- PRO: Looking toward the future of asynchronous, multi-threaded Javascript.
+- PRO: Recoil API surface is tiny
+- PRO: Lightning fast (at least for my application size)
+- CON: Still experimental, few releases, limited contributes - not yet mature or fully tested.
+
+## Atoms
+- PRO: Like their name, atoms let you model your data as small and independent plain types.
+- PRO: Serializing atoms to storage is straightforward compared to trying to rehydrate an observed state tree.
+- PRO: The default state of an atom with the reset() hook is convenient for clearing application state.
+- CON: Atoms are often arrays of items which means needing to look up items. This requires more undefined checks for cases where items are not found.
+- CON: There is additional "param baggage" to pass multiple atoms to different functions vs. passing the root of the application state or a child.
+- HAZARD: Care has to be taken to avoid updating atom state by mutating the current state.  Only a set will trigger change detection.
+
+## Selectors
+- PRO: Simple to shape data for React components. 
+- PRO: Like atoms they are small and can be built as an library of independent data shapes.
+- PRO: The dependency chain from one selector to other selectors or atoms is easy to follow.
+- PRO: Less ceremony compared to Redux's mapStateToProps.
+- UNKNOWN: Unsure if a selector would re-compute its value if no UI is subscribed to it currently but it had been subscribed before.
+- UNKNOWN: The writable selector is not well documented and unclear when to use.
+- HAZARD: Atom (and selector) key names need to be unique across the application. It can be easy to create an atom and selector with the same key.
+
+## useRecoilValue, useSetRecoilValue, useRecoilState hooks
+- PRO: The recoil state hooks cause rendering just like useState.
+- PRO: useRecoilValue and useSetRecoil value clearly denote the read-only and write-only scenarios.
+- PRO: No magic inspecting dot-notation usage like with MobX.  
+- HAZARD: Very easy to accidentally put a recoil hook within a condition. They need to be used as a block at the start of a component just like React hooks.
+
+## useRecoilCallback hook
+- PRO: Allows for creating methods that always get latest state even when called outside the context of a React component render. 
+- CON: It can be a bit of extra effort to get the Typescript types worked out for complex callbacks.
+- UNKNOWN: Unclear when to use the dependencies parameter vs. the getPromise to get latest state
+- UNKNOWN: Unclear when to use the synchronous loader vs the asynchronous getPromise.- UNKNOWN:? The callback documentation indicates is still possible this hook will change. This is likely due to the experimental portions relating to JS async threading.
+- HAZARD: Using getPromise within a callback requires the callback method be async, the return value be a promise, and usage outside UI callbacks use await.
+- HAZARD: Using set within a callback that is called multiple times in succession or called from other callbacks requires using updator function to ensure latest state.  Otherwise arrays may not be updated with the previous call and will be overridden - losing data.
+
+## RecoilRoot component
+- PRO: Simple to understand. It leverages the React context API, but more efficiently tracking atoms and selectors as keys in a dictionary.
+- HAZARD: The React component rendering RecoilRoot cannot use recoil hooks.
+? I didn't try nesting any RecoilRoots so I don't know if that really works.
+
+# Cost/Impact of moving to Recoil
+- Restructuring application state from observable hierarchy to flat arrays is painful and expensive. A major refactoring is likely to introduce bugs if there is insufficient testing on the existing code.
+- Converting deep dependencies on the context API down the component hierarchy is expensive especially when there are few pure functional components.
+- Creating selectors is straightforward. Many existing render calculations can be replaced incrementally with selectors.
+- Updating components to leverage recoil hooks rather than props with React.useEffect or React.useState is straightforward. Data passed through context can be updated with an atom or selector and then context usage can be replace incrementally down the component hierarchy.
+- Updating an existing dispatcher pattern is straightforward.
+
+# Tips and Best Practices
+
+## Follow functional programming paradigms
+- Always make state data immutable: All types used with atoms should have read only fields. 
+- Always write pure update functions: Write function that take read only inputs and return a new modified instance. This avoid repeating the update logic in multiple components (i.e. DRY).
+- If you atoms are related to or party of a hierarchy of other atoms, author lookup functions to help navigate the hierarchy. This will avoid repeating similar logic in multiple selectors. 
+
+## Create atoms like you would independent files or tables in a database
+- Avoid deeply nested types. Prefer to think of each atom containing an array as a set of rows in a table.
+- Prefer to put identity on your items for easy lookup. The uuid NPM package does a great job. Note: You can create a selector to produce a dictionary if you have large numbers of reads compared to writes on an array atom.
+
+## Create selectors based on each React component's needs
+- Prefer to create selectors to make it easy to render your component. Think of selectors as tiny view-models for different pieces of your UI.
+- Prefer to write each selector as purpose built for one rendering/component. Selectors should follow the Single Responsibility Principle.
+- Avoid creating selectors that generate more data than a component uses.  For example, a giant state tree where a component only renders a few parts. Otherwise, the selector is doing extra work and your component will re-render in response to data it doesn't use.
+
+## Prefer to separate common state updates into a dispatcher pattern.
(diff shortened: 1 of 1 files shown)


docs/internal/GeoffsRecoilAssessment.md
```diff
@@ -0,0 +1,176 @@
...
+- Creating selectors is straightforward. Many existing render calculations can be replaced incrementally with selectors.
+- Updating components to leverage recoil hooks rather than props with React.useEffect or React.useState is straightforward. Data passed through context can be updated with an atom or selector and then context usage can be replace incrementally down the component hierarchy.
+- Updating an existing dispatcher pattern is straightforward.
+
+# Tips and Best Practices
+
+## Follow functional programming paradigms
+- Always make state data immutable: All types used with atoms should have read only fields. 
+- Always write pure update functions: Write function that take read only inputs and return a new modified instance. This avoid repeating the update logic in multiple components (i.e. DRY).
+- If you atoms are related to or party of a hierarchy of other atoms, author lookup functions to help navigate the hierarchy. This will avoid repeating similar logic in multiple selectors.
```
srinaath, line 69: If your*

docs/internal/GeoffsRecoilAssessment.md
```diff
@@ -0,0 +1,176 @@
...
+- PRO: The recoil state hooks cause rendering just like useState.
+- PRO: useRecoilValue and useSetRecoil value clearly denote the read-only and write-only scenarios.
+- PRO: No magic inspecting dot-notation usage like with MobX.  
+- HAZARD: Very easy to accidentally put a recoil hook within a condition. They need to be used as a block at the start of a component just like React hooks.
+
+## useRecoilCallback hook
+- PRO: Allows for creating methods that always get latest state even when called outside the context of a React component render. 
+- CON: It can be a bit of extra effort to get the Typescript types worked out for complex callbacks.
+- UNKNOWN: Unclear when to use the dependencies parameter vs. the getPromise to get latest state
+- UNKNOWN: Unclear when to use the synchronous loader vs the asynchronous getPromise.- UNKNOWN:? The callback documentation indicates is still possible this hook will change. This is likely due to the experimental portions relating to JS async threading.
```
a-b-r-o-w-n, line 48:
```suggestion
- UNKNOWN: Unclear when to use the synchronous loader vs the asynchronous getPromise.
- UNKNOWN: The callback documentation indicates is still possible this hook will change. This is likely due to the experimental portions relating to JS async threading.
```

docs/internal/GeoffsRecoilAssessment.md
```diff
@@ -0,0 +1,176 @@
...
+- Restructuring application state from observable hierarchy to flat arrays is painful and expensive. A major refactoring is likely to introduce bugs if there is insufficient testing on the existing code.
+- Converting deep dependencies on the context API down the component hierarchy is expensive especially when there are few pure functional components.
+- Creating selectors is straightforward. Many existing render calculations can be replaced incrementally with selectors.
+- Updating components to leverage recoil hooks rather than props with React.useEffect or React.useState is straightforward. Data passed through context can be updated with an atom or selector and then context usage can be replace incrementally down the component hierarchy.
+- Updating an existing dispatcher pattern is straightforward.
+
+# Tips and Best Practices
+
+## Follow functional programming paradigms
+- Always make state data immutable: All types used with atoms should have read only fields.
```
a-b-r-o-w-n, line 67: Can this be enforced? (ex. `Object.freeze`)

docs/internal/GeoffsRecoilAssessment.md
```diff
@@ -0,0 +1,176 @@
...
+- PRO: Looking toward the future of asynchronous, multi-threaded Javascript.
+- PRO: Recoil API surface is tiny
+- PRO: Lightning fast (at least for my application size)
+- CON: Still experimental, few releases, limited contributes - not yet mature or fully tested.
+
+## Atoms
+- PRO: Like their name, atoms let you model your data as small and independent plain types.
+- PRO: Serializing atoms to storage is straightforward compared to trying to rehydrate an observed state tree.
+- PRO: The default state of an atom with the reset() hook is convenient for clearing application state.
+- CON: Atoms are often arrays of items which means needing to look up items. This requires more undefined checks for cases where items are not found.
```
a-b-r-o-w-n, line 25: Can this be addressed via a selector?

docs/internal/GeoffsRecoilAssessment.md
```diff
@@ -0,0 +1,176 @@
...
+                charRange: charRange,
+                entityId: entity.id,
+                parent: { kind: 'label', id: parent.id },
+            };
+            set(labelListState, [...labels, newLabel]);
+            logMessage(`+label ${entity.name}  ${characterRangeToString(newLabel.charRange)}`);
+            return newLabel;
+        }
+    );
+...
```
a-b-r-o-w-n, line 151: Is this were we would put our actions? Like fetching data from the server, etc?

docs/internal/GeoffsRecoilAssessment.md
```diff
@@ -0,0 +1,176 @@
...
+- CON: There is additional "param baggage" to pass multiple atoms to different functions vs. passing the root of the application state or a child.
+- HAZARD: Care has to be taken to avoid updating atom state by mutating the current state.  Only a set will trigger change detection.
+
+## Selectors
+- PRO: Simple to shape data for React components. 
+- PRO: Like atoms they are small and can be built as an library of independent data shapes.
+- PRO: The dependency chain from one selector to other selectors or atoms is easy to follow.
+- PRO: Less ceremony compared to Redux's mapStateToProps.
+- UNKNOWN: Unsure if a selector would re-compute its value if no UI is subscribed to it currently but it had been subscribed before.
+- UNKNOWN: The writable selector is not well documented and unclear when to use.
```
srinaath, line 35: A selector is a derivative obtained from atoms. This should never be writable right?

tonyanziano:
I just finished watching the recording of the meeting on Thursday.

To me, one of the biggest advantages of Recoil seems to be that we can pull a ton of state code out of the components themselves and abstract it away into the selectors. I think this would make understanding the Composer source a lot easier.

However, I'm only going off of what I saw in the video, so it would be interesting to actually try it out and see what results we end up with. I'm all for trying out new technology!

docs/internal/GeoffsRecoilAssessment.md
```diff
@@ -0,0 +1,176 @@
...
+# Recoil API: Pros, Cons, Unknowns, and Hazards
+
+## Overall
+- PRO: Built for hooks from the start and follows React concepts
+- PRO: Atom and selectors allow for composing applications from small pieces without introducing unnecessary dependencies.
+- PRO: Significantly reduces the complexity and code volume in components.
+- PRO: Looking toward the future of asynchronous, multi-threaded Javascript.
+- PRO: Recoil API surface is tiny
+- PRO: Lightning fast (at least for my application size)
+- CON: Still experimental, few releases, limited contributes - not yet mature or fully tested.
```
beyackle, line 19: This for me is the big reason not to jump fully into Recoil yet. As cool as it is, I'm a little wary of future API changes that would force us to go through and rewrite everything.

docs/internal/GeoffsRecoilAssessment.md
beyackle, line 25: I don't understand this objection. If an atom is defined as an array, then it should be because some natural part of our state is an array, and in that case we know already that we'll need to be smart about undefined checks. That's not part of Recoil, that's part of array handling no matter what system you use.

docs/internal/GeoffsRecoilAssessment.md
```diff
@@ -0,0 +1,176 @@
...
+- PRO: The default state of an atom with the reset() hook is convenient for clearing application state.
+- CON: Atoms are often arrays of items which means needing to look up items. This requires more undefined checks for cases where items are not found.
+- CON: There is additional "param baggage" to pass multiple atoms to different functions vs. passing the root of the application state or a child.
+- HAZARD: Care has to be taken to avoid updating atom state by mutating the current state.  Only a set will trigger change detection.
+
+## Selectors
+- PRO: Simple to shape data for React components. 
+- PRO: Like atoms they are small and can be built as an library of independent data shapes.
+- PRO: The dependency chain from one selector to other selectors or atoms is easy to follow.
+- PRO: Less ceremony compared to Redux's mapStateToProps.
```
beyackle, line 33: I like the word "ceremony" here. It's a good description of what Redux makes you do (and mapDispatchToProps is even worse in that regard).

docs/internal/GeoffsRecoilAssessment.md
```diff
@@ -0,0 +1,176 @@
...
+- UNKNOWN: Unsure if a selector would re-compute its value if no UI is subscribed to it currently but it had been subscribed before.
+- UNKNOWN: The writable selector is not well documented and unclear when to use.
+- HAZARD: Atom (and selector) key names need to be unique across the application. It can be easy to create an atom and selector with the same key.
```
beyackle, line 36: Would it be worth it to stick all of the keys into a giant enum or similar dictionary of constants, just to prevent this from happening? That dictionary could serve as a glossary, too, if we add in comments describing the function of each key.

docs/internal/GeoffsRecoilAssessment.md
```diff
@@ -0,0 +1,176 @@
...
+- CON: It can be a bit of extra effort to get the Typescript types worked out for complex callbacks.
+- UNKNOWN: Unclear when to use the dependencies parameter vs. the getPromise to get latest state
+- UNKNOWN: Unclear when to use the synchronous loader vs the asynchronous getPromise.- UNKNOWN:? The callback documentation indicates is still possible this hook will change. This is likely due to the experimental portions relating to JS async threading.
+- HAZARD: Using getPromise within a callback requires the callback method be async, the return value be a promise, and usage outside UI callbacks use await.
+- HAZARD: Using set within a callback that is called multiple times in succession or called from other callbacks requires using updator function to ensure latest state.  Otherwise arrays may not be updated with the previous call and will be overridden - losing data.
+
+## RecoilRoot component
+- PRO: Simple to understand. It leverages the React context API, but more efficiently tracking atoms and selectors as keys in a dictionary.
+- HAZARD: The React component rendering RecoilRoot cannot use recoil hooks.
```
beyackle, line 54: This seems analogous to React having a "hazard" that anything outside the React root can't use React hooks. That's not a hazard, it's just how it works. On the other hand, I don't know what nesting RecoilRoots will do, so making complicated embedded structures is probably something to avoid (for now).

docs/internal/GeoffsRecoilAssessment.md
beyackle: Type definitions can use `readonly` as a keyword on object fields. I don't know where that gets enforced (I only just learned about it myself), but there's a mechanism already.

docs/internal/GeoffsRecoilAssessment.md
srinaath: Agreed. It seems like we wont find a usage of nesting recoil roots

docs/internal/GeoffsRecoilAssessment.md
boydc2014: Agree, and is it the only CON in overall? sounds too good to be true

docs/internal/GeoffsRecoilAssessment.md
boydc2014: yeah, i was also thinking in pactise, it feels like similar to actionTypes in redux, where usually you should not define keys freely, instead you should refer from a gaint enum

docs/internal/GeoffsRecoilAssessment.md
srinaath: Looks like the dispatchers are like factory functions. We make different dispatcher sets of functions for different purposes and leverage them from the components to update state. So yeah actions and reducers combined I guess.

srinaath: So the official documentation uses selectors to make those calls.
https://recoiljs.org/docs/guides/asynchronous-data-queries/

docs/internal/GeoffsRecoilAssessment.md
cwhitten: Alternatively we can use `immer` (we already have this dependency in the app)

docs/internal/GeoffsRecoilAssessment.md
cwhitten: Yes you can use a selector to handle if for example a list is empty, however if you wanted to look up an element in that list by a unique parameter the constraint still exists

docs/internal/GeoffsRecoilAssessment.md
cwhitten: Let's call this out as an anti-pattern for now.

cwhitten:
Need to add a pattern for how the extension boundaries across plugins can play with recoil state

docs/internal/GeoffsRecoilAssessment.md
GeoffCoxMSFT: Sometimes you can return an empty value when something isn't found, you can enforce something is always there when setting the atom or throwing if there isn't . Anytime you have an array and are using findIndex or find, there is the chance something isn't there.

GeoffCoxMSFT: I think the CON is that you have to deal with items not found in the array - vs if you can express required relationships in a state tree.

If you have a state tree where let's say an order always points at a customer and a product, you can enforce that you don't allow any order to be created with those items null.  If you have an order with a product ID and a customer ID and then have an array of order, an array of products, and and array of customers, then you will have to look them up.  If you have a bug in your state setters you may not find product ID 123 in the list of products.

docs/internal/GeoffsRecoilAssessment.md
GeoffCoxMSFT: You can if you don't mind the singleton dependency on the enum across your whole system.  You may have sets of atoms and selectors that have nothing to do with each other and your don't want to create a tight binding.  You can architect for a distributed approach or a centralized approach.  I think some naming conventions to help avoid collisions could be useful - maybe extensions should start an atom key with their extension name?

docs/internal/GeoffsRecoilAssessment.md
a-b-r-o-w-n: Yes. Unfortunately, Typescript does not yet support something like `@throws` so I'm not sure if we can ever not leak the undefined case out of a selector.

docs/internal/GeoffsRecoilAssessment.md
a-b-r-o-w-n: Is there a runtime error for duplicate ids? If so, I don't think an enum is necessary as the values would just be used in one place. In the case of our action types, those are referenced in both the action creator and the reducer, so it adds some safety by using an enum.

docs/internal/GeoffsRecoilAssessment.md
a-b-r-o-w-n: This could be something to explore for hosted UI plugins, or multiple tabs with different bot projects, etc. I don't think this will be something we do off the bat, but I could see if being an interesting technique to isolate certain parts of the application.

srinaath: @a-b-r-o-w-n I could see the multiple recoil roots being super effective for 3rd party plugins which need to have an isolated scope. 

For multi bot management I was thinking more on the line of an atom for BotProjects which is an array of BotId’s. Our LUFiles, LGFiles, Schemas individually are atoms containing arrays of  files with each object inside them referring to the BotID they  associate with. When a new tab  for a new project is opened or shifting between tabs I see the selector filtering out LUFiles and LGFiles atoms for that specific botID to refresh the view. What do u feel about this approach instead of going for swapping RecoilRoots based on the bot currently selected

a-b-r-o-w-n: Yeah that is a reasonable approach.

PR #3423, opened 2020-06-16
refactor: split types.ts for clarity

beyackle:
## Description

This splits the "State" type into "BotState" and "AppState" for organizational purposes.

## Task Item

#minor


Composer/packages/client/src/store/types.ts
```diff
@@ -88,79 +88,114 @@ export interface RuntimeTemplate {
...
     coachMarkRefs: { [key: string]: any };
     complete: boolean;
   };
   clipboardActions: any[];
-  publishTargets: any[];
+
+  creationFlowStatus: CreationFlowStatus;
+  storageFileLoadingStatus: string;
+
+  templateId: string;
```
srinaath, line 170: With some refactoring, we can get rid of this i beleive

Composer/packages/client/src/store/types.ts
```diff
@@ -88,79 +88,114 @@ export interface RuntimeTemplate {
...
+
   visualEditorSelection: string[];
   onboarding: {
     coachMarkRefs: { [key: string]: any };
     complete: boolean;
   };
   clipboardActions: any[];
-  publishTargets: any[];
+
+  creationFlowStatus: CreationFlowStatus;
```
srinaath, line 167: TODO:// Can be removed with minor refactoring

PR #3427, opened 2020-06-18
fix: #2692 Tabbing / clicking / returning out of a number field round your input

### LINKED ISSUE #2692: Tabbing / clicking / returning out of a number field round your input
1. in Set properties action node, you are adding a property
2. Set the type of a value to number, and type any number, say 0.4
3. Tab, return, or clicking out of that field round 0.4 to 0. 

The same behavior occurs in Switch node as well. You have to use the spin button to set a number value.

LouisEugeneMSFT:
## Description

The field (Composer/packages/extensions/adaptive-form/src/components/fields/NumberField.tsx) was always rounding numbers to be integers since _onValidate_ calls _updateValue(0)_.
In the case of floats, _updateValue(0)_ would call _getFloat(value, 0)_, which would call _.toFixed(0)_.

This is because the _**step**_ variable was used in different places for different purposes. In the case of _getFloat_ it was used both for incrementing **and** rounding, which is incorrect.

The PR therefore makes the rounding "amount" a separate variable (floatNumberOfDecimals).

## Task Item

fixes #2692

## Screenshots

image


Composer/packages/extensions/adaptive-form/src/components/fields/NumberField.tsx
```diff
@@ -1,18 +1,19 @@
...
 
+const floatNumberOfDecimals = 1;
+
 const getInt = (value: string, step: number) => {
   return parseInt(value, 10) + step;
 };
 
 const getFloat = (value: string, step: number) => {
-  const fixed = (parseFloat(value) + step).toFixed(step !== 0 ? `${step}`.split('.')[1].length : step);
+  const fixed = (parseFloat(value) + step).toFixed(floatNumberOfDecimals);
```
GeoffCox, line 16: The issue was that numbers were being rounded to an integer, should we not round at all?

Composer/packages/extensions/adaptive-form/src/components/fields/NumberField.tsx
```diff
@@ -1,18 +1,19 @@
...
 
   return parseFloat(fixed);
```
GeoffCox, line 18: Why parse the float to turn it into a float, then back to a string to round it, then back to a float.  How about .round if we are rounding the float?

Composer/packages/extensions/adaptive-form/src/components/fields/NumberField.tsx
```diff
@@ -36,7 +37,7 @@ const NumberField: React.FC<FieldProps> = (props) => {
     onChange(newValue);
   };
 
-  const step = type === 'integer' ? 1 : 0.1;
+  const step = type === 'integer' ? 1 : 1 / Math.pow(10, floatNumberOfDecimals);
```
beyackle, line 40: This could just be `Math.pow(10, -floatNumberOfDecimals)`.

Composer/packages/extensions/adaptive-form/src/components/fields/NumberField.tsx
```diff
@@ -1,18 +1,19 @@
...
-import React from 'react';
-import { SpinButton } from 'office-ui-fabric-react/lib/SpinButton';
 import { FieldProps } from '@bfc/extension';
 import formatMessage from 'format-message';
-
+import { SpinButton } from 'office-ui-fabric-react/lib/SpinButton';
+import React from 'react';
 import { FieldLabel } from '../FieldLabel';
 
+const floatNumberOfDecimals = 1;
```
beyackle, line 9: Is there a reason we're fixing this at 1? I feel like users might have different use cases where they might need other numbers of decimals (e.g. if they're implementing a "currency" field and need 2, or if they need even more precision).

Composer/packages/extensions/adaptive-form/src/components/fields/NumberField.tsx
LouisEugeneMSFT: Good point!

Composer/packages/extensions/adaptive-form/src/components/fields/NumberField.tsx
LouisEugeneMSFT: Agreed. It was set up to a precision of 1 so I didn't want to break anything since I am still new to the codebase. How do we make that decision? If we don't have clear guidance then I agree 2 is a better default.

cwhitten:
@vishwacsena @tomlm can you comment to the introduction of `multipleOf` to specify precision on a number type? See "Validations" on https://cswr.github.io/JsonSchema/spec/basic_types/

luhan2017:
@cwhitten , It seems I have some misunderstanding here, I was trying to say that
For the NumberInput, we can use the outputFormat(=formatNumber(this.value, 2)) property to define the precision of the number type. 
Here is an example:
image
image

This means user can set precision of the output by formatNumber expression in LG. 
This seems not relate to what you want. 
I don't quite understand what is the `multipleOf` used for. Do you mean to define the precision of a number variable in the component schema?

cwhitten:
@luhan2017 @LouisEugeneMSFT let's take this and continue discussion to determine if we want to allow the the number type to include a precision option, so that you don't need to format using lg.

Issue #3432, opened 2020-06-18
custom runtime error should be user friendly 


hibrenda:
- create a bot
- runtime config to use custom runtime
- either runtime code location or start command is not correctly set, start bot will hit following error
![image](https://user-images.githubusercontent.com/43408889/85003917-7896b500-b189-11ea-856e-5723c2267f9d.png)

we should provide valid error info to user.

Issue #3351, opened 2020-06-08
[MCS-DE-GO] - Make local publishing an Async Operation


joaollq:
When publishing locally from the Composer the application freezes until the dotnet build + dotnet run flow is completed.

There are several problems with this approach, namely:
  - When the dotnet build takes longer >1 min (due to the nuget restore operation) the UI times out and gives an error stating that there was a problem with the runtime. Actually the process is still running server side.
 - The user cannot navigate to other parts of the Composer app. The requests will receive a 502 Bad Gateway error after the timeout
 - **When running the composer in a hosted mode, the local publish process freezes the app for all users and not only for the user publishing the code** 

In my local publish plugin I tried to use the same approach as the other plugins, by putting all the build and run part in a different thread and returning a 202 Created response to the UI in the hope of seeing the UI polling the "get publish status" endpoint until a 200 OK is returned. Nevertheless it doesn't work that way as the local publish seems to be prepared to receive only 200 OK results.

@goergenj @benbrown @sgellock

benbrown:
@VanyLaw we should take this into consideration as we update the runtime abstractions.

benbrown:
This will be fixed by #3198.

cwhitten:
Closing in anticipation of https://github.com/microsoft/BotFramework-Composer/pull/3198

PR #3434, opened 2020-06-18
fix: fix relative path of runtime folder in eject customized

### LINKED ISSUE #3431: Copy default custom runtime hit errors
- create a bot
- under runtime config to enable custom runtime
- click copy default runtime will hit following error
image

VanyLaw:
## Description

- Fix relative path of runtime folder in eject customized runtime, which make copy runtime failure.

- Fix start process will be created even runtime folder not existed by adding `return` after `reject`. and the process have no pid.

image

## Task Item

close #3431


VanyLaw:
Hi @cwhitten @a-b-r-o-w-n @benbrown, I checked this issue and found there had another issue when user input a existed path but not a path of runtime folder; it will caused the start process error few seconds later, but the request already be response, So user could not get the error message. Client would show `test in emulator` but bot could not run actually.
I don't have a good way to solve this, Do you have any suggestion?

FYI.
image
image

hibrenda:
Verified this solved the issue #3431 and #3432 . So we may get this in as we plan to release the electron patch. For new issues raised above, just synced with @luhan2017  and @VanyLaw ,  it will be addressed by another PR(@VanyLaw will work on it)  which will support #3351 as well.

boydc2014:
This PR is verified to be able to fix web app, so i leave an approve.

PR #3440, opened 2020-06-19
fix: fix get history before file persistence

### LINKED ISSUE #3300: Create Publish Profile -> Error: undefined is not a valid publishing target type. There may be a missing plugin.
When creating a new publish profile it always throws this error
image

## Version
ffde16f0

VanyLaw:
## Description

Fix missing plugin issue when creating a new publish profile, and make url redirect to all profile when url error.

Cause: get publish history before settings file persisted to server.
Solution: Add filePersistence.flush after clicking the save button.

## Task Item

close #3300

### DIFF
--- a/Composer/packages/client/src/pages/publish/createPublishTarget.tsx
+++ b/Composer/packages/client/src/pages/publish/createPublishTarget.tsx
@@ -11,6 +11,7 @@ import { Fragment, useState, useMemo } from 'react';
 import { PrimaryButton, DefaultButton } from 'office-ui-fabric-react/lib/Button';
 import { JsonEditor } from '@bfc/code-editor';
 
+import filePersistence from '../../store/persistence/FilePersistence';
 import { PublishTarget, PublishType } from '../../store/types';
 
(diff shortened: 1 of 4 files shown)


Composer/packages/client/src/pages/publish/createPublishTarget.tsx
```diff
@@ -83,6 +84,7 @@ const CreatePublishTarget: React.FC<CreatePublishTargetProps> = (props) => {
     if (targetType) {
       await props.updateSettings(name, targetType, JSON.stringify(config) || '{}');
       props.closeDialog();
+      await filePersistence.flush();
```
a-b-r-o-w-n, line 87: I'm not sure I understand the bug that well, but is it possible to do this somewhere else instead of the component? The file persistence layer needs to remain an implementation detail to components. This violates that principle.

VanyLaw: move it into getHistory action.

PR #3448, opened 2020-06-22
feat: 'Ctrl + Click', 'Shift + Click' behaviors in Flow Editor

### LINKED ISSUE #3249: Add ctrl+click and shift+click support to multi-select
Currently you have to use mouse drag to select multiple nodes in composer. Windows users expect to be able to use `ctrl+click` and `shift+click` to toggle item selection.  `shift+click` performs ranged selection and `ctrl+click` toggles the selection of a single item.  It would be nice if composer supported these key bindings in addition to mouse drag.

Additionally, you don't disable the property editor when multi-selecting. You should either re-scope the property editor to show only common properties or disable it all together.

yeze322:
## Description

Closes #3249 
1. [Ctrl + Click] - Toggle selection
  ctrl-click
2. [Shift + Click] - Range Selection
  shift-click

### DIFF
--- a/Composer/packages/extensions/adaptive-flow/src/adaptive-flow-editor/hooks/useEditorEventApi.ts
+++ b/Composer/packages/extensions/adaptive-flow/src/adaptive-flow-editor/hooks/useEditorEventApi.ts
@@ -81,16 +81,24 @@ export const useEditorEventApi = (
         };
         break;
       case NodeEventTypes.CtrlClick:
-        handler = (e: { id: string }) => {
+        handler = (e: { id: string; tab?: string }) => {
+          if (!focusedId && !selectedIds.length) {
+            return handleEditorEvent(NodeEventTypes.Focus, e);
+          }
+
           // Toggle the selection state of clicked id
           const alreadySelected = selectedIds.some((x) => x === e.id);
           if (alreadySelected) {
             const shrinkedSelection = selectedIds.filter((x) => x !== e.id);
             setSelectedIds(shrinkedSelection);
+            if (focusedId === e.id) {
+              onFocusSteps([shrinkedSelection[0] || '']);
+            }
             announce(ScreenReaderMessage.ActionUnfocused);
           } else {
             const expandedSelection = [...selectedIds, e.id];
             setSelectedIds(expandedSelection);
+            onFocusSteps([e.id], e.tab);
             announce(ScreenReaderMessage.ActionFocused);
           }
         };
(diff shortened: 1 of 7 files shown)


Composer/packages/extensions/adaptive-flow/src/adaptive-flow-editor/renderers/NodeWrapper.tsx
```diff
@@ -75,7 +78,16 @@ export const ActionNodeWrapper: FC<NodeWrapperProps> = ({ id, tab, data, onEvent
       aria-label={generateSDKTitle(data, '', tab)}
       onClick={(e) => {
         e.stopPropagation();
-        onEvent(NodeEventTypes.Focus, { id, tab });
+        e.preventDefault();
+
+        const payload = { id, tab };
+        if (e.ctrlKey) {
```
beyackle, line 84: Do we want to also check for `e.metaKey` so Mac users can use Cmd+Click?

Composer/packages/extensions/adaptive-flow/src/adaptive-flow-editor/hooks/useEditorEventApi.ts
```diff
@@ -80,6 +81,47 @@ export const useEditorEventApi = (
...
+
+          // Toggle the selection state of clicked id
+          const alreadySelected = selectedIds.some((x) => x === e.id);
+          if (alreadySelected) {
+            const shrinkedSelection = selectedIds.filter((x) => x !== e.id);
+            setSelectedIds(shrinkedSelection);
+            if (focusedId === e.id) {
+              onFocusSteps([shrinkedSelection[0] || '']);
+            }
+            announce(ScreenReaderMessage.ActionUnfocused);
```
beyackle, line 98: Is there a way to get the title of the action into these messages? I think it would make for better feedback.

yeze322: This module is blind to the action title, it only interacts with action's id.
If this is an important requirement, we can track it in a separated issue aims at unifing all existing operations' messages.

Composer/packages/extensions/adaptive-flow/src/adaptive-flow-editor/renderers/NodeWrapper.tsx
yeze322: resolved

Composer/packages/extensions/adaptive-flow/src/adaptive-flow-editor/hooks/useEditorEventApi.ts
beyackle: I won't block this particular PR on this issue, but I ran into this a lot fixing a11y bugs - there's no good way to get the title of an action, and it looks like the same problem is happening here. I think we need to do what you said, just to have some internal way that those of us working on the editor can reliably get an action's title for use in announcements and things.

yeze322: @beyackle Yeah, I do remember you wrote some code to query DOM properties.
The action title could be a static map in which each $kind has a constant title string. It would be a good time to do that during the multi locale work. We create multiple title sets for different locales, then announcements could carry title info. What's your opinion?

PR #3453, opened 2020-06-22
Remove language specific features from core and into runtime plugin

benbrown:
## Description

Extend the runtimeTemplate plugin API to support encapsulating the build/run and build/deploy steps in the plugin rather than in the core runtime management.

- [x] Add new runtimeTemplate fields
- [x] Refactor localPublish
- [x] Refactor azurePublish
- [x] Refactor azureFunctionsPublish
- [x] Refactor botProjectDeploy
- [x] Refactor configuration writing
- [x] Capture and store runtime type at eject
- [x] Make sure the localPublish build steps can be async. #3351 

## Details

This pull request pulls the language specific and platform specific details out of the azurePublish and localPublish plugins and into a new plugin called `runtimes`.

The new runtimes plugin now contains the necessary steps to do the dotnet build (during local testing) and the dotnet deploy (during deploy to azure). I also moved the writing of the appsettings.json file into the plugin - previously it was done several times throughout the course of the deploy.  It should only be necessary to use the single `settings/appsettings.json` file now in the runtime as ALL settings are combined and written to this file.  This allows the plugin to handle the settings differently if so desired.

Along the way, I cleaned up some of the code, added a lot of inline comments, and moved some of the classes around so they were located closer to where they are used.

## Next Steps

- [ ] Merge this into @VanyLaw's work with the JS runtime - this should be up to date
- [ ] Continue work on the JS runtime
- [ ] Update the runtime templates to remove the need for multiple appsettings.json files

And some more stuff I want to extract into the runtime definition in a future PR:

- [ ] Extract the actual "run" part of the localPublish into the runtimes - there is a method for this defined, but I wanted to merge before embarking on this.  (This would allow the plugin to define how to interact with the runtime)
- [ ] Extract the "Copy the code" part of these plugins into the runtime - the runtime should be responsible for providing the code into the right location rather than assuming it lives locally.


Composer/packages/client/src/store/reducer/index.ts
```diff
@@ -516,9 +525,16 @@ const publishFailure: (title: string) => ReducerFunc = (title) => (state, { erro
 const getPublishStatus: ReducerFunc = (state, payload) => {
   // the action below only applies to when a bot is being started using the "start bot" button
   // a check should be added to this that ensures this ONLY applies to the "default" profile.
-  if (payload.target.name === 'default' && payload.endpointURL) {
-    state.botStatus = BotStatus.connected;
-    state.botEndpoints[state.projectId] = `${payload.endpointURL}/api/messages`;
+  if (payload.target.name === 'default') {
+    if (payload.status == 200 && payload.endpointURL) {
```
carlosscastro, line 529: Optional: In general I prefer having a centralized status codes enum if we are not already using some http library that has http status codes. more readability, harder to confuse, etc.

Composer/packages/client/src/store/reducer/index.ts
```diff
@@ -516,9 +525,16 @@ const publishFailure: (title: string) => ReducerFunc = (title) => (state, { erro
...
+      state.botEndpoints[state.projectId] = `${payload.endpointURL}/api/messages`;
+      state.botStatus = BotStatus.connected;
+    } else if (payload.status == 202) {
+      state.botStatus = BotStatus.reloading;
+    } else if (payload.status == 500) {
+      state.botStatus = BotStatus.failed;
```
carlosscastro, line 535: Does the payload come with a message in addition to the status? If that is the case, that message could be valuable to surface or log for troubleshooting.

Composer/plugins/runtimes/src/index.ts
```diff
@@ -0,0 +1,186 @@
...
+        const schemaFolderInRuntime = path.join(destPath, 'azurewebapp/schemas');
+        await removeDirAndFiles(schemaFolderInRuntime);
+        return destPath;
+      }
+      throw new Error(`Runtime already exists at ${destPath}`);
+    },
+  });
+
+  composer.addRuntimeTemplate({
+    key: 'javascript-azurewebapp',
```
carlosscastro, line 134: in the directory structure I think we agreed on calling this `node` instead of `javascript`. Should we make this `node-azurewebapp` or `nodejs-azurewebapp`? I'm flexible here but ideally it should be the same as the directory name under `runtime` for the js runtime.

Composer/packages/client/src/store/reducer/index.ts
benbrown: yes, it is being passed into the botLoadErrorMsg one line below this line...

Composer/packages/client/src/store/reducer/index.ts
benbrown: sounds good, will do.

Composer/plugins/runtimes/src/index.ts
benbrown: makes sense to me.

PR #3479, opened 2020-06-24
fix: revert plugins out of workspace

VanyLaw:
## Description

revert[ this PR](https://github.com/microsoft/BotFramework-Composer/pull/3322) and make sure all the relative paths were fixed. For the bot-deploy update problem, currently I just remove yarn.lock in azure plugins before yarn install, and we can polish this issue in the next PR.

## Task Item

close #3322


a-b-r-o-w-n:
I'm not super excited about this change, but this unblocks the electron build so let's :shipit:

benbrown:
This is probably going to cause chaos for me as I've made a bunch of changes in the new location :/

PR #3500, opened 2020-06-25
fix: add logo-clicking to tests

beyackle:
## Description

Some e2e tests are still failing (LGPage and NotificationPage) because Cypress can't click on elements it thinks are covered. Moving the mouse around (by clicking something non-interactive) might fix this.

## Task Item

#minor


a-b-r-o-w-n:
Looks good. @beyackle can you fix conflicts?

Issue #3456, opened 2020-06-22
[test] increase test coverage to 70%


a-b-r-o-w-n:
This issue is meant to be a long running issue that tracks efforts to increase unit test code coverage.

At the time is was created, coverage sits around 49%. Let's target an increase of 20% for R10.

beyackle:
I don't think this was meant to be closed yet - not until we get to 70%, right?

cwhitten:
Oops

boydc2014:
number at 23/9 is 55%.

boydc2014:
As a review guidance, let's make sure new PR have tests.

boydc2014:
I will look at bottle-neck from the coverage report.

@a-b-r-o-w-n  will set a bar for coverage for PR check in, maybe start with -0.5%?

a-b-r-o-w-n:
I set the threshold in coveralls to fail the build if coverage drops more than 0.5% or overall coverage drops below our current 55%

PR #3524, opened 2020-06-29
test: add utility and some tests for uncovered actions

beyackle:
## Description

This adds a quick way to test the simplest form of store-based actions, where there's no logic in the action itself - it just assembles an ActionType and a payload and dispatches that off to the store. It also adds some simple code that shows its use in a few places.

## Task Item

refs #3456


Composer/cypress/integration/NotificationPage.spec.ts
```diff
@@ -9,6 +9,9 @@ context('Notification Page', () => {
   it('can show lg syntax error ', () => {
     cy.visitPage('Bot Responses');
 
+    // click the logo to clear any stray tooltips from page navigation
+    cy.findByAltText('Composer Logo').click();
```
a-b-r-o-w-n, line 13: Do you think its worth moving this into the `visitPage` function?

beyackle: Done in another PR (based on the suggestion here).

beyackle:
@a-b-r-o-w-n I think you're right that moving this into an "expect X to dispatch" pattern would make more sense. I'll close this PR - there's not much here worth saving - and try that approach in a new one.

PR #3529, opened 2020-07-01
fix: LG unit test failure

zhixzhan:
## Description

LGServer.ts use LgWorker do parsing, and LgWorker is created as a forked process, that take times, and would caused a timeout error when running test at some not very powerful PC or old laptop.

Fix it by using LGParser instead of create worker when running test.

## Task Item

#minor


zhixzhan:
Found another FileSelector.test.tsx error
image

I synced with Weitian, he fixed in #3531

PR #3536, opened 2020-07-01
fix: fix deploy issue while zipping the folder

### LINKED ISSUE #3301: Publishing -> Error: Only absolute URLs are supported
## Describe the bug

When creating a new Publishing Profile and copying over the settings from the provisioncomposer script and then running it it always fails with the following error;
image
image

## Version
ffde16f0

zidaneymar:
## Description
Fix an issue that the deploy process will get stuck at the zipping folder.

## Task Item
fixes #3301


zidaneymar:
@boydc2014 @VanyLaw 
I find a way to resolve this issue without replacing the zipping library.
I remove the "@bfc/libs/bot-deploy": "../../packages/lib/bot-deploy" in package.json of azurePublish plugin, and everything starts to work fine, but not sure about the reason.
image

boydc2014:
> @boydc2014 @VanyLaw
> I find a way to resolve this issue without replacing the zipping library.
> I remove the "@bfc/libs/bot-deploy": "../../packages/lib/bot-deploy" in package.json of azurePublish plugin, and everything starts to work fine, but not sure about the reason.
> image


Good to see we don't have to replace a package without knowing the root cause.

zidaneymar:
Revert the previous commit and fix the path importing in azure publish plugin, previous style leads to the unexpetect behavior in zipping runtime process.

lei9444:
I create an electron installer. But electron app can't load the azurePublish plugin. Maybe it's not a good solution

boydc2014:
> I create an electron installer. But electron app can't load the azurePublish plugin. Maybe it's not a good solution

OK, @zidaneymar let's revert again back to use zip-folder and get the release out, then later figure out why `archiver` is not working.

boydc2014:
@cwhitten After digging into this issue whole day trying to figure out why previous "archiver" not working, Qi and Leilei located the issue is related to the way the package is referred, as Qi stated above, but simply change that will break Electron case. So Qi revert back again to use zip-folder. 

I think this one is good to go now, we will leave the investigation later.

zidaneymar:
close, have a better solution

boydc2014:
> close, have a better solution

Can you link the better solution here?

boydc2014:
Close this in favor of https://github.com/microsoft/BotFramework-Composer/pull/3536

PR #3537, opened 2020-07-01
fix: Dialog validate throw error when delete an action

### LINKED ISSUE #3534: Dialog validate error when add a action
## Browser

- [ ] Electron distribution
- [x] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [x] macOS
- [ ] Windows
- [ ] Ubuntu

## To Reproduce

1. Go to 'Home'
2. Click on 'New' ->'Create from template' -> 'Echo Bot' 
3. Go to 'Design'
4. Click 'Unknown intent'
5. Add a `Send a Response`, fill with `hi`
6. Add a `Ask a question` -> `Text`
7. See error

Add a Luis trigger may also trigger this bug, when refresh page, error gone.

## Screenshots
Untitled5

image

lei9444:
## Description
The options in parsed lgfile is undefined rather than empty array if there is no options. 

## Task Item
closes #3534


Composer/packages/lib/indexers/src/validations/expressionValidation/index.ts
```diff
@@ -15,16 +15,17 @@ const ExportsKey = '@exports';
...
-    let namespace = extractOptionByKey(NamespaceKey, options);
-    if (!namespace) namespace = lgFile.id; //if namespace doesn't exist, use file name
-    const funcList = exports.split(',');
-    funcList.forEach((func) => {
-      if (func) {
-        result.push(`${namespace}.${func.trim()}`);
-      }
-    });
-
+    if (options && options.length) {
```
a-b-r-o-w-n, line 18:
```suggestion
    if (options?.length > 0) {
```

Issue #3535, opened 2020-07-01
Bot Gets Created in Default Directory


cxbuilders:
When creating a new bot the location dropdown shows the location of the last used bot in composer. However when creating a new bot it gets created in my default user folder. Only when I change the dropdown one folder down and up again, it gets saved correctly.

Version: 2f655bb

PR #3549, opened 2020-07-02
fix: location initialization

liweitian:
## Description
The location is not initialized.  And an useEffect also reset location to "". Fixed both cases and updated test case to guarantee initialization correct.

## Task Item
refs #3535

## Screenshots
aaaa


boydc2014:
@liweitian the fix looks good, but can you do some investigation which PR or commit introduced this regression, what can we improve to make this more stable?

liweitian:
> @liweitian the fix looks good, but can you do some investigation which PR or commit introduced this regression, what can we improve to make this more stable?

updated in pr description

PR #3555, opened 2020-07-02
test: add toDispatch to expect and runMinorTests

beyackle:
## Description

This adds an "expect.toDispatch" custom matcher, which we can use to test all of the actions that just take in arguments and pass them along into a dispatch. This also adds the `dialog.test.ts` file to demonstrate how these are intended to work.

## Task Item

refs #3456

### DIFF
--- a/Composer/packages/client/__tests__/jest.d.ts
+++ b/Composer/packages/client/__tests__/jest.d.ts
@@ -7,6 +7,7 @@ declare global {
   namespace jest {
     interface Matchers<R> {
       toBeDispatchedWith(type: ActionTypes, payload?: any, error?: any);
+      toDispatch(actionType: ActionTypes, payload: {}, output?: {});
     }
   }
 }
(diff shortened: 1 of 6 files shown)


Composer/packages/client/__tests__/jest.d.ts
```diff
@@ -7,6 +7,7 @@ declare global {
   namespace jest {
     interface Matchers<R> {
       toBeDispatchedWith(type: ActionTypes, payload?: any, error?: any);
```
a-b-r-o-w-n, line 9: So I didn't realize we already have this matcher. Can you use this instead? 

The goal would be to not have the `runMinorTests` function, but instead to have an explicit test for each action so that devs do not need to leave the context of the test to figure out what is actually being asserted.

a-b-r-o-w-n:
@beyackle can this be closed now?

beyackle:
Yes, I'll close it.

PR #3525, opened 2020-06-30
refactor: split some actions off setSettings

beyackle:
## Description

This splits the `setSetting` action apart into a number of smaller actions for clarity.

## Task Item

closes #2753

PR #3556, opened 2020-07-03
fix: fix profile not saving

VanyLaw:
## Description
Fix profile saving and editing issue in publish page.

## Task Item

refs #3550 #3525 

## Screenshots
AnyConv com__Video_2020-07-03_171753


boydc2014:
@VanyLaw can you state more about the root cause here? I thought  #3550 fixed the adding publish profile page failed to render, is this PR fixing something #3550 was missing, or a new regression after we take #3550 ?

boydc2014:
OK, i get it, If i understand correctly, #3550 only fix the page is not rendering issue, but after 3550, the profile is not actually serialized to disk, which is fixed in this PR, right?

boydc2014:
This one looks good and works for me, but as we can see, we have several issues related to publishing, just between the time windows of 1.0.1 to 1.0.2. @VanyLaw how can we improve and guarantee the overall quality of this part? Please re-think about this and create a plan for this.

boydc2014:
Close this dup to #3556 have merged this.

VanyLaw:
> OK, i get it, If i understand correctly, #3550 only fix the page is not rendering issue, but after 3550, the profile is not actually serialized to disk, which is fixed in this PR, right?

Yes, #3550 fix the render broken issue, but not exactly save it to settings file. And I will add more unit test and e2e test about the settings and deploy stuff later.

PR #3557, opened 2020-07-03
fix: deploy issue while zipping the folder

### LINKED ISSUE #3301: Publishing -> Error: Only absolute URLs are supported

lei9444:
## Description
Today plugins use bot-deploy as local dependency in package.json. But the archiver doesn't work in this pattern(I don't know the reason now. I have added archiver to test in the plugins and composer server, it works).

The bot-deploy is only used by plugins. So I move it to plugins' lib folder and add a yarn workspace for plugins. 

I also merged #3556 to fix the profile save issue, this will break the publish workflow.

## Task Item
fixes #3301
refs #3536


boydc2014:
@cwhitten this is the fix that I believe eventually resolve the "archiver" not working issue without have to switching to "zip-folder" (which depends on a lower version of lodash, thus not passing secuirty check). 

we tested this both in web app and in electron and it's working as expected E2E.

a-b-r-o-w-n:
Are the other changes un-releated to the workspace change meant to be in this PR? Specifically the runtime settings changes?

lei9444:
> Are the other changes un-releated to the workspace change meant to be in this PR? Specifically the runtime settings changes?

Hi, @a-b-r-o-w-n I merged @VanyLaw 's PR about profile can not be saved issue. Remove this change will make the publish fail. The PR is https://github.com/microsoft/BotFramework-Composer/pull/3556.

Do you think we need to move un-releated changes to another PR? Maybe I can drop the commit and reopen @VanyLaw 's PR #3556.

boydc2014:
> > Are the other changes un-releated to the workspace change meant to be in this PR? Specifically the runtime settings changes?
> 
> Hi, @a-b-r-o-w-n I just merged @VanyLaw 's PR about profile can not be saved issue. Remove this change will make the publish fail. The PR is #3556.
> 
> Do you think we need to move the un-releated changes to another PR? Maybe I can drop the commit and reopen @VanyLaw 's PR #3556.

I think this one is now considered as an fix to the E2E publishing workflow, it fixes the zipping stucking issue and fix the publishing profile not saving issue. I suggest we take this one. @a-b-r-o-w-n  what do you think

a-b-r-o-w-n:
@boydc2014 I haven't had a chance to test this. So if you feel like it is good, let's merge.

PR #3575, opened 2020-07-06
fix: add ability to update the boilerplate content

### LINKED ISSUE #3102: Composer should upgrade provisioning scripts on version upgrade
## Describe the bug

If a bot is created with a certain version of the BotComposer and an update for BotComposer comes thru, the provisioning scripts never get upgraded.

## Version

v1.0.0

## Browser

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge
- [X] Electron

## OS

- [ ] macOS
- [X] Windows
- [ ] Ubuntu

benbrown:
## Description

On load of the runtime settings page, the client will now make an API call that returns the current package.json version from the scripts/ folder in the project, the LATEST version of that file from the "boilerplate" template, and a boolean indicating whether the project version needs to be updated.

If the project version is missing or out of date, a new box will appear on the page with a button that allows the user to update this content - the scripts and readme files that come with every new project. This option will be display whether or not the runtime has been ejected. It does *not* affect the runtime code itself - only the provisioning scripts.

* Also adds a "loading" indicator during ejection, which can sometimes take a few seconds.

## Task Item

Closes #3102 

## Screenshots

This GIF shows a project where the boilerplate is old. The UI is present.  I click the button, the boilerplate is updated.  I navigate away, navigate back: the warning and the UI are now hidden because the boilerplate is up to date.

ZGoXfCDZam


benbrown:
@sangwoohaan can you check out the design for this new feature and let me know if you've got requested changes?

benbrown:
GREAT QUESTION from @sangwoohaan  -- how does the user know they need to update these scripts?  Do we need to link this to a function that will track which version of Composer was used to create a project?  (And then some mechanism that allows composer to update a project from version x to version y?)

ALSO NOTE: when we move provisioning into composer, we will no longer need the current set of scripts. So maybe we don't need this at all?

ALSO NOTE: no changes have happened to the scripts since 1.0.1, and no changes are currently planned... so this would be useful to a small number of users.

benbrown:
We decided to check against the version number in the package.json file inside scripts/ to determine if this option should be displayed.  I will update the PR.

benbrown:
This PR has been updated with a version check using the package.json file present inside the boilerplate content. The UI ONLY appears if expected boilerplate content is out of date or missing.

PR #3587, opened 2020-07-07
feat: Using Recoil JS for state management

### LINKED ISSUE #3465: Moving Bot specific states to use Recoil JS
We currently use the context API for our state management. As the application has matured we need a robust state management library. Recoil JS will be tried as a new state management library.

Describe the solution you'd like
Our states are split into Application States and Bot States. This ticket addresses moving all Bot specific states to Recoil JS.

Describe alternatives you've considered
We have considered Redux and Mobx and landed with Recoil

srinaath:
## Description
This PR is the first leg of the effort to move away from React's context API for all our state management needs and instead use Recoil JS. More unit tests to follow

## Task Item
Fixes #3465 #3464


Composer/packages/lib/indexers/src/utils/lgUtil.ts
```diff
@@ -32,59 +32,73 @@ export function increaseNameUtilNotExist(templates: LgTemplate[], name: string):
...
-  { name = 'TemplateName', parameters = [], body = '-TemplateBody' }: LgTemplate
-): string {
-  const resource = Templates.parseText(content);
+  { name = 'TemplateName', parameters = [], body = '-TemplateBody' }: LgTemplate,
+  importResolver?: ImportResolverDelegate
+): Templates {
+  const resource = Templates.parseText(content, undefined, importResolver);
   const newName = increaseNameUtilNotExist(resource.toArray(), name);
 
-  return resource.addTemplate(newName, parameters, body).toString();
```
zhixzhan, line 68: Seems all lgUtil methods return value has changed from `string` to `Templates`. That's great! since we don't need pass them cross worker & iframe, which those objects are non-copy-able.

If we step forward, can we return `LgFile` type instead of `Templates` type from lgParser, cause I have saw in lgIndexer there added a method `convertTemplatesToLgFile`, I image all scenario call lgUtil would also call parse to  get a `LgFile`.

Composer/packages/client/src/recoilModel/dispatchers/lg.ts
```diff
@@ -0,0 +1,108 @@
...
+      template,
+    }: {
+      id: string;
+      templateName: string;
+      template: LgTemplate;
+    }) => {
+      set(lgFilesState, (lgFiles) => {
+        const content = lgFiles.find((file) => file.id === id)?.content ?? '';
+        const templates = lgUtil.updateTemplate(content, templateName, template, lgFileResolver(lgFiles));
+        const result = lgIndexer.convertTemplatesToLgFile(id, templates.toString(), templates);
```
zhixzhan, line 38: Maybe not necessary if we return `LgFile` from lgUtil

Composer/packages/client/src/pages/language-generation/table-view.tsx
```diff
@@ -59,44 +65,47 @@ const TableView: React.FC<TableViewProps> = (props) => {
   );
 
   const onCreateNewTemplate = useCallback(() => {
-    const newName = increaseNameUtilNotExist(templates, 'TemplateName');
-    const payload = {
-      file,
-      projectId,
```
zhixzhan, line 65: Remove `projectId` here means we handle it in store layer, like use `current bot's projectId` ?

lei9444: Yes, we will use current bot's project id. And if composer want to switch to another bot, we flush all updates before opening a new bot.

Composer/packages/client/src/shell/lgApi.ts
```diff
@@ -4,14 +4,21 @@
 import { useEffect, useState } from 'react';
```
srinaath, line 4: @zhixzhan review requested on this file

Composer/packages/client/src/shell/luApi.ts
```diff
@@ -4,14 +4,22 @@
 import { useEffect, useState } from 'react';
 import { LuFile, LuIntentSection } from '@bfc/shared';
```
srinaath, line 5: @zhixzhan review requested on this file

Composer/packages/client/src/recoilModel/dispatchers/lg.ts
```diff
@@ -0,0 +1,102 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
```
srinaath, line 2: @zhixzhan review requested on this file

Composer/packages/client/src/recoilModel/dispatchers/lu.ts
```diff
@@ -0,0 +1,114 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
```
srinaath, line 2: @zhixzhan  review requested on this file

Composer/packages/client/src/shell/useShell.ts
```diff
@@ -1,15 +1,31 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-import { useContext, useMemo, useRef } from 'react';
+import { useMemo, useRef } from 'react';
```
srinaath, line 4: @yeze322 could you validate if the shell api operations are working in this PR and you dont see anything wrong with these changes

zhixzhan:
@srinaath @lei9444 I have verified LG LU and some Shell api, looks good to me.

liweitian:
tested creating bot flow and creating trigger flow. LGTM.

srinaath:
> @srinaath @lei9444 I have verified LG LU and some Shell api, looks good to me.

Awesome @zhixzhan

srinaath:
Skills, settings and Onboarding flow has been tested by  @tdurnford  and me

Composer/packages/client/src/recoilModel/dispatchers/application.ts
```diff
@@ -0,0 +1,117 @@
...
+      if (status === AppUpdaterStatus.IDLE) {
+        newAppUpdateState.progressPercent = 0;
+        newAppUpdateState.version = undefined;
+      }
+
+      newAppUpdateState.status = status;
+
+      set(appUpdateState, {
+        ...newAppUpdateState,
+        status,
```
tonyanziano, line 37: I don't think this line is necessary because `status` is already assigned to `newAppUpdateState.status` on line 33.

This is the equivalent of:

```
set(appUpdateState, {
   ...{ status },
   status,
});
```

Composer/packages/client/__tests__/components/appUpdater.test.tsx
```diff
@@ -5,79 +5,110 @@ import React from 'react';
...
 
   it('should render the error dialog', () => {
-    state.appUpdate.status = AppUpdaterStatus.UPDATE_FAILED;
-    state.appUpdate.error = '408 Request timed out.';
-    const { getByText } = renderWithStore(<AppUpdater />, state);
+    const { getByText } = renderWithRecoil(<AppUpdater />, ({ set }) => {
+      set(appUpdateState, {
+        ...baseState,
+        error: '408 Request timed out.',
+        downloadSizeInBytes: 14760000,
```
tonyanziano, line 108: can get rid of this line

boydc2014:
I think @zidaneymar and @VanyLaw  was also looking on more scenarios on plugin, publishing all that, can you guys share some comments?

peterbozso:
Hi!

I am going through the code of the client application and just stumbled upon the usage of Recoil. I am a bit surprised, because I can't really see the reason why the team choose this experimental state management library over more mature and widely known and used ones - like Redux or MobX, just to name a few. I also saw in [one of the referenced issues](https://github.com/microsoft/BotFramework-Composer/issues/3464) that those two have been considered, but ultimately you decided against them in favor of Recoil.

@srinaath as I see, you have been the driving force behind this. Could you please shed some light on the this decision?

Thanks,
Péter

cwhitten:
Hello @peterbozso 

React Hooks has allowed more flexibility and expressiveness when it comes to data flow and state management decisions. You'll see in your architecture that we model a redux-like unidirectional data-flow pattern, but use Recoil to isolate and manage state transitions. We did indeed evaluate various options, new and old but landed on Recoil and so far are happy with this decision. We articulated this with an assessment document here: https://github.com/microsoft/BotFramework-Composer/blob/main/docs/internal/GeoffsRecoilAssessment.md

peterbozso:
Thank you @cwhitten for your quick answer! It's a really informative writing about Recoil itself. What I still struggle to see is the actual reason for choosing Recoil above other, more established alternatives like Redux or MobX.

It seems to me from the assessment that Recoil was chosen primarily because it's more lightweight than the other two and aligns better with the design of the new context API and hooks. I absolutely see the merit of this, but in case of such a big project as the Composer, I see the experimental state of Recoil a much bigger danger than its previously mentioned benefits. What if Facebook decides to scrap it? What if they change it in 1.0.0 so much that almost all the current code that depends on it needs to be changed? In my experience with the JS ecosystem, the chance of either of these is well above 0 for an experimental project. Also bugs, performance problems, etc.: all the common characteristics of beta software. Because let's admit it: that's what Recoil essentially is right now. I think it's okay to choose a library like this for an internal/pet project, but it seems to me a bit too brave to build one of the main aspects (state management) of a public-facing, production-ready application on such an unstable base.
Not mentioning the fact that orders of magnitude more people know and use the other state management libraries. Chosing one of those would have significantly lowered the barrier of entry for contributors outside of the core engineering team. I think the latter is a very important aspect to consider for an open source project where attracting contributors is always a challenge, even for very well established ones.
Unfortunately I can't find any of these points addressed in the document you referenced.

Note that I am not a pro with React myself (more like a confident intermediate :)), so please excuse me my ignorance if there's any! Please also understand that I am not trying to call out or even really challenge anybody for the decision of using of Recoil in the project. Right the contrary: I feel that my understanding of the topic is lacking and I hope somebody can point me at a direction that makes the situation clear to me, so I (and hopefully others who are reading this and are similarly clueless as I am ) can learn from it.

cwhitten:
For quite some time (~years), Redux itself was built on top of the React Context API, something the React team made very clear was unstable until they rewrote the entire thing. I mention this not as a counter-argument to Facebook's future with Recoil and ongoing support, but to point out that large-scale production applications can take a dependency on something not guaranteed to ship backwards compatible changes and still survive. 

In addition to the context API and better composition with hooks, React Suspense will also fundamentally change the way data gets sent to the view components and from our analysis is something Recoil is set up to adopt rather easily if you model your state appropriately and something we're looking forward to integrating once it ships in the framework.

What if Facebook pulls investment from Recoil? They may, but what is important to remember is that our architecture goes beyond the tool chosen to help model & manage state. Refactoring away from Recoil won't mean that we need to rewrite our application because the client architecture is set up in such a way that we can preserve the layering and replacing a dependency won't trickle down through the system nearly as much as you may think. It will be an expensive change, but it won't sink us. I'm optimistic we won't need to do this, though.

Bottom-line: in front-end engineering you often find yourself with the opportunity to take a bet on a new tool and it's up to you to do the due diligence to make sure it isn't a gamble, which this isn't. We made a choice and it has tradeoffs, but it is defendable, and we're OK with this for now.

PR #3597, opened 2020-07-08
chore: upgrade Fabric/Fluent and relatives to 7.121

### LINKED ISSUE #2569: [chore] update fabric ui

beyackle:
## Description

Upgrade to the latest version of Fluent (still using the "fabric" package name, which is a supported alias through 7.x versions). This also updates the one test that failed (NumberInput needs a "value" prop to start with or else the up/down buttons don't work).

## Task Item

closes #2569


beyackle:
Note: the only failing unit test is NumberInput.test.tsx, because for some reason its up and down buttons aren't triggering the updateValue function. I'll return to this later.

PR #3636, opened 2020-07-15
Update qn aui

liweitian:


liweitian:
will reopen in another branch

Issue #3466, opened 2020-06-23
Multi Bot Management Prototype


srinaath:
To decide on an approach for dealing with opening multiple bots inside a single composer instance we are investing in a prototype. This ticket covers both engineering and design work to land a prototype that can be used as an accelerator for our Multi Bot management solution

## Describe the solution you'd like
The prototype is to build multi window approach where each bot is opened in a new Composer browser window but tied to a single Electron instance

## Describe alternatives you've considered
We have considered using a in app tabbed experience for opening multiple bots but feel the best solution at this time is the multi windows model that Microsoft Word follows

cwhitten:
Marking the prototype as complete fo R10

PR #3641, opened 2020-07-15
feature: remodel Composer sidebar for multibot

beyackle:
## Description

This redesigns the ProjectTree component to fit the new designs. The reveal/hide triangles work, although there's currently nothing anywhere but the Core Bot section, and it will re-render (for reasons yet unknown) when it gets revealed from being hidden.

Note: this is still a prototype; some functionality might be missing (like a11y).

## Task Item

refs #3466 

## Screenshots

image


beyackle:
Updated to match the latest designs: 
image

beyackle:
Putting this back into draft for a moment while I fix a bug; deleting something while we're looking at it makes the app crash.

beyackle:
Couldn't reproduce that bug, but at least I fixed the problem with the headers' width.

Composer/packages/client/src/components/ProjectTree/treeItem.tsx
```diff
@@ -38,6 +38,8 @@ const content = css`
   outline: none;
   display: flex;
   align-items: center;
+  justify-items: middle;
```
hatpick, line 41: perhaps you meant 'center'
https://developer.mozilla.org/en-US/docs/Web/CSS/justify-items

PR #3644, opened 2020-07-16
feature: runtime luis settings refactor

zidaneymar:
## Description

* remove luis settings adapter, use customized composer luis settings extensions to load luis settings, previously there exsits a temporary work around which is adding "region" to replace "luis:region", adding "environment" to replace "luis:environment". now remove that implementation.
* remove appsettings.deployment.json and appsettings.development.json, just keep single 'appsettings.json' instead.

## Task Item
#minor


benbrown:
The changes to the way the settings file are handled may be in conflict with changes made in the `wenyluo/jsRuntime` branch.  In this branch, we have already changed how the settings file is written. We should make sure these changes are not in conflict!

benbrown:
Does this change the form of the configuration? Will previously created publishing profiles break?

runtime/dotnet/core/ComposerLuisExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+            var settings = new Dictionary<string, string>();
+            settings["luis:endpoint"] = configuration.GetValue<string>("luis:endpoint") ?? $"https://{luisRegion}.api.cognitive.microsoft.com";
+            settings["BotRoot"] = botRoot;
+            builder.AddInMemoryCollection(settings);
+            if (environment == "Development")
+            {
+                environment = Environment.UserName;
+            }
+
+            var di = new DirectoryInfo(botRoot);
```
benbrown, line 38: Are the dialog-to-luisAppId settings ALWAYS expected to live in the generated/ folder?

If so, we should consider NOT putting them into the main appsettings.json file

runtime/dotnet/azurefunctions/Startup.cs
```diff
@@ -43,19 +43,17 @@ private IConfigurationRoot BuildConfiguration(string rootDirectory)
...
+            config.SetBasePath(rootDirectory);
 
             // Config precedence 2: ComposerDialogs/settings settings which are injected by the composer publish
             // Hard code the settings path to 'ComposerDialogs' for deployment
             var configFile = Path.GetFullPath(Path.Combine(rootDirectory, @"ComposerDialogs/settings/appsettings.json"));
             config.AddJsonFile(configFile, optional: true, reloadOnChange: true);
 
-            // Config Precedence 3: Deployment specific config
-            config.AddJsonFile("appsettings.deployment.json", optional: true, reloadOnChange: true);
+            config.AddJsonFile("appsettings.json", optional: true, reloadOnChange: true);
```
benbrown, line 53: this still appears to be using 2 setttings files.

Changes in the `wenyluo/jsRuntime` branch should make this unnecessary. ALL settings are written to `settings/appsettings.json` in that branch - there should be no secondary settings file.

runtime/dotnet/core/ComposerLuisExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+            foreach (var file in di.GetFiles($"luis.settings.{environment.ToLower()}.{luisRegion}.json", SearchOption.AllDirectories))
```
carlosscastro, line 39: Do we need 'allDirectories'? File I/O is pretty slow and it would affect load time for scenarios like functions where load happens more frequently. We should minimize I/O as much as possible here.

carlosscastro: For example, iterating through our bin and obj folders is just a no-go. Can we be more specific?

runtime/dotnet/core/ComposerBotPathAdapter.cs
```diff
@@ -0,0 +1,29 @@
...
+
+using Microsoft.Extensions.Configuration;
+using System.Collections.Generic;
+
+namespace Microsoft.BotFramework.Composer.Core
+{
+    /// <summary>
+    /// Bot path adapter, for development environment, use '../../' as the bot path, for deployment and production environment, use 'ComposerDialogs' as bot path
+    /// </summary>
+    public static class ComposerBotPathAdapter
```
carlosscastro, line 12: Naming: Should this be called `ComposerBotPathExtensions` or something that ends with `*Extensions`? When we think of Adapter I tend to think of the Adapter object oriented pattern but instead here we have extension methods.

runtime/dotnet/core/ComposerBotPathAdapter.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        public static IConfigurationBuilder UseBotPathAdapter(this IConfigurationBuilder builder, bool isDevelopment = true)
```
carlosscastro, line 14: Naming: consider 'UseBotPathTransformer', 'UseBotPathMapper', 'UseBotPathConverter', since adapter refers to an object oriented pattern that could be confused and is a bit different.

runtime/dotnet/core/ComposerLuisExtensions.cs
carlosscastro: Also, the `ResourceExplorer` instance in the startup already has very optmized code to find files. Can you get the luis settings using the resource explorer? What changed that we need this search? Thanks in advance, I'd like to understand more in general about this scenario.

zidaneymar:
Since some comments are based on the jsruntime branch changes, and many things (botProjectDeploy.ts) have been changed on jsruntime branch, I will recreate a pr based on jsruntime branch.

PR #3649, opened 2020-07-16
ci: allow dependabot PRs to skip checks

beyackle:
## Description

Dependabot is a useful tool, but the PRs it autogenerates don't match the format we restrict our PRs to. This change will detect if a PR's author is Dependabot and, if so, let it skip the usual checks.

## Task Item

#minor


beyackle:
image

Screenshot of the GraphQL explorer showing the author.login field in action

PR #3681, opened 2020-07-22
fix: add missing formatMessage calls

### LINKED ISSUE #3458: Improve error messages for common errors
Let's compile a list of common errors that could use some TLC.

beyackle:
## Description

This adds two things to the codebase:
* formatMessage calls to where they were missing - I attempted to put them everywhere they were needed, including outside the "client" folder, and I have a list of notes with some questions I still have about this.
* A "yarn extract" action in the client's package.json file, which creates the file `locales/en-US.json` from all of the strings it finds in the client.

This also addresses issue #3458 by cleaning up and addressing the phrasing of the error messages that needed it.

This does not yet:
* Implement any mocklations or a way to switch to alternate locales - the way formatMessage works is to substitute localized strings for all of its calls when the format-message CLI tool is used as a build step, so we're going to end up with a full parallel client bundle for each region. This is apparently the way formatMessage is intended to work. This might be cause to re-evaluate formatMessage if trying to serve the right client bundle based on the user's browser locale ends up being difficult.

## Task Item

fixes #3458


Composer/packages/client/locales/en-US.json
```diff
@@ -0,0 +1,1259 @@
+{
+  "0_bytes_a1e1cdb3": {
```
GeoffCoxMSFT, line 2: These IDs are generated when you call formatMessage like this formatMessage('hello'), we could have formatMessage({id: 'mycomponent.helloMessage, default: 'hello'}).  Do you think semantic IDs help give context to localization?

Composer/packages/client/package.json
```diff
@@ -30,16 +31,18 @@
...
     "@reach/router": "^1.2.1",
     "@uifabric/fluent-theme": "^7.1.107",
     "@uifabric/icons": "^7.3.59",
     "@uifabric/react-hooks": "^7.4.12",
     "@uifabric/styling": "^7.13.7",
     "axios": "^0.19.2",
+    "babel-plugin-extract-format-message": "^6.2.3",
     "botbuilder-lg": "4.10.0-preview-147186",
     "format-message": "^6.2.3",
+    "format-message-cli": "^6.2.3",
```
GeoffCoxMSFT, line 45: Should this be in devDependencies?

Composer/packages/client/src/components/CreationFlow/FileSelector.tsx
```diff
@@ -345,7 +347,7 @@ export const FileSelector: React.FC<FileSelectorProps> = (props) => {
           <div data-is-focusable css={tableCell}>
             <div css={content} tabIndex={-1}>
               <IconButton
-                ariaLabel="Edit"
+                ariaLabel={formatMessage('Edit')}
```
GeoffCoxMSFT, line 350: Do ariaLabel's get localized?  I haven't seen that before.

Composer/packages/client/src/components/ErrorPopup/ErrorPopup.tsx
```diff
@@ -47,7 +48,7 @@ export const ErrorPopup = (props: ErrorPopupProps) => {
     >
       <div css={consoleStyle}>{props.error}</div>
       <DialogFooter>
-        <PrimaryButton text="Ok" onClick={closeDialog} />
+        <PrimaryButton text={formatMessage('Ok')} onClick={closeDialog} />
```
GeoffCoxMSFT, line 51: Should this be OK rather than Ok (I think the K is lowercase, but it is hard to tell in this font)?

Composer/packages/client/src/components/Modal/AlertDialog.tsx
```diff
@@ -2,32 +2,19 @@
...
 /** @jsx jsx */
-import { jsx, css } from '@emotion/core';
+import { jsx } from '@emotion/core';
 import { Dialog, DialogType, DialogFooter } from 'office-ui-fabric-react/lib/Dialog';
 import { PrimaryButton } from 'office-ui-fabric-react/lib/Button';
 import ReactDOM from 'react-dom';
+import formatMessage from 'format-message';
 
-import { dialogStyle } from './dialogStyle';
+import { dialogStyle, NORMAL_STYLE, CONSOLE_STYLE } from './dialogStyle';
```
GeoffCoxMSFT, line 11: I've been scrubbing through the code to merge styles into components rather than pick up styles from external modules.

Composer/packages/client/src/components/Modal/AlertDialog.tsx
```diff
@@ -45,9 +32,9 @@ type Props = {
 
 const AlertDialog = (props: Props) => {
   const { setting, onCancel, onConfirm } = props;
-  const { title, subtitle = '', confirmText = 'Ok', style = dialogStyle.normal } = setting;
+  const { title, subtitle = '', confirmText = formatMessage('Ok'), style = dialogStyle.normal } = setting;
```
GeoffCoxMSFT, line 35: Same comment OK vs Ok?

Composer/packages/client/src/components/Modal/AlertDialog.tsx
```diff
@@ -45,9 +32,9 @@ type Props = {
...
   if (!title) {
-    throw new Error('Confirmation modal must have a title');
+    throw new Error(formatMessage('Confirmation modal must have a title'));
```
GeoffCoxMSFT, line 37: Nit: It can be nice to have a period at the end of error message sentences. When they get combined it makes them easier to read.

Composer/packages/client/src/components/Modal/dialogStyle.ts
```diff
@@ -1,10 +1,30 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+import { css } from '@emotion/core';
+
 /**
  * Well known style names for the AlertDialog and ConfirmDialog.
  */
 export const dialogStyle = {
```
GeoffCoxMSFT, line 9: Constants used across dialogs components and shared styles are two different things.  I recommend a constants.ts and putting styles within the components.

Composer/packages/client/src/components/Modal/dialogStyle.ts
```diff
@@ -1,10 +1,30 @@
...
   normal: 'NORMAL',
   console: 'CONSOLE',
 };
+
+export const NORMAL_STYLE = css`
```
GeoffCoxMSFT, line 14: Is this naming casing a standard convention for styles?

Composer/packages/client/src/components/Pagination.tsx
```diff
@@ -45,6 +45,10 @@ const createDropdownOption = (pageCount: number) => {
   }));
 };
 
+// Note: these should be made RTL-aware at some point
+const PREVIOUS_LABEL = '← ' + formatMessage('Previous');
```
GeoffCoxMSFT, line 49: Could we use an icon in the button instead?

Composer/packages/client/src/components/ProjectTree/TriggerCreationModal.tsx
```diff
@@ -413,7 +413,7 @@ export const TriggerCreationModal: React.FC<TriggerCreationModalProps> = (props)
             <TextField
               data-testid="RegExField"
               errorMessage={formData.errors.regEx}
-              label={formatMessage('Please input regex pattern')}
+              label={formatMessage('Please input regEx pattern')}
```
GeoffCoxMSFT, line 416: Should this say 'regular expression pattern' and avoid the abbreviation?

Composer/packages/client/src/pages/language-generation/table-view.tsx
```diff
@@ -329,7 +329,7 @@ const TableView: React.FC<TableViewProps> = (props) => {
           iconProps={{ iconName: 'CirclePlus' }}
           onClick={() => {
             onCreateNewTemplate();
-            setMessage('item added');
+            setMessage(formatMessage('item added'));
```
GeoffCoxMSFT, line 332: Item added ?

Composer/packages/client/src/pages/publish/publishDialog.tsx
```diff
@@ -36,7 +36,9 @@ export const PublishDialog = (props) => {
             multiline
             label={formatMessage('Comment')}
             // styles={styles.textarea}
-            placeholder="Provide a brief description of this publish. It will appear on the publish history list"
+            placeholder={formatMessage(
+              'Provide a brief description of this publish. It will appear on the publish history list'
```
GeoffCoxMSFT, line 40: Nit: list.'

Composer/packages/extensions/adaptive-form/src/components/ErrorInfo.tsx
```diff
@@ -1,25 +1,28 @@
...
   <div style={{ marginRight: '20px' }}>
     <MessageBar isMultiline={false} messageBarType={MessageBarType.error}>
       <p style={{ whiteSpace: 'pre-wrap' }}>
-        <strong>Oops! An error occured in the form editor!</strong>
-        <br />
-        This is likely due to malformed data or missing functionality in Composer.
-        <br />
-        Try navigating to another node in the visual editor.
+        {formatMessage.rich(
+          '<strong>Oops! An error occured in the form editor!</strong><br />This is likely due to malformed data or missing functionality in Composer.<br />Try navigating to another node in the visual editor.',
```
GeoffCoxMSFT, line 13: It seems like it would be pretty easy for a localizer to accidentally mess up the HTML here.  I would consider doing 3 format messages here.

Composer/packages/lib/shared/src/resolverFactory.ts
```diff
@@ -42,7 +44,7 @@ export function importResolverGenerator(
     const targetFile =
       resources.find(({ id }) => id === `${targetId}.${locale}`) || resources.find(({ id }) => id === targetId);
 
-    if (!targetFile) throw new Error(`file not found`);
+    if (!targetFile) throw new Error(formatMessage(`file not found`));
```
GeoffCoxMSFT, line 47: File not found ?

Composer/packages/client/src/components/Modal/AlertDialog.tsx
beyackle: I just saw your email about it - I'll change these back, then.

Composer/packages/client/src/components/Modal/dialogStyle.ts
beyackle: I use CAPS_WITH_UNDERSCORES as a convention for constants. Do we have a team standard for these?

Composer/packages/client/src/components/Pagination.tsx
beyackle: Changed this to use icons. Pending comment from design: 
image

Composer/packages/client/src/components/ProjectTree/TriggerCreationModal.tsx
beyackle: Good question. This should be part of the same discussion as the "Ok/OK/Okay" thing.

Composer/packages/client/src/pages/language-generation/table-view.tsx
beyackle: This is for the screen-reader, so it doesn't need capitalization.

Composer/packages/extensions/adaptive-form/src/components/ErrorInfo.tsx
beyackle: This looks like a good use for that "id" argument so a translator knows they go together, too.

PR #3686, opened 2020-07-23
fix: validate skill empty & add duplicate dialog

### LINKED ISSUE #3284: provide warning if id & host endpoint not set when using skills

 I'm always frustrated when I forget to set them and it takes time to figure out why my skills don't work.

Provide warning if id & host endpoint not set when using skills:
image

### LINKED ISSUE #3401: [server] Issue creating new dialogs with the same name but different casing
## Describe the bug
If you create two dialogs with the same name but with different casing - 'DIALOG' and 'dialog' - the second dialog and dialog files are not created on the disk. Editing the second dialog causes an error. 

## Version
Master

## To Reproduce

1. Add a new dialog titled 'DIALOG'
2. Add a second dialog titled 'dialog'
3. Edit the second dialog created - 'dialog'
4. See error and notice only one dialog - 'DIALOG' - exists in the bot assets

## Expected behavior
Composer should either create both dialog directories on disk or it shouldn't allow users to create a second dialog with the same name regardless of casing.

## Screenshots
Jun-12-2020 12-56-58

### LINKED ISSUE #3687: multi-language validation issue
## Describe the bug

* After adding a language, I author .lg file inline and generate validation error. The error can be removed after refreshing the page. 

image

zhixzhan:
#### Warning style beside Start Bot button
Show warning in toolbar, if both has error and warning, will show error at first priority. 
imageimage

#### When bot call a skill
1. check dialog used skills exist (Error)
2. check skill required info (Warning)
image. 

#### When add a dialog
Prevent create same id dialog, ignore case

## Task Item

close #3284 
close #3401 
close #3687 

## Screenshots
Untitled2

### DIFF
--- a/Composer/packages/client/__tests__/pages/notifications/useNotifications.test.tsx
+++ b/Composer/packages/client/__tests__/pages/notifications/useNotifications.test.tsx
@@ -30,6 +30,7 @@ const state = {
           source: 'test',
         },
       ],
+      skills: ['https://yuesuemailskill0207-gjvga67.azurewebsites.net/manifest/manifest-1.0.json'],
     },
   ],
   luFiles: [
@@ -91,6 +92,14 @@ const state = {
       source: 'server',
     },
   ],
+  settings: {
+    skill: [
+      {
+        manifestUrl: 'https://yuesuemailskill0207-gjvga67.azurewebsites.net/manifest/manifest-1.0.json',
+        name: 'Email Skill',
+      },
+    ],
+  },
 };
 
 const initRecoilState = ({ set }) => {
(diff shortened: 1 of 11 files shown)


zhixzhan:
I feel a little tricky is the `endpoint url` is produced after start bot, and it's required by skill, so the user have to 
1. Start bot get `endpoint url`   
2. Fill it in composer skill page form, restart bot. 

First start bot would cause error when it run into call skill step, we should avoid this. but if we forbidden start bot, user can never get `endpoint url`. Discussed with @hibrenda, issue filed by @xieofxie

Issue #3480, opened 2020-06-24
Code hygiene: Rename index.ts and index.tsx files to specific names.


GeoffCoxMSFT:
Scrub through the code to avoid index.ts (barrel files) and name index.tsx components to semantically meaningful names.

This will

- Help devs learn and find code faster
- Allow bundling optimizations and HMR to function better

GeoffCoxMSFT:
This is an ongoing thing across releases, so closing for R11

PR #3691, opened 2020-07-23
chore: Gcox/toolbar hygiene

GeoffCoxMSFT:
## Description

Updated to move styles inside Toolbar component
Updated naming to Toolbar (rather than ToolBar) for consistency
Removed index.ts barrel file, Toolbar moved out of folder

## Task Item

#minor
[#3480](https://github.com/microsoft/BotFramework-Composer/issues/3480)

## Screenshots

N/A


Composer/packages/client/src/components/Toolbar.tsx
```diff
@@ -2,19 +2,62 @@
...
+
+export const actionButton = css`
+  font-size: 16px;
+  margin-top: 2px;
+  margin-left: 15px;
+`;
+
+// -------------------- IToolbarItem -------------------- //
+
+export type IToolbarItem = {
```
hatpick, line 43: nit: ISomething is interface convention.

PR #3708, opened 2020-07-24
draft: Bot Projects Initial Schema Representation

srinaath:
## Description
This is a first draft of the schema for the bot projects file and the schema for the provisioning profile


srinaath:
@cwhitten I figured we can use this branch to collaborate on schema and sample provisioning profile and bot projects file

srinaath:
@beyackle . Yeah, that decision is tricky. Essentially, the one click deployment is something at the core of this feature. I have abstracted all the keys and sensitive stuff into this provisioning profile file. We are persisting these keys currently in our appSettings.json when we create a publish target. I felt providing these deployment config in a more structured provisioningProfile file would carry the same risks. The core botproject file has an optional reference to the provisioning profile.

srinaath:
Should update the schema to support locales. Especially, how QNA config would change to support multiple locales

PR #3712, opened 2020-07-26
feat: Add Image Upload and Vision API Sample

bradrlaw:
This is a sample bot the demonstrates being able to upload a file or provide a URL that can be sent to the Computer Vision API.

## Description

Sample bot project that allows an image to be processed by the Azure Computer Vision service.  The bot will accept an image through a local file upload or by specifying a URL.  The image will be processed by the Computer Vision service and analyzed.  The results will presented back in a thumbnail card.

The sample has a carousel of cards for the welcome message that guides users through the pre-requisites to use the sample including creating a Computer Vision service and getting keys / endpoints.  In addition, it provides information on a SDK bug workaround necessary for the example to work ("the parent is missing" issue).

## Task Item

#minor


cwhitten:
@bradrlaw we're going to close this as we deprecate the samples directory in favor of providing samples & templates through an updated component model that we call packages. You can preserve this sample but expose it over the package manifest. It would not be stored in Composer directly but consumed from a nuget or npm feed.

PR #3713, opened 2020-07-26
feat: Add a form submission sample

bradrlaw:
## Description

Sample bot for composer to demonstrate handling data from multiple forms in a single bot.  The sample creates two forms with adaptive cards and has two dialogs to process the data from the forms.  The sample shows how to use the unknown intent to catch the form submissions and distinguish between which form had its submit action triggered.

## Task Item

#minor


bradrlaw:
@cwhitten Got it, will update the example when the new release hits.

cwhitten:
@bradrlaw  we're going to close this as we deprecate the samples directory in favor of providing samples & templates through an updated component model that we call packages. You can preserve this sample but expose it over the package manifest. It would not be stored in Composer directly but consumed from a nuget or npm feed.

PR #3743, opened 2020-07-29
feature: Intellisense for Composer 

### LINKED ISSUE #3663: Add intellisense to LG/LU in Composer Server
- [ ] Determine if intellisense can be added to existing servers or if a new server has to be created
- [ ] Add intellisense with UTs into composer server
- [ ] Ensure documentation (MD | code comments) of any modifications to the standard language protocol or client/server shared dependencies (like on enum values)

### LINKED ISSUE #3664: Add shared control infrastructure for intellisense/auto-complete
- [ ] As needed, define and implement the control that will show auto-complete suggestions
- [ ] Call out here which aspects of Sangwoo's design will/won't be included or modifications
- [ ] Unit test (or test) against control types (input, textarea, etc.) where you expect the control used.

### LINKED ISSUE #3665: Update property grid controls to support scoped variable auto-complete using intellisense
- [ ] List the control types from the property grid that will support auto-complete/intelli-sense
- [ ] Update each control in turn to support showing auto-complete options
- [ ] Wire up controls to the intellisense information from the composer server
- [ ] Update UTs to stub out auto-complete for existing tests

LouisEugeneMSFT:
## Description

Adding Intellisense to Composer textfields.

This PR includes:
- a new language server (see commit 1)
- a new ui textfield component with a hook that can communicate through LSP to the server (see commit 2)
- integration of intellisense into Composer for expressions and for setting a property (see commit 3)

## Task Item

fixes #3663 
fixes #3664 
fixes #3665

## Screenshots

image

image


cwhitten:
I love the `useLanguageServer`, this can accrue value anywhere we want to wire into the language servers.

What about `useIntellisense(...)` instead of `<IntellisenseTextField />`?

`<IntellisenseTextField />` houses multiple useEffects and event listeners that I believe could be managed in a `useIntellisense(...)` hook, and I'd like to decouple this from the DOM if possible for a few reasons:

1. As a developer, I'd much prefer to be able to use Fluent's docs directly with `<TextField />` and other input controls and augment my code with `useIntellisense(...)` hook. 
2. Going with a hooks approach means that as you want to wrap new input controls with intellisense, you only need to pull in the hook instead of creating a new component that wraps the input control and copies the intellisense logic into it.

Going from:
```jsx
<IntellisenseTextField {...} />
<IntellisenseInputControl2 {...} />
<IntellisenseInputControl3 {...} />
<IntellisenseInputControl4 {...} />
<IntellisenseInputControl5 {...} />
```

to
```jsx
const [foo, setFoo] = useIntellisense(...)
<TextField {...} />

const [foo, setFoo] = useIntellisense(...)
<InputControl2 {...} />

const [foo, setFoo] = useIntellisense(...)
<InputControl3 {...} />

const [foo, setFoo] = useIntellisense(...)
<InputControl4 {...} />

const [foo, setFoo] = useIntellisense(...)
<InputControl5 />
```

LouisEugeneMSFT:
Hi @cwhitten 

There are a few challenges with this approach, I am curious to have your opinion on them:
- The IntellisenseTextField component relies on some textfield specific props such as _**onKeyDown**_. Additionally everything is wrapped in a div that has its own handlers. All of that is necessary to react to keyboard events (such as navigating the list of suggestions using up and down, selecting something using "Enter" or closing the list of suggestions using "Escape").
- Additionally, having references to the components is necessary for click handlers (aka closing the list of suggestions if you click away from the textfield or one of the suggestions). Is is also useful in order to show the list of suggestions in the right place (although I see easy workarounds for that).

Happy to chat about those things in person

Composer/packages/extensions/adaptive-form/src/components/fields/ExpressionField/ExpressionEditor.tsx
```diff
@@ -1,20 +1,24 @@
...
 
-import React from 'react';
-import { FieldProps } from '@bfc/extension';
-import { TextField } from 'office-ui-fabric-react/lib/TextField';
+import { FieldProps, useShellApi } from '@bfc/extension';
+import { IntellisenseTextField } from '@bfc/intellisense';
 import { Icon } from 'office-ui-fabric-react/lib/Icon';
+import React from 'react';
+
+let intellisenseWsUrl = window.location.origin.replace(/^https/, 'wss').replace(/^http/, 'ws');
```
GeoffCoxMSFT, line 9: Consider if the URL change should happen outside a UI component.

Composer/packages/extensions/adaptive-form/src/components/fields/ExpressionField/ExpressionEditor.tsx
```diff
@@ -1,20 +1,24 @@
...
 
-  const handleChange = (e: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>, newValue?: string) => {
-    onChange(newValue);
-  };
+  const { projectId } = useShellApi();
 
   return (
-    <TextField
+    <IntellisenseTextField
+      url={intellisenseWsUrl}
```
GeoffCoxMSFT, line 19: Consider avoiding abbreviation in the name - intellisenseServerUrl

Composer/packages/extensions/adaptive-form/src/components/fields/ExpressionField/ExpressionEditor.tsx
```diff
@@ -25,7 +29,7 @@ const ExpressionEditor: React.FC<FieldProps> = (props) => {
         errorMessage: { display: 'none' },
       }}
       value={value}
-      onChange={handleChange}
+      onChange={onChange}
```
GeoffCoxMSFT, line 32: I'm not sure you can get rid of the handleChange - it was sending the 2nd param as the value.  I think this would change the FormField<T>.onChange<T> to have the T be the event rather than the value.

Composer/packages/extensions/adaptive-form/src/components/fields/index.ts
```diff
@@ -16,3 +16,4 @@ export * from './RegexIntentField';
 export * from './SelectField';
 export * from './StringField';
 export * from './UnsupportedField';
+export * from './IntellisenseField';
```
GeoffCoxMSFT, line 19: We should remove index files except at the package/extension/plug-in entry point.

Composer/packages/extensions/intellisense/src/components/CompletionElement.tsx
```diff
@@ -0,0 +1,106 @@
...
+    display: 'flex',
+    alignItems: 'center',
+  },
+  selectedElement: { backgroundColor: '#ddd' },
+  text: { fontSize: '15px' },
+  icon: { marginRight: '5px' },
+};
+
+const renderIcon = (kind: CompletionItemKind | undefined): JSX.Element => {
+  let icon: JSX.Element = <> </>;
```
GeoffCoxMSFT, line 27: Nit: You don't have to declare a mutable let variable here if you put a default: in your switch

Composer/packages/extensions/intellisense/src/components/CompletionElement.tsx
```diff
@@ -0,0 +1,106 @@
...
+    case CompletionItemKind.Enum:
+      icon = <FontIcon iconName="BulletedList" style={styles.icon} />;
+      break;
+  }
+
+  return icon;
+};
+
+const renderLabelWithCharacterHighlights = (matches: FuseJsMatch[]): JSX.Element => {
+  const renderMatch = (match: FuseJsMatch, segmentIndex: number): JSX.Element => {
```
GeoffCoxMSFT, line 45: Nit: This method can be declared outside this parent as it doesn't use any closure.

Composer/packages/extensions/intellisense/src/components/CompletionElement.tsx
```diff
@@ -0,0 +1,106 @@
...
+      icon = <FontIcon iconName="VariableGroup" style={styles.icon} />;
+      break;
+    case CompletionItemKind.Enum:
+      icon = <FontIcon iconName="BulletedList" style={styles.icon} />;
+      break;
+  }
+
+  return icon;
+};
+
```
GeoffCoxMSFT, line 43: Nit: You generally can let the return value be inferred.  Sometimes this requires putting the HTML in parens.  Consider either always having it or not - for consistency in this file

Composer/packages/extensions/intellisense/src/components/CompletionElement.tsx
```diff
@@ -0,0 +1,106 @@
...
+const renderLabelWithCharacterHighlights = (matches: FuseJsMatch[]): JSX.Element => {
+  const renderMatch = (match: FuseJsMatch, segmentIndex: number): JSX.Element => {
+    let firstIndex = 0;
+    const lastIndex = match.value.length;
+
+    const items = match.indices.map((m, spanIndex) => {
+      const firstSpan = <span>{match.value.slice(firstIndex, m[0])}</span>;
+      const secondSpan = <span style={{ color: 'blue' }}>{match.value.slice(m[0], m[1] + 1)}</span>;
```
GeoffCoxMSFT, line 51: Are you certain that there is a m[0] and m[1]?

Composer/packages/extensions/intellisense/src/components/CompletionElement.tsx
```diff
@@ -0,0 +1,106 @@
...
+          {completionItem.data.matches
+            ? renderLabelWithCharacterHighlights(completionItem.data.matches)
+            : completionItem.label}
+        </div>
+      </div>
+    </TooltipHost>
+  );
+};
+
+export default CompletionElement;
```
GeoffCoxMSFT, line 106: Why not just export CompletionElement?  I don't understand why so many Composer files have export default, but maybe I'm missing something.

Composer/packages/extensions/intellisense/src/components/CompletionList.tsx
```diff
@@ -0,0 +1,47 @@
...
+    width: '100%',
+    backgroundColor: 'white',
+    overflowY: 'auto',
+    overflowX: 'hidden',
+    boxShadow: '0 4px 8px 0 rgba(0, 0, 0, 0.2), 0 6px 20px 0 rgba(0, 0, 0, 0.19)',
+    zIndex: 2000,
+  },
+};
+
+const CompletionList = React.forwardRef<
```
GeoffCoxMSFT, line 23: It is worth a comment why you are exposing the completionList as a div element for manipulation outside the component.  Consider if you want to provide an explicit method on this rather than expose the root is a div.

Composer/packages/extensions/intellisense/src/components/IntellisenseTextField.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+export const IntellisenseTextField = (
+  props: {
+    url: string;
+    scopes: string[];
+    projectId?: string;
+    id: string;
+    value?: string;
+    onChange: (newValue: string) => void;
+  } & Omit<Partial<ITextFieldProps>, 'onChange'>
```
GeoffCoxMSFT, line 18: This feels strange to omit an expected event and then redefine it with a different prototype.

Composer/packages/extensions/intellisense/src/components/IntellisenseTextField.tsx
```diff
@@ -0,0 +1,170 @@
...
+      }
+
+      // This makes sure we do not show the completion items after a value is picked from the list
+      didComplete.current = true;
+      setShowCompletionList(false);
+    }
+  };
+
+  return (
+    <div onKeyUp={onKeyUpMainComponent} ref={mainContainerRef} style={{ position: 'relative' }}>
```
GeoffCoxMSFT, line 148: Consider if inline styles are required here or if you can use emotion/styled-components.  You may want to provide the className in the props to allow callers to override styles.

Composer/packages/extensions/intellisense/src/hooks/useLanguageServer.ts
```diff
@@ -0,0 +1,118 @@
...
+import { w3cwebsocket as W3CWebSocket } from 'websocket';
+import {
+  getCompletionRequestMessage,
+  getConfigurationChangedMessage,
+  getDocumentChangedMessage,
+  getInitializeMessage,
+  getTextDocumentOpenedMessage,
+} from '../utils/lspMessagesUtils';
+
+const LANGUAGE_NAME = 'intellisense';
```
GeoffCoxMSFT, line 15: I don't think UPPER_CASE is the right style for constants in TS.  The style across the codebase seems inconsistent, but this feels like a holdover from some C++ stuff.

Composer/packages/extensions/intellisense/src/hooks/useLanguageServer.ts
```diff
@@ -0,0 +1,118 @@
...
+    ws.current.onopen = () => {
+      ws.current.send(JSON.stringify(getInitializeMessage(scopes, projectId)));
+      ws.current.send(
+        JSON.stringify(
+          getTextDocumentOpenedMessage(documentUri, LANGUAGE_NAME, latestDocumentVersion.current, textFieldValue)
+        )
+      );
+    };
+    ws.current.onmessage = (messageText) => {
+      handleMessage(messageText);
```
GeoffCoxMSFT, line 52: Usage before declaration

Composer/packages/extensions/intellisense/src/hooks/useLanguageServer.ts
```diff
@@ -0,0 +1,118 @@
...
+  React.useEffect(() => {
+    if (ws.current.readyState === WebSocket.OPEN) {
+      ws.current.send(JSON.stringify(getConfigurationChangedMessage(scopes, projectId)));
+    }
+  }, [scopes, projectId]);
+
+  // When textField value changes, update backend memory and get latest completion results
+  React.useEffect(() => {
+    if (ws.current.readyState === WebSocket.OPEN) {
+      updateBackendMemory(textFieldValue);
```
GeoffCoxMSFT, line 66: Usage before declaration

Composer/packages/extensions/intellisense/src/hooks/useLanguageServer.ts
```diff
@@ -0,0 +1,118 @@
...
+      getCompletionItems();
+    }
+  }, [textFieldValue]);
+
+  // Get completion results when selection changes
+  React.useEffect(() => {
+    if (ws.current.readyState === WebSocket.OPEN) {
+      getCompletionItems();
```
GeoffCoxMSFT, line 74: Usage before declaration

Composer/packages/tools/language-servers/intellisense/package.json
```diff
@@ -0,0 +1,21 @@
...
+    "vscode-ws-jsonrpc": "^0.1.1",
+    "ws": "^7.2.0",
+    "@bfc/lg-languageserver": "*"
+  },
+  "scripts": {
+    "dev": "ts-node src/server.ts",
+    "start": "ts-node lib/server.js",
+    "build": "rimraf ./lib && tsc --build tsconfig.build.json"
+  },
+  "author": "Louis Eugene",
```
GeoffCoxMSFT, line 19: Microsoft

Composer/packages/tools/language-servers/intellisense/src/resolvers/expressions.ts
```diff
@@ -0,0 +1,23 @@
...
+import { buildInfunctionsMap } from '@bfc/lg-languageserver';
+import { CompletionItem, CompletionItemKind } from 'vscode-languageserver';
+
+export const expressionsResolver = (): CompletionItem[] => {
+  return Array.from(buildInfunctionsMap).map((item) => {
+    const [key, value] = item;
+    return {
+      label: key,
+      kind: CompletionItemKind.Function,
+      insertText: key + '(' + removeParamFormat(value.Params.toString()) + ')',
```
GeoffCoxMSFT, line 12: Consider template formatting over string concats.

Composer/packages/tools/language-servers/intellisense/src/utils/intellisenseServerUtils.ts
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { Position, Range, TextDocument } from 'vscode-languageserver';
+
+export const getRangeAtPosition = (document: TextDocument, position: Position): Range | undefined => {
+  const text = document.getText();
+  const line = position.line;
+  const pos = position.character;
+  const lineText = text.split('\n')[line];
```
GeoffCoxMSFT, line 10: Consider a check that line is not outside the range of the split result.

Composer/packages/tools/language-servers/intellisense/src/utils/intellisenseServerUtils.ts
```diff
@@ -0,0 +1,33 @@
...
+  let match: RegExpMatchArray | null;
+  const wordDefinition = /[a-zA-Z0-9_/.-]+/g;
```
GeoffCoxMSFT, line 12: Nit: Declare const outside function

Composer/packages/tools/language-servers/intellisense/tsconfig.build.json
```diff
@@ -0,0 +1,7 @@
+{
+  /* Options used for building production code (tests excluded) */
```
GeoffCoxMSFT, line 2: Unless something changed, I don't think JSON supports comments.

Composer/packages/extensions/intellisense/src/components/CompletionElement.tsx
hatpick: Actually all this function is doing is figuring out the iconName, and passing it to FontIcon, so DRY?

Composer/packages/tools/language-servers/intellisense/tsconfig.build.json
LouisEugeneMSFT: Hmmm. I re-used the same tsconfig.build.json found in every folder.

Composer/packages/tools/language-servers/intellisense/src/utils/intellisenseServerUtils.ts
LouisEugeneMSFT: done :)

Composer/packages/tools/language-servers/intellisense/src/utils/intellisenseServerUtils.ts
LouisEugeneMSFT: ok!

Composer/packages/tools/language-servers/intellisense/src/resolvers/expressions.ts
LouisEugeneMSFT: done

Composer/packages/tools/language-servers/intellisense/package.json
LouisEugeneMSFT: oops

Composer/packages/extensions/intellisense/src/hooks/useLanguageServer.ts
LouisEugeneMSFT: fixed

Composer/packages/extensions/intellisense/src/hooks/useLanguageServer.ts
LouisEugeneMSFT: fixed

Composer/packages/extensions/intellisense/src/hooks/useLanguageServer.ts
LouisEugeneMSFT: fixed

Composer/packages/extensions/intellisense/src/hooks/useLanguageServer.ts
LouisEugeneMSFT: I tried to be consistent with the rest of the app. Check out lg.ts and lu.ts.

Composer/packages/extensions/adaptive-form/src/components/fields/ExpressionField/ExpressionEditor.tsx
LouisEugeneMSFT: Renamed to intellisenseWebSocketUrl

Composer/packages/extensions/adaptive-form/src/components/fields/ExpressionField/ExpressionEditor.tsx
LouisEugeneMSFT: Hmm. I could potentially remove url as a prop of IntellisenseTextField and move the generation of this url to a new file under packages/extensions/intellisense/src/utils. The drawback of this is that then the url is "hardcoded" into the Intellisense package, therefore communicating to 1 single server. If I wanted to create an util where the component is used then I would have to either duplicate the util per package, or create new dependency between packages in the case of 1 util. What do you think is best?

Composer/packages/extensions/adaptive-form/src/components/fields/ExpressionField/ExpressionEditor.tsx
LouisEugeneMSFT: Actually no. The only thing that handleChange did is get the newValue and call onChange. The event was not used.

Composer/packages/extensions/adaptive-form/src/components/fields/index.ts
LouisEugeneMSFT: Should this be done as part of a separate PR? Otherwise many more files would be included in this PR..

Composer/packages/extensions/intellisense/src/components/CompletionElement.tsx
LouisEugeneMSFT: Good points. I refactored to address both your comments.

Composer/packages/extensions/intellisense/src/components/CompletionElement.tsx
LouisEugeneMSFT: done

Composer/packages/extensions/intellisense/src/components/CompletionElement.tsx
LouisEugeneMSFT: The other comment addressed this :)

Composer/packages/extensions/intellisense/src/components/CompletionElement.tsx
LouisEugeneMSFT: Yes, based on the returned results from a FUSE.js search.

Composer/packages/extensions/intellisense/src/components/CompletionElement.tsx
LouisEugeneMSFT: Is your comment about having everything inline aka export default on line 69. Or about not having default aka importing as {CompletionElement} instead of CompletionElement?

Composer/packages/extensions/intellisense/src/components/IntellisenseTextField.tsx
LouisEugeneMSFT: This was to conform to the FieldProps onChange definition (packages/extensions/extension/src/types/form.ts) while enabling some ITextField customization if needed.

Composer/packages/extensions/adaptive-form/src/components/fields/index.ts
GeoffCoxMSFT: Sure.

Composer/packages/extensions/intellisense/src/components/CompletionElement.tsx
GeoffCoxMSFT: It is worth a comment to say that Fuse searches always return at least 2 items then.  Otherwise, anyone reading this code fresh will have the same concern.

Composer/packages/extensions/intellisense/src/components/CompletionElement.tsx
GeoffCoxMSFT: I think there shouldn't be an export default.  I would export the component directly.

Composer/packages/extensions/intellisense/src/components/IntellisenseTextField.tsx
GeoffCoxMSFT: This goes back to the comment I had on FieldProps<T> and the onChange expecting the param to be of type T.  Other fields expose an onChange with the event which the wrapping component then has handleChange to raise the onChange<T>.  In this case you are exposing the onChange<T> directly so you don't need the handleChange.  This is fine, but there isn't consistency between the intellisense fields and the other fields.

Composer/packages/extensions/intellisense/src/hooks/useLanguageServer.ts
GeoffCoxMSFT: Blurg.

Composer/packages/extensions/intellisense/src/components/IntellisenseTextField.tsx
```diff
@@ -0,0 +1,179 @@
...
+  // When textField value is changed
+  const onValueChanged = (newValue: string) => {
+    setTextFieldValue(newValue);
+
+    onChange(newValue);
+  };
+
+  // Set textField value to completion item value
+  const setValueToSelectedCompletionItem = (index: number) => {
+    if (completionItems.length >= index) {
```
a-b-r-o-w-n, line 97:
```suggestion
    if (completionItems.length > index) {
```

Composer/packages/tools/language-servers/intellisense/src/intellisenseServer.ts
```diff
@@ -0,0 +1,124 @@
...
+  IConnection,
+  InitializeParams,
+  TextDocuments,
+} from 'vscode-languageserver';
+import { CompletionItem, CompletionList } from 'vscode-languageserver-types';
+import { expressionsResolver } from './resolvers/expressions';
+import { userVariablesResolver } from './resolvers/userVariables';
+import { variableScopesResolver } from './resolvers/variableScopes';
+import { getCompletionString, getRangeAtPosition } from './utils/intellisenseServerUtils';
+const Fuse = require('fuse.js');
```
a-b-r-o-w-n, line 16: Why is this not imported?

Composer/packages/extensions/intellisense/src/components/IntellisenseTextField.tsx
LouisEugeneMSFT: thanks!

Composer/packages/tools/language-servers/intellisense/src/intellisenseServer.ts
LouisEugeneMSFT: There is a namespace conflict error...

a-b-r-o-w-n: Can you expand on this?

LouisEugeneMSFT: Soroush helped me fix this. "esModuleInterop" needed to be true.

PR #3751, opened 2020-07-30
fix: composer loading bot forever due to unexpected main dialog

### LINKED ISSUE #3428: composer loading bot forever due to unexpected main dialog
Based on the discussion with @lei9444  . This is the "forever loading" issue.

- in case ".dialog" file exists under bot root directory as below
image
- open this bot will not loading any main dialog and will be in loading  state forever
image

we should pop up error to user instead of in loading state forever.

lei9444:
## Description
newdialog

In this PR if the name is empty, disable the "OK" button

Do we need to use { validateOnMount: true }? This will show error message when we open the dialog first time

## Task Item
closes #3428

### DIFF
--- a/Composer/packages/client/src/pages/design/createDialogModal.tsx
+++ b/Composer/packages/client/src/pages/design/createDialogModal.tsx
@@ -43,9 +43,11 @@ export const CreateDialogModal: React.FC<CreateDialogModalProps> = (props) => {
           return formatMessage('Duplicate dialog name');
         }
       },
+      defaultValue: '',
     },
     description: {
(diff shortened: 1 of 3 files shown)


boydc2014:
I see the root cause, but for the fix, i think we should also validate in server side, not just in client, so that any invalid input, for whatever reason, that flows into server is rejected, make sense?

lei9444:
> let's also validate in server and make sure it throws error

Yes, server has validate now, in main branch will throw error, I reproduced this in electron version 1.0.2

PR #3755, opened 2020-07-31
feat: provide error detail in popup.

### LINKED ISSUE #3409: Something went wrong
Something went wrong immediately after creating the app. I tried to access the trigger and it show that something has went wrong

### LINKED ISSUE #3771: Something went wrong in editing SendResponse for HeroCard
Reporting per message.
(ver Release: 1.0.2)

```
#StartSurvey
[Herocard
    title = ''
    subtitle = Hello. Help us improve our product. Please take our survey!  
    text = The survey contains 9 questions and should only take a few minutes.
    images = https://sec.ch9.ms/ch9/7ff5/e07cfef0-aa3b-40bb-9baa-7c9ef8ff7ff5/buildreactionbotframework_960.jpg
    buttons = Start
]
```

image

zhixzhan:
## Task Item
close #3409 
close #3771 

## Screenshots
Untitled4


Composer/packages/client/src/components/ErrorBoundary.tsx
```diff
@@ -11,16 +11,28 @@ import { StateError } from '../recoilModel/types';
...
-    a: ({ children }) => (
-      <a key="a" href={githubIssueUrl} rel="noopener noreferrer" style={{ color: `greenyellow` }} target="_blank">
-        {children}
-      </a>
-    ),
-  }),
-  summary: formatMessage('Something went wrong!'),
+const genericErrorTitle = 'Something went wrong!';
+
+const errorToShow = (error: any): StateError => {
```
yeze322, line 16: It can have a better name as a function

Composer/packages/client/src/components/ErrorBoundary.tsx
```diff
@@ -11,16 +11,28 @@ import { StateError } from '../recoilModel/types';
...
+  const msg: string = typeof error === 'string' ? error : error?.message || error?.detail;
+  let messageHtml = 'If this problem persists, please file an issue on <a>GitHub</a>';
+  if (msg) {
+    messageHtml += `<details>${msg}</details>`;
+  }
+  const summary: string = msg ? msg.substring(0, 20) + '...' : genericErrorTitle;
+
+  return {
+    message: formatMessage.rich(messageHtml, {
```
yeze322, line 25: `formatMessage` only receives static strings and thus the eslint rule throws an error. Seems we don't need that `formatMessage.rich` when error messages are dynamic.

We can only wrap line 18 with formatMessage and manually build a message (either a string or JSX).

Composer/packages/client/src/components/ErrorBoundary.tsx
```diff
@@ -11,16 +11,28 @@ import { StateError } from '../recoilModel/types';
...
+      // eslint-disable-next-line react/display-name
```
yeze322, line 26: Another thing I'm wondering here is, shall we define the render logic of error messages in this module `ErrorBoundary`?

Ideally, the duty of `ErrorBoundary` should be only collecting errors; it's another component `ErrorPopup` who should care about how to render an error message.

So we can move those error transforming logic to ErrorPopup.

zhixzhan:
Hi @vivekkshankar , I'm adding a 'Details' collapse in the black area, to provide more information if user reported bug.
Do you think it make sense or should have a better design ?

cwhitten:
@zhixzhan @vivekkshankar I went ahead and took this but I agree that we should take a design/UX pass on rendering error messages

PR #3763, opened 2020-08-03
fix: Remove the logic of forcing the autoEndDialog to be false

luhan2017:
## Description
Fix the bug that autoEndDialog is forced to be false when it is not skillMode.
Now, after this fix, the autoEndDialog can be set by the FormDialog in Composer. 

We only force it to true when it is skillMode.

#minor


boydc2014:
@cwhitten  I think this one is good to go to provide a temp mitigation to at least not force override user's intention.  What do you think?

PR #3784, opened 2020-08-05
fix: use different shellData in Form and Flow's ExtensionContext

### LINKED ISSUE #3783: LgEditor shows wrong content and is not editable
## Describe the bug

(reported by @lei9444 )

lg-issue

## Browser

- [ ] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

yeze322:
## Description

fixes #3783 
refs #3721 

Let Flow Editor consumes shellData from `useShell('FlowEditor')`;
let Property Editor consumes shellData from `useShell('PropertyEditor')`

The difference is when an Action is selected, Property Editor's shellData should be the Action json, compared with Flow's data is whole dialog json.

image


lei9444:
I have tested it and it works

PR #3785, opened 2020-08-05
fix: fix lubuild error while publishing to azure

zidaneymar:
## Description
fix lubuild in azure publishing, same as pr: https://github.com/microsoft/BotFramework-Composer/pull/3764

## Task Item
#minor

### DIFF
--- a/Composer/packages/server/package.json
+++ b/Composer/packages/server/package.json
@@ -22,7 +22,7 @@
   },
   "author": "",
   "nodemonConfig": {
-    "exec": "cross-env TS_NODE_FILES=true node --inspect=9228 -r ts-node/register src/init.ts",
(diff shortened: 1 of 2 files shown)


Composer/packages/server/package.json
```diff
@@ -22,7 +22,7 @@
   },
   "author": "",
   "nodemonConfig": {
-    "exec": "cross-env TS_NODE_FILES=true node -r ts-node/register src/init.ts",
+    "exec": "cross-env TS_NODE_FILES=true node --inspect=9228 -r ts-node/register src/init.ts",
```
a-b-r-o-w-n, line 25:
```suggestion
    "exec": "cross-env TS_NODE_FILES=true node -r ts-node/register src/init.ts",
```

Issue #3809, opened 2020-08-07
Localization/Internationalization


beyackle:
## Describe the bug

Composer is currently only usable in English; users who need support for other languages are being left out. Note that this refers to the Composer app itself, not the bots it produces; the effort for multi-language bot authoring is already well underway. You should be able to write a bot in any supported language and produce a bot that speaks any supported language independently of each other.

## Expected behavior

Composer should match a user's locale settings (whether it gets them from the browser or somewhere in its own Settings page).

PR #3810, opened 2020-08-07
feature: create l10n-extract script

beyackle:
## Description

While formatMessage is a good way to extract strings from the JavaScript parts of Composer, we did not have a way to extract strings from the schema files. This provides a new script to extract just the relevant parts of a schema and produce something we can send off to be translated without a translator needing to see the rest of the schema. The intent is for these translated files to get merged in as plugins.

The l10n-extract script will take a single file (in JSON format, just like our schemas are) and produce a stripped-down version of it consisting only of the fields "title", "subtitle", "label", and "description" and any object containing another object with at least one of those fields. The new file will be placed in the same place, with ".en-US.json" appended to its name.

## Task Item

refs #3809 

## Screenshots

image


a-b-r-o-w-n:
Also, I wonder if this PR should include the server support for loading the locale uischemas.

beyackle:
The schema file contains text like the sidebar's various titles and descriptions, so those fields will need to be localized. Loading a localized schema from the server is something I'd like to handle in a future PR; this one is just providing a convenient way to extract all the fields we care about.
image

Composer/packages/server/schemas/locales/en-US/sdk.uischema
```diff
@@ -0,0 +1,326 @@
+{
```
a-b-r-o-w-n, line 1: The naming convention for these will just be `sdk.en-us.uischema`

beyackle:
Superseded by #3913.

PR #3738, opened 2020-07-29
fix: LG/LU editing performance optimize

zhixzhan:
## Description
Optimize LG/LU editing performance by:
1. LG/LU parser side delta-parsing
2. Composer side refine utils methods, decrease parsing invoke as low frequency as possible.

LU update time is 1/5 ~ 1/10 compared to previous, LG remain no obvious gain but data flow are optimized.

## Task Item
#minor

## Screenshots
Current LG whole-file & inline editing:
Untitled1
Current LU whole-file & inline editing:
Untitled2

Previous LG whole-file & inline editing:
Untitled3

Previous LU whole-file & inline editing:
Untitled4

PR #3337, opened 2020-06-05
fix: improve the user experience about lg editor

lei9444:
## Description
The lg parser takes a long time to parse a large file(20k lines will take 115s). And some lg template related functions are in UI thread. So the ui is blocked by the parser.

Now

- move all parser function to web worker. complex calculation will not block the UI thread 
- update the botbuild-lg version. The parse time decreases from 115s to 10s (20k lines).
- add worker for lg lsp to parse the lg file
But there is still an issue, the text in dialog will take a long time to show the update: 
test1

## Task Item
refs #3295 
refs #3314

PR #3816, opened 2020-08-10
fix: lg editor validation error & intelligence

### LINKED ISSUE #3815: Inline lg validation is wrong on import file
## Browser

- [ ] Electron distribution
- [x] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [x] macOS
- [ ] Windows
- [ ] Ubuntu

## Screenshots
image

zhixzhan:
## Description

fix lg editor validation error & intelligence, bug introduced by #3738 & #3337 

## Task Item
close #3815

## Screenshots
image

### DIFF
--- a/Composer/packages/tools/language-servers/language-generation/src/LGServer.ts
+++ b/Composer/packages/tools/language-servers/language-generation/src/LGServer.ts
@@ -16,7 +16,7 @@ import {
 import { TextDocumentPositionParams, DocumentOnTypeFormattingParams } from 'vscode-languageserver-protocol';
 import get from 'lodash/get';
 import { filterTemplateDiagnostics, isValid, lgUtil } from '@bfc/indexers';
-import { MemoryResolver, ResolverResource, LgFile } from '@bfc/shared';
+import { MemoryResolver, ResolverResource, LgFile, importResolverGenerator } from '@bfc/shared';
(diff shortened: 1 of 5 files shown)


boydc2014:
Can you add some details about the cause and the fix to better understand the context

boydc2014:
I can repro the issue without this fix and confirm this PR fixed it.

zhixzhan:
intelligence not work is because this line only return templates of current lg file, but we need is allTemplate(include imported file)
https://github.com/microsoft/BotFramework-Composer/pull/3337/files#r467668623

Composer/packages/tools/language-servers/language-generation/src/LGServer.ts
```diff
@@ -166,7 +166,8 @@ export class LGServer {
         const lgTextFile = lgTextFiles.find((item) => item.id === fileId);
         if (lgTextFile) {
           const lgFile = lgUtil.parse(lgTextFile.id, lgTextFile.content, lgTextFiles);
-          return lgUtil.updateTemplate(lgFile, templateId, { body: content });
```
zhixzhan, line 169: lgUtil.updateTemplate now accept `resolver` and return `lgFile`, here miss `resolver` then diagnostic show imported file not find

PR #3830, opened 2020-08-11
feat: use more specific placeholders

### LINKED ISSUE #3565: form: improve example resolution
Currently, Composer uses the leaf node of the schema to get the placeholder. This isn't useful if a property defines its own `examples` but also includes a $ref with `examples`. We should prefer the `examples` on the property schema and fallback to the $ref's `examples` if there are none defined.

The example for the `Property` field below is bad because we're expecting a property path not prompt text as in the example. A better example would be `ex. dialog.result`.
 
image

LouisEugeneMSFT:
## Description

When getting placeholder text for a field, Composer tries to grab the most specific examples from the schema tree. 
The problem though is that those leafs are references to basic types such as "string" or "expression", which have generic examples (then used for placeholder text). To take an example, every textfield in the app will have a placeholder of "Hello ${user.name}", which is not what we want. 

This PR makes sure that we get the placeholder of the parent before the ref to a generic type. This is an improvements but still has its own set of challenges:
A) The sdk needs to be updated in multiple places because it currently has some wrong examples (see screenshot). @vishwacsena 
B) Even if those examples were correct, they would still not be to the highest level of specificity possible. For example if a field can be a string or an expression, we would want an example for each. But this is currently not supported by the sdk schema (and would be hard to - see note i))

Note i): To take the example from the screenshot, "root.property" has a placeholder of "ex. $birthday, dialog.${user.name}, =f(x)". The possible types (string and expression) are known through a reference. Therefore "root.property" cannot have a mapping between types and examples without resolving that reference, which defeats the purpose of having refs in the first place. We are left with having to pick between either the top level example ("ex. $birthday, dialog.${user.name}, =f(x)"), which could be more specific based on type or examples from generic types ("Hello ${user.name}" and "=concat('x','y','z')") which are specifc but generic (aka that's not what we want)...

## Task Item

closes #3565 

## Screenshots

image


a-b-r-o-w-n:
@LouisEugeneMSFT Can you rebase this to only contain the commits for this PR? It looks like there are a bunch of extra commits.

Composer/packages/extensions/adaptive-form/src/utils/uiOptionsHelpers.ts
```diff
@@ -39,15 +39,22 @@ export function getUiPlaceholder(props: FieldProps): string | undefined {
...
+    const regex = `${schema.pattern}`;
+    const placeholderExamples = fieldUIPlaceholder.split(',').map((example) => example.trim());
+    const filteredExamples = placeholderExamples.filter((example) => example.match(regex));
+    fieldUIPlaceholder = filteredExamples.join(', ');
   }
 
-  if (schema && Array.isArray(schema.examples) && schema.examples.length > 0) {
-    return `ex. ${schema.examples.join(', ')}`;
-  }
+  return fieldUIPlaceholder !== '' ? fieldUIPlaceholder : undefined;
```
a-b-r-o-w-n, line 59: Can you show me an example schema that you were thinking about when coding this? I don't see how this fixes the problem yet.

LouisEugeneMSFT: The screenshot from the PR description has a good example. You can see there than the schema has a regex pattern for strings vs expressions and that filtering the examples with those patterns greatly help getting more refined examples.

a-b-r-o-w-n: Ah ok. So this uses the $ref's pattern to match against the "parent's" examples? That makes more sense.

PR #3835, opened 2020-08-12
feat: Generate skill manifest

### LINKED ISSUE #3380: Skill Manifest experience under par
The whole setup of the skill manifest editor feels like it is lacking behind the rest quite a bit. Below my thoughts on this:

**Export as skill**
Having it under the export dropdown and calling it export as skill sounds strange to me. You're actually not exporting anything, you're just creating a manifest which makes it a skill. So I would just expect it to be called something like "edit skill manifests"

**Naming of Manifests**
The naming of the manifests are fixed to {Botname}{schema version}.json
This makes no sense to me, why is the schema version part of the filename? Also having the botname in the filename seems not needed, as the url of the manifest will contain the botname, so it only gets unnecessarily complicated. To me the best way of doing this would be to use the actual version of the manifest itself like so: manifest-v1.0.json
image

**Editing Manifests**
It should be possible to manage multiple manifests to allow for different versions. Currently when creating a new one it just overwrites whatever is there..

### LINKED ISSUE #3756: The generated skill manifest should have at least a placeholder for the endpoints field

When we generate a skill manifest, it does not include all the required fields. We should include placeholders at a minimum for all the fields necessary to publish a valid skill.

tdurnford:
## Description
- New Steps:
  - Configure endpoints
    - [#3756] Previously, users had add and configure the manifest's endpoint property in the JSON Editor. This new steps uses the Adaptive Form package to create a schema driven form that allows users to easily modify endpoints and add new ones.
  - Select dialogs
    - In the select dialogs step, users select which dialogs they want to include in the skill manifest. The dialogs get add to the manifest as event activities in the `activities` property. If the dialog added has a dialog schema, the `value` and `resultValue` schema definitions are included in the manifest.
  - Select triggers
     - This step is very simple to the previous step - select triggers; however, instead of selecting which dialogs to include, the user selects which triggers from the root dialog they want to include. These triggers are also added to the skill manifest's `activities` property. 
  - Save manifest
    - The previous version of the Skill Manifest wizard generated a name for the manifest that user couldn't edit. This new steps now suggests a name for the manifest that user's can edit. It also allows users to save the manifest to a new file if they are editing an existing manifest.

## Task Item
Closes #3380
Closes #3756

### Configure endpoints step
image

### Select dialogs step
image

### Select triggers step
image

### Save manifest step
image

### DIFF
--- a/Composer/packages/client/package.json
+++ b/Composer/packages/client/package.json
@@ -26,10 +26,10 @@
     "@bfc/indexers": "*",
     "@bfc/shared": "*",
     "@bfc/ui-plugin-composer": "*",
+    "@bfc/ui-plugin-dialog-schema-editor": "*",
     "@bfc/ui-plugin-lg": "*",
     "@bfc/ui-plugin-luis": "*",
     "@bfc/ui-plugin-prompts": "*",
-    "@bfc/ui-plugin-dialog-schema-editor": "*",
     "@bfc/ui-plugin-select-dialog": "*",
     "@bfc/ui-plugin-select-skill-dialog": "*",
     "@emotion/core": "^10.0.27",
@@ -56,6 +56,7 @@
     "react-timeago": "^4.4.0",
     "recoil": "^0.0.10",
     "styled-components": "^4.1.3",
+    "uuid": "^8.3.0",
     "webpack-bundle-analyzer": "^3.8.0"
   },
   "browserslist": [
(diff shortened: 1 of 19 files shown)


cwhitten:
@tdurnford from the screenshots it looks like some columns aren't wrapped in the uischema. It shouldn't say "Microsoft.OnIntent", etc.

PR #3852, opened 2020-08-17
fix: refresh page client state lost error.

### LINKED ISSUE #3845: Reloading via ctrl-r on the Settings page of a bot with an ejected runtime throws an error
Screen Shot 2020-08-13 at 6 47 41 PM

zhixzhan:
## Description

Current if re-fresh page in home/setting page, all client state are lost. Notice `navigation` are disabled:
Untitled5

other page (design/lg/lu) which url include `:projectId` don't have this issue, cause when refresh page, `projectId` is still in url.

### Changes made to fix this
1. Cache projectId at client session storage.
2. Auto-fetch project if current no project is opened.

## Task Item
close #3845 

## Screenshots

Refresh in home/setting page wouldn't lost states

Untitled4

### DIFF
--- a/Composer/packages/client/src/pages/home/Home.tsx
+++ b/Composer/packages/client/src/pages/home/Home.tsx
@@ -3,7 +3,7 @@
 
 /** @jsx jsx */
 import { jsx } from '@emotion/core';
-import React, { useCallback, useEffect } from 'react';
+import React, { useCallback } from 'react';
 import formatMessage from 'format-message';
 import { Link } from 'office-ui-fabric-react/lib/Link';
 import { Icon } from 'office-ui-fabric-react/lib/Icon';
@@ -62,13 +62,9 @@ const Home: React.FC<RouteComponentProps> = () => {
   const recentProjects = useRecoilValue(recentProjectsState);
   const projectId = useRecoilValue(projectIdState);
   const templateId = useRecoilValue(templateIdState);
-  const {
-    openBotProject,
-    fetchRecentProjects,
-    setCreationFlowStatus,
-    onboardingAddCoachMarkRef,
-    saveTemplateId,
-  } = useRecoilValue(dispatcherState);
+  const { openBotProject, setCreationFlowStatus, onboardingAddCoachMarkRef, saveTemplateId } = useRecoilValue(
+    dispatcherState
+  );
 
   const onItemChosen = async (item) => {
     if (item && item.path) {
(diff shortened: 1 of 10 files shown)


Composer/packages/client/src/pages/setting/SettingsPage.tsx
```diff
@@ -50,6 +52,14 @@ const SettingPage: React.FC<RouteComponentProps<{ '*': string }>> = () => {
...
   const { navigate } = useLocation();
 
+  // when fresh page, projectId in store are empty, no project are opened at client
+  // use cached projectId do fetch.
+  const cachedProjectId = useProjectIdCache();
+  useEffect(() => {
+    if (!projectId && cachedProjectId) {
+      fetchProjectById(cachedProjectId);
+    }
+  }, [projectId]);
```
yeze322, line 62: since it's duplicated, we can move it inside the shared hook.

zhixzhan: I think it's not been used in a lot place or complex enough. maybe leave it here would be more flexible.

yeze322: If the fix only applies to the page refresh scenario, the deps of `useEffect` can become `[]`?

zhixzhan: I moved fetch project logic to root router, better than do this in components.

Composer/packages/client/src/pages/home/Home.tsx
```diff
@@ -136,10 +132,6 @@ const Home: React.FC<RouteComponentProps> = () => {
     },
   ];
 
-  useEffect(() => {
-    fetchRecentProjects();
```
zhixzhan, line 140: move fetchRecentProjects to creationFlow to centralize resource fetch in sequence.

Composer/packages/client/src/pages/setting/dialog-settings/DialogSettings.tsx
```diff
@@ -37,9 +36,6 @@ export const DialogSettings: React.FC<RouteComponentProps> = () => {
   const { setSettings, setLocale, addLanguageDialogBegin } = useRecoilValue(dispatcherState);
 
   const { languages, defaultLanguage } = settings;
-  useEffect(() => {
-    if (!botName) navigateTo('/home');
```
zhixzhan, line 41: No bot opened , navigate to Home. No needed anymore cause we also fetchProject in setting page.

boydc2014:
verified on my side

PR #3858, opened 2020-08-19
feat: use uischema to define Flow Actions

### LINKED ISSUE #2955: Customize the appearance of custom action nodes in Visual Editor
**Objective**:
After ejecting custom runtime instances with custom action $kinds inside, users should also be able to control those new nodes' appearances.

**Description**:
In R9, we already supported the custom schema story that users can use their custom sdk.schema file to write a bot. However, those $kinds of actions appear under 'Custom Action' label don't have good appearances in the Flow Editor.

Next step we should make the appearance of those custom action nodes be configurable instead of only showing a gray node.

**Plan**:
To achieve that goal, we should:
- [ ] Hoist Visual Editor's flow schema to bot project level under a proper folder (BOT_NAME\schemas?)
- [ ] Instead of verbose visual editor schema, provide a concise schema (5 lines, title - body - footer - icon - color)

yeze322:
## Description

closes #2955 

Convert Flow Editor's schema to pure JSON format. ([builtinSchema.ts](https://github.com/microsoft/BotFramework-Composer/blob/b3a115b7a8a5a9d17aae009379e3a4f0ba4c0268/Composer/packages/extensions/adaptive-flow/src/adaptive-flow-renderer/configs/builtinSchema.ts))

**Tasks**
- [x] Integrate with `adaptive-expressions`
- [x] convert flow schema from TSX to pure JSON format
- [x] inject Flow Actions schema from uischema

**Following tasks on other repos**
- [ ] Update the JSON schema of uischema.
- [ ] Move 'flow' uischema from 'ui-plugins/composer' to SDK component schema.

**Schema Preview**
  `IfCondition`
  image
  ```ts
  [SDKKinds.IfCondition]: {
    widget: 'IfConditionWidget',
    nowrap: true,
    judgement: {
      widget: 'ActionCard',
      body: '=coalesce(action.condition, "<condition>")',
    },
  }
  ```

  `BeginDialog`
  image
  ```
  [SDKKinds.BeginDialog]: {
    widget: 'ActionCard',
    body: {
      widget: 'DialogRef',
      dialog: '=action.dialog',
    },
    footer: {
      widget: 'PropertyDescription',
      property: '=action.resultProperty',
      description: '= Return value',
    },
    hideFooter: '=!action.resultProperty',
  }
  ```

  `SetProperties`
  image
  ```
  [SDKKinds.SetProperties]: {
    widget: 'ActionCard',
    body: {
      widget: 'ListOverview',
      items: '=foreach(action.assignments, x => concat(x.property, " : ", x.value))',
    },
  }
  ```
  `SetProperty` (string interpolation in LG format)
  image
  ```
  [SDKKinds.SetProperty]: {
    widget: 'ActionCard',
    body: '${coalesce(action.property, "?")} : ${coalesce(action.value, "?")}',
  },
  ```

## Screenshots

Use `sdk.override.uischema` to demo the scenario.

- Original EditArray node:
  image
- Define the overrided Flow schema in `sdk.override.uischema`:
```json
{
  "$schema": "https://schemas.botframework.com/schemas/ui/v1.0/ui.schema",
  "Microsoft.EditArray": {
    "flow": {
      "widget": "ActionCard",
      "header": {
        "widget": "ActionHeader",
        "title": "Overrided Edit Array",
        "icon": "Edit",
        "colors": {
          "theme": "#168686",
          "color": "white",
          "icon": "white"
        }
      },
      "body": "=concat('Change Type: ', action.changeType, '\nItems Property: ', action.itemsProperty)",
      "footer": "=string(action.value)",
      "hideFooter": "=!action.resultProperty"
    }
  }
}
```

- With the overrided Flow schema, EditArray will look like this:
  image


yeze322:
Taking the $kind `Microsoft.EditArray` as an example.

In Flow Editor, the EditArray action is describe as below:
image

We can schematize it stage by stage.

**Stage 1** - Regardless text color, text placeholder ("?"), EditArray's schema could be defined as below (body, footer are LG strings)
```yaml
Microsoft.EditArray:
  widget: ActionCard
  body: '${action.changeType} ${action.itemsProperty}'
  footer: '${action.resultProperty} = Result'
```

**Stage 2** - With text placeholder ("?") included, when `changeType` not set, show a "?" as placeholder
Solution:
- In JS, it can be written as `changeType || "?"`
- In LG, it can be written in two ways
  - Using `coalesce`, it's `coalesce(action.changeType, '?')`
  - Using `if`, it's `if(action.changeType, action.changeType, '?')`

Then EditArray's schema could be defined as
```yaml
Microsoft.EditArray:
  widget: ActionCard
  # when changeType, itemsProperty not set, shows '?'
  body: '${coalesce(action.changeType, '?')} ${coalesce(action.itemsProperty, '?')}'
  # when resultProperty not set, returns null
  footer: 'if(action.resultProperty, concant(action.resultProperty, ' = Result'), null)'
```

yeze322:
@cwhitten Chris I posted the example schema above

Composer/packages/extensions/adaptive-flow/package.json
```diff
@@ -28,6 +28,8 @@
     "@bfc/ui-shared": "*",
     "@emotion/core": "^10.0.27",
     "@emotion/styled": "^10.0.27",
+    "adaptive-expressions": "4.10.0-preview-147186",
+    "botbuilder-lg": "^4.10.0-preview-150886",
```
Danieladu, line 32: "botbuilder-lg" already contains the corresponding "adaptive-expressions" library

Composer/packages/extensions/adaptive-flow/src/adaptive-flow-renderer/utils/expression/widgetExpressionEvaluator.ts
```diff
@@ -0,0 +1,45 @@
...
+import { Templates } from 'botbuilder-lg';
+import { ExpressionParser } from 'adaptive-expressions';
+
+// Capture patterns that are strictly '${EXPRESSION}'
+//   - '${action.condition}'
+//   - '${if(action.value, concat(action.value, "(Event)"))}'
+const ValueAccessPattern = new RegExp(/^\$\{([^}]+)\}$/);
+
+// Catpure patterns that include '...${...}...'
+const StringTemplatePattern = new RegExp(/\$\{.+\}/);
```
Danieladu, line 13: Personally think it is not the best solution to achieve the template result.
Language generation is aimed to help people to generate the result based on the template and data, so, if there are some gaps between the requirements and the features provided by LG, please file an issue on SDK side, and the LG team would follow up.

Composer/packages/extensions/adaptive-flow/package.json
yeze322: Are expression utils exported from the botbuilder-lg pkg as well? I didn't find them

Composer/packages/extensions/adaptive-flow/src/adaptive-flow-renderer/utils/expression/widgetExpressionEvaluator.ts
yeze322: I understand it could (and indeed) be ambigous to mix LG and Expression together. The confusion comes from LG users always expects `${expression}` to return a string; while Expression users expects to write `expression` directly.

Actually inspired from our discussion last week, I do have a backup solution that using `=if(x, a, b)` to mark a field as an Expression. How do you think about that?

Composer/packages/extensions/adaptive-flow/package.json
yeze322: As a discussion result, we only use the `adaptive-expressions` package now. mark as resovled

Composer/packages/extensions/adaptive-flow/src/adaptive-flow-renderer/utils/expression/widgetExpressionEvaluator.ts
yeze322: After a offline discussion, we finally decided to use a pure expression plan, where `"=action.x"` is evaluated by the `ValueExpression` class.

yeze322:
Thanks Hongyang's review and inputs. Now we use a pure expression solution to build the flow schema

Composer/packages/ui-plugins/composer/src/defaultFlowSchema.ts
```diff
@@ -0,0 +1,195 @@
...
+  [SDKKinds.DeleteActivity]: {
+    widget: 'ActionCard',
+    header: {
+      widget: 'ActionHeader',
+      title: 'Delete activity',
+    },
+    body: {
+      widget: 'PropertyDescription',
+      property: '=coalesce(action.activityId, "?")',
+      description: '=string("= ActivityId")',
```
a-b-r-o-w-n, line 118: Is this really necessary? Couldn't this just be a string?

yeze322: It begins with '=', causes the expression engine will recognize it as a valid expression and returns `undefined` while `= Result property` won't.

To escapt this case, either using `=concat("= ActivityId")` or using `=string()`

yeze322:
> We should try to allow string interpolation as well.
> 
> Instead of `'=coalesce(action.changeType, "?")'` we could use `'${action.changeType}?'`.

@a-b-r-o-w-n That's definitely what I've tried to do. What about this one

Composer/packages/ui-plugins/composer/src/defaultFlowSchema.ts
a-b-r-o-w-n: So this is a special case where we want the string to start with a `=` but that is seen as an expression? Can we escape it? `"\= ActivityId"`? Or attempt to parse as an expression then fallback to this string if that fails?

Having to write an expression for this isn't very nice.

a-b-r-o-w-n:
@yeze322 It's weird that the string interpolation comes from the lg library, not the expression one. Is that the case with the dotnet libraries?

yeze322:
> @yeze322 It's weird that the string interpolation comes from the lg library, not the expression one. Is that the case with the dotnet libraries?

Yeah, it is. LG receives a format like `Hello, ${user.name}` while Expression receives `concat("a", "b")`. @Danieladu Hongyang is that correct?

Danieladu:
> > @yeze322 It's weird that the string interpolation comes from the lg library, not the expression one. Is that the case with the dotnet libraries?
> 
> Yeah, it is. LG receives a format like `Hello, ${user.name}` while Expression receives `concat("a", "b")`. @Danieladu Hongyang is that correct?

Yes. `${Welcome}, ${user.name}` in LG is same with `concat(Welcome, ',', user.name)` in Expression, and also same with ``` `${Welcome},${user.name}` ``` in Expression with string interpolation. Except for some different evaluation result of `null`

yeze322:
@a-b-r-o-w-n We have multiple approaches to write a string interpolation
```
// 1. LG
'Hello ${user.name}.'

// 2. Expression
'=concat("Hello ", user.name)'

// 3. Expression - with string interpolation
'=`Hello ${user.name}`'
```
In Expression, string interpolation needs to be wrapped in ``.
```
// Can be evaluated by Expression -> Expression class
'`Hello ${user.name}`'

// Can be evaluated by Expression -> ValueExpression class
'=`Hello ${user.name}`'

// Can be evaluated by LG
'Hello ${user.name}'
```

Composer/packages/ui-plugins/composer/src/defaultFlowSchema.ts
yeze322: `"\= ActivityId"` won't be recognized as expression but will be displayed as it is with a prefix `\`, so seems we can't write it that way.
Another tricky point here is, `= Activity` can be evaluated successfully by Exp while `= Activity property` throws an error and can fallback, so as a result, we don't need to wrap  `"= Result property"` but we have to wrap `"= ActivityId"`.

I think the `"="` problem is the cost of introducing Expression engine. There will always be an indicater to mark a plain string as Expression, either the `"="` or other characters. Under current expression engine's capability, the most concise way to write that is "=\`= Activity\`"

yeze322: Another possible solution here is we detect the expression pattern with Regex, only when an expression contains `action.xxx` pattern do we evaluate it with Expression.
The assumption is, in the Flow schema, an meaningful expression have to read action data by writing `=...action.xxx`  .

```ts
if (exp.beginWith("=") && exp.indexOf("action.") > -1) {
  return Expression.evaluate(exp, context);
}
```

yeze322: resolved, '=string()' no longer needed

a-b-r-o-w-n:
@yeze322 I think matching the sdk properties as closely as possible would be best. So to me that looks like we should use the LG parser for string interpolation.

Composer/packages/ui-plugins/composer/src/defaultFlowSchema.ts
```diff
@@ -0,0 +1,195 @@
+// Copyright (c) Microsoft Corporation.
```
a-b-r-o-w-n, line 1: Will we want to move this over to the sdk component schemas in this release?

yeze322: yes we do, plan to do that in another PR

yeze322:
> @yeze322 I think matching the sdk properties as closely as possible would be best. So to me that looks like we should use the LG parser for string interpolation.

I updated the schema, the Action ‘SetProperty' is an example on how string interpolation be written LG format.

One thing to note is writting string interpolation in LG format can simplify the schema in some case but not all the cases. Expression engine is still required regarding sometimes we want to take the raw value instead of getting a string. (For example, in BeginDialog, `dialog` field could be a string or a IDialog object. We have to write `=action.dialog` instead of `${action.dialog}` because of the latter one will return a string. This is LG parser's default behavior.)

a-b-r-o-w-n:
That sounds reasonable and is the same as when authoring bots, so hopefully the learning curve will be small.

I think we should consider adding this functionality to the form properties in the ui schema as well.

a-b-r-o-w-n:
@yeze322 sorry for all the conflicts. I removed the extensions directory 😕

yeze322:
@a-b-r-o-w-n that's fine Andy, I just need to do some copy/paste

cwhitten (approved):

PR #3864, opened 2020-08-19
feat: Design tweaks to Skill Manifest Generation Wizard

### LINKED ISSUE #3333: improve skill manifest editing experience

 I'm always frustrated when skill manifest editing experience needs improvements.

* This section should be included in the default template for being used as a skill (some automation of msAppId would be better)
```
"endpoints": [
      {
        "name": "production",
        "protocol": "BotFrameworkV3",
        "description": "Production endpoint",
        "endpointUrl": "https://YOUR_DEPLOYED.azurewebsites.net/api/messages",
        "msAppId": "YOUR_ID"
      }
    ]
```
* Provide a place for displaying and copying the full url for manifest like https://YOUR_DEPLOYED.azurewebsites.net/manifests/YOUR_BOT-0-2-1-preview-1-manifest.json. The url for manifest is not obvious now and Composer can't display the name fully:
image
* Provide a bigger place for manifest editing. Here are some almost real world manifest examples: https://github.com/microsoft/botframework-skills/blob/master/skills/csharp/todoskill/wwwroot/manifest/manifest-1.1.json, https://github.com/microsoft/botframework-skills/blob/master/skills/csharp/calendarskill/wwwroot/manifest/manifest-1.1.json
They are easy to be over several hundreds lines long.

tdurnford:
## Description
- Endpoints Step
  - Hide the `protocol` property since it can only be set to 'BotFrameworkv3' at the moment
    - There was a bug in the Adaptive Form package that didn't set the default/const for hidden properties. To resolve this, we shifted the responsibility for hiding fields from the `ObjectField` to the `SchemaField` where the default/const values are set.
- Dialog Selection Step
  - Added a description column with an editable field to change the dialog description
  - Added a 'Select all' checkbox and hid the select all checkbox in the table
- Select Trigger Step
  - Added a 'Select all' checkbox and hid the select all checkbox in the table

## Task Item
closes #3333

### Endpoints Step
image
### Select Dialogs
image

### Select Triggers Step
image

### DIFF
--- a/Composer/packages/extensions/adaptive-form/src/utils/resolveRef.ts
+++ b/Composer/packages/extensions/adaptive-form/src/utils/resolveRef.ts
@@ -8,6 +8,10 @@ export function resolveRef(
   definitions: { [key: string]: JSONSchema7Definition } = {}
 ): JSONSchema7 {
   if (typeof schema?.$ref === 'string') {
+    if (!schema?.$ref?.startsWith('#/definitions/')) {
+      return schema;
+    }
+
     const defName = schema.$ref.replace('#/definitions/', '');
     const defSchema = definitions?.[defName] as JSONSchema7;
 
--- a/Composer/packages/client/src/pages/design/exportSkillModal/__tests__/generateSkillManifest.test.ts
+++ b/Composer/packages/client/src/pages/design/exportSkillModal/__tests__/generateSkillManifest.test.ts
@@ -79,7 +79,7 @@ describe('generateOtherActivities', () => {
 describe('generateActivity', () => {
   it('should return simple activity when there is no dialog schema', () => {
     const dialogSchemas = [];
-    const dialog: any = { id: 'test' };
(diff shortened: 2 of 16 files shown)


Composer/packages/extensions/adaptive-form/src/utils/resolveRef.ts
```diff
@@ -8,6 +8,10 @@ export function resolveRef(
   definitions: { [key: string]: JSONSchema7Definition } = {}
 ): JSONSchema7 {
   if (typeof schema?.$ref === 'string') {
+    if (!schema?.$ref?.startsWith('#/definitions/')) {
```
cwhitten, line 11: why was this needed?

tdurnford: Some of the refs in the Dialog Schema refer to 'http://json-schema.org/draft-07/schema#', so it throws the function throws an error since 'http://json-schema.org/draft-07/schema#' is not included in the `definitions`.

Issue #3022, opened 2020-05-13
state.designPageLocation is not persistence with state


zhixzhan:
some value in `state.designPageLocation` e.g. `projectId` `dialogId` not always sync with it's initial value in `state`

zhixzhan:
Since we migrate to Recoil, this issue was no longer valid

lei9444:
I think we can fix it in R11

PR #3871, opened 2020-08-20
fix: open a new bot will navigate to the old bot

### LINKED ISSUE #3865: Issues in samples
## Describe the bug

* **Empty Bot**
Every time after you create an _Empty Bot_ first, then creating the following three samples will load the _Empty Bot_ template: _ToDoWithLuis_,  _SimpleToDo_ and _Interruptions_ samples. 

* **ToDoWithLuis, QnAMakerAndLuis**
Warning message.

* **AskingQuestionsSample**
Picture2

## Version

The [main](https://github.com/microsoft/BotFramework-Composer) branch. 

## Browser

- [x] Chrome

## OS

- [x] Windows

## To Reproduce

1. Check out the main branch.
2. Build Composer.
3. Open samples. 

## Additional context
The rest of the samples worked as expected.

lei9444:
## Description

The reason is the design page will use the location state to navigate to the begin dialog action. The location will be updated by the @reach\router, then we use the location to update the location in our state management. So the first open will use the old location to navigate and get the old project id.

In this PR, check the project to avoid the page navigate to the old bot. 

**I think in R11 we need to make the router and state one way data flow.** #3022

## Task Item
closes #3865

### DIFF
--- a/Composer/packages/client/src/pages/design/DesignPage.tsx
+++ b/Composer/packages/client/src/pages/design/DesignPage.tsx
@@ -172,7 +172,7 @@ const DesignPage: React.FC<RouteComponentProps<{ dialogId: string; projectId: st
 
   useEffect(() => {
     const index = currentDialog.triggers.findIndex(({ type }) => type === SDKKinds.OnBeginDialog);
-    if (index >= 0 && !designPageLocation.selected) {
+    if (index >= 0 && !designPageLocation.selected && designPageLocation.projectId === projectId) {
       selectTo(createSelectedPath(index));
     }
   }, [currentDialog?.id]);


Composer/packages/client/src/pages/design/DesignPage.tsx
```diff
@@ -172,7 +172,7 @@ const DesignPage: React.FC<RouteComponentProps<{ dialogId: string; projectId: st
 
   useEffect(() => {
     const index = currentDialog.triggers.findIndex(({ type }) => type === SDKKinds.OnBeginDialog);
-    if (index >= 0 && !designPageLocation.selected) {
+    if (index >= 0 && !designPageLocation.selected && designPageLocation.projectId === projectId) {
```
zhixzhan, line 175: To keep `projectId` `dialogId` always consistence, I think we should remove them from `designPageLocation` state. let components get from shell state.

zhixzhan: Previous related issue https://github.com/microsoft/BotFramework-Composer/issues/3022

lei9444: yes, These may take some time, We can do it in R11

boydc2014 (approved):

PR #3887, opened 2020-08-24
feat: Plumbs UpdateActivity and DeleteActivity into the app

### LINKED ISSUE #3874: Update Activity and Delete Activity support
Are there plans to support the [Update Activity and Delete Activity](https://docs.microsoft.com/en-us/azure/bot-service/adaptive-dialog/adaptive-dialog-prebuilt-actions?view=azure-bot-service-4.0#activities) actions? It would be great to be able to e.g. update an adaptive card after a user performed an action.

cwhitten:
closes #3874

note: it still isn't clear where UI is getting component information (label, title, description, etc)

### DIFF
--- a/Composer/packages/server/schemas/sdk.uischema
+++ b/Composer/packages/server/schemas/sdk.uischema
@@ -145,6 +145,18 @@
       "subtitle": "Delete Property"
     }
   },
+  "Microsoft.DeleteActivity": {
+    "form": {
+      "label": "Delete activity"
+    }
+  },
+  "Microsoft.DeleteProperty": {
+    "form": {
+      "helpLink": "https://aka.ms/bfc-using-memory",
+      "label": "Delete a property",
+      "subtitle": "Delete Property"
+    }
+  },
   "Microsoft.EditActions": {
     "form": {
       "label": "Modify active dialog",
--- a/Composer/packages/server/schemas/sdk.uischema
+++ b/Composer/packages/server/schemas/sdk.uischema
@@ -130,7 +130,6 @@
       "label": "Debug Break"
     }
   },
-  "Microsoft.DeleteActivity": {},
   "Microsoft.DeleteProperties": {
     "form": {
       "helpLink": "https://aka.ms/bfc-using-memory",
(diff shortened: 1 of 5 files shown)


Composer/packages/server/schemas/sdk.uischema
```diff
@@ -145,6 +145,18 @@
...
+      "label": "Delete activity"
+    }
+  },
+  "Microsoft.DeleteProperty": {
+    "form": {
+      "helpLink": "https://aka.ms/bfc-using-memory",
+      "label": "Delete a property",
+      "subtitle": "Delete Property"
+    }
+  },
```
a-b-r-o-w-n, line 159: This is duplicated.

Composer/packages/server/schemas/sdk.uischema
```diff
@@ -145,6 +145,18 @@
       "subtitle": "Delete Property"
     }
   },
+  "Microsoft.DeleteActivity": {
+    "form": {
+      "label": "Delete activity"
+    }
+  },
```
a-b-r-o-w-n, line 152: Let's make sure this gets added to the SDK component schemas.

Composer/packages/ui-plugins/composer/src/defaultMenuSchema.ts
```diff
@@ -163,4 +163,12 @@ export const DefaultMenuSchema: MenuUISchema = {
     label: formatMessage('Emit a telemetry track event'),
     submenu: [SubmenuNames.Debugging],
   },
+  [SDKKinds.UpdateActivity]: {
+    label: formatMessage('Update activity'),
+    submenu: [SubmenuNames.PropertyManagement],
```
a-b-r-o-w-n, line 168: Is this in the right submenu?

cwhitten: There wasn't an exiting menu that felt great. Any ideas?

Composer/packages/server/schemas/sdk.uischema
cwhitten: https://github.com/microsoft/botbuilder-dotnet/issues/4514

yeze322:
@cwhitten we are going to retire the `ConceptLabels` under 'labelMaps.ts' after the UI Schema's Trigger part is finished. Then all kinds of component information could be aggregated together and be much clearer.

sschutten:
How should we use these new actions? We need to get the activity ID from an earlier activity, but the suggested dialog.lastActivityId doesn't exist.

PR #3914, opened 2020-08-27
fix: Fix dialog redirect/refresh behavior

### LINKED ISSUE #3770: always redirected to the first trigger when loading a bot
## Describe the bug

When navigating directly to a trigger (something like `/bot/:id/dialogs/todobotwithluissample-0?selected=triggers[3]`), the page always redirects to the first trigger (`selected=triggers[0]`).

## Version

84b305d6c

## Browser

- [ ] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [x] Edge

## OS

- [x] macOS
- [ ] Windows
- [ ] Ubuntu

## To Reproduce

1. Open a bot
2. Navigate to any trigger, not the first
3. Refresh the page

## Expected behavior

The page should not redirect.

### LINKED ISSUE #3911: [shell] Breadcrumb is incorrect when the user creates a new dialog
## Describe the bug
Currently when the user creates a new dialog, the design page navigates to the 'OnBeginDialog' trigger, and the breadcrumb is just set to 'BeginDialog.' However, it should be 'DialogName > BeginDialog.'

## Version
Latest build from Master

## To Reproduce

1. Add a new dialog
4. See error in breadcrumb

## Expected behavior
The Breadcrumb should be  'DialogName > BeginDialog'

## Screenshots
image

tdurnford:
## Description
  Currently, the `<DesignPage />` component uses the `useEffect` hook to select the `OnBeginDialog` trigger when the user selects a dialog in the Project Tree. There are two issues with this approach. First, when the user refreshes the page or modifies the url, the page returns to the `OnBeginDialog` trigger. Secondly, the page navigates to the `AdaptiveDialog` editor first and then redirects to `OnBeginDialog` trigger. This causes the screen to flash between the two editors instead of just rendering the `OnBeginDialog` trigger.
 
To resolve this, we shifted the responsibility to select the `OnBeginDialog` trigger from the `<DesignPage />` to the `navTo` dispatcher, which is called when the user selects a different dialog from the Project Tree. If the selected dialog has an `OnBeginDialog` trigger, the dispatcher appends the selection to the url before navigating to the appropriate editor. This allows the user to freely refresh the browser and modify the url without navigating away from their intended editor. It also doesn't cause the design page to switch between the `AdaptiveDialog` and the `OnBeginDialog` editors. 

## Task Item
Closes #3770
Closes #3911


Composer/packages/client/src/recoilModel/dispatchers/__tests__/navigation.test.tsx
```diff
@@ -3,13 +3,27 @@
 
 import { useRecoilValue } from 'recoil';
 import { act } from '@bfc/test-utils/lib/hooks';
+import { SDKKinds } from '@bfc/shared/src/types';
```
a-b-r-o-w-n, line 6:
```suggestion
import { SDKKinds } from '@bfc/shared';
```

Composer/packages/client/src/recoilModel/dispatchers/navigation.ts
```diff
@@ -46,7 +59,21 @@ export const navigationDispatcher = () => {
...
       const designPageLocation = await snapshot.getPromise(designPageLocationState);
-      const currentUri = `/bot/${projectId}/dialogs/${dialogId}`;
+      const dialogs = await snapshot.getPromise(dialogsState);
+      const currentDialog = dialogs.find(({ id }) => id === dialogId);
+
+      let path;
+      if (dialogId !== designPageLocation.dialogId) {
+        // Redirect to Microsoft.OnBeginDialog trigger if it exists on the dialog
+        const beginDialogIndex = currentDialog?.triggers.findIndex(({ type }) => type === SDKKinds.OnBeginDialog);
+        if (typeof beginDialogIndex !== 'undefined') {
```
a-b-r-o-w-n, line 69: Is it ok if `beginDialogIndex` is `-1`?

tdurnford: Good catch. It should be.
```suggestion
        if (typeof beginDialogIndex !== 'undefined' && beginDialogIndex >= 0) {
```

a-b-r-o-w-n: I think just checking `>= 0` is enough.

PR #3932, opened 2020-08-28
fix qna bug

liweitian:
## Description
urls with one more trailing / should be duplicated.
http://abc/  is duplicated with http://abc

When creating a qna bot, we can see the create from scratch button. But when we only want to import qna file, we should hide the button.

## Screenshots
duplicatedUrl
hidden button


liweitian:
already fixed through #3927

PR #3938, opened 2020-08-28
fix: Fixed Design Page redirecting when an action is deleted

tdurnford:
## Description
Currently, the `saveData` function in the `useShell` hook checks if the focus path exists in the updated data. If the focused value is `undefined`, it navigates the design page to the Adaptive Dialog editor. However, the `saveData` method uses the `selected` and `focused` values in the `designPageLocation` value, which are not updated until the next render,  to retrieve the value. Consequently, when the `saveData` function is called, the `selected` and `focused` have not been updated and the page incorrectly redirects the page.

To resolve this, we moved the logic to check if the value is `undefined` from the `saveData` function to the `<DesignPage />` component. 

## Task Item
Closes #3929
Closes #3939

PR #3941, opened 2020-08-31
fix: change atomic value 'breadcrumb' correctly

yeze322:


yeze322:
Duplicated with changes in TJ's on-going PR #3938. Mark as closed.
Based on TJ's PR, the fix is at https://github.com/tdurnford/BotFramework-Composer/pull/1 .

PR #3976, opened 2020-09-01
fix: split lubuild and luis publish

### LINKED ISSUE #3969: Got an authentication error while the luis.authoringKey is not configured in the publish profile
I didn't use luis in my bot, but while publishing I got such error.

401 - "{ \"statusCode\": 401, \"message\": \"Unauthorized. Access token is missing, invalid, audience is incorrect (https://cognitiveservices.azure.com), or have expired.\" }"

The error is gone after i configure the luis.authoringKey

VanyLaw:
## Description
**Cause**: do luis build and luis publish even not luis bot in azurePublish plugin.
**Solution**: split lubuild and luis publish. Publishing luis when lu file not empty.

## Task Item

close #3969

## Screenshots

EchoBot can deploy without authoringKey.
image

### DIFF
--- a/Composer/packages/client/src/components/ErrorPopup/ErrorPopup.tsx
+++ b/Composer/packages/client/src/components/ErrorPopup/ErrorPopup.tsx
@@ -30,7 +30,9 @@ const formatErrorDetail = (error: StateError): React.ReactElement => {
   return (
     <section>
       {helpText}
-      <a href={'https://github.com/microsoft/BotFramework-Composer/issues'}>GitHub</a>
+      <a href={'https://github.com/microsoft/BotFramework-Composer/issues'} rel="noopener noreferrer" target="_blank">
+        GitHub
+      </a>
       {message && <details>{message}</details>}
     </section>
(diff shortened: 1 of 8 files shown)


luhan2017:
I've tested this and found 
- it is working when no content in lu files and no luis.authoringKey configured.
- there is an error when there is content in lu files and luis.authortingKey and endpointKey configured.
image

I will work with Wenyi tomorrow to figure this out.

luhan2017:
There is a new issue that, then both luis and qna contents exist in the bot, sometimes the qna can be successfully published but sometime not,  we have investigated and found the root cause is:

It is not recommended to use async and await in the forEach loop. 
 crossTrainResult.forEach(async (value, key) => {
      const fileName = path.basename(key);
      const newFileId = path.join(this.interruptionFolderPath, fileName);
      await fs.writeFile(newFileId, value.Content);
    });

we will update on both localPublish and azurePublish side on this.

luhan2017 (approved):

PR #3979, opened 2020-09-02
fix: merge settings from generated files in js runtime

chon219:
## Description

Merge settings from generated settings files (e.g. QnA, Luis settings) in js runtime.

## Task Item

#minor


VanyLaw:
close due to security check failed.

PR #4033, opened 2020-09-04
fix: update bl for security reasons

beyackle:
## Description

DevOps tests are failing because the NPM package "bl" has security issues, and Dependabot can't see all of them. This should fix those.

## Task Item

#minor


a-b-r-o-w-n (approved):
Thanks for doing this!

PR #4043, opened 2020-09-05
feat: allows to define a custom static port with custom runtime command

### LINKED ISSUE #4044: Define a static port for your bot/skill

To setup a dev env to test bot skills with skill consumers locally. It is very useful to be able to define a static local port for a bot.

I want to be able to set a static port in in the custom runtime startup command.

Always checking/changing the skill endpoints before starting the skill consumer -> cumbersome

ltwlf:
## Description

To setup a dev env to test bot skills with skill consumers locally. It is very useful to be able to define a static local port for a bot. 
This PR allows to configure a custom port with -port 5000 in the custom runtime command. Without this PR the port gets always overwritten by the dynamic port finder.    

e.g. 
Start command*
dotnet run --port 5001 --project azurewebapp

## Issue
fixes #4044


a-b-r-o-w-n:
@ltwlf thanks for the PR! I think this may be more useful to expose as a separate configuration option. @benbrown what do you think?

benbrown:
I like the proposed solution.

I'm not sure exposing another "official" setting makes sense since those settings are getting written to the appsettings file and would then potentially affect non-local environments.

a-b-r-o-w-n (approved):

PR #4050, opened 2020-09-06
fix: Fixes various issue in the web app deployment script

ltwlf:
## Description
The azure webapp deployment script has various issues e.g. bf luis:build has wrong --dialog argument.
appsettings.deployment.json will never be used by the published bot and exposes local secrets to the remote folder.

## Task Item
fixes #4049

PR #4053, opened 2020-09-07
fix: remove deprecated script in runtime template

VanyLaw:
## Description

remove deprecated scripts in dotnet runtime template

## Task Item
 
#minor


ltwlf:
I would not suggest to delete this scrip and rather improve or rewrite it. The current approach to deploy via UI is not sufficient for professionals. Additionally from IT security it is really bad. Customers need a build script for azure devops pipeline and must be able to define the secrets securely in Azure key vault.   
I use/fix this script and going to use it for deployment / release via Azure pipelines for an enterprise customer. I can't imagine any enterprise customer using the deploy via UI publishing.

ltwlf:
This PR fixes the azurewebapp deploy script #4050 and makes it ready for enterprise deployment. The other scripts are not needed, only deploy.ps1.

VanyLaw:
@ltwlf Thanks for the reminder. Because getting keys from "Azure key vault" is still in out todo list and have not been implemented yet. Considering the security, we should keep the `deploy.ps1`. because `create.ps1` was replaced by `provisionComposer.js` script, so we will remove the rest deprecated scripts and take you fix.

ltwlf:
@VanyLaw sounds good to me. I'm preparing a PR for optimizing the script for CI/CD deployment and a short readme. I think the readme is also outdated because running the bot directly (not directly via composer ui) will not generate/copy the dialogs.

boydc2014:
@VanyLaw can we just close this one?

PR #4065, opened 2020-09-08
feat: Refer to settings.skill inside dialog files

### LINKED ISSUE #4032: Reference skills object in dialog
We will represent the skills as an object keyed by the skill name or the custom name given by the user. This ticket handles backward compat by converting older bots skills array to a dictionary. The work will assist the BotProject work to start skills and root bot in a single composer instance. The properties in .dialog files referring to skills would be an expression leveraging the settings rather than a hardcoded value. Ex

```
Inside .dialog file's BeginSkill action
{
  "id": ""https://xyz/OneNoteSync-2-1-preview-1-manifest.json"",
  "skillEndpoint": "https://xyz/api/messages",
  "skillAppId": "123-234-345-34543"
}

changes to
{
  "id": "=settings.skill['GoogleKeepSync'].manifestUrl",
  "skillEndpoint": "=settings.skill['GoogleKeepSync'].endpointUrl",
  "skillAppId": "=settings.skill['GoogleKeepSync'].appId"
}
```

srinaath:
## Description

This PR transforms settings.skill into a dictionary. We now let the BeginDialog action's id, skillAppId and skillEndpoint to refer to skills object inside settings. This makes sure that we have an expression instead of hardcoded values inside BeginDialog action

Fixes #4032


Composer/packages/lib/indexers/src/botIndexer.ts
```diff
@@ -38,7 +47,11 @@ const checkSkillSetting = (assets: BotAssets): Diagnostic[] => {
   dialogs.forEach((dialog) => {
     // used skill not existed in setting
     dialog.skills.forEach((skillId) => {
-      if (skill.findIndex(({ manifestUrl }) => manifestUrl === skillId) === -1) {
+      const manifestUrlCollection = map(skill, ({ manifestUrl }) => manifestUrl);
+      if (
+        manifestUrlCollection.findIndex((manifestUrl) => manifestUrl === fetchFromSettings(skillId, assets.setting)) ===
```
beyackle, line 52: Use the built-in `.includes()` method instead of checking if an index is -1.

beyackle (dismissed):

Composer/packages/lib/shared/src/skillsUtils/index.ts
```diff
@@ -0,0 +1,29 @@
...
+  const mappedSkills = skills.map(({ msAppId, endpointUrl, manifestUrl, name }: Skill) => {
+    return {
+      name,
+      msAppId,
+      endpointUrl,
+      manifestUrl,
+    };
+  });
+
+  const mapped = keyBy(mappedSkills, 'name');
```
tdurnford, line 27: Single use variable.
```suggestion
  return keyBy(mappedSkills, 'name');
```

Composer/packages/lib/shared/src/types/settings.ts
```diff
@@ -33,9 +33,13 @@ export interface DialogSetting {
   defaultLanguage: string;
   languages: string[];
   skill?: {
-    name: string;
-    manifestUrl: string;
-  }[];
+    [skillName: string]: {
+      name?: string;
```
tdurnford, line 37: Should name be required?

Composer/packages/ui-plugins/select-skill-dialog/src/BeginSkillDialogField.tsx
```diff
@@ -6,38 +6,88 @@ import { FieldProps, JSONSchema7, useShellApi } from '@bfc/extension';
...
+  const endpointOptions = useMemo(() => {
+    return (manifest?.endpoints || []).map(({ name }) => name);
+  }, [manifest]);
 
-  const handleIdChange = ({ key }) => {
+  const handleIdChange = ({ key, text }) => {
     if (!manifest || key !== manifest.manifestUrl) {
       const { skillEndpoint, skillAppId, ...rest } = value;
-      onChange({ ...rest, id: key });
+      onChange({ ...rest, id: referBySettings(text, 'manifestUrl') });
```
tdurnford, line 65: Previously, we were only using the `id` field to correctly populate the 'Skill endpoint' dropdown; however, since the 'Skill endpoint' is mostly static, I think we can move away from setting the `id` field and populate the dropdown options based on the `skillName` in `=settings.skill['${skillName}'].skillEndpoint`.

Composer/packages/ui-plugins/select-skill-dialog/src/BeginSkillDialogField.tsx
```diff
@@ -46,6 +96,12 @@ export const BeginSkillDialogField: React.FC<FieldProps> = (props) => {
...
 
+  const skillIdUiSchema = uiOptions.properties?.id || {};
+  skillIdUiSchema.serializer = {
+    get: (value) => skillsInSettings.get(value),
+    set: (value) => value,
+  };
+
   return (
     <React.Fragment>
       <SchemaField
```
tdurnford, line 107: I think we should replace this `SchemaField` with a combobox that doesn't set the `id` in the dialog file.

Composer/packages/ui-plugins/select-skill-dialog/src/SkillEndpointField.tsx
```diff
@@ -26,8 +26,10 @@ export const SkillEndpointField: React.FC<FieldProps> = (props) => {
   const deserializedValue = typeof uiOptions?.serializer?.get === 'function' ? uiOptions.serializer.get(value) : value;
 
   const handleChange = (newValue: any) => {
-    const serializedValue =
-      typeof uiOptions?.serializer?.set === 'function' ? uiOptions.serializer.set(newValue) : newValue;
+    const serializedValue = newValue;
+    if (typeof uiOptions?.serializer?.set === 'function') {
+      uiOptions.serializer.set(newValue);
```
tdurnford, line 31: I'm confused about what the intent is here. `uiOptions.serializer.set` returns a value; however, that value isn't being used. Do we still need the serializers?

Composer/packages/client/src/recoilModel/dispatchers/setting.ts
```diff
@@ -89,12 +89,32 @@ export const settingsDispatcher = () => {
     }
   );
 
+  const updateSkillsInSetting = useRecoilCallback(
+    ({ set, snapshot }: CallbackInterface) => async (skillName: string, skillInfo: Partial<Skill>) => {
+      const currentSettings: DialogSetting = await snapshot.getPromise(settingsState);
+      const matchedSkill = get(currentSettings, `skill[${skillName}]`, {});
+      if (matchedSkill) {
```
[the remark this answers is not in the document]
srinaath: _Ohh. nice catch. I put the default value as empty object as an oversight. Setting it to undefined_

Composer/packages/lib/shared/src/skillsUtils/index.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export const convertSkillsToDictionary = (skills: Skill[]) => {
+  const mappedSkills = skills.map(({ msAppId, endpointUrl, manifestUrl, name }: Skill) => {
+    return {
+      name,
+      msAppId,
+      endpointUrl,
+      manifestUrl,
+    };
+  });
```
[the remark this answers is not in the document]
srinaath: THe skill has other properties. As a result i have to map and then keyBy

Composer/packages/ui-plugins/select-skill-dialog/src/SkillEndpointField.tsx
srinaath: `=`setting.skill\['skillName'\].endpointUrl\` is in the schema. It gets transformed to the endpointName through get fn

srinaath: `=`setting.skill\['skillName'\].endpointUrl\` is in the schema. It gets transformed to the endpointName through get fn

Composer/packages/ui-plugins/select-skill-dialog/src/BeginSkillDialogField.tsx
srinaath: `Id` the non mandatory field is no longer set in the schema

Composer/packages/client/src/recoilModel/dispatchers/setting.ts
```diff
@@ -89,12 +89,32 @@ export const settingsDispatcher = () => {
     }
   );
 
+  const updateSkillsInSetting = useRecoilCallback(
+    ({ set, snapshot }: CallbackInterface) => async (skillName: string, skillInfo: Partial<Skill>) => {
+      const currentSettings: DialogSetting = await snapshot.getPromise(settingsState);
+      const matchedSkill = get(currentSettings, `skill[${skillName}]`, undefined);
```
tdurnford, line 95: Isn't `get`'s third parameter `undefined` by default?
```suggestion
      const matchedSkill = get(currentSettings, `skill[${skillName}]`);
```

Composer/packages/lib/indexers/src/botIndexer.ts
```diff
@@ -38,7 +47,8 @@ const checkSkillSetting = (assets: BotAssets): Diagnostic[] => {
   dialogs.forEach((dialog) => {
     // used skill not existed in setting
     dialog.skills.forEach((skillId) => {
-      if (skill.findIndex(({ manifestUrl }) => manifestUrl === skillId) === -1) {
+      const endpointUrlCollection = map(skill, ({ endpointUrl }) => endpointUrl);
+      if (!endpointUrlCollection.includes(fetchFromSettings(skillId, assets.setting))) {
         diagnostics.push(
           new Diagnostic(`skill '${skillId}' is not existed in appsettings.json`, dialog.id, DiagnosticSeverity.Error)
```
tdurnford, line 53: The grammar on this doesn't make sense. Now might be a good time to fix it while were making changes.
```suggestion
          new Diagnostic(`skill '${skillId}' does not exist in appsettings.json`, dialog.id, DiagnosticSeverity.Error)
```

Composer/packages/lib/indexers/src/dialogIndexer.ts
```diff
@@ -166,8 +166,8 @@ function extractReferredSkills(dialog): string[] {
   const visitor: VisitorFunc = (path: string, value: any): boolean => {
     // it's a valid schema dialog node.
     if (has(value, '$kind') && value.$kind === SDKKinds.BeginSkill) {
-      const skillId = value.id;
-      skills.push(skillId);
+      const skillEndpoint = value.skillEndpoint;
+      skills.push(skillEndpoint);
```
tdurnford, line 170:
```suggestion
      skills.push(value.skillEndpoint);
```

Composer/packages/lib/shared/src/types/settings.ts
```diff
@@ -33,9 +33,13 @@ export interface DialogSetting {
   defaultLanguage: string;
   languages: string[];
   skill?: {
-    name: string;
-    manifestUrl: string;
-  }[];
+    [skillName: string]: {
+      name: string;
+      manifestUrl: string;
+      msAppId?: string;
```
tdurnford, line 39: I think `msAppId` and `endpointUrl` should both be optional or both required since they go hand-in-hand.

Composer/packages/ui-plugins/select-skill-dialog/src/BeginSkillDialogField.tsx
```diff
@@ -1,66 +1,132 @@
...
+        if (result.endpointName) {
+          loadEndpointForOldBots(result.endpointName);
+        }
+      }
     }
-  };
+  }, []);
+
+  const matchedSkill: Skill | undefined = useMemo(() => {
+    const skill = skills.find(({ name }) => name === selectedSkill);
```
tdurnford, line 65: Single use variable 
```suggestion
   return skills.find(({ name }) => name === selectedSkill);
```

Composer/packages/ui-plugins/select-skill-dialog/src/BeginSkillDialogField.tsx
```diff
@@ -1,66 +1,132 @@
...
-        depth={depth + 1}
-        id={`${id}.id`}
-        name="id"
-        rawErrors={{}}
-        schema={(schema?.properties?.id as JSONSchema7) || {}}
-        uiOptions={uiOptions.properties?.id || {}}
-        value={value?.id}
-        onChange={handleIdChange}
-      />
+      <SelectSkillDialog depth={depth + 1} value={selectedSkill} onChange={onSkillSelectionChange} />
```
tdurnford, line 127: I don't think we need to pass `depth` as a prop here. We can just use a value greater than 0 in the component.

Composer/packages/ui-plugins/select-skill-dialog/src/SelectSkillDialogField.tsx
```diff
@@ -1,25 +1,32 @@
...
 
 import { ComboBoxField } from './ComboBoxField';
 
 const ADD_DIALOG = 'ADD_DIALOG';
 
-export const SelectSkillDialog: React.FC<FieldProps> = (props) => {
-  const { value, onChange } = props;
+export const SelectSkillDialog: React.FC<{
+  value: string;
+  depth: number;
```
tdurnford, line 18:
```suggestion
```

Composer/packages/ui-plugins/select-skill-dialog/src/SelectSkillDialogField.tsx
```diff
@@ -53,5 +60,17 @@ export const SelectSkillDialog: React.FC<FieldProps> = (props) => {
     }
   };
 
-  return <ComboBoxField {...props} comboboxTitle={comboboxTitle} options={options} onChange={handleChange} />;
+  return (
+    <div css={schemaField.container(depth)}>
```
tdurnford, line 64:
```suggestion
    <div css={schemaField.container(1)}>
```

Composer/yarn.lock
```diff
@@ -18283,10 +18283,10 @@ tr46@^2.0.2:
   dependencies:
     punycode "^2.1.1"
 
-tree-kill@^1.2.1:
-  version "1.2.1"
-  resolved "https://registry.yarnpkg.com/tree-kill/-/tree-kill-1.2.1.tgz#5398f374e2f292b9dcc7b2e71e30a5c3bb6c743a"
-  integrity sha512-4hjqbObwlh2dLyW4tcz0Ymw0ggoaVDMveUB9w8kFSQScdRLo0gxO9J7WFcUBo+W3C1TLdFIEwNOWebgZZ0RH9Q==
+tree-kill@^1.1.0:
```
tdurnford, line 18286: Do we need to make changes to the lock file?

Composer/packages/client/src/recoilModel/dispatchers/setting.ts
srinaath: It is. But just wanted to be more explicit.

Composer/packages/lib/indexers/src/botIndexer.ts
srinaath: Any specific suggestions. How about `The skill ${skillName} does not exist in appsettings.json`?

Composer/yarn.lock
srinaath: We dont need to. So, if i run the yarn pulling in the latest main it sets it to this. I think a stale yarn.lock was committed on main.

Composer/packages/lib/indexers/src/botIndexer.ts
tdurnford: Yea that's perfect

Composer/packages/lib/shared/src/types/settings.ts
srinaath: This is basically in the settings. We cannot enforce those two props . But it makes sense to make them mandatory

tdurnford (dismissed):

tdurnford (approved):

boydc2014 (approved):

boydc2014 (approved):

Issue #3872, opened 2020-08-20
Align QnA with the new design


liweitian:
The final design is here. https://www.figma.com/file/deFhMPBnRWxTYiYDYqYxrc/Composer-Vivek?node-id=7561%3A0

1. if expanded, no number needed: questions(number) => questions
2. color on selected and hovered should be different
3. auto-grow when typing question.
4.no (0), 0 is meaningless.
5. add QnA Pair, the new field should get auto focused
6. chevron icon and question space.
7. make the chevron icon in the middle. (left edge to question column)
8. spacing between two icons to both edge.
9. header top margin.
10. question and answer should have no hover and click state since nothing will do on that.
11. editing questions and answer should expand the row.
12. Something more on the vivke's design like how to delete the QA fields, show warning etc.

PR #4080, opened 2020-09-10
feat: add notification center for composer (QnA url import)

lei9444:
## Description
The notification flow is base on @vivekkshankar 's design
https://www.figma.com/file/LlTlh5vXwq91zjGnFtrUUR6l/Composer-main-design-spec-%2F-UI-library?node-id=313%3A15006

1. support default cards
2. support customized notification cards
3. Add time management for every single card.

## Task Item
refs #3872

## Screenshots
notifications2


cwhitten:
@lei9444 @GeoffCoxMSFT I'd like wire in the existing error tooltip that hangs off of the start bot button into this notifications module. We don't need to block this PR on doing that but I'd like it in R11 so we can remain consistent.

@vivekkshankar @sangwoohaan I assume the above is the expectation from a UX side.

hibrenda:
Seems too short for error notification. After this pop up disappear, user can not find the error message.

lei9444:
> Seems too short for error notification. After this pop up disappear, user can not find the error message.

I removed the time for error notification, customers can close it by click the close button

Composer/packages/client/src/recoilModel/DispatcherWrapper.tsx
```diff
@@ -55,7 +55,7 @@ const wrapDispatcher = (dispatchers, forceUpdate) => {
   return Object.keys(dispatchers).reduce((boundDispatchers, dispatcherName) => {
     const dispatcher = async (...args) => {
       forceUpdate([]); //gurarantee the snapshot get the latset state
-      await dispatchers[dispatcherName](...args);
+      return await dispatchers[dispatcherName](...args);
```
a-b-r-o-w-n, line 58: We should not be returning values from dispatchers. They should only be used to update state, then components subscribe to state updates to re-render.

Composer/packages/client/src/recoilModel/dispatchers/notification.ts
```diff
@@ -0,0 +1,43 @@
...
+import { CallbackInterface, useRecoilCallback } from 'recoil';
+import { v4 as uuid } from 'uuid';
+
+import { notificationsState } from '../atoms/appState';
+import { ICardProps } from '../../components/NotificationCard';
+
+export const addNotificationInternal = ({ set }: CallbackInterface, notification: ICardProps) => {
+  const id = uuid(6);
+  set(notificationsState, (notifications) => [...notifications, { id, cardProps: notification }]);
+  return id;
```
a-b-r-o-w-n, line 14: This is an anti-pattern. Instead, extract a helper method to create a notification and then invoke this dispatcher with the created notification.

Composer/packages/client/src/recoilModel/dispatchers/qna.ts
```diff
@@ -89,22 +90,31 @@ export const qnaDispatcher = () => {
...
   const importQnAFromUrls = useRecoilCallback(
     (callbackHelpers: CallbackInterface) => async ({ id, urls }: { id: string; urls: string[] }) => {
-      const { set, snapshot } = callbackHelpers;
+      const { snapshot } = callbackHelpers;
       const qnaFiles = await snapshot.getPromise(qnaFilesState);
+      const projectId = await snapshot.getPromise(projectIdState);
       const qnaFile = qnaFiles.find((f) => f.id === id);
-      set(qnaAllUpViewStatusState, QnAAllUpViewStatus.Loading);
+
+      const notificationId = await addNotificationInternal(callbackHelpers, getQnAPending(urls));
```
a-b-r-o-w-n, line 98: With my proposed change, this could look like:

```ts
const notification = createNotification(getQnAPending(urls));
await addNotificationInternal(callbackHelpers, notification);

// later
deleteNotificationInternal(callbackHelpers, notification.id);
```

a-b-r-o-w-n (changes requested):

lei9444:
Thanks @a-b-r-o-w-n, I have removed the return in dispatcher and updated the notification creation flow

Composer/packages/client/src/components/NotificationCard.tsx
```diff
@@ -0,0 +1,259 @@
...
+  ],
+  shimmerGradient: [
+    {
+      backgroundImage: 'radial-gradient(at 50% 50%, #0078D4 0%, #EDEBE9 100%);',
+    },
+  ],
+};
+// -------------------- NotificationCard -------------------- //
+
+export enum NotificationType {
```
hatpick, line 116: I highly recommend using union types for two reasons:
1. Union types are easier to debug since enums produce integers in transpiled code.
2.  You can use string literal type to do the same but much easier, because you dont have to import string literal type, but you have to import enums.

export NotificationType = 'info' | 'warning' | 'error' | 'pending' | 'success'

Composer/packages/client/src/components/NotificationCard.tsx
```diff
@@ -0,0 +1,259 @@
...
+      timer.resume();
+    }
+  };
+
+  const handleAnimationEnd = () => {
+    if (!show) onDismiss(id);
+  };
+
+  let renderCard = renderCardContent;
+  if (cardProps.onRenderCardContent) renderCard = cardProps.onRenderCardContent;
```
hatpick, line 238: nit: ```const renderCard = cardProps.onRenderCardContent || defaultCardContentRenderer```

Composer/packages/client/src/components/NotificationCard.tsx
```diff
@@ -0,0 +1,259 @@
...
+
+  return (
+    <div
+      ref={containerRef}
+      css={cardContainer(show, containerRef.current)}
+      role="presentation"
+      onAnimationEnd={handleAnimationEnd}
+      onFocus={() => {}}
```
hatpick, line 246: ?

Composer/packages/client/src/components/NotificationCard.tsx
```diff
@@ -0,0 +1,259 @@
...
+        )}
+      </div>
+    </div>
+  );
+};
+
+export const NotificationCard = (props: INotificationProps) => {
+  const { cardProps, id, onDismiss } = props;
+  const [show, setShow] = useState(true);
+  const containerRef = useRef(null);
```
hatpick, line 203: Please add the HTMLDivElement type to the ref.

Composer/packages/client/src/components/NotificationCard.tsx
```diff
@@ -0,0 +1,259 @@
...
+  onRenderCardContent?: (props: ICardProps) => JSX.Element;
+}
+
+export interface INotificationProps {
+  id: string;
+  cardProps: ICardProps;
+  onDismiss: (id: string) => void;
+}
+
+export class Timer {
```
hatpick, line 144: if this needs to be exported perhaps better to put it in its own file.

Composer/packages/client/src/recoilModel/atoms/appState.ts
```diff
@@ -152,6 +153,11 @@ export const boilerplateVersionState = atom<BoilerplateVersion>({
   },
 });
 
+export const notificationsState = atom<Notification[]>({
```
hatpick, line 156: Have you considered using an atomFamily?

Composer/packages/client/src/components/NotificationIcon.tsx
```diff
@@ -0,0 +1,69 @@
...
+const numberText = css`
+  transform: scale(0.5);
+  font-size: 20px;
+  color: white;
+`;
+
+// -------------------- NotificationIcon -------------------- //
+
+interface INotificationIconProps {
+  number: number;
```
hatpick, line 53: nit: count

Composer/packages/client/src/components/NotificationIcon.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+export const NotificationIcon = (props: INotificationIconProps) => {
+  const { number } = props;
+  return (
+    <div css={container} role="presentation">
+      <Icon css={ringer} iconName="Ringer" />
+      {number > 0 && <Icon css={circleMask} iconName="FullCircleMask" />}
+      {number > 0 && (
+        <div css={numberContainer}>
+          <span css={numberText}>{number > 99 ? '...' : number}</span>
```
hatpick, line 64: is the '...' from a UX design? If not, how about ```99+``` instead

Composer/packages/client/src/recoilModel/types.ts
```diff
@@ -128,3 +130,8 @@ export enum QnAAllUpViewStatus {
   Success,
   Failed,
 }
+
+export type Notification = {
```
hatpick, line 134: nit: ```export type Notification = ICardProps & { id: string }```

Composer/packages/client/src/utils/notifications.ts
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+import formatMessage from 'format-message';
+
+import { NotificationType } from './../components/NotificationCard';
+import { ICardProps } from './../components/NotificationCard';
```
hatpick, line 6: new line

Composer/packages/client/src/utils/notifications.ts
```diff
@@ -0,0 +1,34 @@
...
+export const getQnAPending = (urls: string[]) => {
+  return {
+    title: formatMessage('Creating your knowledge base'),
+    description: formatMessage('Extacting QnA pairs from ') + urls.join(' '),
```
hatpick, line 10: ```formatMessage('Extracting QNA pairs from {urls}', { urls: urls.join (' ') })```

Composer/packages/client/src/utils/notifications.ts
```diff
@@ -0,0 +1,34 @@
...
+    type: NotificationType.loading,
+  };
+};
+
+export const getQnASuccess = (callback: () => void): ICardProps => {
+  return {
+    title: formatMessage('Your knowledge base Surface go FAQ is ready!'),
+    description: '',
```
hatpick, line 18: make this optional

Composer/packages/client/src/utils/notifications.ts
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+import formatMessage from 'format-message';
+
+import { NotificationType } from './../components/NotificationCard';
+import { ICardProps } from './../components/NotificationCard';
+export const getQnAPending = (urls: string[]) => {
```
hatpick, line 7: not: getQnaPendingNotification (same for the other function names)

Composer/packages/client/src/components/NotificationCard.tsx
```diff
@@ -0,0 +1,259 @@
...
+
+export enum NotificationType {
+  info = 1,
+  warning,
+  error,
+  loading,
+  success,
+}
+
+export interface ILink {
```
hatpick, line 124: prefer types over interfaces

Composer/packages/client/src/components/NotificationCard.tsx
```diff
@@ -0,0 +1,259 @@
...
+  label: string;
+  onClick: () => void;
+}
+
+export interface ICardProps {
```
hatpick, line 129: prefer types over interfaces

Composer/packages/client/src/components/NotificationCard.tsx
```diff
@@ -0,0 +1,259 @@
...
+  title: string;
+  description: string;
+  type: NotificationType;
+  retentionTime?: number;
+  link?: ILink;
+  onRenderCardContent?: (props: ICardProps) => JSX.Element;
+}
+
+export interface INotificationProps {
```
hatpick, line 138: prefer types over interfaces

Composer/packages/client/src/recoilModel/atoms/appState.ts
hatpick: If there are multiple notifications showing and user deletes one of them or a new notification gets added to the list, the entire list gets re-rendered. I suggest doing an atom to keep the notificationIds and an atomFamily for the notifications content using that id, then use React.memo on the notification component to prevent re-rendering if that atomFamily hasn't changed (aka non of the existing notifications gets re-rendered when a new notification comes in or one gets deleted)

Composer/packages/client/src/components/NotificationCard.tsx
a-b-r-o-w-n: This is a good suggestion.

Composer/packages/client/src/components/NotificationIcon.tsx
lei9444: This icon is not used in composer now, this is not the final design and I will remove it first.

Composer/packages/client/src/components/NotificationCard.tsx
lei9444: this is a eslint rule onMouseOver must be accompanied by onFocus for accessibility, will have a warn. I add this to pass the rule.

lei9444:
Hi @hatpick,  Thanks for you suggestions and it's very useful for me. I have updated it.

Composer/packages/client/src/recoilModel/dispatchers/notification.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export const addNotificationInternal = ({ set }: CallbackInterface, notification: Notification) => {
+  set(notificationsState(notification.id), notification);
+  set(notificationIdsState, (ids) => [...ids, notification.id]);
+};
+
+export const deleteNotificationInternal = ({ reset, set }: CallbackInterface, id: string) => {
+  reset(notificationsState(id));
+  set(notificationIdsState, (items) => {
+    return [...items].filter((item) => item !== id);
```
hatpick, line 25: nit: filter will create a new ref to the array, no need to spread it.

hatpick (dismissed):

hatpick (approved):

cwhitten (approved):

vivekkshankar:
@cwhitten Missed your message earlier. We do not plan to merge validation errors with notifications.
Notifications are reserved on system progress, publish, runtime errors etc.

The interaction model for validations is a bit different by design. 

https://www.figma.com/file/deFhMPBnRWxTYiYDYqYxrc/Composer-Vivek?node-id=8206%3A0

cwhitten:
@vivekkshankar you're talking about something different. I do mean the existing runtime & publish errors that we show currently below the start button in a popover element. It would certainly be a bad UX to preserve this & the notification panel.

PR #4086, opened 2020-09-10
feat: Feature authorization middleware

### LINKED ISSUE #4084: Bot to bot authorization for skills is missing
This is a security issue for skills build with composer (secure by design principle)

I want to set a feature flag in the bot settings to activate the an authorization middleware, so that I can configure which bot/app ids are allowed to consume my skills.  

I added the auth middleware in my custom runtime, but this should be an out of box feature -> secure by design

## Additional context
You can find my draft PR here #4086

ltwlf:
## Description
This PR adds a claim validation middleware for secure bot to bot communication. Users should whitelist which consumers can call a composer skill bot. By default the feature is activated (secure by config) and adds AllowedCallers=["*"] to the appsettings.json. The middleware implementation is included in the runtime as C# file so that devs can customize the validation.

This implementation should be compatible with older bot versions. Middleware will not be activated when the feature flag in the setting is missing. 

## Task Item
closes #4084


cwhitten:
Thank you for the contribution @ltwlf - @carlosscastro will review soon.

Are you able to add any tests with this change?

ltwlf:
I will check tests when the draft is confirmed as an useful feature for composer.

runtime/dotnet/azurewebapp/Authorization/AllowedCallersClaimsValidator.cs
```diff
@@ -0,0 +1,55 @@
+using System;
```
carlosscastro, line 1: Copyright header here.

runtime/dotnet/azurewebapp/Startup.cs
```diff
@@ -84,9 +85,14 @@ public IStorage ConfigureStorage(BotSettings settings)
             return storage;
         }
 
+        public bool ShouldUseAuthorizationMiddleWare(BotSettings settings)
+        {
+            return settings?.Feature?.UseAuthorizationMiddleware == true;
```
carlosscastro, line 90: Naming: Features should be higher level feature ideally. A more high level, user facing feature I'm thinking here would be _is Is this a Skill?_ or _Is this a Skill Host?_. In that context, we could have 2 settings: IsSkill and IsSkillHost, and when IsSkill is true, then one of the consequences is that we enable this authorization claims validator.

I encourage you to provide feedback on this suggestion, would love to hear your thoughts on it.

carlosscastro (changes requested):
Thanks for your contribution! This is great and perfect timing. The change is very close to being in a mergeable state. Just 2 things that need to be addressed: 1) I added a comment about naming and the granularity of this feature. Please take a look. and 2) Tomorrow, we're having a discussion on how allowed claims lists will be configured in the UI. From that meeting, there might be some extra requirements around this. If that is the case, I'll update the Pr and we can discuss.

ltwlf:
Will check the feedback tomorrow in detail. The idea with the higher level feature make sense. As far as I know claims validation makes only sense for skills and not for consumers. As soon you activate that a bot can be a skill the claim validation should be activated and the skill manifest should become part of the composer project (not sure how this is handled) currently. Let’s see what the others say.

ltwlf:
@carlosscastro do you have feedback from the meeting? Would like to work on this at the weekend. 
And do you have any hint for me where is the best place to integrate the feature on a higher level?

carlosscastro:
Hello @ltwlf! After the meeting, the direction we discussed looks good. Some more details below: 

- Addition of a new IsSkill setting
- If IsSkill is set to `false`, in startup.cs we do not register the bot as a skill here: https://github.com/microsoft/BotFramework-Composer/blob/main/runtime/dotnet/azurewebapp/Startup.cs#L141 and in line 143. Line 142 is about being a skill host which is a different story and we'll likely hide it behind other settings in a near future but you don't need to worry about it.
- If IsSkill is set to `true`, in startup.cs we register the bot as a skill https://github.com/microsoft/BotFramework-Composer/blob/main/runtime/dotnet/azurewebapp/Startup.cs#L141 and in line 143
- If IsSkill is set to `true`, we configure your claim validator.
- The AllowedCallers list is empty by default, as opposed to '*'. We'll add some Ui and docs around this in order to improve the experience and avoid frustration. However the current * configuration is allowing a lot of access skills and bot developers may not be aware which we think is not ideal.

Let me know if any of that doesn't make sense, happy to discuss. I'll be around over the weekend so just tag me if you have questions!

ltwlf:
@carlosscastro here is my new draft for skill authorization. I've created an own class for skill settings, because there may will be more skill specific settings. I create default settings like this 
```
skill: {
  isSkill:false,
  allowedCallers:['*']
} 
```
is skill could then either be configured via settings.json or UI

Looking forward to your feedback.

Best wishes,
Christian

runtime/dotnet/azurewebapp/Authorization/AllowedCallersClaimsValidator.cs
carlosscastro: Please add a copyright header like other files have.

runtime/dotnet/azurewebapp/Startup.cs
```diff
@@ -84,9 +85,14 @@ public IStorage ConfigureStorage(BotSettings settings)
...
 
+        public bool IsSkill(BotSettings settings)
+        {
+            return settings?.Skill?.IsSkill == true;
+        }
+
         public BotFrameworkHttpAdapter GetBotAdapter(IStorage storage, BotSettings settings, UserState userState, ConversationState conversationState, IServiceProvider s, TelemetryInitializerMiddleware telemetryInitializerMiddleware)
         {
-            var adapter = new BotFrameworkHttpAdapter(new ConfigurationCredentialProvider(this.Configuration));
+            var adapter = IsSkill(settings) ? new BotFrameworkHttpAdapter(new ConfigurationCredentialProvider(this.Configuration), s.GetService<AuthenticationConfiguration>()) : new BotFrameworkHttpAdapter(new ConfigurationCredentialProvider(this.Configuration));
```
carlosscastro, line 95: In general we want to move away from doing GetService in startup.cs. If we could optionally do a services.AddSingleton<AuthenticationConfiguration> and have DI wire it into the adapter's constructor , and also do services.AddSingleton<ICredentialProvider>(new ConfigurationCredentialProvider...), then we could avoid the getService. Not 100% sure that would work because of the numerous overloads of BotFrameworkHttpAdapter constructor. If that doesn't work, then as-is is good. We are currently refactoring the adapter and will soon refactor this startup.cs, so we should clean this up soon.

runtime/dotnet/azurewebapp/Startup.cs
```diff
@@ -123,8 +129,11 @@ public void ConfigureServices(IServiceCollection services)
             services.AddSingleton<ICredentialProvider, ConfigurationCredentialProvider>();
             services.AddSingleton<BotAdapter>(sp => (BotFrameworkHttpAdapter)sp.GetService<IBotFrameworkHttpAdapter>());
 
-            // Register AuthConfiguration to enable custom claim validation.
-            services.AddSingleton<AuthenticationConfiguration>();
+            // Register AuthConfiguration to enable custom claim validation for skills.
+            if (IsSkill(settings))
+            {
+                services.AddSingleton(sp => new AuthenticationConfiguration { ClaimsValidator = new AllowedCallersClaimsValidator(settings.Skill) });
```
carlosscastro, line 135: `settings` could be null, right?

carlosscastro: If yes, we cannot throw NullRef

carlosscastro (commented):
Looking good. Last details on my comments: 1) copyright header missing, 2) possible null ref exception in startup.cs and 3) potential improvement in registration (which may not have a happy solution, just a suggestion). Once we have these, we should be ready to go.

Separately, FYI we will create items to port this to the (in progress) js runtime and azure functions. But that can be done by other contributors, or by you in a separate PR -- your call. Definitely let's focus on this in this PR, since it's almost ready to go. Thanks again for iterating on this.

runtime/dotnet/azurewebapp/Startup.cs
ltwlf: I couldn't find a way to inject the Authorization. That's why I used GetService.

runtime/dotnet/azurewebapp/Startup.cs
ltwlf: IsSkill returns false when null:
```csharp 
        public bool IsSkill(BotSettings settings)
        {
            return settings?.Skill?.IsSkill == true;
        }
```
So it should be fine

ltwlf:
@carlosscastro I changed from PR draft and added some unit tests. Looking forward to your feedback,

runtime/dotnet/core/Settings/BotSkillSettings.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.BotFramework.Composer.Core.Settings
+{
+    public class BotSkillSettings
+    {
+        // Is skill
```
carlosscastro, line 8: Minor, optional: Ideally we strive for either xml docs (`///` format) or no docs.

runtime/dotnet/tests/AllowedCallersClaimsValidationTests.cs
```diff
@@ -0,0 +1,98 @@
+﻿using System;
```
carlosscastro, line 1:
```suggestion
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License.

﻿using System;
```

carlosscastro (approved):

cwhitten (approved):

ltwlf:
Hi, can I do anything to finalize this PR? When this PR is merged, I would like to the same for Azure Funcs and then for JS...
@carlosscastro @cwhitten

cwhitten:
@ltwlf for some reason the unit tests are failing so I've re-run CI. We need that to pass before we can merge.

cwhitten:
Summary of all failing tests
FAIL packages/server/src/controllers/__tests__/project.test.ts (8.628 s)
  ● skill operation › should retrieve skill manifest

    expect(jest.fn()).toHaveBeenCalledWith(...expected)

    Expected: 200
    Received: 404

    Number of calls: 1

      339 |     } as Request;
      340 |     await ProjectController.getSkill(mockReq, mockRes);
    > 341 |     expect(mockRes.status).toHaveBeenCalledWith(200);
          |                            ^
      342 |   }, 10000);
      343 | });
      344 | 

      at Object.toHaveBeenCalledWith (src/controllers/__tests__/project.test.ts:341:28)
          at runMicrotasks (<anonymous>)


Test Suites: 1 failed, 217 passed, 218 total
Tests:       1 failed, 3 skipped, 1 todo, 840 passed, 845 total
Snapshots:   5 passed, 5 total
Time:        410.851 s
Ran all test suites in 13 projects.
Force exiting Jest: Have you considered using `--detectOpenHandles` to detect async operations that kept running after all tests finished?
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
Error: Process completed with exit code 1.

cwhitten:
@ltwlf The test failure is unrelated to this change. We'll merge this and address the failure in a follow-up PR

cwhitten (approved):

PR #3913, opened 2020-08-26
feature: localization framework and pseudolocalization

beyackle:
## Description

This adds in the ability to swap between locales in User Settings, along with a pseudo-locale to be used for testing purposes.

- [x] Should the "pseudo" locale files be removed? They're not needed by the client, but might be useful for other people developing new features (so they can run the yarn:l10n script and check they didn't forget any formatMessage calls themselves) -- removed the pseudo locale files, but left in the script to generate them. Their entry in the user settings is now commented out by default. Developers who are adding strings should run `yarn l10n` to regenerate the locale files; there's a check now to add the pseudo-locale and "does not exist" locale to the settings box in non-prod settings.
- [x] We need a new icon for the settings menu's "User Locale" setting.
-- Done, thanks to Vivek on the design team
- [x] I had to comment out some error handling in the server. This should be added back in a way that doesn't break things.
-- Fixed (I discovered how to do this on my own)

## Task Item

closes #3809 

## Screenshots

MicrosoftTeams-image (2)

PR #4113, opened 2020-09-14
refactor: remove lib folder under client extensions

VanyLaw:
## Description
This PR remove the `/extensions` folder in client.
because all extension under `/extensions` were move to `/packages` in #4075, but its lib folders are still in main branch.

## Task Item
#minor


VanyLaw:
Hi @a-b-r-o-w-n, I saw you rename all extension in `/extensions`, So I think all the `/extensions` can be removed. What do you think?

a-b-r-o-w-n:
It looks like the lib directories were committed in #3913. They were removed in #4075, so this looks good to me.

a-b-r-o-w-n (approved):

PR #4117, opened 2020-09-14
feat: State management support for loading mutiple projects

### LINKED ISSUE #4031: FilePersistence and Undo/Redo Support for BotProjects
Support for filepersistence and Undo/redo for multiple projects

srinaath:
This is the first leg of the work to support multiple projects in Composer. This work lets our state management to use atomFamilies in Recoil for botstates. This PR does not add net new functionality but sets up the Bot Project work which wil coming in succesive PR's

Fixes #4031


Composer/packages/client/src/recoilModel/undo/__test__/undoHistory.test.ts
```diff
@@ -1,29 +1,32 @@
-// Copyright (c) Microsoft Corporation.
-// Licensed under the MIT License.
+// // Copyright (c) Microsoft Corporation.
+// // Licensed under the MIT License.
```
beyackle, line 2: How'd these get double-commented?

Composer/packages/client/src/recoilModel/undo/__test__/undoHistory.test.ts
```diff
@@ -47,12 +50,12 @@ describe('undoHistory class', () => {
     expect(undoHistory.stack.length).toBe(0);
   });
 
-  it('should only support 30 history', () => {
+  it('should only support 40 actions in history', () => {
```
beyackle, line 53: This test still looks like it's only testing for 30 actions (it generates 40, but then checks that 10 through 39 are present, so it's throwing out 0 through 9).

Composer/packages/client/src/recoilModel/dispatchers/__tests__/MeetingVA.botProject.json
```diff
@@ -0,0 +1,12 @@
+{
+  "$schema": "https://schemas.botframework.com/schemas/botprojects/v0.1/worskspace.json",
+  "name": "MeetingVA",
+  "workspace": "/Users/srravich/Desktop/samples/Parent",
```
beyackle, line 4: Should we change this to something less personal?

beyackle (changes requested):

lei9444:
I tested the create bot, delete bot, save as, copy, paste,  update dialog, lu, lg and setting. all this parts work well

Composer/packages/client/src/recoilModel/dispatchers/skill.ts
```diff
@@ -66,7 +66,39 @@ export const skillDispatcher = () => {
...
+
+      set(showAddSkillDialogModalState(projectId), false);
+      set(onAddSkillDialogCompleteState(projectId), {});
+    }
+  );
+
+  const removeSkill = useRecoilCallback(
+    (callbackHelpers: CallbackInterface) => async (projectId: string, manifestUrl?: string) => {
+      const { snapshot } = callbackHelpers;
+      const skills = [...(await snapshot.getPromise(skillsState(projectId)))].filter(
```
beyackle, line 89: I don't think you need the [...] here - `filter` creates a new array anyway.

Composer/packages/client/src/recoilModel/undo/__test__/undoHistory.test.ts
srinaath: Nice catch. So we push a max of 30 items on the stack. As we add beyond that we do a stack.shift() to keep pushing the oldest elements out of the stack. I will revert the test name

beyackle (approved):

PR #4119, opened 2020-09-14
feat: electron splash screen

### LINKED ISSUE #4047: Composer needs a splash screen which shows immediately.
## Describe the bug

It takes way too long to load the composer.  I click on the icon and there is zero feedback that the app is loading until almost a minute, when the splash screen shows for a second then the app shows.  

The point of splash screens is to solve this problem.  If we can't get a display immediately because of the electron app we should create a tiny launcher app which shows the splash screen and then launches the real electron app.  Once the electron app has loaded it signals back to the splash screen so that the splash screen can go away.

## Version

1.1

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

start the app

## Expected behavior

It should show the splash screen with 1-2 seconds, certainly no longer then 5 seconds.

hatpick:
## Description
Adds splash screen to composer.

## Task Item
closes #4047 

## Screenshots

Windows 10
win_ss

Mac OS (Catalina)
mac_ss

Linux (Ubuntu 20.04 LTS)
linux_ss


Composer/packages/electron-server/src/main.ts
```diff
@@ -200,13 +205,30 @@ async function run() {
...
+      logo: `file://${microsoftLogoPath}`,
+      productName: 'Bot Framework Composer',
+      productFamily: 'Microsoft Azure',
+      status: 'Initializing...',
+      website: 'www.botframework.com',
+      width: 500,
+      height: 300,
+    });
+
+    updateStatus('Starting server...');
```
hatpick, line 221: Messages like this should be localized, since the app menu was not localized, I assumed the effort is still going on, please feel free to let me know if things are in place for electron localization and how can I use it.

tonyanziano:
This is awesome! 👍👍👍

Composer/packages/electron-server/src/main.ts
cwhitten: @beyackle can you comment?

cwhitten:
Very nice @hatpick

Composer/packages/electron-server/src/main.ts
beyackle: Looks like we're not localizing anything in the `electron-server` directory yet, but we ought to be, with our standard `formatMessage` calls. Let's open another issue for that and handle this string along with all the others.

hatpick: @beyackle sounds good, thanks

beyackle: As a side note, the l10n script does look in `electron-server` already, so no change is needed there. Just get `formatMessage` around all your strings and you'll be good to go.

hatpick: #4120 Here's the issue.

srinaath:
This looks so branded and official. Love it!

Composer/packages/electron-server/src/main.ts
```diff
@@ -150,7 +153,7 @@ async function loadServer() {
   log(`Server started at port: ${serverPort}`);
 }
 
-async function main() {
+async function main(show = false) {
```
GeoffCoxMSFT, line 156: Nit: I know with the default value the boolean type is inferred, but it would be nice to specify the type of the show param here.

Composer/packages/electron-server/src/splash/splashScreen.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  splashScreenWindow.on('ready-to-show', () => {
+    splashScreenWindow.show();
+  });
+
+  const file = 'data:text/html;charset=UTF-8,' + encodeURIComponent(getSplashScreenContent(args));
+  await splashScreenWindow.loadURL(file);
+
+  const startApp = () => {
+    setTimeout(() => splashScreenWindow.destroy(), 500);
```
GeoffCoxMSFT, line 72: Any chance that splashScreenWindow could be undefined/null here after the timeout?  Should it be set to undefined after being destroyed?

GeoffCoxMSFT (dismissed):

Composer/packages/electron-server/src/splash/splashScreen.ts
hatpick: 1. It's guaranteed to be "not-null"
2. That's a good code practice if it was exposed to the caller side, but since no one else is going to consume the window, it's fine.

hatpick:
@mareekuh I talked to Sangwoo about the "brand family" (Microsoft Azure), and he wasn't sure if Composer falls under that family. What do you think? Should I remove the "Microsoft Azure" wording?

cwhitten:
@hatpick let's take this into the nightly and rev the UI if need be. Please move out of Draft when you are ready. Thanks!

GeoffCoxMSFT (approved):

PR #4133, opened 2020-09-15
fix: security: downgrade node-forge to 0.9.0

boydc2014:
#minor


boydc2014:
@beyackle can you help check why this src/locales/en-us.json is changed after just yarn install and build?

boydc2014:
> @beyackle can you help check why this src/locales/en-us.json is changed after just yarn install and build?

It turns out that the changes to src/locales/en-us.json is not caused by this package update, so i revert the change and created another ticket for it https://github.com/microsoft/BotFramework-Composer/issues/4134

a-b-r-o-w-n (approved):

a-b-r-o-w-n:
@boydc2014 Thanks! I was looking at this yesterday and was confused. I didn't realize we needed a downgrade! It's nice to have all of CI passing again. 🐼

PR #4140, opened 2020-09-15
feature: Adding VA template behind feature flag in Composer

### LINKED ISSUE #4142: Add VA template behind feature flag in Composer

pavolum:
## Description

This is the first of a series of PRs for VA Creation in Composer. This PR covers the addition of the VA Core template to Composer behind an env variabe (VA_CREATION) feature flag. fixes #4142 

To test, pull this branch, set the VA_CREATION env variable to true in the command line and run `yarn start:dev` in that same command line. This will result in 'va-core' appearing in the list of bot templates. 

At current state the VA template is functional in Composer and can run locally if you have preprovisioned LUIS and QNA resources that you can publish to.

**NOTE:** 
- Most of this PR is comprised of the VA core template itself which is maintained [here](https://github.com/microsoft/botframework-virtual-assistant). **I would not review this code (everything in the vacre/template folder) at this point and instead review the "glue" code that loads this template as a plug in in Composer.** This template code will be maintained and reviewed in its own repo (linked above) and we will periodically update the template code in Composer to reflect the latest until we determine a more ideal long term solution.
- Initially the VA template had a corresponding custom runtime associated with it. I have removed this from the PR and set it so the VA Template uses the Composer base runtime. The custom runtime discussion is still in the air but it seems like we are straying away from the concept and will instead want everything to work on a single base runtime. Until we decide otherwise I am removing the custom runtime code for the time being.


Composer/plugins/vacore/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "vacore",
+  "version": "1.0.0",
+  "description": "provide functionality around using VA core",
```
GeoffCoxMSFT, line 4: Nit: Provides
How about:  A plug-in to provide Virtual Assistant creation ? (or something like that)

Composer/plugins/vacore/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "main": "lib/index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
```
GeoffCoxMSFT, line 7: I'd remove the test script if it doesn't do anything.

Composer/plugins/vacore/src/index.ts
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+import path from 'path';
+
+export default async (composer: any): Promise<void> => {
+  if (process.env.VA_CREATION) {
+    // register the base template which will appear in the NEw Bot modal
```
GeoffCoxMSFT, line 7: Nit: new

Composer/plugins/vacore/template/.gitignore
```diff
@@ -0,0 +1 @@
+/generated/
```
GeoffCoxMSFT, line 1: Consider putting all the .gitignore paths in the top level one.  It is painful to track down .gitignores at levels lower than the package.json.

Composer/plugins/vacore/template/assistant.dialog
```diff
@@ -0,0 +1,118 @@
+{
+  "$kind": "Microsoft.AdaptiveDialog",
+  "$designer": {
+    "$designer": {
```
GeoffCoxMSFT, line 4: Do you want the $designer nodes in these templates?  I've seen where the id get reused when starting projects from templates and although nothing really cares about the dialog IDs in the $designer node, it appears strange when in text editing mode.

Composer/plugins/vacore/template/assistant.dialog
```diff
@@ -0,0 +1,118 @@
...
+      "$kind": "Microsoft.OnConversationUpdateActivity",
+      "$designer": {
+        "id": "376720"
+      },
+      "actions": [
+        {
+          "$kind": "Microsoft.Foreach",
+          "$designer": {
+            "id": "518944",
+            "name": "Loop: for each item"
```
GeoffCoxMSFT, line 22: Will the templates be localized to the different languages?

Composer/plugins/vacore/template/schemas/update-schema.ps1
```diff
@@ -0,0 +1,30 @@
+param (
```
GeoffCoxMSFT, line 1: Do we need to ship these scripts with each template? They are already in the runtime with composer?

Composer/plugins/vacore/template/scripts/VAScripts/deploy_cognitive_models.ps1
```diff
@@ -0,0 +1,385 @@
+#Requires -Version 6
+
+Param(
+	[string] $name,
```
GeoffCoxMSFT, line 4: Nit: alignment of items

GeoffCoxMSFT (dismissed):
Looks good to me, but you should have a more knowledgeable person with the plug-ins and templating approve before committing.

Composer/plugins/vacore/template/scripts/DeploymentTemplates/function-template-with-preexisting-rg.json
```diff
@@ -0,0 +1,356 @@
+{
```
benbrown, line 1: Is this the same file as is currently present in the "boilerplate" that is copied into every new project?  If so, we should remove it from this template.  If not, we should make sure that there is not a file naming collision, otherwise the boilerplate copy may overwrite it.

https://github.com/microsoft/BotFramework-Composer/tree/main/Composer/plugins/samples/assets/shared/scripts

Composer/plugins/vacore/template/language-generation/en-us/general.en-us.lg
```diff
@@ -0,0 +1,24 @@
+[import](common.lg)
```
benbrown, line 1: what is this file for?  It doesn't seem to be included anywhere.

benbrown (changes requested):
There are some extraneous files here that should probably be removed. 

Also, duplicating the boilerplate content is probably not good - we should make sure we ONLY include files that are NOT including in the boilerplate or runtime folders, and make sure they merge properly during creation.  (/scripts folder, /schemas folder)

Composer/plugins/vacore/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "vacore",
+  "version": "1.0.0",
```
hatpick, line 3: Check with @cwhitten on the versioning.

Composer/plugins/vacore/template/scripts/DeploymentTemplates/function-template-with-preexisting-rg.json
pavolum: I have cleaned up the template a good bit, to note this is a version of the declarative VA template that is maintained and reviewed in a separate repo.

pavolum: @GeoffCoxMSFT @benbrown  I am not going to address all of the template comments in this PR. The reason being that this template is a placeholder for the time being and will change as a whole in the next iteration of the creation work.

Composer/plugins/vacore/src/index.ts
```diff
@@ -0,0 +1,15 @@
...
+// Licensed under the MIT License.
+import path from 'path';
+
+export default async (composer: any): Promise<void> => {
+  if (process.env.VA_CREATION) {
+    // register the base template which will appear in the new bot modal
+    composer.addBotTemplate({
+      id: 'va-core',
+      name: 'VA Core',
+      description: 'The core of your new VA - ready to run, just add skills!',
```
cwhitten, line 11: we need to localize these strings using formatMessage

cwhitten (changes requested):

Composer/plugins/vacore/src/index.ts
pavolum: Okay, I have added formatMessage

cwhitten (approved):

PR #4161, opened 2020-09-16
feat: electron app localization 

### LINKED ISSUE #4120: Localize electron app

Localize electron app (app menu, messages, etc.)

Follow the same approach used in the web app.

hatpick:
## Description

Adds localization to the electron server.

## Task Item
closes #4120 

## Screenshots
l10n


Composer/packages/server/src/router/api.ts
```diff
@@ -68,7 +68,7 @@ router.post('/runtime/eject/:projectId/:template', EjectController.eject);
 //assets
 router.get('/assets/projectTemplates', AssetController.getProjTemplates);
 
-router.use('/assets/locales/', express.static(path.join(__dirname, '..', '/locales')));
+router.use('/assets/locales/', express.static(path.join(__dirname, '../../src/locales')));
```
beyackle: I think it would help to break this out as `'..','..','src','locales'`, just so we're not assuming the OS's path separators.

beyackle (approved):

PR #4163, opened 2020-09-16
feat: Add endpoint dropdown to the add skill modal

### LINKED ISSUE #4087: The value in the Manifest Url field in the add skill modal persists between adding multiple skills
## Description
If you try to add a second skill right after you added the first, the Manifest Url field is populated with the previous skill's url.

## To Reproduce

1. Go to the 'Skills' page
2. Click the 'Connect to a new skill' button
3. Add a valid Skill Manifest Url - https://yuesuemailskill0207-gjvga67.azurewebsites.net/manifest/manifest-1.0.json
2. Click the 'Connect to a new skill' button again
1. Notice the 'Manifest url' field is already populated

## Screenshots
Sep-10-2020 10-38-50

### LINKED ISSUE #4122: Users should be able to select the skill endpoint when they add a new skill
When the user adds a new skill to the parent bot, they should be able to select which endpoint they want to use. The `msAppId` and `endpointUrl` for the endpoint the user selects should serialize to the bot settings.

tdurnford:
## Description
Adds an endpoint dropdown to the add skill modal which allows the user to select which messaging endpoint they want to use when they add the skill. The `msAppId` and `endpointUrl` for the endpoint the user selects is serialized to the `skills` object in the bot settings with the other skill data - `name` and `manifestUrl` - provided in the modal.

## Task Item
Closes #4122
Closes #4087

## Screenshots

image

### DIFF
--- a/Composer/packages/client/src/components/CreateSkillModal.tsx
+++ b/Composer/packages/client/src/components/CreateSkillModal.tsx
@@ -15,7 +15,7 @@ import { Skill } from '@bfc/shared';
 
 import { addSkillDialog } from '../constants';
 import httpClient from '../utils/httpUtil';
-import { skillsState, projectIdState } from '../recoilModel';
+import { skillsState } from '../recoilModel';
 
 import { DialogWrapper, DialogTypes } from './DialogWrapper';
 
@@ -30,6 +30,7 @@ export const skillNameRegex = /^\w[-\w]*$/;
 export const msAppIdRegex = /^[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{12}$/;
 
 export interface CreateSkillModalProps {
+  projectId: string;
   onSubmit: (data: Skill) => void;
   onDismiss: () => void;
 }
@@ -114,10 +115,7 @@ export const validateName = ({
   }
 };
 
-export const CreateSkillModal: React.FC<CreateSkillModalProps> = (props) => {
-  const { onSubmit, onDismiss } = props;
-
-  const projectId = useRecoilValue(projectIdState);
+export const CreateSkillModal: React.FC<CreateSkillModalProps> = ({ projectId, onSubmit, onDismiss }) => {
   const skills = useRecoilValue(skillsState);
 
   const [formData, setFormData] = useState<Partial<Skill>>({});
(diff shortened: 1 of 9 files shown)


Composer/packages/client/src/components/CreateSkillModal.tsx
```diff
@@ -1,265 +1,279 @@
...
+  setFormDataErrors,
+  setValidationState,
+  validationState,
+}) => {
+  const { name } = formData;
+  const { name: _, ...errors } = formDataErrors;
+
+  if (name && !skillNameRegex.test(name)) {
+    setFormDataErrors({ ...errors, name: formatMessage('Name cannot include special characters or spaces') });
+  } else if (name && skills.some((skill) => skill.name === name)) {
```
srinaath, line 109: Do you think we should allow same skill name with different case? Hello-skill and hello-skill. I feel like we should'nt

Composer/packages/client/src/pages/design/DesignPage.tsx
```diff
@@ -676,16 +670,7 @@ const DesignPage: React.FC<RouteComponentProps<{ dialogId: string; projectId: st
             onSubmit={handleCreateDialogSubmit}
           />
         )}
-        {showAddSkillDialogModal && (
-          <CreateSkillModal
```
srinaath, line 680: Can we pass the projectId along with the CreateSkillModal. The reason being when we support multiple bots we are calling the dispatcher function with the project ID rather than fetching the current project ID asynchronously in the dispatcher.

Composer/packages/client/src/recoilModel/dispatchers/skill.ts
```diff
@@ -39,50 +40,61 @@ export const skillDispatcher = () => {
...
+    const skills = [...(await snapshot.getPromise(skillsState))].filter((skill) => skill.manifestUrl !== manifestUrl);
+    await updateSkillState(callbackHelpers, skills);
+  });
+
+  const updateSkill = useRecoilCallback(
+    (callbackHelpers: CallbackInterface) => async ({ targetId, skillData }: { targetId: number; skillData?: any }) => {
+      const { snapshot } = callbackHelpers;
       const originSkills = [...(await snapshot.getPromise(skillsState))];
 
-      // add
```
srinaath, line 48: Nice work! Separating out the logic into functions

Composer/packages/client/src/pages/design/DesignPage.tsx
srinaath: https://github.com/microsoft/BotFramework-Composer/pull/4117/files#diff-566cb1ffd02936030db44dde3e418609L20

Composer/packages/client/src/components/CreateSkillModal.tsx
tdurnford: I agree. We should probably do the same thing for the manifest url.

srinaath (approved):

PR #4169, opened 2020-09-17
fix: qna migration bug

### LINKED ISSUE #4168: qna files migration errors.
## Describe the bug
Create a new todoluisbot. Then check global qnaFile state. You can see the number of qnafiles is not the same as dialog files.

## Browser

- [ ] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

liweitian:
## Description
We should do migration on server. This pr will check missing qna files, create them and return all files to client.

## Task Item
Closes #4168

### DIFF
--- a/Composer/packages/server/src/models/bot/botProject.ts
+++ b/Composer/packages/server/src/models/bot/botProject.ts
@@ -24,7 +24,7 @@ import values from 'lodash/values';
 import { Path } from '../../utility/path';
 import { copyDir } from '../../utility/storage';
 import StorageService from '../../services/storage';
-import { getDialogFileName, getQnAFileName } from '../utilities/util';
+import { getDialogNameFromFile } from '../utilities/util';
 import { ISettingManager, OBFUSCATED_VALUE } from '../settings';
 import { DefaultSettingManager } from '../settings/defaultSettingManager';
 import log from '../../logger';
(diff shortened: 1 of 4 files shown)


Composer/packages/server/src/models/bot/botProject.ts
```diff
@@ -728,6 +729,46 @@ export class BotProject implements IBotProject {
       fileList.set(file.name, file);
     });
 
+    const migrationFiles = await this._createQnAFilesForOldBot(fileList);
+
+    return new Map<string, FileInfo>([...fileList, ...migrationFiles]);
+  };
+
+  // migration: create qna files for old bots
+  private _createQnAFilesForOldBot = async (files: Map<string, FileInfo>) => {
```
srinaath, line 738: I like that this work has been moved to the server to create the qna files for the bot. The code that was replaced in the DesignPage created a new qna file when a new dialog file is created on the client. It looks like this will only handle existing dialogs of the bot.

Composer/packages/server/src/models/utilities/util.ts
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+export const getDialogFileName = (file: string) => {
+  if (file.endsWith('.dialog')) {
```
srinaath, line 5: nitpick

I think u can simplify this `file.match(/(.+?)(\.[^.]*$|$)/).`This way u can get the extension and the name of the file in 1 line rather than endsWith and another substring function

Composer/packages/server/src/models/bot/botProject.ts
srinaath: The qna files for the new dialogs will are handled in dispatcher/dialogs.

srinaath (dismissed):

Composer/packages/server/src/models/bot/botProject.ts
liweitian: 216db8343122f5dbac554463c1fde8c
In dispatchers/dialog.ts line 78. It already handled it.

Composer/packages/server/src/models/utilities/util.ts
liweitian: The dialog file is a.dialog, while qna file is a.en-us.qna. The locale cannot be removed. So I change it in another way.

boydc2014 (approved):

PR #4171, opened 2020-09-17
fix: Typing too fast in composer results in dropped characters

### LINKED ISSUE #4102: Typing too fast in composer results in dropped characters
## Describe the bug

I'm on windows and I found that typing in the expression editors is both laggy and buggy. Below is a screenshot of an expression I was just trying to update.  I deleted the characters to the right of "taskId" leaving the string `{"taskId": }` in the editor. I then typed in `skill.tasks.cardTaskId` but ended up with the string below. Every time I type in a reasonably long property path or expression I have to double check it because its either dropped a character or done something wacky like below.

image

## Version

1.1.1

lei9444:
### Root Cause

Our simplified component structure is like
   L1: TextField's parent (property editor and\or it's parent)
      L2: TextField (hold a local state, uncontrolled value)
         L3: Native browser's input which also have a value
 
A normal workflow is L3's state get changed and propagated to L2 and L2 use local state to set value again to L3, and it's synced. 
In our case, L1 component is re-rendered occasionally, which is causing L2 to re-render also, and if it's timing is just at the time L3's value is changed but not being propagated to L2 yet.  the changes in L3 will lost and will be override by L2's local state and cursor is moved to the end.
 
### Solution
The first step is a quick fix for this type issue: we fork the TextField component from FluentUI and replace the uncontrolled logic with defaultValue of the native <input/>. 

For further plan: we need to avoid elements over rendering

## Task Item
closes #4102


Composer/packages/adaptive-form/src/components/fields/TextField/TextField.base.tsx
```diff
@@ -0,0 +1,615 @@
...
+  warnMutuallyExclusive,
+} from '@uifabric/utilities';
+import {
+  ITextField,
+  ITextFieldProps,
+  ITextFieldStyleProps,
+  ITextFieldStyles,
+} from 'office-ui-fabric-react/lib/TextField';
+import debounce from 'lodash/debounce';
+import isEqual from 'lodash/isEqual';
```
hatpick, line 30: Do you really need this? you are comparing two strings, right?

LouisEugeneMSFT:
I am curious, is there a specific reason for wanting to replicate all of Fabric textfield functionalities? Since we are rebuilding the component from scratch, I feel it is our opportunity to add only what we need.

My opinion would be to keep a streamlined input field, like the initial prototypes. In the meantime we can convince Fabric team to update their component / PR against their repo.

This PR feels like a weird mix since we are importing all the interfaces from office-ui-fabric-react/lib but then re-hardcoding the entire component in Composer. It is also hard to know what is forked and what is "us".

Composer/packages/adaptive-form/src/components/fields/TextField/TextField.base.tsx
```diff
@@ -0,0 +1,610 @@
...
+    };
+
+    this._delayedValidate = this._async.debounce(this._validate, this.props.deferredValidationTime);
+    this._lastValidation = 0;
+
+    this._syncData = debounce((defaultValue: string | undefined) => {
+      if (this._textElement.current && defaultValue !== undefined && defaultValue !== this._textElement.current.value) {
+        this._textElement.current.value = defaultValue;
+      }
+    }, 300);
```
LouisEugeneMSFT, line 115: This should probably come from a prop with a default value

lei9444: Hi, @LouisEugeneMSFT this is a quick fix for the issue,  we will remove this if we have a final solution. I want to keep the props the same as the FluentUI's TextField, then if we can use FluentUI again, we can replace it quickly. I have add some comments in this part of the code.

a-b-r-o-w-n:
@LouisEugeneMSFT can you look at this again today?

Composer/packages/adaptive-form/src/components/fields/TextField/TextField.base.tsx
```diff
@@ -0,0 +1,624 @@
...
+        className={this.componentClassNames.field}
+        defaultValue={this.props.defaultValue ?? ''}
+        readOnly={this.props.readOnly}
+        onBlur={this._onBlur}
+        onChange={this._onInputChange}
+        onFocus={this._onFocus}
+        onInput={this._onInputChange}
+      />
+    );
+  };
```
LouisEugeneMSFT, line 495: When the component is uncontrolled I believe you have _**defaultValue**_ 3 times: 

- Line 464 from getNativeProps
- Line 472 from valueProps 
- Line 487 from props directly

When the component is controlled you have it two times:

- Line 464 from getNativeProps
- Line 487 from props directly

When the component is uncontrolled I believe you have _**value**_ 1 time: 

- Line 464 from getNativeProps

When the component is controlled you have it two times:

- Line 464 from getNativeProps
- Line 472 from valueProps 


Could you explain why you removed the third argument of getNativeProps from the Fabric code? This 3rd argument specifies what properties to exclude:

`(props: Record<string, any>, allowedPropNames: string[] | Record<string, number>, excludedPropNames?: string[])`

Therefore it looks like both _**value**_ and _**defaultValue**_ are always passed along and valueProps does nothing.

Have you explored something like this instead?

image

Composer/packages/adaptive-form/src/components/fields/TextField/TextField.base.tsx
```diff
@@ -0,0 +1,624 @@
...
+
+  private _renderInput = (): React.ReactElement<React.HTMLAttributes<HTMLInputElement>> => {
```
LouisEugeneMSFT, line 497: Same comment as _**_renderTextArea**_

Composer/packages/adaptive-form/src/components/fields/TextField/TextField.base.tsx
lei9444: Thanks @LouisEugeneMSFT .I think you are right and I updated this part.

LouisEugeneMSFT (approved):
This looks good to me.
@a-b-r-o-w-n , any other comments?

boydc2014 (approved):

PR #4173, opened 2020-09-17
fix: Unify skill and skill setting dispatchers

### LINKED ISSUE #4172: The setSettings dispatcher should update the skills state
Since the skills state is dependent on the settings states, the setSettings dispatcher should updated the skills state when the skills related settings are updated.

tdurnford:
## Description
- Added a skill indexer that recreates the skill state when the skills related settings are updated.
- dispatchers
  - `setSettings`: Modified the dispatcher to update the `skillsState` when the settings are changed
  - `addSkill`, `updateSkill`, `removeSkill`: Updated the dispatchers to only set the skill's settings and not the `skillState`
  - `updateSkillsInSettings`: Removed in favor of `updateSkill`

## Task Item
Closes #4172


Composer/packages/client/src/components/CreateSkillModal.tsx
```diff
@@ -83,7 +83,9 @@ export const validateManifestUrl = async ({
   } else {
     try {
       setValidationState({ ...validationState, manifestUrl: ValidationState.Validating });
-      const { data } = await httpClient.post(`/projects/${projectId}/skill/check`, { url: manifestUrl });
+      const { data } = await httpClient.post(`/projects/${projectId}/skill/retrieve-skill-manifest`, {
```
srinaath, line 86: nitpick. Feel free to ignore if a lot of work.

The endpoint is a Post. It's name is retrieve-skill-manifest. Ideally this would be GET with the query param being the url encoded manifest URL.

Composer/packages/client/src/recoilModel/dispatchers/skill.ts
```diff
@@ -84,29 +63,37 @@ export const skillDispatcher = () => {
...
-      }
-
-      updateSkillState(callbackHelpers, projectId, originSkills);
+      const settings = await snapshot.getPromise(settingsState(projectId));
+      const newSettings = cloneDeep(settings);
+      setSettingState(callbackHelpers, projectId, {
+        ...newSettings,
+        skill: {
+          ...newSettings?.skill,
+          [key]: {
```
srinaath, line 89: nitpick

We have been updating the specific property in the state like this in all dispatchers. But, we actually have immer. We can use the produce function and this section of the code would be more readable without using the spread operators.

Composer/packages/client/src/components/CreateSkillModal.tsx
tdurnford: Yea, I thought about it that lol..

Composer/packages/client/src/recoilModel/dispatchers/setting.ts
```diff
@@ -4,26 +4,66 @@
...
+              },
+            });
+            return { id, content };
+          } catch (error) {
+            return { id };
+          }
+        }
+
+        const { content = {} } = skills.find(({ id: key }) => id === key) || ({} as Skill);
+
```
srinaath, line 45: No need for whitespace

Composer/packages/client/src/components/CreateSkillModal.tsx
```diff
@@ -83,7 +83,11 @@ export const validateManifestUrl = async ({
   } else {
     try {
       setValidationState({ ...validationState, manifestUrl: ValidationState.Validating });
-      const { data } = await httpClient.post(`/projects/${projectId}/skill/check`, { url: manifestUrl });
+      const { data } = await httpClient.get(`/projects/${projectId}/skill/retrieve-skill-manifest`, {
```
srinaath, line 86: Nice!!

srinaath (approved):

srinaath (dismissed):

srinaath (approved):

PR #4176, opened 2020-09-18
feature: Adding VA template custom creation UI

### LINKED ISSUE #4142: Add VA template behind feature flag in Composer

pavolum:
## Description

This PR consists of the custom creation UI portion of the VA template (template code in PR #4140 ). fixes #4142.

Feature break down:

- Moved DialogWrapper component outside of client workspace and into ui-shared workspace for cross workspace reuse
- Added virtual-assistant-creation workspace which is referenced from Client workspace and stores custom creation UI for va-core template
  - VA creation workspace is grouped with existing UI-Plugin extensions and follows same build steps
- Altered client creation flow to call virtual-assistant-creation workspace components when user select the va-core template
- Created ShellAPI entries for virtual-assistant-creation extension
- Put VA Creation UI behind env variable feature flag


Composer/packages/client/src/components/CreationFlow/CreationFlow.tsx
```diff
@@ -150,43 +159,61 @@ const CreationFlow: React.FC<CreationFlowProps> = () => {
     navigate(`./create/${data}`);
   };
 
+  const pluginConfig: PluginConfig = useMemo(() => {
```
hatpick, line 162: this can be a const outside the component code, right?

Composer/packages/ui-plugins/virtual-assistant-creation/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "@bfc/adaptive-form": "*",
+    "@uifabric/fluent-theme": "^7.1.4",
+    "@uifabric/icons": "^7.3.0",
+    "@uifabric/styling": "^7.7.4",
+    "format-message": "^6.2.3",
+    "office-ui-fabric-react": "^7.71.0",
+    "react": "16.13.0",
+    "react-dom": "16.13.0"
+  },
+  "devDependencies": {
```
hatpick, line 31: please clean up dev dependencies

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/VirtualAssistantCreationModal.tsx
```diff
@@ -0,0 +1,95 @@
...
+// -------------------- Styles -------------------- //
+
+// -------------------- BotTypeTile -------------------- //
+interface VirtualAssistantCreationModalProps
+  extends RouteComponentProps<{
+    location: string;
+  }> {
+  // handleCreateNew: (formData: any, );
+  onDismiss: () => void;
+  handleCreateNew: (formData: any, templateId: string) => Promise<void>;
```
hatpick, line 26: can the form data be typed?

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/VirtualAssistantCreationModal.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { AppContextDefaultValue } from '../models/stateModels';
+import { useShellApi } from '@bfc/extension-client';
+import { updatePersonalityQnaFile } from '../shared/util';
+import ProvisionSummaryPage from './provisionSummaryPage';
+import { RouterPaths } from '../shared/constants';
+import { navigate } from '@reach/router';
+// -------------------- Styles -------------------- //
+
+// -------------------- BotTypeTile -------------------- //
+interface VirtualAssistantCreationModalProps
```
hatpick, line 20: prefer type over interface

Composer/packages/ui-plugins/virtual-assistant-creation/src/shared/constants.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+export enum RouterPaths {
```
hatpick, line 4: this can be a simple object, why enum?

Composer/packages/ui-plugins/virtual-assistant-creation/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "../../../tsconfig.base.json",
+  "compilerOptions": {
+    "allowJs": true,
```
hatpick, line 4: why?

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/provisionSummaryPage.tsx
```diff
@@ -0,0 +1,131 @@
...
+      name: formatMessage('Notes'),
+      fieldName: 'Notes',
+      minWidth: 300,
+      maxWidth: 500,
+      isResizable: false,
+      isMultiline: true,
+    },
+  ];
+
+  const items = [
```
hatpick, line 47: Please move this outside of the component.

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/provisionSummaryPage.tsx
```diff
@@ -0,0 +1,131 @@
...
+import { DialogWrapper, DialogTypes } from '@bfc/ui-shared';
+
+import { Pivot, PivotItem } from 'office-ui-fabric-react/lib/Pivot';
+import { DialogFooterWrapper } from './dialogFooterWrapper';
+import { RouterPaths } from '../shared/constants';
+import { DetailsList, DetailsListLayoutMode, IColumn, SelectionMode } from 'office-ui-fabric-react/lib/DetailsList';
+import { TextField } from 'office-ui-fabric-react/lib/TextField';
+
+// -------------------- BotTypeTile -------------------- //
+interface ProvisionSummaryPageProps extends RouteComponentProps {
```
hatpick, line 18: prefer type over interface

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/newBotPage.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+// -------------------- NewBotPage -------------------- //
+interface NewBotPageProps extends RouteComponentProps {
+  onDismiss: () => void;
+}
+
+export const NewBotPage: React.FC<NewBotPageProps> = (props) => {
+  const { state, setState } = useContext(AppContext);
+
+  let assistantSelectionChanged = (event: any, option?: IChoiceGroupOption) => {
```
hatpick, line 25: why let?

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/newBotPage.tsx
```diff
@@ -0,0 +1,76 @@
...
+    var selectedAssistant = state.availableAssistantTemplates.find((assistant: IAssistant) => {
+      return assistant.name.toLowerCase() == option?.key.toLowerCase();
+    });
+    if (selectedAssistant) {
+      setState({ ...state, selectedAssistant: selectedAssistant });
+    }
+  };
+
+  let getAssistantsToRender = (): IChoiceGroupOption[] => {
```
hatpick, line 34: why let?

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/newBotPage.tsx
```diff
@@ -0,0 +1,76 @@
...
+    var result: IChoiceGroupOption[] = [];
+    state.availableAssistantTemplates.forEach((assistant: IAssistant) => {
+      result.push({
+        key: assistant.name,
+        text: assistant.name + ' : ' + assistant.description,
```
hatpick, line 39: prefer: `${assistant.name} : ${assistant.description}`

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/newBotPage.tsx
```diff
@@ -0,0 +1,76 @@
...
+      });
+    });
+    return result;
+  };
+
+  return (
+    <Fragment>
```
hatpick, line 46: fragment is not needed

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/dialogFooterWrapper.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const navForwardOrSubmit = () => {
+    if (onSubmit && !nextPath) {
+      onSubmit();
+    } else if (nextPath) {
+      navigate(nextPath);
+    }
+  };
+
+  return (
+    <Fragment>
```
hatpick, line 41: fragment is not needed

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/customizeBotPage.tsx
```diff
@@ -0,0 +1,130 @@
...
+  const { state, setState } = useContext(AppContext);
+
+  const personalityOptions = (): IDropdownOption[] => {
+    return Object.values(AvailablePersonalities).map((personality) => {
+      return { key: personality, text: formatMessage(personality) };
+    });
+  };
+
+  return (
+    <Fragment>
```
hatpick, line 48: fragment is not needed

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/configSummaryPage.tsx
```diff
@@ -0,0 +1,87 @@
+// Copyright (c) Microsoft Corporation.
```
hatpick, line 1: prefer type over interface

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/VirtualAssistantCreationModal.tsx
```diff
@@ -0,0 +1,95 @@
...
+    return '- ' + text;
+  };
+
+  const onModalDismiss = () => {
+    navigate('./');
+    onDismiss();
+  };
+
+  return (
+    <Fragment>
```
hatpick, line 78: fragment is not needed

Composer/packages/ui-plugins/virtual-assistant-creation/src/shared/util.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import axios from 'axios';
+import { ShellApi } from '@bfc/shared';
+import { AvailablePersonalities } from '../models/creationOptions';
+
+// GitHub API Utility Functions
+export const updatePersonalityQnaFile = async (shellApi: ShellApi, selectedPersonality: AvailablePersonalities) => {
+  const GitHubRepoApiRootUrl =
+    'https://api.github.com/repos/microsoft/BotBuilder-PersonalityChat/contents/CSharp/Datasets/qnaFormat/';
+  const personalityFileName = 'qna_chitchat_' + selectedPersonality + '.qna';
```
hatpick, line 12: prefer ```const personalityFileName = `qna_chitchat_${selectedPersonality}.qna';```

Composer/packages/ui-plugins/virtual-assistant-creation/src/shared/util.ts
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import axios from 'axios';
+import { ShellApi } from '@bfc/shared';
+import { AvailablePersonalities } from '../models/creationOptions';
+
+// GitHub API Utility Functions
+export const updatePersonalityQnaFile = async (shellApi: ShellApi, selectedPersonality: AvailablePersonalities) => {
+  const GitHubRepoApiRootUrl =
```
hatpick, line 10: please move this out of the function, it's a constant.

Composer/packages/ui-plugins/virtual-assistant-creation/src/shared/util.ts
```diff
@@ -0,0 +1,26 @@
...
+    'https://api.github.com/repos/microsoft/BotBuilder-PersonalityChat/contents/CSharp/Datasets/qnaFormat/';
+  const personalityFileName = 'qna_chitchat_' + selectedPersonality + '.qna';
+
+  // TODO: handle different markets once creation supports multi-local
+  const endpoint = GitHubRepoApiRootUrl + 'english/' + personalityFileName;
```
hatpick, line 15: prefer ```const endpoint = `${GitHubRepoApiRootUrl}english/${personalityFileName}```;

Composer/packages/ui-plugins/virtual-assistant-creation/src/shared/util.ts
```diff
@@ -0,0 +1,26 @@
...
+  axios({
+    method: 'get',
+    url: endpoint,
+    responseType: 'json',
+  }).then(async function (response) {
+    console.log(response);
```
hatpick, line 21: remove

Composer/packages/ui-plugins/virtual-assistant-creation/src/shared/util.ts
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) Microsoft Corporation.
```
hatpick, line 1: What happens if the fetch call fail?

Composer/packages/client/src/components/CreationFlow/CreationFlow.tsx
```diff
@@ -129,6 +133,11 @@ const CreationFlow: React.FC<CreationFlowProps> = () => {
       navigate(`./QnASample/importQnA`);
       return;
     }
+    if (templateId === 'va-core') {
```
hatpick, line 136: now that this function is not only handling the QnA, please rename the function accordingly.

hatpick: Is there a particular reason that the QnA route and templateId values are campleCase but va-core is hyphen-case?

hatpick (changes requested):

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/VirtualAssistantCreationModal.tsx
a-b-r-o-w-n: Is there an eslint rule we could use for this?

hatpick: @a-b-r-o-w-n This is a candidate, https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/consistent-type-definitions.md
The ```prefer-interface``` (discussion: https://github.com/typescript-eslint/typescript-eslint/issues/433) was originally added because types were not very powerful at the beginning.

Composer/packages/ui-plugins/virtual-assistant-creation/src/shared/constants.ts
hatpick: @a-b-r-o-w-n 
```
{
  "rules": {
    "no-restricted-syntax": [
      "error",
      {
        "selector": "TSEnumDeclaration",
        "message": "Enums are ugh!"
      }
    ]
  }
}
```

a-b-r-o-w-n: Enums are fine to me, especially for shared keys. In this case, I think an object is fine though.

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/provisionSummaryPage.tsx
a-b-r-o-w-n: Note that the `formatMessage` calls will need to be invoked at render time. It's fine to move this out of the component, but those properties need to be functions instead. ex: `() => formatMessage('...')`

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/provisionSummaryPage.tsx
```diff
@@ -0,0 +1,131 @@
...
+      isMultiline: true,
+    },
+  ];
+
+  const items = [
+    {
+      key: '0',
+      Resource: formatMessage('Azure Bot Service'),
+      Notes:
+        'The Azure Bot Service resource stores configuration information that allows your Virtual Assistant to be accessed on the supported Channels and provide OAuth authentication.',
```
a-b-r-o-w-n, line 52: `formatMessage`

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/provisionSummaryPage.tsx
```diff
@@ -0,0 +1,131 @@
...
+import formatMessage from 'format-message';
+import React, { Fragment } from 'react';
+import { RouteComponentProps } from '@reach/router';
+import { DialogWrapper, DialogTypes } from '@bfc/ui-shared';
+
+import { Pivot, PivotItem } from 'office-ui-fabric-react/lib/Pivot';
+import { DialogFooterWrapper } from './dialogFooterWrapper';
+import { RouterPaths } from '../shared/constants';
+import { DetailsList, DetailsListLayoutMode, IColumn, SelectionMode } from 'office-ui-fabric-react/lib/DetailsList';
+import { TextField } from 'office-ui-fabric-react/lib/TextField';
```
a-b-r-o-w-n, line 15: This should have been caught by the linter, but the import order here is incorrect.

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/provisionSummaryPage.tsx
a-b-r-o-w-n: Can a type extend an interface? Or do you need to define an intersection type?

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/newBotPage.tsx
```diff
@@ -0,0 +1,76 @@
...
+// -------------------- NewBotPage -------------------- //
+interface NewBotPageProps extends RouteComponentProps {
+  onDismiss: () => void;
+}
+
+export const NewBotPage: React.FC<NewBotPageProps> = (props) => {
+  const { state, setState } = useContext(AppContext);
+
+  let assistantSelectionChanged = (event: any, option?: IChoiceGroupOption) => {
+    var selectedAssistant = state.availableAssistantTemplates.find((assistant: IAssistant) => {
```
a-b-r-o-w-n, line 26: no `var`.

a-b-r-o-w-n (commented):
It doesn't look like you have your editor to apply formatting on save. A lot of lint errors I've seen can automatically be fixed when saving.

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/provisionSummaryPage.tsx
hatpick: RouteComponentProps itself is a type that accepts a T for props: RouteComponentProps<TProps>

Composer/packages/client/src/components/CreationFlow/CreationFlow.tsx
pavolum: I believe since it uses the `useMemo` hook it has to be declared in the functional component

Composer/packages/ui-plugins/virtual-assistant-creation/tsconfig.json
pavolum: removed

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/provisionSummaryPage.tsx
pavolum: I moved `items` outside of the component and turned it into a function called `provisioningDetailItems` which returns the array of content with strings leveraging the formatMessage function.

Composer/packages/ui-plugins/virtual-assistant-creation/src/shared/util.ts
pavolum: I added a console log if there is an error on the fetch, are there any additional recommendations? Note the long term solution will probably not have this GitHub API call as we determine the long term hosting plan for these personality .qna files.

Composer/packages/client/src/components/CreationFlow/CreationFlow.tsx
pavolum: Fixed function naming and route syntax

Composer/packages/ui-plugins/virtual-assistant-creation/.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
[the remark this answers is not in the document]
pavolum: I have removed this change and added it to be gitignored.

Composer/packages/ui-plugins/virtual-assistant-creation/.eslintrc.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  extends: ['../../../.eslintrc.react.js'],
+  parserOptions: {
+    project: './tsconfig.json',
+    tsconfigRootDir: __dirname,
+  },
+  rules: {
+    noUnusedVars: 'false',
```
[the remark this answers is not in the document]
pavolum: Accidentally left from testing, I have removed the rule and fixed resulting errors.

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/provisionSummaryPage.tsx
pavolum: I have converted all props interfaces to types using the method @hatpick outlined. VirtualAssistantCreationModalProps example:
```
type VirtualAssistantCreationModalProps = RouteComponentProps<{
  onDismiss: () => void;
  handleCreateNew: (formData: formData, templateId: string) => Promise<void>;
  formData: formData;
}>;
```

However `RouteComponentProps` is set up in such a way that the partial props values passed to it have a type of `DECLAREDTYPE | undefined`. 

Which resulted in having to add the following undefined checking code at the start of each functional component: 
```
 const { onDismiss, handleCreateNew, formData } = props;
  if (onDismiss === undefined || handleCreateNew === undefined || formData === undefined) {
    console.log('invalid props passed to VirtualAssistantCreationModal');
    return null;
  }
```
This does not happen with the previous interface definition. So converting these RouteComponentProps interfaces to types does add a bit of additional null checking code. This has nothing to do with the type conversion but more so for types that "extend" RouteComponetProps. 

- Should I keep this as types regardless (change is currently in latest commit in this branch)?
- Alternatively I can go back to using interfaces that extend `RouteComponetProps`?
- Or get rid of reach/router in this workspace all together and handle navigation state in appState, this would remove the need to have props extend `RouteComponetProps` and may be something we want to do regardless of this issue.

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/provisionSummaryPage.tsx
pavolum: Locally I have it formatting on save but it applies 'formatting' based on the TS standard, not based on the lint rules in the Composer/eslintrc.js. Im not sure how to set it up to use the linting rules from Composer/eslintrc.js, let alone how to auto fix linting errors on format. Any tips on setting this up? This workspace follows the linting approach of other UI plugins so as I understand it all of them would have this issue.

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/provisionSummaryPage.tsx
hatpick: @pavolum that's a weird thing they do, I suggest keeping it as type (in oppose to interface) but do an intersection type:
```
type MyProps = {
  onDismiss: () => void;
  handleCreateNew: (formData: formData, templateId: string) => Promise<void>;
  formData: formData;
} & ReactComponentProps<{}>;
```

Composer/packages/ui-plugins/virtual-assistant-creation/src/shared/util.ts
hatpick: No matter how reliable the backend, the call still can fail, like a slow internet connection can cause a timeout or etc.

hatpick (changes requested):
I've answered your follow up questions. Please make the changes accordingly, they are minor changes. 😃

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/provisionSummaryPage.tsx
pavolum: Great, thanks for the tip. I have implemented this across the repo with the latest iteration.

Composer/packages/client/src/components/CreationFlow/CreationFlow.tsx
```diff
@@ -113,7 +113,7 @@ const CreationFlow: React.FC<CreationFlowProps> = () => {
   };
 
   const handleCreateNew = async (formData, templateId: string, qnaKbUrls?: string[]) => {
-    createProject(
+    await createProject(
```
pavolum, line 116: @srinaath I notice you removed this 'await' when you submitted a PR a few days ago. My code relies on this call being awaited. What was the reason you removed the 'await' previously? I can refactor my code to account for this call not being awaited but it would be a rather large refactor.

As I see it, this should not be an issue if it is awaited. the createProject function handles triggering UI loading state, then it does all the work to create the project and then triggers UI state changes so the user is in the editing view.

In a non VA create flow, If it as awaited it will do all these, then return to this function and end execution, I could see it being slightly more optimal to not await but it creates an issue for my implementation. My code calls await handleCreateNew which calls createProject (assuming it is awaited) and then calls code that alters the bot files that it assumes are now created. If we dont await this call then my code runs bot asset altering code prior to the bot assets being stored in memory. 

Options
1. Revert this call to be awaited as it was prior to your change
2. Conditionally await depending on template

srinaath: So the way all the dispatcher functions need to work is that they should not be awaited inside the component. The way we built it is that it would be a "fire and forget" approach. We call the dispatcher, it does some updates to the state. We need to let the component subscribe to the change and do something in the useEffect of the change.

pavolum: That makes a lot of sense, many thanks for the tip and explanation. I will make the change now :)

Composer/packages/ui-plugins/virtual-assistant-creation/package.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "@bfc/ui-plugin-va-creation",
+  "version": "0.0.0",
```
GeoffCoxMSFT, line 3: Consider 0.0.1

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/VirtualAssistantCreationModal.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const { shellApi, ...shellData } = useShellApi();
+  const { projectId } = shellData;
+  const [initialProjectId] = useState(projectId);
+
+  const createAndConfigureBot = async () => {
+    handleCreateNew(formData, 'vaCore');
+  };
+
+  const updateBotResponses = async () => {
+    const generatedLgFileId = formData.name.toLowerCase() + '.en-us';
```
GeoffCoxMSFT, line 42: Hard-coding the locale?  Probably deserves a comment about why.

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/VirtualAssistantCreationModal.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { NewBotPage } from './newBotPage';
+import { CustomizeBotPage } from './customizeBotPage';
+import { ConfigSummaryPage } from './configSummaryPage';
+import ProvisionSummaryPage from './provisionSummaryPage';
+
+// -------------------- VirtualAssistantCreationModal -------------------- //
+type VirtualAssistantCreationModalProps = {
+  onDismiss: () => void;
+  handleCreateNew: (formData: formData, templateId: string) => Promise<void>;
+  formData: formData;
```
GeoffCoxMSFT, line 25: Why is the formData type camelCased rather than PascalCased?

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/VirtualAssistantCreationModal.tsx
```diff
@@ -0,0 +1,99 @@
...
+} & RouteComponentProps<{}>;
+
+export const AppContext = React.createContext(AppContextDefaultValue);
+
+export const VirtualAssistantCreationModal: React.FC<VirtualAssistantCreationModalProps> = (props) => {
+  const { onDismiss, handleCreateNew, formData } = props;
+  const [state, setState] = useState(AppContextDefaultValue.state);
+  const { shellApi, ...shellData } = useShellApi();
+  const { projectId } = shellData;
```
GeoffCoxMSFT, line 34: Doesn't spreading the shellData on the previous line already define projectId?

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/VirtualAssistantCreationModal.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const [initialProjectId] = useState(projectId);
+
+  const createAndConfigureBot = async () => {
+    handleCreateNew(formData, 'vaCore');
+  };
+
+  const updateBotResponses = async () => {
+    const generatedLgFileId = formData.name.toLowerCase() + '.en-us';
+    await shellApi.updateLgTemplate(
```
GeoffCoxMSFT, line 43: Could these template updates be done in parallel (await all) ? or must they be sequenced?

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/VirtualAssistantCreationModal.tsx
```diff
@@ -0,0 +1,99 @@
...
+      createLgFromString(state.selectedFallbackText)
+    );
+  };
+
+  const updateQnaFiles = async () => {
+    await updatePersonalityQnaFile(shellApi, state.selectedPersonality);
+  };
+
+  const createLgFromString = (text: string) => {
+    return '- ' + text;
```
GeoffCoxMSFT, line 66: Consider `- {text}`
Does this really need to be an independent function?

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/VirtualAssistantCreationModal.tsx
```diff
@@ -0,0 +1,99 @@
...
+          path={RouterPaths.provisionSummaryPage}
+          onDismiss={onModalDismiss}
+          onSubmit={createAndConfigureBot}
+        />
+      </Router>
+    </AppContext.Provider>
+  );
+};
+
+export default VirtualAssistantCreationModal;
```
GeoffCoxMSFT, line 99: Consider removing this.  You already exported it.  Plus, we don't need to export default in every file as we should be doing types imports.

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/configSummaryPage.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { DialogFooterWrapper } from './dialogFooterWrapper';
+
+// -------------------- Styles -------------------- //
+
+const categoryTextStyling = css`
+  display: block;
+  margin-top: 10px;
+`;
+
+const entryText = (bold: boolean) => css`
```
GeoffCoxMSFT, line 26: Prefer suffix of Style (or Css) for any css variable and then using a name for any styled component.

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/configSummaryPage.tsx
```diff
@@ -0,0 +1,89 @@
...
+const separatorStyles: Partial<ISeparatorStyles> = {
+  root: { color: 'black' },
+};
+
+// -------------------- ConfigSummaryPage -------------------- //
+type ConfigSummaryPageProps = {
+  onDismiss: () => void;
+} & RouteComponentProps<{}>;
+
+export const ConfigSummaryPage: React.FC<ConfigSummaryPageProps> = (props) => {
```
GeoffCoxMSFT, line 40: You don't need to declare the type this way, instead put the props type on the props parameter.

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/configSummaryPage.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const { state } = useContext(AppContext);
+  const onDismiss = props.onDismiss;
+
+  const categoryText = (text: string) => {
```
GeoffCoxMSFT, line 44: Prefer prefix of render (e.g renderCategoryText) to avoid confusion between data variable and rendering helper functions.

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/customizeBotPage.tsx
```diff
@@ -0,0 +1,130 @@
+/* eslint-disable format-message/literal-pattern */
```
GeoffCoxMSFT, line 1: Name of file should be CustomizeBotPage.tsx (PascalCase for React components)

GeoffCoxMSFT, line 1: Move eslint disable after copyright.

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/customizeBotPage.tsx
```diff
@@ -0,0 +1,130 @@
...
+import { Checkbox } from 'office-ui-fabric-react/lib/Checkbox';
+import { Stack } from 'office-ui-fabric-react/lib/Stack';
+
+import { RouterPaths } from '../shared/constants';
+import { AvailablePersonalities } from '../models/creationOptions';
+
+import { AppContext } from './VirtualAssistantCreationModal';
+import { DialogFooterWrapper } from './dialogFooterWrapper';
+
+// -------------------- Styles -------------------- //
```
GeoffCoxMSFT, line 23: Thanks for putting styles inline! (instead of styles.ts)

Composer/packages/ui-plugins/virtual-assistant-creation/src/models/creationOptions.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+export enum AvailablePersonalities {
```
GeoffCoxMSFT, line 4: Prefer string unions to enums
@hatpick - what are the specific problem with enums again?

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/customizeBotPage.tsx
```diff
@@ -0,0 +1,130 @@
...
+  onDismiss: () => void;
+} & RouteComponentProps<{}>;
+
+export const CustomizeBotPage: React.FC<CustomizeBotPageProps> = (props) => {
+  const { state, setState } = useContext(AppContext);
+  const onDismiss = props.onDismiss;
+
+  const personalityOptions = (): IDropdownOption[] => {
+    return Object.values(AvailablePersonalities).map((personality) => {
+      return { key: personality, text: formatMessage(personality) };
```
GeoffCoxMSFT, line 46: I'm uncertain if this will end up putting each personality into the en-us.json file.
@beyackle - does formatMessage handle runtime calls in a loop to formatMessage?

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/customizeBotPage.tsx
```diff
@@ -0,0 +1,130 @@
...
+    });
+  };
+
+  return (
+    <DialogWrapper
+      isOpen
+      dialogType={DialogTypes.CreateFlow}
+      subText={formatMessage(
+        'Give your bot personality, multi-language capabilities and more. You can edit this later in Bot Settings.'
```
GeoffCoxMSFT, line 55: Nit: Oxford comma

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/customizeBotPage.tsx
```diff
@@ -0,0 +1,130 @@
...
+        </Stack.Item>
+        <Stack.Item>
+          <Dropdown
+            label={formatMessage('Personality')}
+            options={personalityOptions()}
+            placeholder={formatMessage('Select an option')}
+            selectedKey={state.selectedPersonality}
+            styles={dropdownStyles}
+            onChange={(ev: any, option?: IDropdownOption) => {
+              setState({ ...state, selectedPersonality: AvailablePersonalities[option?.text as string] });
```
GeoffCoxMSFT, line 98: Wouldn't option.key be the index into the personalities?

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/customizeBotPage.tsx
```diff
@@ -0,0 +1,130 @@
...
+// -------------------- CustomizeBotPage -------------------- //
+type CustomizeBotPageProps = {
+  onDismiss: () => void;
+} & RouteComponentProps<{}>;
+
+export const CustomizeBotPage: React.FC<CustomizeBotPageProps> = (props) => {
+  const { state, setState } = useContext(AppContext);
+  const onDismiss = props.onDismiss;
+
+  const personalityOptions = (): IDropdownOption[] => {
```
GeoffCoxMSFT, line 44: Does this need to be a function called with every render?

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/customizeBotPage.tsx
```diff
@@ -0,0 +1,130 @@
...
+  dropdown: { width: 300 },
+};
+
+// -------------------- CustomizeBotPage -------------------- //
+type CustomizeBotPageProps = {
+  onDismiss: () => void;
+} & RouteComponentProps<{}>;
+
+export const CustomizeBotPage: React.FC<CustomizeBotPageProps> = (props) => {
+  const { state, setState } = useContext(AppContext);
```
GeoffCoxMSFT, line 41: Are you using the single large context object rather than the recoil state here?
@a-b-r-o-w-n How should UI plug-ins wire plug-in wide state or integrate with app state?

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/customizeBotPage.tsx
```diff
@@ -0,0 +1,130 @@
...
+            }}
+          />
+        </Stack.Item>
+        <Stack.Item>
+          <Label>{formatMessage('Greeting Message')}</Label>
+          <TextField
+            css={textFieldStyling}
+            value={formatMessage(state.selectedGreetingMessage)}
+            onChange={(ev?: any, newValue?: string) => {
+              setState({ ...state, selectedGreetingMessage: newValue as string });
```
GeoffCoxMSFT, line 108: With this approach, every state change will update AppContext which will re-render every component that uses AppContext.

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/index.ts
```diff
@@ -0,0 +1,3 @@
+// Copyright (c) Microsoft Corporation.
```
GeoffCoxMSFT, line 1: Avoid barrel files (other index.ts exports) other than the top level one for the library.

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/newBotPage.tsx
```diff
@@ -0,0 +1,77 @@
+// Copyright (c) Microsoft Corporation.
```
GeoffCoxMSFT, line 1: File name shoudl be NewBotPage.tsx

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/newBotPage.tsx
```diff
@@ -0,0 +1,77 @@
...
+      <Label>{formatMessage('Choose one:')}</Label>
+      <ChoiceGroup
+        required
+        defaultSelectedKey={state.selectedAssistant.name}
+        options={getAssistantsToRender()}
+        styles={{
+          root: {
+            width: '100%',
+          },
+        }}
```
GeoffCoxMSFT, line 65: Consider defining a const styles so that it doesn't instantiate a new object every render.

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/provisionSummaryPage.tsx
```diff
@@ -0,0 +1,136 @@
...
+import { Pivot, PivotItem } from 'office-ui-fabric-react/lib/Pivot';
+import { DetailsList, DetailsListLayoutMode, IColumn, SelectionMode } from 'office-ui-fabric-react/lib/DetailsList';
+import { TextField } from 'office-ui-fabric-react/lib/TextField';
+
+import { RouterPaths } from '../shared/constants';
+
+import { DialogFooterWrapper } from './dialogFooterWrapper';
+
+// Content Data
+const provisioningDetailItems = () => {
```
GeoffCoxMSFT, line 19: Does this need to be function or can it be a const value?

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/provisionSummaryPage.tsx
```diff
@@ -0,0 +1,136 @@
...
+      Resource: formatMessage('QnA Maker Azure Search Service'),
+      Notes: formatMessage('Search index for your QnA Maker knowledgebases.'),
+    },
+  ];
+};
+
+const columns: IColumn[] = [
+  {
+    key: 'column1',
+    name: formatMessage('Resource'),
```
GeoffCoxMSFT, line 81: Here is a case where columns is a const value rather than a function, but it calls formatMessage.  We should be consistent depending on what Ben Y says above.

Composer/packages/ui-plugins/virtual-assistant-creation/src/fileTypes.d.ts
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation.
```
GeoffCoxMSFT, line 1: Could this not be done a .ts file and have the compiler generate the d.ts?

Composer/packages/ui-plugins/virtual-assistant-creation/src/models/stateModels.ts
```diff
@@ -0,0 +1,98 @@
...
+  name: string;
+  description: string;
+  location?: string;
+}
+
+export const initialAppState: IAppState = {
+  availableHostedSkills: [
+    {
+      name: 'POI',
+      description: 'Find points of interest and directions',
```
GeoffCoxMSFT, line 46: Should these have formatMessage?

Composer/packages/ui-plugins/virtual-assistant-creation/src/shared/util.ts
```diff
@@ -0,0 +1,30 @@
...
+// Licensed under the MIT License.
+
+import axios from 'axios';
+import { ShellApi } from '@bfc/shared';
+
+import { AvailablePersonalities } from '../models/creationOptions';
+
+// Utility constants
+const GitHubRepoApiRootUrl =
+  'https://api.github.com/repos/microsoft/BotBuilder-PersonalityChat/contents/CSharp/Datasets/qnaFormat/';
```
GeoffCoxMSFT, line 11: @cwhitten - Should we use aka.ms links to avoid hard-coding any specific URLs into Composer? In the case it changes, we could then update the URL on aka.ms and not have to redeploy and upgrade everyone.  What do you think?

Composer/packages/ui-plugins/virtual-assistant-creation/src/shared/util.ts
```diff
@@ -0,0 +1,30 @@
...
+
+// GitHub API Utility Functions
+export const updatePersonalityQnaFile = async (shellApi: ShellApi, selectedPersonality: AvailablePersonalities) => {
+  const personalityFileName = `qna_chitchat_${selectedPersonality}.qna`;
+  // TODO: handle different markets once creation supports multi-local
```
GeoffCoxMSFT, line 16: If you want to check-in with a TODO, consider creating a GitHub issue and referencing it here.

Composer/packages/ui-plugins/virtual-assistant-creation/src/shared/util.ts
```diff
@@ -0,0 +1,30 @@
...
+    method: 'get',
+    url: endpoint,
+    responseType: 'json',
+  })
+    .then(async (response) => {
+      await shellApi.updateQnaContent('Chit-chat.en-us', atob(response.data.content));
+      return response;
+    })
+    .catch((err) => {
+      console.error('Error occurred while grabbing personality chat files from GitHub ', err);
```
GeoffCoxMSFT, line 28: @a-b-r-o-w-n How should plug-ins surface errors up to the top level UI?

Composer/yarn.lock
```diff
@@ -3150,6 +3150,16 @@
     react-lifecycles-compat "^3.0.4"
     warning "^3.0.0"
 
+"@reach/router@^1.3.4":
```
GeoffCoxMSFT, line 3153: Are we changing our routing to use this over react-router?

Composer/packages/ui-plugins/virtual-assistant-creation/src/shared/types.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Microsoft Corporation.
```
GeoffCoxMSFT, line 1: You don't need to create a shared folder, consider putting types and utils directly under src/

Composer/packages/ui-plugins/virtual-assistant-creation/src/shared/types.ts
```diff
@@ -0,0 +1,8 @@
...
+// Licensed under the MIT License.
+
+export type formData = {
```
GeoffCoxMSFT, line 4: FormData

GeoffCoxMSFT (changes requested):
I have some non-nits  that should be fixed before approval as well as some open questions.

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/configSummaryPage.tsx
hatpick: @GeoffCoxMSFT  RouteComponentProps makes the passed T partial, that causes lots of undefined checking

Composer/packages/ui-plugins/virtual-assistant-creation/src/models/creationOptions.ts
hatpick: lol, 1. union types makes debugging much easier, 2. union types dont need importing.

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/customizeBotPage.tsx
pavolum: I believe we want to limit the visibility a plug in has to recoil state via the shellAPI so plugins should not be able to access recoil state directly. 
The UI plug in for creation implements its own state management to handle state within its experience as any UI plug in would. I believe client should largely not care what a ui plugin uses for its own state management as long as it follows a strict contract when interacting with client state via the ShellAPI. @a-b-r-o-w-n  is that correct? 

Granted the state management I have chosen is resulting in unnecessary rerenders as your next comment indicates so I will fix that.

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/provisionSummaryPage.tsx
pavolum: changed to const

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/provisionSummaryPage.tsx
pavolum: Okay I have changed above to const, will wait on what Ben says to determine if they should stay that way.

Composer/packages/ui-plugins/virtual-assistant-creation/src/fileTypes.d.ts
pavolum: unused file, removing

Composer/packages/ui-plugins/virtual-assistant-creation/src/shared/util.ts
pavolum: Yeah that would make sense. If its any consolation this hosting method/url is not the long term hosting plan for the personality chat assets. As we move to the asset library model the hosting and storage of these personality .qna files will change. 
This initial implementation works and assumes a hosted github scenario but they will change on the next iteration so I dont think it matters that it is not behind an aka.ms link.

Composer/packages/ui-plugins/virtual-assistant-creation/src/shared/util.ts
pavolum: Changed the comment from a todo to a more descriptive comment

Composer/yarn.lock
pavolum: When I started working in the Composer code base the set router was already Reach/router. 

When choosing a routing library to help with navigation in my UI plug in I chose reach/router to stay consistent with the routing libs used in the larger codebase.

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/customizeBotPage.tsx
pavolum: I was wondering if this was the case. I initially had this implemented with redux when I was working on it outside composer and refactored it with hooks when moving it into Composer. I now realize with the refactor that it is calling excessive rerenders, good catch!

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/configSummaryPage.tsx
GeoffCoxMSFT: @hatpick - so (props : ConfigSummaryPageProps) => is going to end up being different than React.FC<ConfigSummaryPageProps> ?  I don't understand what you mean by undefined checking - runtime or compile time?

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/customizeBotPage.tsx
a-b-r-o-w-n: It does not. `formatMessage` must be called with a string literal.

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/customizeBotPage.tsx
a-b-r-o-w-n: I agree with @pavolum. Plugins will be able to access state through the shell api. Composer internal state management should not leak into plugins.

And I also agree that plugins can bring whatever state management library they prefer.

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/provisionSummaryPage.tsx
a-b-r-o-w-n: This needs to be wrapped in a function in order for `formatMessage` to pick up the current locale.

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/provisionSummaryPage.tsx
a-b-r-o-w-n: See my comment above.

Composer/packages/client/public/react-bundle.js.LICENSE.txt
```diff
@@ -0,0 +1,14 @@
+/*
```
a-b-r-o-w-n, line 1: Do we know why these files are being generated? @tonyanziano  any ideas?

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/configSummaryPage.tsx
```diff
@@ -0,0 +1,89 @@
+/* eslint-disable format-message/literal-pattern */
```
a-b-r-o-w-n, line 1: Can you move this below the copyright?

Composer/packages/ui-plugins/virtual-assistant-creation/src/models/stateModels.ts
```diff
@@ -0,0 +1,98 @@
...
+  isTextEnabled: true,
+  isSpeechEnabled: false,
+  selectedLanguages: [],
+  selectedPersonality: 'professional',
+  selectedWelcomeImage: '',
+  selectedFallbackText: formatMessage("I am sorry, I didn't understand that"),
+  selectedGreetingMessage: formatMessage('Hi there! Here are some things that I can do!'),
+  selectedSkills: [],
+  selectedQnaFile: '',
+};
```
a-b-r-o-w-n, line 93: All of these `formatMessage` calls will not get localized since they are invoked when the module is loaded.

Composer/packages/client/public/react-bundle.js.LICENSE.txt
tonyanziano: I noticed this on my branch as well when I rebased onto yesterday's main. I suspect there was a recent PR that went in that is somehow generating this.

a-b-r-o-w-n: I've ignored it in another PR, but couldn't figure out why there were being generated.

tonyanziano: [This is the only thing I could find that might be generating it](https://webpack.js.org/plugins/terser-webpack-plugin/#extractcomments) as we do use the `Terser` webpack plugin, but I don't see this option configured, nor do I see any recent changes to the webpack configs.

beyackle: I've been getting these files cropping up too. I don't know why either.

a-b-r-o-w-n: Might be due to this: https://github.com/microsoft/BotFramework-Composer/pull/4315/files#diff-a1975c6de2cf5658b3b91350573c8791R16

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/provisionSummaryPage.tsx
pavolum: Changed to function :D

Composer/packages/client/public/react-bundle.js.LICENSE.txt
tonyanziano: I have confirmed that just running `yarn build:react-bundles` in the `/client/` package creates these files, but the webpack configs are [incredibly minimalistic](https://github.com/microsoft/BotFramework-Composer/blob/main/Composer/packages/client/config/webpack-react-dom.config.js) with [no plugins](https://github.com/microsoft/BotFramework-Composer/blob/main/Composer/packages/client/config/webpack-react.config.js) so I'm a little bit stumped.

a-b-r-o-w-n: I've found the issue: https://webpack.js.org/plugins/terser-webpack-plugin/#extractcomments.

I'll fix in #4336

pavolum: I have removed them from the PR for the time being

Composer/packages/ui-plugins/virtual-assistant-creation/package.json
pavolum: I was following the standard set in the other ui plugin workspaces. Is there a functional reason they are all 0.0.0? When these get referenced in client package.json they have the "*" version indicator

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/VirtualAssistantCreationModal.tsx
pavolum: Added a comment

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/VirtualAssistantCreationModal.tsx
pavolum: These particular calls manipulate the same .lg file via state so they have to be awaited to prevent clashing when writing to that file. 
This can be done in parallel to the .qna file creation since it is a separate file so I have made that change.

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/VirtualAssistantCreationModal.tsx
pavolum: Made the syntax change and added additional string checking logic that I believe justifies the function

Composer/packages/ui-plugins/virtual-assistant-creation/src/components/customizeBotPage.tsx
pavolum: Changed the syntax to support variable string

GeoffCoxMSFT (approved):

GeoffCoxMSFT (approved):

hatpick (approved):

PR #4206, opened 2020-09-18
feature: add CSV creation to locale scripts

beyackle:
## Description

There was a request from people working on our localization to provide the localization files in a tabular text form, so this script adds that transformation. It also adds the CSV files to our .gitignore.

## Task Item

#minor


hatpick (commented):
Please add the same for the electron-server package.

Composer/package.json
```diff
@@ -72,14 +72,16 @@
...
     "l10n:transform": "node scripts/l10n-transform.js",
     "l10n:babel": "babel ./packages --extensions \".ts,.tsx,.jsx,.js\" --out-dir l10ntemp --presets=@babel/react,@babel/typescript --plugins=@babel/plugin-proposal-class-properties --ignore \"packages/electron-server\",\"packages/**/__tests__\",\"packages/**/node_modules\",\"packages/**/build/**/*.js\"",
-    "l10n": "yarn l10n:babel && yarn l10n:extract && yarn l10n:transform packages/server/src/locales/en-US.json && yarn l10n:extractJson packages/server/schemas"
+    "l10n": "yarn l10n:babel && yarn l10n:extract && yarn l10n:makeTSV && yarn l10n:transform packages/server/src/locales/en-US.json && yarn l10n:extractJson packages/server/schemas"
   },
   "husky": {
     "hooks": {
-      "pre-commit": "lint-staged"
+      "pre-commit": "lint-staged",
+      "pre-push": "yarn l10n"
```
a-b-r-o-w-n, line 84: Can we instead make this a github action that runs on a cron? This is a little disruptive to the dev experience.

Composer/scripts/l10n-makeTSV.js
```diff
@@ -0,0 +1,22 @@
...
+
+const inPath = process.argv[2];
+
+// eslint-disable-next-line security/detect-non-literal-fs-filename
+const inJson = JSON.parse(fs.readFileSync(inPath));
+const outPath = path.join(path.dirname(inPath), 'en-US.tsv');
+
+// eslint-disable-next-line security/detect-non-literal-fs-filename
+const outTSV = fs.openSync(outPath, 'w');
+fs.writeSync(outTSV, 'Key\tMessage\n');
```
a-b-r-o-w-n, line 16: We should consider using (fast-csv)[https://c2fo.io/fast-csv] instead to handle escaping properly. We could then use the csv format instead.

a-b-r-o-w-n (changes requested):

Composer/scripts/l10n-makeTSV.js
a-b-r-o-w-n: I just saw the tsv was requested. I still suggest using the `fast-csv` package with the tab delimeter.

beyackle: What they requested was an Excel file. I recommended TSV because none of our strings have tabs in them, making it a little more natural than CSV. You're right, though, that it's probably cleaner to just import a library.

Composer/package.json
beyackle: This is intended to stop another disruptive behavior of someone getting a PR merged that hasn't had the localization script run followed by the next person to pick up the build wondering why `en-US.json` changes when they haven't done anything to strings. I'll look into making it into an action, though.

a-b-r-o-w-n: My suggestion is to stop running the script as a part of normal dev so that it is not disruptive in those ways. The action could run every night and if the en-US.json file changes, it can commit the result directly into main.

Composer/scripts/l10n-makeTSV.js
a-b-r-o-w-n: Since it is not a requirement to use a tab delimiter, just stick with a comma and allow fast-csv to properly escape the values.

Composer/packages/server/src/locales/en-US.tsv
```diff
@@ -0,0 +1,802 @@
+Key	Message
```
a-b-r-o-w-n, line 1: It looks like this file should be ignored, but got committed.

Composer/.gitignore
```diff
@@ -11,5 +11,9 @@ l10ntemp/
 
 packages/server/schemas/*.schema
 packages/server/schemas/*.uischema
+packages/server/schemas/*.csv
+packages/server/src/locales/en-US.csv
 !packages/server/schemas/sdk.schema
 !packages/server/schemas/sdk.uischema
+
+packages/electron-server/locales/en-US.tsv
```
a-b-r-o-w-n, line 19: Does this need to be updated to `csv`?

Composer/scripts/l10n-extractJson.js
```diff
@@ -23,15 +24,19 @@ try {
...
+const uiOutput = keep(uiSchema, L10N_FIELDS_UI, (x) => x, uiArray);
 const outputTransformed = keep(output, L10N_FIELDS, transFn);
-const uiOutputTransformed = keep(uiOutput, L10N_FIELDS, transFn);
+const uiOutputTransformed = keep(uiOutput, L10N_FIELDS_UI, transFn);
 const outputFilename = path.join(schemaDir, 'sdk.en-US.schema');
 const outputFilenameTrans = path.join(schemaDir, 'sdk.en-US-pseudo.schema');
 const uiOutputFilename = path.join(schemaDir, 'sdk.en-US.uischema');
 const uiOutputFilenameTrans = path.join(schemaDir, 'sdk.en-US-pseudo.uischema');
+const outputFilenameTSV = path.join(schemaDir, 'sdk.en-us.schema.csv');
+const uiOutputFilenameTSV = path.join(schemaDir, 'sdk.en-US.uischema.csv');
```
a-b-r-o-w-n, line 39: nit: these still reference TSV

a-b-r-o-w-n (changes requested):

beyackle:
I'm going to temporarily close this one and come back to it as part of a later effort to automate more of the documentation tasks (once that becomes a priority for me again). It doesn't need to stay open that whole time.

PR #4207, opened 2020-09-18
fix: refactored select skill ui-plugin

### LINKED ISSUE #4143: Changing the skill endpoint in the Begin Skill property editor does not properly update the skill settings
## Describe the bug
Changing the 'Skill endpoint' in the Begin Skill property editor updates the settings correctly; however, it does not update the selected key in the dropdown. 

## Version
Last build from main

## To Reproduce

1. Add a skill with multiple endpoints
2. Configure a Begin Skill action
3. Try changing the endpoint

## Expected behavior
The dropdown should reflect the skills settings

## Screenshots
Sep-15-2020 08-15-29

tdurnford:
## Description
- Refactored the select skill ui-plugin to simplify logic
  - Previously, the `<BeginSkillDialogField />`component was responsible for updating the dialog data and the skills setting.
  - This moves the responsibility for updating the dialog to the `<SelectSkillDialogField />` component and the responsibility to manage the settings to the `<SelectEndpointField />`component.
- Removes logic to support backward compatibility since it will be addressed in another in PR in R11 by @srinaath

## Task Item
Closes #4143


Composer/packages/lib/shared/src/types/shell.ts
```diff
@@ -58,6 +58,7 @@ export interface ShellData {
   qnaFiles: QnAFile[];
   userSettings: UserSettings;
   skills: any[];
+  skillsSettings: { [name: string]: SkillSetting };
```
hatpick, line 61: nit: Record<string, SkillSetting>

Composer/packages/ui-plugins/select-skill-dialog/src/SelectSkillDialogField.tsx
```diff
@@ -41,21 +56,19 @@ export const SelectSkillDialog: React.FC<{
...
         addSkillDialog().then((skill) => {
           if (skill?.manifestUrl && skill?.name) {
-            onChange({ key: skill?.manifestUrl, text: skill?.name });
+            onChange({ ...value, ...settingReferences(skill.name) });
           }
           setComboboxTitle(null);
         });
       } else {
-        onChange(option);
+        onChange({ ...value, ...option?.data });
```
hatpick, line 70: unnecessary null conditional operator because of L60 check

Composer/packages/ui-plugins/select-skill-dialog/src/SkillEndpointField.tsx
```diff
@@ -3,50 +3,48 @@
...
+  const id = getSkillNameFromSetting(value);
+  const skill = skills.find(({ id: skillId }) => skillId === id) || ({} as Skill);
+  const { endpointUrl, msAppId } = skillsSettings[id] || {};
+
+  const { endpoints = [] } = skill;
 
-  const deserializedValue = typeof uiOptions?.serializer?.get === 'function' ? uiOptions.serializer.get(value) : value;
+  const options = useMemo(
+    () =>
+      endpoints.map(({ name, endpointUrl, msAppId }, key) => ({
```
hatpick, line 25: Isn't endpointUrl is shadowed by L19 endpointUrl?

tdurnford: The `endpointUrl` on L19 is the current setting. On L25 we are creating the array of options from all the available endpoints which each have their own `endpointUrl`. The map function uses the `endpointUrl` from the mapper function's scope, which is the correct behavior.

hatpick: I understand they are different variables, but the one inside the mapper is shadowing the outer scope one, we currently don't have this on: https://eslint.org/docs/rules/no-shadow so take this comment as nit.

srinaath (approved):

PR #4210, opened 2020-09-21
[Do not merge] fix: Typing too fast in composer results in dropped characters. (by solving over rendering)

### LINKED ISSUE #4102: Typing too fast in composer results in dropped characters

lei9444:
## Description
This is a quick fix for TextField over rendering.

The first step is use class shouldComponentUpdate to trigger a render only  when the value and error changed.

## Task Item
refs #4171
closes #4102

### DIFF
--- a/Composer/packages/adaptive-form/src/components/fields/ArrayField.tsx
+++ b/Composer/packages/adaptive-form/src/components/fields/ArrayField.tsx
@@ -3,7 +3,6 @@
 /** @jsx jsx */
 import { jsx } from '@emotion/core';
 import React, { useState } from 'react';
-import { TextField } from 'office-ui-fabric-react/lib/TextField';
 import { IconButton } from 'office-ui-fabric-react/lib/Button';
 import { TooltipHost } from 'office-ui-fabric-react/lib/Tooltip';
 import { SharedColors, NeutralColors, FontSizes } from '@uifabric/fluent-theme';
@@ -16,6 +15,7 @@ import { FieldLabel } from '../FieldLabel';
 import { arrayField } from './styles';
 import { ArrayFieldItem } from './ArrayFieldItem';
 import { UnsupportedField } from './UnsupportedField';
+import { WrappedTextField } from './WrappedTextField';
 
 const ArrayField: React.FC<FieldProps<unknown[]>> = (props) => {
   const {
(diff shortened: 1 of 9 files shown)


boydc2014:
Do you have a data than, how much over rendering is saved by this fix?

lei9444:
> Do you have a data than, how much over rendering is saved by this fix?
If I only type one character you can see that the the first one only render once, but the second one render 10 times
image
image

lei9444:
> > > Do you have a data than, how much over rendering is saved by this fix?
> > 
> > 
> > If I only type one character you can see that the the first one only render once(this PR), but the second one render 10 times
> > image
> > image
> 
> So, it's 1/10 render times comparison to previous version. Another question is what else props (other than value & error) would trigger this render, i mean, what you are guarding for?

The default behavior is to re-render every prop change. So if the onChange is re-created in the parent, it will trigger a re-render. But I think  in most of the cases the onChange is no need to change. Some props like disabled maybe changed dynamically, we can't prevent it to trigger a re-render.

lei9444:
> > > > Do you have a data than, how much over rendering is saved by this fix?
> > > 
> > > 
> > > If I only type one character you can see that the the first one only render once(this PR), but the second one render 10 times
> > > image
> > > image
> > 
> > 
> > So, it's 1/10 render times comparison to previous version. Another question is what else props (other than value & error) would trigger this render, i mean, what you are guarding for?
> 
> The default behavior is to re-render every prop change. So if the onChange is re-created in the parent, it will trigger a re-render. But I think in most of the cases there is no need to re-create the onChange function. Some props like disabled maybe changed dynamically, we can't prevent it to trigger a re-render.

boydc2014:
This branch seems to have a bug that, the updates in one text field will appear in another textField.

Repro step
* create a setProperty action
* update property field
* update value field
* clean value field, then you will see property filed also gets cleared

lei9444:
this is not a good approach to avoid re-render by detecting the value change. Because if we want to use some props' value in the onChange, we need to re-create the function by react or use ref

Issue #3449, opened 2020-06-22
Flow Editor zoom in / zoom out


yeze322:
In Flow Edtior, the canvas could be zoomed in / zoomed out by either keyboard shortcuts or a minimap.

1. Use Ctrl + Mouse wheel to zoom in / zoom out;
2. Use Mouse Right Click + drag to move the focus area.

yeze322:
@sangwoohaan @DesignPolice guys do we have a UX design for that?
What's the expected behavior to control the zoom of the action canvas?
Shall we also provide a way to focus on different area such as mouse drag?

sangwoohaan:
I don't think we are ready to move forward with this feature yet.

bradrlaw:
@sangwoohaan Another option could be using the design we have in viso, powerpoint, etc...  Leave the area scrollable as it is now but have a slider for the zoom:

<img width="223" alt="Screen Shot 2020-08-03 at 2 32 49 AM" src="https://user-images.githubusercontent.com/10415799/89152823-c72dc200-d531-11ea-9dae-40d50d556489.png">

As I am building more complex flows this would be very useful.  With decision flows I spend quite a bit of time scrolling left / right.  Having the separate zoom allows for quicker navigation while still keeping the property sheet easy to read.  Could also have buttons to fit the flow to the window or have a full screen view option as well.

mareekuh:
@yeze322 please work with Vivek on this.
@vivekkshankar  Please connect with Thad if there are any updates to what is currently in PVA. If not, enable the following (functionality & look and feel)

![Screen Shot 2020-09-14 at 2 40 34 PM](https://user-images.githubusercontent.com/48503434/93141251-b88e0b00-f698-11ea-969c-719a884abfdb.png)
![Screen Shot 2020-09-14 at 2 40 28 PM](https://user-images.githubusercontent.com/48503434/93141255-b926a180-f698-11ea-9e00-f5dd2c07c4c1.png)
![Screen Shot 2020-09-14 at 2 40 16 PM](https://user-images.githubusercontent.com/48503434/93141258-b9bf3800-f698-11ea-9838-952214d06e71.png)

DesignPolice:
Always nice to be CC’d on things – I can help til Thursday. 😊
m

sangwoohaan:
It's a ghost!!!

yeze322:
cc @alanlong9278

vivekkshankar:
[...]

PR #4217, opened 2020-09-21
feat: zoom in out

alanlong9278:
### Zoom Rate info
- list is the same with Chrome. `[0.25, 0.33, 0.5, 0.75, 0.8, 0.9, 1, 1.1, 1.25, 1.75, 2, 2.5, 3, 4, 5],`, 
- max: 3
- min: 0.5
zoom

### Done
- Use Ctrl + Mouse wheel to zoom in / zoom out
- Add zoom in/out and home button

### TODO
- Create a miniMap for Editor.
## Task Item

ref #3449


hatpick:
@alanlong9278 is this feature going to be paired with panning?

alanlong9278:
> @alanlong9278 is this feature going to be paired with panning?

This feature only includes zoom in/out and focus, not pair with panning

Composer/packages/adaptive-flow/src/adaptive-flow-editor/AdaptiveFlowEditor.tsx
```diff
@@ -73,8 +72,11 @@ const VisualDesigner: React.FC<VisualDesignerProps> = ({ onFocus, onBlur, schema
     data: inputData,
     hosted,
     schemas,
+    zoomRateInfo,
   } = shellData;
 
+  const { updateZoomRate } = shellApi;
```
yeze322, line 78: it can be more specific, such as `updateFlowZoom`

yeze322: `zoomRateInfo` -> `flowZoomRate`

Composer/packages/adaptive-flow/src/adaptive-flow-editor/components/ZoomZone.tsx
```diff
@@ -0,0 +1,131 @@
...
+              fill="white"
+            />
+          </svg>
+        </IconButton>
+      </div>
+    );
+  };
+  useEffect(() => {
+    if (zoomRateInfo) {
+      divRef.current?.addEventListener('wheel', onWheel, { passive: false });
```
yeze322, line 120: nit: we can add a comment to explain why a ref is necessary here

Composer/packages/adaptive-flow/src/adaptive-flow-editor/components/ZoomZone.tsx
```diff
@@ -0,0 +1,131 @@
...
+// Licensed under the MIT License.
+
+/** @jsx jsx */
+import { jsx, css } from '@emotion/core';
+import { useRef, useEffect, ReactNode } from 'react';
+import { ZoomInfo } from '@bfc/shared';
+import { IconButton, IButtonStyles } from 'office-ui-fabric-react/lib/Button';
+import { IIconProps } from 'office-ui-fabric-react/lib/Icon';
+
+function ScrollZoom(
```
yeze322, line 11: the function name needs to be camel-cased

yeze322: And we can make this function pure by not operating DOM inside it. The DOM operation can be hoist to `handleZoome` which can consume component context.

yeze322: nit: another suggestion is, we can wrap the zoom rate calculation logic as a recoil action.
where it exposes two interfaces `zoomIn()` and `zoomOut`, they manipulates the recoil value directly.

alanlong9278: I think that the zoom logics are wraped in the zoomZone is more flexable because we need use the newest rate to excute zoom in/out for the flow.  Recoil action just set recoil state.

cwhitten (approved):

PR #4225, opened 2020-09-21
ci: make new GitHub action to run l10n script

beyackle:
## Description

This adds a new GitHub action to run our string-extraction script and, if it finds anything new, commit the results.

## Task Item

#minor


.github/workflows/l10n.yml
```diff
@@ -0,0 +1,27 @@
...
+      # Runs a set of commands using the runners shell
+      - name: run l10n scripts
+        run: |
+          cd Composer
+          yarn l10n
+
+      - uses: stefanzweifel/git-auto-commit-action@v4
+        with:
+          commit_message: find and localize new strings
+          commit_author: Localization Bot c/o Ben Yackley <beyackle@microsoft.com>
```
a-b-r-o-w-n, line 26: Consider using the `file_patterns` option here to only commit files we expect to have changed.

.github/workflows/l10n.yml
```diff
@@ -0,0 +1,27 @@
+name: daily l10n run
+
+on:
+  schedule:
+    - cron: "0 0 * * 1-5"
```
a-b-r-o-w-n, line 5: Let's also enable this to be run manually using the `workflow_dispatch` event: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_dispatch

a-b-r-o-w-n, line 5: Can you add a comment with what this cron means? I almost always have to look it up!

a-b-r-o-w-n:
@beyackle is this still something we want to do? I think it would be beneficial to run nightly, but I'll defer to you. I'm going to close this one now, but re-open if this is an effort we want to pursue for R14.

PR #3847, opened 2020-08-14
fix: multi-lang all up view refinement

zhixzhan:
## Description

Some refinement in multiple language & all up view

1. Search box for multiple languages
2. Editable table list view
3. Pretty column space.

## Task Item

#minor 

## Screenshots
Search box for multiple languages

Untitled

Editable table list view

Untitled1

Columns space in 1920px
Untitled2

Columns space in 1280px
Untitled3

PR #4230, opened 2020-09-22
fix: New LG template not sync to other locale files

### LINKED ISSUE #4228: New LG templates aren't get synced to other languages
Repro step,

1. create an echo bot
2. add a new language "zh" for example
3. add a send activity action
4. test in emulator with en-us. 

Expected:
the content get synced over en-us and the bot replies fine

Action:
Error saying, there is no evaluator

zhixzhan:
## Task Item
close #4228

### DIFF
--- a/Composer/packages/client/src/recoilModel/dispatchers/__tests__/lg.test.tsx
+++ b/Composer/packages/client/src/recoilModel/dispatchers/__tests__/lg.test.tsx
@@ -40,6 +40,9 @@ const lgFiles = [
   {
     id: 'common.en-us',
     content: `\r\n# Hello\r\n-hi`,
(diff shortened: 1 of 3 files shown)


boydc2014:
I think this is a regression, can you also investigate when this regression is introduced, and how we can better avoid this?

zhixzhan:
> I think this is a regression right? can you also investigate when this regression is introduced, and how we can better avoid this?

Yes, It's a regression introduced by #3847 , when refactor this, I didn't aware formEditor is always call `updateTemplate` do create, and updateTemplate won't sync cross multi-locale.

A possible approach to avoid this side effect may be centralize all operations in action center.

boydc2014 (approved):

Issue #3588, opened 2020-07-07
Support `fieldsets` in ui schema


a-b-r-o-w-n:
## Is your feature request related to a problem? Please describe.

As a developer of a custom sdk component, I would like to be able to have better control over the organization of the form UI by being able to group fields together in fieldsets.

## Describe the solution you'd like

```json
{
  "fieldsets": [
    {
      "title": "Advanced Settings",
      "fields": ["autoEndDialog", "disabled", "..."],
      "defaultCollapsed": true
    }
  ]
}
```

## Additional context

This could also support the tabbed view that we do with the prompts today.

a-b-r-o-w-n:
@sangwoohaan @mareekuh Getting this on your radar to start a discussion. @tomlm has requested this feature in the ui schema that we currently don't support. 

Since we have talked about hiding some fields behind an "Advanced" section, I think this would be a good way to achieve that. Let's use this thread to discuss.

mareekuh:
Could you share some examples of the scenario - what is the dev trying to accomplish? And what would be common examples for which fieldsets are needed/used?

tomlm:
Any time you build a form UX which edits a number of fields you almost always end up wanting to organize them to make it easier to understand.  For example

Color Properties
* BackgroundColor
* ForegroundColor
* TextColor
* Border Color

Another example is cluster things so that less used fields are below the fold or only accessible for advanced users.

Field1
Field2
Field3

Advanced Options<--collapsed -->
* Advanced Field1
* Advanced Field2
* Advanced Field3

This has been part of GUIs toolkits and design since they were invented and every GUI system since then has supported this as a design pattern.

**FieldSet** is the name of the way that w3 HTML defined it all the way back to HTML 1 as the way to group input fields in a form 
[www.w3.org](https://www.w3.org/WAI/tutorials/forms/grouping/)

It a key way of reducing the cognitive load of someone who is presented with a number of fields, and so shows up in many design style best practices:
* [form design](https://www.nngroup.com/articles/form-design-white-space/)
* [Form design best practices](https://www.ventureharbour.com/form-design-best-practices/)

> If your form has more than six fields, it’s considered good practice to group questions into logical sections or steps.

PR #4231, opened 2020-09-22
feat: Added fieldSets to UIOptions

tdurnford:
## Description
Added `fieldSets` to the UIOptions. `fieldSets` is an array of `FieldSet` objects that have properties `title`, `fields`, and `defaultExpanded`. `title` is a required field while `fields` and `defaultExpanded` are optional. If `fields` is `undefined`, it defaults to an array with a wildcard - `['*']`. Each `fieldSet` can only have one wild card and fields cannot be used multiple times. `defaultExpanded` defaults to falsy. When `defaultExpanded` is truthy, the collapsible group is expanded; otherwise, it is collapsed. 

In a future PR, I plan to add another UIOption - `pivotFieldSet` - to support the tabbed view used in the prompts property editor. I would have included that work in this PR; however, UIOptions needs to support `additionalFields` before we can migrate the prompts to `fieldSets` because the prompts ui-plugin currently adds a field to support LU.

## Task Item
refs #3588

## Screenshots
Screen Shot 2020-09-21 at 12 45 34 PM


a-b-r-o-w-n (approved):
This is great! Can you update the ui schema json schema in the botframework-sdk repo?

a-b-r-o-w-n (approved):

PR #4240, opened 2020-09-23
feat: Adds support for additional fields in UIOptions

### LINKED ISSUE #4208: Consider adding additional fields to ui schema
## Description
Currently, to add additional fields to the Property Editor, you have to override the root field with a component that renders the additional fields and the `<ObjectField />` that renders the rest of the form. For example, the Select Skill Dialog's plugin config overrides the root field with the `<BeginSkillDialogField />` component, which renders an additional field - `<SelectSkillDialog />` - and the `<ObjectField />` to render the rest of the form.  

There is also currently no way to add additional fields in the middle of the form - the additional fields either have to be at the beginning or the end of the form.

#### PluginConfig
```js
const config: PluginConfig = {
  uiSchema: {
    [SDKKinds.BeginSkill]: {
      form: {
        field: BeginSkillDialogField,
      },
    },
  },
};
```

#### Field Override 
```jsx
export const BeginSkillDialogField: React.FC<FieldProps> = (props) => {
  const { value, onChange } = props;

  return (
    <React.Fragment>
      <SelectSkillDialog value={value} onChange={onChange} />
      <ObjectField {...props} />
    </React.Fragment>
  );
};
```
## Proposed solution
We should consider adding an `additionalFields` option to the the `UIOptions`. This would allow us to easily add additional fields and custom behavior to the form without having to override the root field. Additionally, we could have some mechanism to add the `additionalFields` to the `order` UI Option -  `order: ['skillAppId', 'additionalFields[0]', '*', 'activityProcessed']` - so we can add fields anywhere in the form. 
#### UIOptions
```js
export interface UIOptions {
...
  // Custom fields added to the form
  additionalFields: FieldWidget[],
...
}
```

## Additional Context
https://github.com/microsoft/BotFramework-Composer/blob/9ba5e8412eb81852fdd140b692ad46528f9c7b1b/Composer/packages/ui-plugins/select-skill-dialog/src/index.ts#L15

https://github.com/microsoft/BotFramework-Composer/blob/9ba5e8412eb81852fdd140b692ad46528f9c7b1b/Composer/packages/ui-plugins/schema-editor/src/formSchema.ts#L12
@a-b-r-o-w-n thoughts?

tdurnford:
## Description
Adds the ability to add additional fields to the Adaptive Form. To add an additional field, developers only have to add a new property with the `additionalField` option set to `true` and a `field` override to the `properties` option in the UIOptions. The `<ObjectField />` will recognize the new field and add it to the schema.

```js
const uiOptions = {
  order: ['schemaField1', 'schemaField2', '*', 'additionalField']
  properties: {
    additionalField: {
      additionalField: true,
      field: () => <div>Additional Field</div>,
    }
  }
}
```

#### Order
Additional fields can be included in the UI Option order override by adding the additional field's `name` to the order array. Note, the additional field's `name` can not exist as a property name in the schema.
```js
const uiOptions = {
  order: ['schemaField1', 'schemaField2', '*', 'additionalField']
  properties: {
    additionalField: {
      additionalField: true,
      field: () => <div>Additional Field</div>,
    }
  }
}
```

#### Field sets
Additional fields can also be included in field sets by adding the additional field's `name` to the `fields` array in the field set.
```js
const uiOptions = {
  fieldset: [{ titile: 'field set 1', fields: ['schema field 1', 'additionalField', '*' }],
  properties: {
    additionalField: {
      additionalField: true,
      field: () => <div>Additional Field</div>,
    }
  }
}
```
## Task Item
Closes #4208


Composer/packages/adaptive-form/src/components/fields/ObjectField.tsx
```diff
@@ -31,18 +33,37 @@ const ObjectField: React.FC<FieldProps<object>> = function ObjectField(props) {
...
+              row={row}
+              schema={schema}
+              uiOptions={uiOptions}
+              value={value}
+              onChange={handleChange}
+            />
+          );
+        } else {
+          return (
+            <AdditionalField
```
tdurnford, line 53: Does it make more sense to move this into the FormRow component?

a-b-r-o-w-n:
This isn't directly related to this PR, but I'm curious how fieldsets and order work together?

Composer/packages/adaptive-form/src/components/fields/AdditionalField.tsx
```diff
@@ -0,0 +1,29 @@
...
+  description?: string;
+  label?: string | false;
+  required?: boolean;
+}
+
+export const AdditionalField: React.FC<AdditionalFieldProps> = (props) => {
+  const { id, depth, description, field: Field, helpLink, label, required } = props;
+
+  return (
+    <div css={schemaField.container(depth)}>
```
a-b-r-o-w-n, line 24: Is `depth` used anywhere currently? Should we remove it?

Composer/packages/adaptive-form/src/components/fields/FieldSets.tsx
```diff
@@ -5,25 +5,22 @@ import React from 'react';
 import { FieldProps } from '@bfc/extension-client';
 
 import { CollapseField } from '../CollapseField';
-import { getFieldSets } from '../../utils';
+import { getFieldsets } from '../../utils';
 
 import { ObjectField } from './ObjectField';
 
-export const FieldSets: React.FC<FieldProps<object>> = (props) => {
+export const Fieldsets: React.FC<FieldProps<object>> = (props) => {
```
a-b-r-o-w-n, line 12: I'm not sure why this happens, but exporting this way does not set the component's display name (which is helpful when using the react debugger). Instead do this: `export { Fieldsets }`

Composer/packages/adaptive-form/src/components/fields/ObjectArrayField.tsx
```diff
@@ -80,7 +80,10 @@ const ObjectArrayField: React.FC<FieldProps<any[]>> = (props) => {
     itemSchema && typeof itemSchema !== 'boolean' ? itemSchema : {},
     uiOptions,
     value
-  ).filter((property) => Array.isArray(property) || !isPropertyHidden(uiOptions, value, property));
+  ).filter(
+    (property) =>
+      Array.isArray(property) || (typeof property !== 'object' && !isPropertyHidden(uiOptions, value, property))
```
a-b-r-o-w-n, line 85: Can this be extracted to a named function for clarity?

Composer/packages/adaptive-form/src/components/fields/ObjectField.tsx
```diff
@@ -31,18 +33,37 @@ const ObjectField: React.FC<FieldProps<object>> = function ObjectField(props) {
...
+              definitions={definitions}
+              depth={newDepth}
+              row={row}
+              schema={schema}
+              uiOptions={uiOptions}
+              value={value}
+              onChange={handleChange}
+            />
+          );
+        } else {
```
a-b-r-o-w-n, line 51: What is `row` when this condition is met? I would think that an additional field gets merged into the schema and resolved like every other field.

Composer/packages/adaptive-form/src/utils/__tests__/getOrderedProperties.test.ts
```diff
@@ -78,4 +78,36 @@ describe('getOrderedProperties', () => {
...
+  it('can include additional fields in the order', () => {
+    const uiOptions = {
+      additionalFields: [{ name: 'additionalField', field: 'field' }],
+      order: ['one', 'two', 'additionalField', '*'],
+    };
+
+    expect(getOrderedProperties(schema, uiOptions, data)).toEqual([
+      'one',
+      'two',
+      { name: 'additionalField', field: 'field' },
```
a-b-r-o-w-n, line 105: Ah. This answers my previous question.

What if instead this just returns the name of the field and the field resolution logic would know to check for additional fields? That would isolate the knowledge of additional fields a little more to the ui options utility functions.

Composer/packages/ui-plugins/select-skill-dialog/src/ComboBoxField.tsx
```diff
@@ -3,7 +3,6 @@
```
a-b-r-o-w-n, line 3: Can this component move into the form library? I think we would need to support the `widget` ui option first.

a-b-r-o-w-n (changes requested):
I think that when processing ui options, we take `additionalFields` and merge that into the schema and ui options so that

```ts
{
  SomeKind: {
    additionalFields: [{ name: 'newField', field: SomeFieldWidget, description: 'some description' }]
  }
}
```

becomes

```ts
// schema
{
  SomeKind: {
    properties: {
      // ... other properties
      newField: { type: 'string' }
    }
  }
}

// ui options
{
  SomeKind: {
    properties: {
      newField: {
        description: 'some description'
      }
    }
  }
}
```

I'm not sure about modifying the schema, but I think this approach would work nicely with the ui options. What do you think?

Composer/packages/adaptive-form/src/components/fields/AdditionalField.tsx
tdurnford: Depth is really only used for padding the root level properties. I'd be alright with removing it.

Composer/packages/adaptive-form/src/components/fields/FieldSets.tsx
tdurnford: Weird. I'll update it.

Composer/packages/adaptive-form/src/components/fields/ObjectField.tsx
tdurnford: When `row` isn't a `string` or an `array`, it is presumably an `additionalField`. I thought about adding the additional fields to the schema, and I do like how clean it would make things; however, the `onChange` handler for each individual property does not meet the requirements for some of the additional fields we currently have in the ui-plugins. For example, the `SelectSkillEndpoint` additional field needs to set multiple properties - `skillEndpoint` and `skillAppId`. Thoughts?

Composer/packages/adaptive-form/src/utils/__tests__/getOrderedProperties.test.ts
tdurnford: We could just return the name, but if we don't add the additional field to the schema, we would have to map it to the additional field object at some point. `getOrderedProperties` just seemed like a convenient place to handle that.

tdurnford:
> This isn't directly related to this PR, but I'm curious how fieldsets and order work together?

Currently, the the ui order take priority over the order provided in the fieldset. Not sure if I like that behavior. Thoughts?

Composer/packages/ui-plugins/select-skill-dialog/src/ComboBoxField.tsx
tdurnford: This component has some logic specific to the plugin, so I don't know if moving it to the component library makes sense.

Composer/packages/extension-client/src/types/formSchema.ts
```diff
@@ -9,13 +9,15 @@ type UIOptionValue<R = string, D = any> = R | UIOptionFunc<R, D>;
...
-interface FieldSet {
+export interface Fieldset {
   title: string;
   fields?: string[];
   defaultExpanded?: boolean;
 }
 
 export interface UIOptions {
+  /** Specifies if the property is an additional field */
+  additionalField?: true;
```
a-b-r-o-w-n, line 20: I think this should be represented by a union type.

```ts
type AdditionalField = {
  additionalField: true;
  field: FieldWidget;
  // other properties
}

// uiOptions.properties
type UiOptions = {
  properties: {
    [key: string]: UIOptions | AdditionalField;
  }
}
```

a-b-r-o-w-n (commented):
Just one suggestion.

a-b-r-o-w-n (approved):

a-b-r-o-w-n (approved):

PR #4224, opened 2020-09-21
feat: install remote extensions from npm

a-b-r-o-w-n:
## Description

Adds apis to install/uninstall extensions from npm. 

Includes other various refactors as well.

## Task Item

closes #3992

PR #4251, opened 2020-09-23
fix: fix loading of extensions by removing sample-ui-plugin

a-b-r-o-w-n:
## Description

In #4224, the way builtin plugins are loaded changed which breaks when the `extensions.json` file already existed, due to a failure to load the `sample-ui-plugin`. This just removes that from the manifest.

## Task Item

#minor

### DIFF
--- a/Composer/packages/extension/src/manager/manager.ts
+++ b/Composer/packages/extension/src/manager/manager.ts
@@ -245,12 +245,15 @@ class ExtensionManager {
       const extensionInstallPath = path.dirname(fullPath);
       const packageJson = (await readJson(fullPath)) as PackageJSON;
       const isEnabled = packageJson?.composer && packageJson.composer.enabled !== false;
+      const metadata = getExtensionMetadata(extensionInstallPath, packageJson);
       if (packageJson && (isEnabled || packageJson.extendsComposer === true)) {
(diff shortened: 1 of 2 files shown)


beyackle (approved):

beyackle (approved):

Composer/packages/extension/src/manager/manager.ts
```diff
@@ -245,12 +245,15 @@ class ExtensionManager {
       const extensionInstallPath = path.dirname(fullPath);
       const packageJson = (await readJson(fullPath)) as PackageJSON;
       const isEnabled = packageJson?.composer && packageJson.composer.enabled !== false;
+      const metadata = getExtensionMetadata(extensionInstallPath, packageJson);
```
tonyanziano, line 248: why was this moved out of the `if` statement if the `else if` does not contain any reference to `metadata`?

a-b-r-o-w-n: Good question. I had a version where I was just disabling the extension but setting the other metadata fields. I felt like removing the extension is better but forgot to move this back.

tonyanziano: we can just move it back on some other PR

PR #4263, opened 2020-09-24
feat: Update Composer logic to leverage the uiSchema when initializing the IntellisenseField.

### LINKED ISSUE #4188: Update Composer logic to leverage the uiSchema when initializing the IntellisenseField.
Part of updating schema to support better intellisense

LouisEugeneMSFT:
## Description

This is the 3rd and final step to make Composer leverage the uiSchema for Intellisense mappings between components and options (for example "_Intellisense for Microsoft.DeleteProperty should show user variables_").

Step 1 was modifying the schema of the uiSchema: https://github.com/microsoft/botframework-sdk/pull/6047
Step 2 was updating the uiSchema of individual components: https://github.com/microsoft/botbuilder-dotnet/pull/4704

This PR is removing the hardcoded mappings and injecting the projectId into IntellisenseTextField to make it work with user variables.

## Task Item

fixes #4188

## Screenshots

image


Composer/packages/adaptive-form/src/components/FormRow.tsx
```diff
@@ -35,13 +33,7 @@ export const getRowProps = (rowProps: FormRowProps, field: string) => {
   const { required = [] } = schema;
   const fieldSchema = resolvePropSchema(schema, field, definitions);
 
-  const intellisenseScopes: string[] = [];
-  if (field === 'property') {
-    intellisenseScopes.push('variable-scopes');
```
GeoffCoxMSFT, line 40: I see the logic is deleted from FormRow, but where is it added in IntellisenseField?  I only see the use of projectId in IntellisenseField.tsx.

LouisEugeneMSFT: Hi @GeoffCoxMSFT, I made sure that Intellisense can consume the data from the uiSchema in the 1st Intellisense PR. 
This is what `uiOptions.properties?.[field] `does.
I only needed the hardcoding as a temporary thing until the data was in the schema.

a-b-r-o-w-n: One thing to be aware of is that users with old ui schemas will not get this behavior. Maybe we keep the hardcoded path for now, but prefer the UI Schema options if they are set?

LouisEugeneMSFT: Is the uiSchema not bundled with the code? Aka won't users with the latest app version not also have the latest uiSchema? I would rather avoid any hardcoding when we can. Any field with the "property" value will get one type of Intellisense, which might lead to unexpected behaviors if that value is used for a different purpose (aka not a variable setter) somewhere in the schema.

GeoffCoxMSFT (approved):

PR #4265, opened 2020-09-25
feat: Adds UI pivotFieldsets to UI Options

### LINKED ISSUE #3393: [DCR] Extensions to component schema for dialogs
Working with PowerVA there are 3 area that have surfaced as missing functionality for describing the contract of dialog components and their consumption.

1. **result shape** We don't have a way of describing the shape of results for a dialog. For the most part we have been ok because we have been dealing with simple types in Input dialogs, but as we start to build more complex components we need a way of saying, AddressDialog returns Street, City, Zip, Country. 
2. **dispatch** - We don't have a way of discovering language assets for dispatching for dialog components.  We rely on people getting the language assets and stitching them together by hand.
3. **public vs internal** - We don't have a way of distinguishing which assets in a library should be consumed vs which assets are internal definitions.

This DCR addresses all 3 items by adding properties to the component .schema definition.

# add $result property
Proposal is to add new property **$result** which is a json schema object definition for the **shape of the result** of the dialog.

Example of simple type:
```
   "$result": {
        "type":"string",
        "title":"name"
    }
```
Example of multiple properties that returns **age** and **name**
```
    "$result": {
        "type": "object",
        "properties": {
            "age": {
                "type": "number",
                "title": "Age",
                "description": "The age."
            },
            "name": {
                "$role": "expression",
                "type": "string",
                "title": "Name",
                "description": "The name."
            }
        }
    },
```

# add $activation property
We should add a **$activation** property to describe metadata useful for deciding when it's appropriate to activate a component.  
Currently we should define metadata useful for building orchestration/dispatch models
* **languages** : array of assets in LU files per language
* **intents** : array of strings which define intents from LU files which are appropriate to dispatch.

```
    "$activation": {
        "languages": {
            "en": [
                {
                    "id": "foo-en",
                    "name": "Foo LU (English)",
                    "contentType": "application/lu",
                    "url": "Foo-en.lu",
                    "description": "English language model for the skill"
                },
                {
                    "id": "SkillQnAModel-en",
                    "name": "Foo QnA LU (English)",
                    "contentType": "application/qna",
                    "url": "Foo-QnA-en.lu",
                    "description": "English language model for the skill (QnAMaker"
                }
            ]
        },
        "intents": [
            "trigger",
            "getWeather"
        ]
    },
```
# Add $public property
**NOT NOW** - We have talked and think it is not appropriate to do this via .schema files. Instead, if we want a declarative way for a library to define exported assets there should be some sort of explicit exports.json format.  We are not going to define a mechanism now.

## Skill Manifest generation
Skill manifests can be auto-generated using the metadata in activation.

## Skill Manifest consumption as dialogs
Skill Manifests can auto-generate .dialog and .dialog.schema which allow them to be added to a orchestration component.

## Full example
Here is an example for an AdaptiveDialog which consumes Age, Name and produces Age,Name
```
{
    "$schema": "https://raw.githubusercontent.com/microsoft/botbuilder-dotnet/master/schemas/component.schema",
    "$role": "implements(Microsoft.IDialog)",
    "title": "Custom Action",
    "description": "My Custom Action.",
    "type": "object",
    "$public" : true,    
    "properties": {
        "age": {
            "$role": "expression",
            "type": "number",
            "title": "Age",
            "description": "The age."
        },
        "name": {
            "$role": "expression",
            "type": "string",
            "title": "Name",
            "description": "The name."
        }
    },
    "$result": {
        "type": "object",
        "properties": {
            "dogYears": {
                "type": "number",
                "title": "Dog Years",
                "description": "The users age in dog years."
            }
    },
    "$activation": {
        "languages": {
            "en": [
                {
                    "id": "foo-en",
                    "name": "Foo LU (English)",
                    "contentType": "application/lu",
                    "url": "Foo-en.lu",
                    "description": "English language model for the skill"
                }
            ]
        },
        "intents": [
            "getAgeInDogYears"
        ]
    }
}
```

### LINKED ISSUE #3588: Support `fieldsets` in ui schema

tdurnford:
## Description
Adds `pivotFieldsets` to UI Options. When the ui options has a `fieldset` and `pivotField` is true, the fieldset is rendered in a tabbed view.

This PR also updates the prompts ui-plugin to leverage `fieldsets` and `additionalFields` instead of custom fields. 
 
## Task Item
Closes #3588


Composer/packages/adaptive-form/src/AdaptiveFormContext.ts
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { createContext, useContext } from 'react';
+import { JSONSchema7 } from '@bfc/extension-client';
+
+interface IAdaptiveFormContext {
+  focusedTab?: string;
+  baseSchema: JSONSchema7;
+  onFocusedTab?: (_: string) => void;
```
a-b-r-o-w-n, line 10: Can you name this argument?

a-b-r-o-w-n (changes requested):
This is cool! Let's make sure this gets added to the upstream ui schema definition as well.

a-b-r-o-w-n (approved):

PR #4269, opened 2020-09-25
fix: Monaco editor links opened in blank window in electron

### LINKED ISSUE #4070: [Electron ] When clicking an url in LG, it opens a strange window instead of default browser
## Describe the bug

When clicking an url in LG, it opens a strange window instead of default browser

## Browser

- [x] Electron distribution 1.1.1
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Write url in LG
2. Ctrl+click

## Expected behavior

Open default browser

## Screenshots

image

zhixzhan:
## Description

Monaco editor has option config to auto detect urls in text. ctrl + click on it will call vscode.open() -> window.open() open url.
Our current electron setting can handle `<a target="_blank">open in browser</a>`, but electron@8.2.4 seems has issue accept correct arguments. then we will see a empty page, case the accepted arguments `url` is `about:blank`.

### Alternative solution
If we upgrade electron@10.1.1, the electron opened window would be correctly, but still failed to open in default browser.

### Current solution
So before find a better way, current fix is disable link detecting in BaseEditor.text for electron environment.

## Task Item
close #4070

## Screenshots

Before (electron): 

image

After (electron):

image

### DIFF
--- a/Composer/packages/lib/code-editor/src/utils/common.ts
+++ b/Composer/packages/lib/code-editor/src/utils/common.ts
@@ -4,3 +4,8 @@
 export function processSize(size) {
   return !/^\d+$/.test(size) ? size : `${size}px`;
 }
+
+export function isElectron(): boolean {
+  // eslint-disable-next-line no-prototype-builtins
+  return window.hasOwnProperty('__IS_ELECTRON__');
+}
--- a/Composer/packages/lib/code-editor/src/utils/common.ts
+++ b/Composer/packages/lib/code-editor/src/utils/common.ts
@@ -6,6 +6,5 @@ export function processSize(size) {
 }
 
 export function isElectron(): boolean {
-  // eslint-disable-next-line no-prototype-builtins
-  return window.hasOwnProperty('__IS_ELECTRON__');
+  return !window.__IS_ELECTRON__;
 }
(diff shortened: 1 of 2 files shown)


Composer/packages/lib/code-editor/src/utils/common.ts
```diff
@@ -4,3 +4,8 @@
 export function processSize(size) {
   return !/^\d+$/.test(size) ? size : `${size}px`;
 }
+
+export function isElectron(): boolean {
+  // eslint-disable-next-line no-prototype-builtins
+  return window.hasOwnProperty('__IS_ELECTRON__');
```
a-b-r-o-w-n, line 10:
```suggestion
  return !window.__IS_ELECTRON__;
```

a-b-r-o-w-n (dismissed):

Composer/packages/lib/code-editor/src/utils/common.ts
```diff
@@ -4,3 +4,7 @@
 export function processSize(size) {
   return !/^\d+$/.test(size) ? size : `${size}px`;
 }
+
+export function isElectron(): boolean {
+  return !window.__IS_ELECTRON__;
```
a-b-r-o-w-n, line 9:
```suggestion
  return !(window as any).__IS_ELECTRON__;
```

cwhitten (approved):

PR #4270, opened 2020-09-25
fix: Upgrade to recoil to version 0.0.13

### LINKED ISSUE #4223: Upgrade to recoil to version 0.0.13
This will help improve typing and fix several runtime warnings.
It may affect some of the dispatch calls and require a little refactoring.

@boydc2014 and @srinaath - @hatpick has been looking into this. @cwhitten mentioned that upgrades should generally be considered  platform work so to include you, but that recoil is the area where platform and authoring meet :-)  We could set up a feature branch to work through the upgrade together.

### LINKED ISSUE #4267: Browser console: Error/Warning: Cannot update a component from...
## Describe the bug

When run composer at dev mode, open dev-tool will see those annoy error log.

## Browser

- [ ] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

## Screenshots
image

lei9444:
## Description
Upgrade the recoil to version 0.0.13
The recoil depends on react 16.13.1, so update the react to the latest

## Task Item
closes #4223
closes #4267


a-b-r-o-w-n:
@srinaath do you think you could pick this up today? There are some TS errors in the client build, otherwise this looks good to me. Since you have the most familiarity with recoil in our codebase, would you mind doing a functional test of this?

srinaath:
Yep. Will do this morning.

srinaath:
@lei9444 

I tried LU/LG/QNA editing, Creation, Opening, Delete bot, Skills, Publish workflows. They all functionally look good. In addition, the upgrade got rid of the warnings when using selectors.

Composer/packages/client/src/recoilModel/DispatcherWrapper.tsx
```diff
@@ -72,7 +72,9 @@ const InitDispatcher = ({ onLoad }) => {
 
   const [currentDispatcherState, setDispatcher] = useRecoilState(dispatcherState);
 
-  useEffect(() => {
+  //The render order is different with 0.0.10, the local state will trigger a render before atom value
+  //so use the useLayoutEffect here
+  useLayoutEffect(() => {
```
hatpick, line 77: Do you mind explaining this change?

lei9444: Hi, @hatpick , it looks like that in 0.0.13, recoil removes the queue update in setRecoilState. So the recoil root state's update doesn't happen immediately. 
https://codesandbox.io/s/inspiring-dew-4r5xi?file=/src/App.js
this is the demo, you can change the recoil version to 0.0.10 and see the log message.
maybe this is updated by the performance https://recoiljs.org/blog/#performance

srinaath: @lei9444 I thought reason you added the change was inside InitDispatcher component the onLoad was fired before the dispatcher state is set using the setDispatcher fn. Essentially, we are left with a state where we load the children(entire app) before dispatcher is initialized. By moving it to the useLayout effect we synchronously set the values before rendering any children. Is that fair explanation?

hatpick: @srinaath That's matches what I understood.

lei9444: Thanks, @srinaath, Yes,  I just want to  sync the set function to make sure the children will get the dispatchers

srinaath (approved):

cwhitten (approved):

PR #4284, opened 2020-09-28
fix: Action flyout menu only shows defined $kinds in sdk.schema

### LINKED ISSUE #4107: Actions appears in the menu flyout even when it's not in sdk.schema
## Describe the bug

**Problem**: some actions are showing up in the menu flyouts (httpaction, etc) even when this component does not exist in the schema or the uischema

**Solution**:
Menu UI Schema should filter out those $kinds not implemented in sdk.schema.

## Browser

- [ ] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

yeze322:
## Description
closes #4107 

**Problem**: If a $kind is defined in uischema but not defined in sdk.schema, it will still be shown in the Action flyout menu.

**Solution**: This PR fixes that problem by filtering those $kinds not exist in sdk.schema, only when a $kind is defined correctly in both sdk.schema and uischema will it appear as a flyout option.

To validate this PR, try removing a $kind's definition in sdk.schema, you will find that the removed Action is also removed from Action flyout menu.

## Task Item
#4107


Composer/packages/extension-client/src/hooks/useMenuConfig.ts
```diff
@@ -8,7 +8,20 @@ import { EditorExtensionContext } from '../EditorExtensionContext';
...
-  const { plugins } = useContext(EditorExtensionContext);
+  const { plugins, shellData } = useContext(EditorExtensionContext);
+  const uiSchema = plugins.uiSchema || {};
+  const sdkSchema = shellData.schemas?.sdk;
+  const sdkDefinitions = sdkSchema?.content?.definitions || {};
 
-  return useMemo(() => mapValues(plugins.uiSchema, 'menu'), [plugins.uiSchema]);
+  return useMemo(() => {
+    const menuSchema = mapValues(uiSchema, 'menu') as MenuUISchema;
+    const implementedMenuSchema = Object.entries(menuSchema).reduce((result, [$kind, menuOpt]) => {
```
GeoffCoxMSFT, line 18: Do you need the reduce here? Could this be a map that is setting value in an associative array?

yeze322: Yeah, I made it obscure, that will be a more readable way.

cwhitten (approved):

PR #4295, opened 2020-09-29
fix: Disable the default timeout for requests on the LU build

### LINKED ISSUE #4294: "Network Error" when publishing a big Luis app
## Describe the bug

When starting a bot using a .lu file with over 200 intents and over 6000 utterances, I get a Network error always in the call to "build" exactly after 2 minutes. 
Luis app is created and version contains all intents, but the training didn't happen. 

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [x] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Create a bot with over 200 intents and over 6000 utterances (I did it using luToBot tool)
2. Click on 'StartBot'

## Screenshots

image

benbrown:
## Description

This removes the timeout behavior on the /build endpoint, which can result in a network error when publishing a large language model. By default, Node's HTTP server class is set to timeout after 2 minutes. 

https://nodejs.org/dist/latest-v6.x/docs/api/http.html#http_server_settimeout_msecs_callback

## Task Item

fixes #4294


a-b-r-o-w-n:
Although this addresses #4294, we should attempt to refactor this endpoint to better fit long processes, like polling or web socket.

benbrown:
I agree @a-b-r-o-w-n and have a proposal that I will follow up with ASAP.

In short: we already implement a polling procedure for the publishing. I would like to formalize this so that publishing, provisioning, and building can take advantage of the same subsystem.

a-b-r-o-w-n:
@benbrown that's good to hear.

cwhitten:
Test environment is failing with: 

TypeError: req.setTimeout is not a function

      325 | 
      326 |   // Disable Express' built in 2 minute timeout for requests. Otherwise, large models may fail to build.
    > 327 |   req.setTimeout(0, () => {
          |       ^
      328 |     throw new Error('LUIS publish process timed out.');
      329 |   });
      330 | 


Maybe we need to shim this api in the test runner?

cwhitten:
@benbrown i don't know why tests are failing but it seems specific to this change

benbrown:
@cwhitten I think I found it. Just pushed a small change to the test.

cwhitten (approved):

PR #4312, opened 2020-10-01
fix: Move skill settings into settings.skillConfiguration

srinaath:
## Task Item

Moved skill settings from settings.skill into settings.skillConfiguration. The reason being settings.skill is being used by Composer to maintain the list of skills consumed by the Root Bot. This fix would just allow the C# runtime to get the skills config.

@carlosscastro is tracking this functionality to be ported into the JS runtume

@tonyanziano  This is the fix for the faulty skill config

#minor

### DIFF
--- a/Composer/packages/server/src/models/settings/defaultSettingManager.ts
+++ b/Composer/packages/server/src/models/settings/defaultSettingManager.ts
@@ -2,7 +2,7 @@
 // Licensed under the MIT License.
 
 import set from 'lodash/set';
-import { SensitiveProperties } from '@bfc/shared';
+import { DialogSetting, SensitiveProperties } from '@bfc/shared';
 import { UserIdentity } from '@bfc/extension';
 
 import { Path } from '../../utility/path';
@@ -16,7 +16,7 @@ export class DefaultSettingManager extends FileSettingManager {
     super(basePath, user);
   }
 
-  protected createDefaultSettings = (): any => {
+  protected createDefaultSettings = (): DialogSetting => {
     return {
       feature: {
         UseShowTypingMiddleware: false,
@@ -69,7 +69,7 @@ export class DefaultSettingManager extends FileSettingManager {
         maxImbalanceRatio: 10,
         maxUtteranceAllowed: 15000,
       },
-      skillConfig: {
+      skillConfiguration: {
         isSkill: false,
         allowedCallers: ['*'],
       },
--- a/Composer/packages/server/src/models/settings/defaultSettingManager.ts
+++ b/Composer/packages/server/src/models/settings/defaultSettingManager.ts
@@ -69,10 +69,11 @@ export class DefaultSettingManager extends FileSettingManager {
         maxImbalanceRatio: 10,
         maxUtteranceAllowed: 15000,
       },
-      skill: {
+      skillConfig: {
         isSkill: false,
         allowedCallers: ['*'],
       },
+      skill: {},
       defaultLanguage: 'en-us',
       languages: ['en-us'],
     };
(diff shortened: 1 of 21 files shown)


Composer/packages/server/src/models/settings/defaultSettingManager.ts
```diff
@@ -69,10 +69,11 @@ export class DefaultSettingManager extends FileSettingManager {
         maxImbalanceRatio: 10,
         maxUtteranceAllowed: 15000,
       },
-      skill: {
+      skillConfig: {
         isSkill: false,
         allowedCallers: ['*'],
       },
+      skill: {},
```
beyackle, line 76: Is there a reason to keep this around even though it's empty?

Composer/packages/server/src/models/settings/defaultSettingManager.ts
```diff
@@ -69,10 +69,11 @@ export class DefaultSettingManager extends FileSettingManager {
         maxImbalanceRatio: 10,
         maxUtteranceAllowed: 15000,
       },
-      skill: {
+      skillConfiguration: {
```
tdurnford, line 72: ❤️

runtime/dotnet/azurewebapp/Startup.cs
```diff
@@ -87,7 +87,7 @@ public IStorage ConfigureStorage(BotSettings settings)
 
         public bool IsSkill(BotSettings settings)
         {
-            return settings?.Skill?.IsSkill == true;
+            return settings?.SkillConfig?.IsSkill == true;
```
tdurnford, line 90: Does this need to be 
```suggestion
            return settings?.SkillConfiguration?.IsSkill == true;
```

runtime/dotnet/azurewebapp/Startup.cs
```diff
@@ -132,7 +132,11 @@ public void ConfigureServices(IServiceCollection services)
             // Register AuthConfiguration to enable custom claim validation for skills.
             if (IsSkill(settings))
             {
-                services.AddSingleton(sp => new AuthenticationConfiguration { ClaimsValidator = new AllowedCallersClaimsValidator(settings.Skill) });
+                services.AddSingleton(sp => new AuthenticationConfiguration { ClaimsValidator = new AllowedCallersClaimsValidator(settings.SkillConfig) });
```
tdurnford, line 135: Same comment as above
```suggestion
                services.AddSingleton(sp => new AuthenticationConfiguration { ClaimsValidator = new AllowedCallersClaimsValidator(settings.SkillConfiguration) });
```

runtime/dotnet/core/Settings/BotSettings.cs
```diff
@@ -24,7 +24,7 @@ public class BotSettings
 
         public string Bot { get; set; }
 
-        public BotSkillSettings Skill { get; set; }
+        public BotSkillConfig SkillConfiguration { get; set; }
```
carlosscastro, line 27: I'd consider calling this either 1) `SkillConfiguration` or 2) `BotSkillConfiguration`.

Composer/packages/server/src/models/settings/defaultSettingManager.ts
srinaath: Yeah. Its the object to which we append new skills from the UI. I presonally prefer to keep it so that when new bots are created by default it shows as an empty object and we just keep adding to the dictionary.

carlosscastro (approved):

tdurnford (approved):

cwhitten (approved):

PR #4315, opened 2020-10-02
refactor: change method of installing extensions

a-b-r-o-w-n:
## Description

Directly download and unpack extensions using the npm registry api. This enables a unified approach to loading built-in vs remote extensions and results in more transparent installation output.

## Task Item

#minor

### DIFF
--- a/Composer/packages/extension/package.json
+++ b/Composer/packages/extension/package.json
@@ -20,6 +20,7 @@
     "@types/path-to-regexp": "^1.7.0",
     "@types/tar": "^4.0.3",
     "json-schema": "^0.2.5",
+    "rimraf": "^3.0.2",
     "typescript": "^3.8.3"
   },
   "dependencies": {
--- a/Composer/packages/extension/package.json
+++ b/Composer/packages/extension/package.json
@@ -18,6 +18,7 @@
     "@types/fs-extra": "^9.0.1",
     "@types/passport": "^1.0.3",
     "@types/path-to-regexp": "^1.7.0",
+    "@types/tar": "^4.0.3",
     "json-schema": "^0.2.5",
     "rimraf": "^3.0.2",
     "typescript": "^3.8.3"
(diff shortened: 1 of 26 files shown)


Composer/packages/extension/src/utils/npm.ts
```diff
@@ -1,65 +1,81 @@
...
+      } as ExtensionSearchResult;
     });
+  } catch (err) {
+    log('%O', err);
+
+    return [];
+  }
+}
+
+export async function downloadPackage(name: string, versionOrTag: string, destination: string) {
```
tonyanziano, line 50: try / catch error handling?

Composer/packages/extension/src/manager/manager.ts
```diff
@@ -190,14 +169,16 @@ class ExtensionManager {
...
-      const { stdout } = await npm('uninstall', id, { '--prefix': this.remoteDir }, { cwd: this.remoteDir });
+    const metadata = this.find(id);
 
-      log('%s', stdout);
+    if (metadata) {
+      if (metadata.builtIn) {
+        throw new Error('Cannot remove builtin extension.');
+      }
 
+      await rimraf(metadata.path);
```
tonyanziano, line 179: use [`remove()`](https://github.com/jprichardson/node-fs-extra/blob/HEAD/docs/remove.md) from `fs-extra` instead?

tonyanziano (changes requested):

Composer/packages/extension/src/manager/manager.ts
a-b-r-o-w-n: Nice! I'll update.

Composer/packages/extension/src/utils/npm.ts
a-b-r-o-w-n: I think I'm ok having the caller handle errors for this one. What do you think?

tonyanziano: That's fine by me. Looks like its only invocation right now is wrapped in a try 👍

Composer/packages/extension/package.json
```diff
@@ -20,7 +20,6 @@
     "@types/path-to-regexp": "^1.7.0",
     "@types/tar": "^4.0.3",
     "json-schema": "^0.2.5",
-    "rimraf": "^3.0.2",
```
tonyanziano, line 22: shouldn't it remain a dev dependency since it's used in the build scripts? counting on it being in the workspace's hoisted deps seems wonky

tonyanziano (approved):
Nice!

tonyanziano (dismissed):

tonyanziano (approved):

Issue #3961, opened 2020-08-31
Project Tree: QnA Page 


srinaath:
Implementing the All up Qna view for botprojects with the new project tree implementation.

sangwoohaan:
QnA will share the same navigation design change with LG below
https://www.figma.com/file/LlTlh5vXwq91zjGnFtrUUR6l/Composer-main-design-spec-UI-library?node-id=4868%3A1

Issue #3960, opened 2020-08-31
Project Tree: User Input page


srinaath:
Implementing the All up User Input view (left nav) for botprojects with the new project tree implementation.

sangwoohaan:
LU will share the same navigation design change with LG below
https://www.figma.com/file/LlTlh5vXwq91zjGnFtrUUR6l/Composer-main-design-spec-UI-library?node-id=4868%3A1

Issue #3959, opened 2020-08-31
Project Tree: LG page


srinaath:
Implementing the All up LG view (left nav) for botprojects with the new project tree implementation.

https://www.figma.com/file/LlTlh5vXwq91zjGnFtrUUR6l/Composer-main-design-spec-UI-library?node-id=4868%3A1

Issue #3958, opened 2020-08-31
Context menu in Project tree


srinaath:
- [ ] create menu with stubs
- [ ] run this bot
- [ ] create/edit skill manifest
- [ ] export
- [ ] settings

![](https://user-images.githubusercontent.com/13004779/91777728-80210400-eba5-11ea-90dd-ad228320d649.png)

Context menu operations on a single bot inside the bot project

Left nav contextual menu - https://www.figma.com/file/LlTlh5vXwq91zjGnFtrUUR6l/Composer-main-design-spec-UI-library?node-id=4511%3A32188

cwhitten:
Please make sure this reflects the current R11 POR with @vivekkshankar

PR #4320, opened 2020-10-05
feature: new ProjectTree (not fully working yet)

beyackle:
## Description

This is a kind of "checkpoint" PR; I'm putting up what I have to get it merged into a feature branch so I can get some eyes on it. This new tree *does not fully work yet*. You can click on links in Design View and make triggers appear selected, but currently this doesn't open the dialog. The LG/LU/QnU pages have the tree in them too, but those links don't work either. Actually finishing this PR will close at least some of the related issues (if not all; the LU, LG, and QnA issues are tightly coupled to each other and all 3 will be solved at once when they are).

## Task Item

refs #3961 
refs #3960 
refs #3959 
refs #3958 
refs #3956 

All of these are "refs" for now; they'll move to "closes" when those features are actually finished.

## Screenshots

image
image


hatpick:
@beyackle I think the gitignore is failing to prevent pushing many files.

beyackle:
It's not the .gitignore that's the problem, it's that somehow this change isn't based on the correct thing. All that noise going on inside the extensions folder, especially the compiled .js files in the /lib directories, are things that should exist already in the thing this change is being compared against.

Composer/packages/client/src/components/AppComponents/RightPanel.tsx
```diff
@@ -45,7 +59,16 @@ export const RightPanel = () => {
         setApplicationLevelError={setApplicationLevelError}
       >
         <RequireAuth>
-          <Routes component={Content} />
+          <div css={{ display: 'flex', flexDirection: 'row' }}>
```
GeoffCoxMSFT, line 62: Consider a styled component (or css template)

Composer/packages/client/src/components/AppComponents/RightPanel.tsx
```diff
@@ -33,10 +37,20 @@ const content = css`
 
 const Content = forwardRef<HTMLDivElement>((props, ref) => <div css={content} {...props} ref={ref} />);
 
+const SHOW_TREE = ['design'];
```
GeoffCoxMSFT, line 40: Nit:  My preference is to avoid the UPPER_CASE.  I camelCase them in my own projects because they are module level variables (that happen to be const). Variables that are scoped to a function are camelCase and often const.

Composer/packages/client/src/components/AppComponents/RightPanel.tsx
```diff
@@ -45,7 +59,16 @@ export const RightPanel = () => {
         setApplicationLevelError={setApplicationLevelError}
       >
         <RequireAuth>
-          <Routes component={Content} />
+          <div css={{ display: 'flex', flexDirection: 'row' }}>
+            {SHOW_TREE.includes(currentMode) ? (
+              <LeftRightSplit initialLeftGridWidth="200px" minLeftPixels={200} minRightPixels={800}>
```
GeoffCoxMSFT, line 64: Consider % as it works well with whatever size the window is when it opens.

PR #4322, opened 2020-10-05
fix: Update en-US.json

beyackle:
## Description

Another round of l10n strings that need to get picked up into main.

## Task Item

#minor


Composer/packages/server/src/locales/en-US.json
```diff
@@ -2420,6 +2462,9 @@
   "view_kb_c382e495": {
     "message": "View KB"
   },
+  "view_on_npm_2051324d": {
+    "message": "View on npm"
+  },
   "vishwac_sena_45910bf0": {
```
srinaath, line 2468: Where is this string being used? Maybe John Doe?

beyackle: It's in `src/pages/setting/extensions/ExtensionSearchResults.tsx`. It looks like we're using npm as a place for people to host extensions.

srinaath (approved):

PR #4323, opened 2020-10-05
fix: Fixed useArrayItems and useObjectItems

### LINKED ISSUE #4276: Property page for new action not displayed
## Describe the bug

This probably isn't an easy repo as I think it's the first time I've ever seen it happen but I just added a new `SetProperties` action and composer is still showing the previous actions property editor. Clicking to another action and back resolved the issue:

image

## Version

1.1.1

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

tdurnford:
## Description
When the Property Editor switches between two actions of the same type, the `useArrayItems` and `useObjectItems` caches were not resetting. 

## Task Item
Closes #4276

### DIFF
--- a/Composer/packages/adaptive-form/src/utils/arrayUtils.ts
+++ b/Composer/packages/adaptive-form/src/utils/arrayUtils.ts
@@ -4,7 +4,7 @@
 import { generateUniqueId } from '@bfc/shared';
 import { ChangeHandler } from '@bfc/extension-client';
 import { useEffect, useState } from 'react';
-import { isEqual } from 'lodash';
+import isEqual from 'lodash/isEqual';
 
 type ArrayChangeHandler<ItemType> = (items: ArrayItem<ItemType>[]) => void;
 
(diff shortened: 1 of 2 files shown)


Composer/packages/adaptive-form/src/utils/arrayUtils.ts
```diff
@@ -78,6 +79,19 @@ export function useArrayItems<ItemType = unknown>(
 ): ArrayItemState<ItemType> {
   const [cache, setCache] = useState(generateArrayItems(items));
 
+  useEffect(() => {
+    const newCache = generateArrayItems(items);
+
+    if (
+      !isEqual(
+        cache.map(({ value }) => value),
```
srinaath, line 87: This could yield undefined === undefined right?

srinaath, line 87: This could yield undefined === undefined right?

tdurnford: I don't think so. If a value is `undefined`, the Adaptive Form removes it from the array.

srinaath (approved):

PR #4328, opened 2020-10-05
feat: Allow undo/redo in form dialog schema builder

### LINKED ISSUE #4327: Allow undo/redo in form dialog schema builder
Enable users to undo/redo their actions while building a form dialog schema

hatpick:
## Description

Enable users to undo/redo their actions while building a form dialog schema

## Task Item

closes #4327 

## Screenshots

N/A


Composer/packages/form-dialogs/src/undo/UndoRoot.tsx
```diff
@@ -0,0 +1,173 @@
+// Copyright (c) Microsoft Corporation.
```
GeoffCoxMSFT, line 1: Nit: UndoRoot reads like a command.  RecoilRootWithUndo?  UndoableRecoilRoot? There may not be a better name...

Composer/packages/form-dialogs/src/components/FormDialogPropertiesEditor.tsx
```diff
@@ -78,6 +80,7 @@ export const FormDialogPropertiesEditor = React.memo((props: Props) => {
 
   const schemaIdRef = React.useRef<string>(schema.id);
 
+  const { undo, redo, canUndo, canRedo } = useUndo();
```
GeoffCoxMSFT, line 83: You could return undo as undefined when you can't undo but maybe having the booleans is semantically clearer.

Composer/packages/form-dialogs/src/components/common/CommandBarUpload.tsx
```diff
@@ -27,6 +27,7 @@ export const CommandBarUploadButton = (props: Props) => {
   const onChange = () => {
     if (inputFileRef.current.files) {
```
GeoffCoxMSFT, line 28: Should this be current?.files  - current for a ref can be null right (although for a UI callback very unlikely)

Composer/packages/form-dialogs/src/undo/UndoRoot.tsx
```diff
@@ -0,0 +1,173 @@
...
+// Licensed under the MIT License.
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
```
GeoffCoxMSFT, line 5: This is a significant amount of work to support undo.  Maybe we can lift this up to shared across Composer for other plug-ins and extensions.

Composer/packages/form-dialogs/src/undo/UndoRoot.tsx
```diff
@@ -0,0 +1,173 @@
...
+  }
+  return atomMap;
+};
+
+const didAtomMapsChange = (prevMap: AtomMap, currMap: AtomMap) => {
+  if (prevMap.size !== currMap.size) {
+    return true;
+  }
+
+  for (const key of prevMap.keys()) {
```
GeoffCoxMSFT, line 31: Is it possible for the sizes to be the same, but then currMap has a key that prevMap does not?  You could use lodash intersection/union methods to determine if there are any key differences.

Composer/packages/form-dialogs/src/undo/UndoRoot.tsx
```diff
@@ -0,0 +1,173 @@
...
+  return current.map((pendingSnap) => {
+    for (const [atom, loadable] of atomMap.entries()) {
+      if (loadable.state === 'hasValue') {
+        pendingSnap.set(atom, loadable.contents);
+      }
+    }
+  });
+};
+
+type ContextState = {
```
GeoffCoxMSFT, line 73: Do you want to export the ContextState type so that callers can useContext with it?

GeoffCoxMSFT (dismissed):

Composer/packages/form-dialogs/src/undo/UndoRoot.tsx
hatpick: It's pretty much this one file.

Composer/packages/form-dialogs/src/undo/UndoRoot.tsx
hatpick: I think it's good enough.

Composer/packages/form-dialogs/src/components/common/CommandBarUpload.tsx
hatpick: If the ref is not set that's a major bug in react.

Composer/packages/form-dialogs/src/undo/UndoRoot.tsx
hatpick: Good catch, I added that there when the trackedAtom size was fixed.

GeoffCoxMSFT (approved):

PR #4336, opened 2020-10-07
refactor: add @bfc/types package

### LINKED ISSUE #4274: Add support for one publish extension registering multiple publish UIs
For example:

The Azure publish extension registers two publishing plugins:
  - azure web
  - azure functions

Each plugin wants to render its own UI. Composer should support this. Currently it renders 1 and 404s for the other.

The workaround for now is to split them up into their own extensions.

a-b-r-o-w-n:
## Description

Extracts shared types from `@bfc/shared` and `@bfc/extension*` into a single package that only exports types and interfaces. This package will eventually be published to npm and consumed by the extension packages.

The public interface of `@bfc/shared` does not change, as it re-exports the types.

## Task Item

closes #4274


Composer/packages/client/config/extensions.config.js
```diff
@@ -0,0 +1,103 @@
...
+      },
+      mode: isEnvProduction ? 'production' : 'development',
+      // export react-dom globally under a variable named ReactDOM
+      output: {
+        path: path.resolve(__dirname, '../public'),
+        library: 'ReactDOM',
+        libraryTarget: 'var',
+      },
+      externals: {
+        // ReactDOM depends on React, but we need this to resolve to the globally-exposed React variable in react-bundle.js (created by webpack-react.config.js).
```
tonyanziano, line 48: update or remove the `(created by webpack-react.config.js)` or so that it doesn't confuse people 😄

Composer/plugins/sample-ui-plugin/src/node/index.ts
```diff
@@ -5,7 +5,7 @@
 function initialize(registration) {
   const plugin = {
     customDescription: 'Publish using custom UI',
-    hasView: true /** we have custom UI to host */,
+    bundleId: 'publish' /** we have custom UI to host */,
```
tonyanziano, line 8: I like this a lot better 👍

Composer/packages/types/src/schema.ts
```diff
@@ -139,19 +139,27 @@ export interface DefinitionSummary {
...
 
-export interface OBISchema extends JSONSchema6 {
-  $schema?: string;
-  $role?: string;
-  $kind?: SDKKinds;
+export type SchemaDefinitions = {
+  [key: string]: AdaptiveSchema;
+};
+
+interface AdaptiveSchema extends Omit<JSONSchema7, 'definitions' | 'properties' | 'additionalProperties'> {
```
tonyanziano, line 146: whoa :O Had no idea `Omit` was a thing

Composer/packages/types/src/diagnostic.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export enum DiagnosticSeverity {
+  Error = 0,
+  Warning = 1,
+  Information = 2,
+  Hint = 3,
+}
+
+export interface IDiagnostic {
```
tonyanziano, line 21: convert these interfaces to types?

Composer/packages/client/src/pages/plugin/pluginPageContainer.tsx
```diff
@@ -6,10 +6,14 @@ import { RouteComponentProps } from '@reach/router';
...
 import { PluginHost } from '../../components/PluginHost/PluginHost';
 
-const PluginPageContainer: React.FC<RouteComponentProps<{ pluginId: string }>> = (props) => {
-  const { pluginId } = props;
+const PluginPageContainer: React.FC<RouteComponentProps<{ pluginId: string; bundleId: string }>> = (props) => {
+  const { pluginId, bundleId } = props;
 
-  return <PluginHost pluginName={pluginId} pluginType={'page'}></PluginHost>;
+  if (!pluginId || !bundleId) {
+    return <div />;
```
tonyanziano, line 13: would it be better to just return something falsy here instead of adding an extra div to the DOM?

tonyanziano (changes requested):
Minor comments. Other than that, great job, I think this is much cleaner

Composer/plugins/sample-ui-plugin/src/node/index.ts
a-b-r-o-w-n: This is to support multiple publish plugins (which I snuck into this PR).

Composer/packages/types/src/diagnostic.ts
a-b-r-o-w-n: This needs to be an interface because a class is implemented in `@bfc/shared`

tonyanziano: I looked into this after I saw those classes and it turns out you can actually implement types with classes:

```ts
type MyType = {
  someProp: boolean;
};

class MyClass implements MyType {
  someProp: boolean = true;
}
```

a-b-r-o-w-n: Oh nice! I'll update then.

tonyanziano (approved):
LGTM! 🚢

PR #4351, opened 2020-10-12
feat: Taking control of recognizer dialog generation and crosstrain config

### LINKED ISSUE #3370: Composer should write out cross train configuration
In order to deploy composer generated dialogs via CI/ CD pipeline (using BF-CLI), composer should write out a cross-train.config.json. This can sit as a peer to main.dialog (root dialog) file.

This enables composer customers to use CI/CD pipelines to build and publish their LU/ QnA applications. This also enables right integration of composer generated dialogs with a code based V4 SDK bot. 

Without this, we expect composer users to always use `(re)start bot` button within composer UI to get the LUIS and QnA Maker applications set up correctly.

### LINKED ISSUE #3811: Preserving *.lu.dialog files if they exist
Currently Composer will remove any *.lu.dialog files when running lubuild. If *.lu.dialog files exist the tooling should respect this file and merge the values it needs into it.

### LINKED ISSUE #4205: Composer overwrites recognizer settings
## Describe the bug

Recognizers have settings, in particular LUIS Recognizer has settings.  Composer overwrites the .lu.dialog file instead of maintaining the settings.

## Version

1.1.1

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Create a LU file
2. Change the settings on the lu file
3. starting the bot will blow away your settings
```
{
    "$kind": "Microsoft.LuisRecognizer",
    "id": "LUIS_msmeeting-actions",
    "applicationId": "=settings.luis.msmeeting_actions_en_us_lu",
    "endpoint": "=settings.luis.endpoint",
    "endpointKey": "=settings.luis.endpointKey",
    "predictionOptions": {
        "includeInstanceData": true
    },
}
```

THIS IS WHAT THE COMPOSER GENERATES:
```
{
    "$kind": "Microsoft.LuisRecognizer",
    "id": "LUIS_msmeeting-actions",
    "applicationId": "=settings.luis.msmeeting_actions_en_us_lu",
    "endpoint": "=settings.luis.endpoint",
    "endpointKey": "=settings.luis.endpointKey"
}
```
## Expected
1. It should not rewrite the file if it doesn't need to
2. It should maintain settings if it does.

### LINKED ISSUE #4259: Taking control of recognizer dialog generation and make composer authored bot CI/CD friendly
#### Context
Composer leverages bf-cli, in particularly, bf cross-train, bf luis:build and bf qna:build to do cross-train, luis and qna publishing. Besides that bf-cli does one more thing is generating those .dialog for recognizers, 3 type of recognizers usually 
* .\<locale\>-lu.dialog, which is a LUIS recognizer
* .lu.dialog which is a MultiLanguageLuis recognizer
* .lu.qna.dialog which is a CrossTrainedRecognizerSet

#### Issue
The issue is those "recognizer dialogs" are generating along with the build processing. It's been generated every time a publish happens and it leave user no control on that, for example user want to make some adjustment on LUIS recognizer https://github.com/microsoft/BotFramework-Composer/issues/4205. Because it will be overridden every time a publishing happens again. 

#### Solution
If we look at the recognizer dialogs, it's actually all fixed values or reference to settings, so conceptually it's not tighten to the build\publishing result. 

So in Composer's POV, it make sense to take control of this processing in composer. Composer will generate that first and support user to author those .dialog files even before a real publishing is happening. 

This will also relief bf-cli from have to do something as a side-car, and help keep cli focus, along with this change the cli can be purely 
* cross-train will take a config + .lu + .qna and generate cross-trained .lu and .qna files
* luis:build will take .lu and keys, and generate a .lu.settings as result
* qna:build will take .qna and keys, and generate a .qna.settings as result

none of those will have to touch recognizer dialogs.

#### Details
Composer will create a new folder called "recognizers" along with the dialogs\runtime\settings folder, and store those .dialogs and also the cross-train config.

Composer will generated those recognizer dialogs when recognizer is chosen (before publishing).

Composer will call bf-cli to do cross-train, luis publish, and qna publish and have the result in generated folder, which will now only contain cross-trained lu and qna files and the lu and qna settings.

This "recognizers" folder will be checked in as bot assets, this generated folder remains to be not checked in.

lei9444:
## Description
#4259

1. write out cross train configuration to the recognizer folder
2. write the generated dialog files to the folder.
3. preserving the luisRecognizer if they exist.

## Task Item
closes #4259 
closes #3811 
closes #3370
closes #4205


Composer/packages/lib/indexers/src/utils/dialogCheckUtil.ts
```diff
@@ -21,3 +22,15 @@ export function autofixReferInDialog(dialogId: string, content: string): string
     return content;
   }
 }
+
+export function getRecognizerType(recognizer): RecognizerType {
+  if (typeof recognizer === 'string') {
+    return 'DefaultRecognizer';
+  }
```
yeze322, line 29: We need to consider the custom recognizer case here

Composer/packages/types/src/indexers.ts
```diff
@@ -212,3 +213,5 @@ export interface BotProjectFile {
   content: BotProjectSpace;
   lastModified: string;
 }
+
+export type RecognizerType = 'DefaultRecognizer' | 'RegexRecognizer' | 'CustomRecognizer';
```
yeze322, line 217: I think that this type is not necessary here, we can use $kind as the constant which is compatible to the custom recognizer scenario

Composer/packages/lib/indexers/src/utils/dialogCheckUtil.ts
yeze322: FYI - useRecognizerConfig, ui-plugins/luis index.ts, ui-plugins/cross-train index.ts

cwhitten:
@lei9444 in an attempt to resolve conflicts, lint is failing with:

Error: [typecheck:server] src/models/bot/__tests__/botProject.test.ts(296,52): error TS2345: Argument of type '{ luisConfig: { authoringEndpoint: string; authoringKey: string; authoringRegion: string; defaultLanguage: string; endpoint: string; endpointKey: string; environment: string; name: string; }; ... 4 more ...; recognizerTypes: RecognizerTypes; }' is not assignable to parameter of type 'IBuildConfig'.

lei9444:
The azurePublish plugin has the build logic. I will remove the build from the plugin and guarantee the bot project build complete before publishing .

cwhitten (approved):

Issue #4150, opened 2020-09-15
Update navigation to display form dialog assets


GeoffCoxMSFT:
This is to extend the improve navigation with the tree of forms/properties/triggers and wire them to the correct page/editor.

PR #4353, opened 2020-10-12
feat: Added jsonSchema and trigger content

GeoffCoxMSFT:
## Description

In order to group triggers by property name, the JSON schema files referenced by dialog must be returned as assets to the bot from the server.  As well the trigger content needs to be included to inspect conditions and properties.

This first PR get the data to the client.

## Task Item

#minor
#4150 

## Screenshots
N/A


Composer/packages/lib/indexers/__tests__/jsonSchemafileIndexer.test.ts
```diff
@@ -0,0 +1,87 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { jsonSchemaFileIndexer } from '../src/jsonSchemaFileIndexer';
+import { FileInfo } from '../src/type';
+
+const { index } = jsonSchemaFileIndexer;
+
+describe('index', () => {
```
hatpick, line 9: better string here to indicate which indexer is running in logs

Composer/packages/lib/indexers/src/jsonSchemaFileIndexer.ts
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { FileInfo, JsonSchemaFile } from '@bfc/shared';
+import has from 'lodash/has';
+
+import { getBaseName } from './utils/help';
+
+const index = (jsonFiles: FileInfo[]) => {
+  return jsonFiles.reduce((jsonSchemaFiles: JsonSchemaFile[], { content, name }) => {
```
hatpick, line 10: since the input type is an array of fileinfo, that can be any file basically, it's better to filter for the interesting extension (i.e. .json) within the body of index.

It makes sure that the JSON.parse doesn't blow up

Composer/packages/lib/indexers/src/skillManifestIndexer.ts
```diff
@@ -2,14 +2,23 @@
...
 const index = (skillManifestFiles: FileInfo[]) => {
   return skillManifestFiles.reduce((manifests: SkillManifestInfo[], { content, name, lastModified }) => {
     try {
       const jsonContent = JSON.parse(content);
-      return [...manifests, { content: jsonContent, id: getBaseName(name, '.json'), lastModified }];
+
+      if (has(jsonContent, '$schema')) {
+        const schema = jsonContent.$schema.toLowerCase().trim();
+        if (schema.startsWith('https://schemas.botframework.com/schemas/skills')) {
+          return [...manifests, { content: jsonContent, id: getBaseName(name, '.json'), lastModified }];
```
hatpick, line 17: use FileExtensions enum instead of .json string

Composer/packages/server/src/models/bot/__tests__/botProject.test.ts
```diff
@@ -44,7 +44,7 @@ beforeEach(async () => {
 describe('init', () => {
   it('should get project successfully', () => {
     const project: { [key: string]: any } = proj.getProject();
-    expect(project.files.length).toBe(15);
+    expect(project.files.length).toBe(16);
```
hatpick, line 47: I don't understand these changes, can you elaborate?

Composer/packages/lib/indexers/src/jsonSchemaFileIndexer.ts
GeoffCoxMSFT: Agree, but none of the other indexers do that (yet).  I guess to allow for indexing multiple extensions that happen to contain JSON.

Composer/packages/lib/indexers/__tests__/jsonSchemafileIndexer.test.ts
```diff
@@ -0,0 +1,89 @@
...
+
+import { jsonSchemaFileIndexer } from '../src/jsonSchemaFileIndexer';
+import { FileInfo } from '../src/type';
+
+const { index } = jsonSchemaFileIndexer;
+
+describe('jsonSchemaFileIndexer', () => {
+  describe('index', () => {
+    it('should return json schema files', () => {
+      const input: FileInfo[] = [
```
hatpick, line 12: nit: Maybe move this to outside and reuse it in both of the cases?

Composer/packages/server/src/models/bot/__tests__/botProject.test.ts
GeoffCoxMSFT: We pick up JSON files now, so this added appsettings.json to the project files list.  The JSON files are filters to skills manifests and JSON schemas in the indexers so it won't hurt anything.

Composer/packages/lib/indexers/__tests__/jsonSchemafileIndexer.test.ts
GeoffCoxMSFT: The schemas are different in the different cases - one to ensure we include those that are schemas, one to ensure we exclude others.  I guess I could do it in one test, but it felt like 2 logical cases.

Composer/packages/types/src/indexers.ts
```diff
@@ -32,6 +33,7 @@ export type ITrigger = {
   displayName: string;
   type: string;
   isIntent: boolean;
+  content: any;
```
hatpick, line 36: why is this?

hatpick: ah never mind, it shows up as a part of FileInfo but its for Trigger.

hatpick (approved):

Issue #3956, opened 2020-08-31
Project Tree for Bot Project


srinaath:
- [x] Client-side URLs include both botProject and bot ids
- [x] Make indentation correct
- [x] Correct click behavior (or workarounds)
- [x] Display error state icons
- [x] Search whole tree for dialog/trigger names
- [x] handle opening bot projects

Implement the new project tree with error states on Design View of Composer

1. Screen state - https://www.figma.com/file/LlTlh5vXwq91zjGnFtrUUR6l/Composer-main-design-spec-UI-library?node-id=4511%3A30622
2. Left nav interactions - https://www.figma.com/file/LlTlh5vXwq91zjGnFtrUUR6l/Composer-main-design-spec-UI-library?node-id=4511%3A30879
3. Left nav contextual menu - https://www.figma.com/file/LlTlh5vXwq91zjGnFtrUUR6l/Composer-main-design-spec-UI-library?node-id=4511%3A32188
4. Errors and warnings, in-between states - https://www.figma.com/file/LlTlh5vXwq91zjGnFtrUUR6l/Composer-main-design-spec-UI-library?node-id=4511%3A33187
5. Errors and warnings, missing LUIS/QnA Key and endpoints - https://www.figma.com/file/LlTlh5vXwq91zjGnFtrUUR6l/Composer-main-design-spec-UI-library?node-id=4511%3A36872

PR #4361, opened 2020-10-13
feature: implement new ProjectTree design with popout menus

beyackle:
## Description

This swaps the Design View project tree to the new design, hopefully keeping all functionality intact while clearing space for the new form dialogs, added menu commands, and multi-skill bot handling.

## Task Item

refs #3956

## Screenshots

image


Composer/packages/server/src/locales/en-US.json
```diff
@@ -125,8 +125,8 @@
   "add_additional_url_bdfac25d": {
     "message": "Add additional URL"
   },
-  "add_alternative_phrasing_47234f77": {
-    "message": "add alternative phrasing"
+  "add_alternative_phrasing_17e0304c": {
```
srinaath, line 128: should this file be changed?

Composer/packages/client/src/components/AppComponents/RightPanel.tsx
```diff
@@ -5,12 +5,21 @@
 import { jsx, css } from '@emotion/core';
 import { useRecoilValue } from 'recoil';
 import { forwardRef } from 'react';
+// import formatMessage from 'format-message';
 
 import { RequireAuth } from '../RequireAuth';
 import { ErrorBoundary } from '../ErrorBoundary';
+import { Conversation } from '../Conversation';
+//import { ProjectTree } from '../ProjectTree/ProjectTree';
+//import { LeftRightSplit } from '../Split/LeftRightSplit';
```
srinaath, line 14: Can remove comments

Composer/packages/server/src/locales/en-US.json
beyackle: It looks like we have another l10n bubble. I'll get a new PR out to fix that first.

Composer/packages/client/src/components/AppComponents/RightPanel.tsx
beyackle: In this case, I don't want to. These are there to remind me where the stubs are for future work.

Composer/packages/client/src/recoilModel/atoms/appState.ts
```diff
@@ -30,6 +30,18 @@ export type CurrentUser = {
   sessionExpired: boolean;
 };
 
+export type PageMode =
+  | 'home'
```
srinaath, line 34: is there a reason why we are not using enums

PageMode = { home = 'home', design = 'design'....}?

Composer/packages/client/src/recoilModel/dispatchers/__tests__/qna.test.tsx
```diff
@@ -95,7 +95,7 @@ describe('QnA dispatcher', () => {
       });
     });
 
-    expect(renderedComponent.current.qnaFiles[0].content).toContain(content);
+    expect(renderedComponent.current.qnaFiles[0].content.replace(/\s/g, '')).toContain(content.replace(/\s/g, ''));
```
srinaath, line 98: is there a reason why this test changed?

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
```diff
@@ -92,97 +112,270 @@ function sortDialog(dialogs: DialogInfo[]) {
...
 
-  const sortedDialogs = useMemo(() => {
-    return sortDialog(dialogs);
-  }, [dialogs]);
+  const notificationMap: { [projectId: string]: { [dialogId: string]: Diagnostic[] } } = {};
+  for (const bot of projectCollection) {
+    notificationMap[bot.projectId] = {};
+    // this seems to be the only way to get these notifications
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    const dialogMessages = useRecoilValue(validateDialogSelectorFamily(bot.projectId));
```
srinaath, line 157: This goes against the hooks rules.  U can leverage the botprojectspace selector to get this value. You would just do

```
const botProjectspace = useRecoilValue(botProjectspaceselector) at the top.

Now on line 157 u can do this

const dialogmessage = botProjectspace.filter((projectId) => projecId === bot.projectId)
```

Composer/packages/client/src/recoilModel/atoms/appState.ts
beyackle: Good idea - I'll switch that over.

Composer/packages/client/src/pages/design/DesignPage.tsx
```diff
@@ -201,6 +203,11 @@ const DesignPage: React.FC<RouteComponentProps<{ dialogId: string; projectId: st
   useEffect(() => {
     if (location && props.dialogId && props.projectId) {
       const { dialogId, projectId } = props;
+
+      // TODO: swap to the commented-out block once we're working on skills for real
+      // let { skillId } = props;
+      // if (skillId == null) skillId = projectId;
```
srinaath, line 209:
```
 const { location, dialogId, projectId = '' } = props;
```

So i dont see references to the skillId being set from the props. Essentially, with this current Design Page how would skills be supported. It seems like the schema, shellApi are all going to be only in the context of the rootBot. Seems like it should be changed to

```javascript
const { location, dialogId, projectId = '', skillId } = props;
const currentProjectId = projectId || skillId

Now all the useRecoilValues would use something like useRecoilValue(schemasState(currentProjectId))
```

Composer/packages/client/src/components/ProjectTree/treeItem.tsx
```diff
@@ -173,16 +194,26 @@ const onRenderItem = (item: IOverflowSetItemProps) => {
           />
         )}
         {item.displayName}
+        {item.errorContent && (
```
srinaath, line 197: Are these are the error indicators inside the project tree.

@zhixzhan This might be useful for the work your doing

Composer/packages/client/src/components/ProjectTree/treeItem.tsx
```diff
@@ -113,58 +111,81 @@ const navItem = (isActive: boolean, isSubItemActive: boolean) => css`
...
   isSubItemActive?: boolean;
-  depth: number | undefined;
-  onDelete: (id: string) => void;
-  onSelect: (id: string) => void;
+  menu?: TreeMenuItem[];
+  onSelect?: (link: TreeLink) => void;
+  icon?: string;
+  dialogName?: string;
+  showProps?: boolean;
+  shiftOut?: number; // needed to make an outline look right; should be the size of the "details" reveal arrow
```
srinaath, line 149: wht does shiftOut mean? Better name to this prop?

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
```diff
@@ -92,97 +112,270 @@ function sortDialog(dialogs: DialogInfo[]) {
...
+  const dialogHasErrors = (dialog: DialogInfo) => {
+    notificationMap[currentProjectId][dialog.id].some((diag) => diag.severity === DiagnosticSeverity.Error);
+  };
 
-  const onRenderHeader = (props: IGroupHeaderProps) => {
-    const toggleCollapse = (): void => {
-      groupRef.current?.toggleCollapseAll(true);
-      props.onToggleCollapse?.(props.group!);
-      onSelect(props.group!.key);
+  const botHasErrors = (bot: BotInProject) => {
```
srinaath, line 176: nit: You could move it inside the selector. It would be cached once and recalculted only when dialogs change rather than being recalculated multiple times here

Composer/packages/client/src/components/ProjectTree/ExpandableNode.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  function handleClick(ev: MouseEvent) {
+    if ((ev.target as Element)?.tagName.toLowerCase() === 'summary') {
+      setOpen(!isOpen);
+    }
+    ev.preventDefault();
+  }
+
+  function handleKey(ev: KeyboardEvent) {
+    if (ev.key === 'Enter' || ev.key === 'Space') setOpen(!isOpen);
```
srinaath, line 37: Maybe use KeyCodes typescript interface?

ev.keyCode === Keycodes.Enter?

Composer/packages/client/__tests__/testUtils/sampleDialog.ts
```diff
@@ -0,0 +1,93 @@
+// Copyright (c) Microsoft Corporation.
```
srinaath, line 1: Maybe moving this file inside \_\_mocks folder rather than directly inside testUtils

Composer/packages/client/src/recoilModel/atoms/appState.ts
hatpick: I'm more in favor of union types. It makes debugging easier and you don't have to import them. We are in typescript land, why do you say it's error prone?

Composer/packages/client/src/components/ProjectTree/ExpandableNode.tsx
hatpick: keyCodes are deprecated, recommended is the way Ben has it:
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent

Composer/packages/client/src/components/ProjectTree/ExpandableNode.tsx
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+/** @jsx jsx */
+import { jsx, css } from '@emotion/core';
+import { useState, MouseEvent, KeyboardEvent } from 'react';
+
+type Props = {
+  children: JSX.Element;
```
hatpick, line 9: nit: React.ReactNode instead of JSX

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
```diff
@@ -55,28 +46,57 @@ const root = css`
...
-function createItem(trigger: ITrigger, index: number, warningContent: string) {
-  return {
-    ...trigger,
-    index,
-    warningContent,
-    displayName: trigger.displayName || getFriendlyName({ $kind: trigger.type }),
-  };
+export type TreeMenuItem = {
+  icon?: string;
+  label: string; // leave this blank to place a separator
```
hatpick, line 94: this is not discoverable, how about adding a kind: 'normal' | 'separator'?

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
```diff
@@ -55,28 +46,57 @@ const root = css`
...
+  action?: (link: TreeLink) => void;
```
hatpick, line 95: is this a click action? If yes, please name it accordingly. If not, shouldn't it be an array of actions?
```
type Action {
   label: string;
   onClick: (link: TreeLink) => void;
}
```

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
```diff
@@ -92,97 +112,270 @@ function sortDialog(dialogs: DialogInfo[]) {
...
   dialogs: DialogInfo[];
-  dialogId: string;
-  selected: string;
-  onSelect: (id: string, selected?: string) => void;
+  projectId: string;
+  name: string;
+  isRemote: boolean;
+};
+
+type IProjectTreeProps = {
```
hatpick, line 122: nit: Interfaces usually start with I (capital i), types are just PascalCase

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
```diff
@@ -92,97 +112,270 @@ function sortDialog(dialogs: DialogInfo[]) {
...
+  onSelect?: (link: TreeLink) => void;
+  onAllSelected?: () => void;
```
hatpick, line 124: nit: onSelectAll (to follow your other callbacks naming pattern)

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
hatpick: If you want to keep it here, please use useMemo

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
```diff
@@ -92,97 +112,270 @@ function sortDialog(dialogs: DialogInfo[]) {
...
+          showProps
+          icon={icons.DIALOG}
+          isSubItemActive={isLinkEqual(link, selectedLink)}
+          link={link}
+          menu={[
+            {
+              label: formatMessage('Remove this dialog'),
+              icon: 'Delete',
+              action: (link) => {
+                onDeleteDialog(link.dialogName ?? '');
```
hatpick, line 264: what does it mean to delete a dialog with blank name?

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
```diff
@@ -92,97 +112,270 @@ function sortDialog(dialogs: DialogInfo[]) {
...
+              },
+            },
+          ]}
+          shiftOut={showTriggers ? 0 : 28}
```
hatpick, line 268: nit: Please keep the number in a variable with a descriptive name, otherwise it seems like a magic number

hatpick:
**Function declarations:**
Sometimes arrow function style is used, and sometimes "function" keyword. Would you please stick to arrow functions?

Composer/packages/client/src/components/ProjectTree/treeItem.tsx
zhixzhan: @srinaath Sure, It helps. 
One more requirements from my side is a missing style.
image

Composer/packages/client/src/recoilModel/atoms/appState.ts
srinaath: It's just a personal preference. Since its a string enum atleast the debugging part would be the same. The reason i prefer enums is there are cases where I needed to enumerate over the enum. But I think union type or enum are fine in this case.

Composer/packages/client/src/components/ProjectTree/ExpandableNode.tsx
srinaath: I didn't realize that. I had always been using keycodes instead of the string. Good to know :)

Composer/packages/client/src/recoilModel/atoms/appState.ts
beyackle: After considering this, I think I'm going to stick with the union type. It really is just a type annotation that gets transpiled out, and we only ever need to use it for a single Recoil atom (and whatever reads from it).

Composer/packages/client/src/recoilModel/dispatchers/__tests__/qna.test.tsx
beyackle: I don't know why it happened, but the results were coming out with a space before each newline, so I decided to just squish all the whitespace out.

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
beyackle: Got it. Should be fixed now.

Composer/packages/client/src/pages/design/DesignPage.tsx
beyackle: I left skillId out of the props on purpose because, at this stage, we don't need it yet. This is a minimal change to get the new ProjectTree design in without overcomplicating things. The older branch I'm pulling stuff from did read skillId, just like you're suggesting.

Composer/packages/client/src/components/ProjectTree/treeItem.tsx
beyackle: Interesting. I won't make that change here, but if I could learn more about the specifics of this I'll make another PR for that. Are skills the only things that can be greyed out? How will the client know if a skill is broken/missing?

Composer/packages/client/src/components/ProjectTree/treeItem.tsx
beyackle: Yeah, I admit it's not a great name. It should be something like "forceIndent".

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
beyackle: I'm not sure about that. useMemo uses reference equality to check if it needs to update, and this is a live view of whether a bot or dialog has an error. If I type something into a trigger's actions that causes an error (like "=[", which doesn't parse), this produces error icons on the trigger and the dialog it belongs to. The work it's doing is only linear in the number of triggers as well, which I don't think is costly enough to warrant useMemo.

Composer/packages/client/__tests__/testUtils/sampleDialog.ts
beyackle: That's a good idea.

Composer/packages/client/src/components/ProjectTree/treeItem.tsx
srinaath: Lets discuss this in the evenings status sync

Composer/packages/client/src/components/ProjectTree/ExpandableNode.tsx
beyackle: 👍

Composer/packages/client/src/components/ProjectTree/ExpandableNode.tsx
```diff
@@ -0,0 +1,51 @@
...
+  padding-left: 12px;
+  padding-top: 6px;
+`;
+
+const nodeStyle = (depth: number) => css`
+  margin-left: ${depth * 16}px;
+`;
+
+export const ExpandableNode = ({ children, summary, detailsRef, depth = 0 }: Props) => {
+  const [isOpen, setOpen] = useState(true);
```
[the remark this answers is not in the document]
beyackle: I don't see that it makes that much difference, but okay.

Composer/packages/client/src/components/ProjectTree/ExpandableNode.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  function handleClick(ev: MouseEvent) {
```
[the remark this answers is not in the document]
beyackle: Not according to TypeScript. That's a standard HTML <summary> tag, and following the chain of type declarations that VSCode gives me leads to this line in the TS lib:
`onclick: ((this: GlobalEventHandlers, ev: MouseEvent) => any) | null;`

Composer/packages/client/src/pages/design/DesignPage.tsx
```diff
@@ -201,6 +203,11 @@ const DesignPage: React.FC<RouteComponentProps<{ dialogId: string; projectId: st
   useEffect(() => {
     if (location && props.dialogId && props.projectId) {
       const { dialogId, projectId } = props;
+
+      // TODO: swap to the commented-out block once we're working on skills for real
```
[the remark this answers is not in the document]
beyackle: Good idea. Will do.

Composer/packages/client/src/components/ProjectTree/ExpandableNode.tsx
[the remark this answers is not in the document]
hatpick: Screen Shot 2020-10-19 at 2 41 53 PM

Composer/packages/client/src/pages/design/DesignPage.tsx
```diff
@@ -455,7 +462,7 @@ const DesignPage: React.FC<RouteComponentProps<{ dialogId: string; projectId: st
   function handleBreadcrumbItemClick(_event, item) {
     if (item) {
       const { dialogId, selected, focused, index } = item;
-      selectAndFocus(projectId, dialogId, selected, focused, clearBreadcrumb(breadcrumb, index));
+      selectAndFocus(projectId, null, dialogId, selected, focused, clearBreadcrumb(breadcrumb, index));
```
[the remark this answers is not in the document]
beyackle: That's a good idea, but refactoring all of these should be its own PR later (which I don't being assigned to).

Composer/packages/client/src/pages/design/DesignPage.tsx
```diff
@@ -528,31 +535,31 @@ const DesignPage: React.FC<RouteComponentProps<{ dialogId: string; projectId: st
...
+      updateDialog({ id: dialogId, content, projectId });
       const match = /\[(\d+)\]/g.exec(selected);
       const current = match && match[1];
       if (!current) return;
       const currentIdx = parseInt(current);
       if (index === currentIdx) {
         if (currentIdx - 1 >= 0) {
           //if the deleted node is selected and the selected one is not the first one, navTo the previous trigger;
-          selectTo(projectId, createSelectedPath(currentIdx - 1));
+          selectTo(projectId, null, dialogId, createSelectedPath(currentIdx - 1));
```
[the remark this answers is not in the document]
beyackle: Same reply.

Composer/packages/client/src/components/ProjectTree/ExpandableNode.tsx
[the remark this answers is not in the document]
hatpick: In the context of react, these are SyntheticEvent. However, as long as the type check is happy I'm fine.

Composer/packages/client/src/recoilModel/dispatchers/application.ts
```diff
@@ -68,6 +75,10 @@ export const applicationDispatcher = () => {
     set(announcementState, message);
   });
 
+  const setCurrentMode = useRecoilCallback(({ set }: CallbackInterface) => (mode: PageMode) => {
```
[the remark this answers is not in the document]
beyackle: Done.

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
```diff
@@ -55,28 +46,57 @@ const root = css`
...
+  isRoot: boolean;
+  warningContent?: string;
+  errorContent?: string;
+  projectId: string;
+  skillId: string | null;
+  dialogName?: string;
+  trigger?: number;
+};
+
+function isLinkEqual(link1: TreeLink | undefined, link2: TreeLink | undefined) {
```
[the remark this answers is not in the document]
beyackle: Only because I tend to forget that exists. Changed.

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
beyackle: No, you're right, it should be `onClick`.

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
beyackle: I almost named it that, but that reads to me like there's some kind of "select-all" action  - as in "select everything in the tree somehow" when what I really mean is that the literal link named "All" got selected.

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
beyackle: Fixed (my usual convention is to just name this `Props`, because we already know what component we're looking at the props of - it's the file name).

hatpick: sure thing, if it's not exported Props will do just fine.

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
beyackle: I think I'll call this `onSelectAllLink`. That's different enough from "select all" that I think it reads better as "select the 'all' link".

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
beyackle: That's only there to make type-checking happy; this is inside the `renderDialogHeader` function, so we know that link.dialogName is never going to be undefined here, but TS doesn't necessarily know that.

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
beyackle: Good call. Will fix this.

cwhitten (approved):

PR #4368, opened 2020-10-14
feat: Update LG library

### LINKED ISSUE #3982: LG namespace option should not be required
Expected, use exports in LG file, namespace should not be a required field, should be an optional field. Or at least the error message should be clear about it. 

Actual: Error pop up when there is only exports without namespace, and the error message is not clear
image

Danieladu:
## Description
- Update LG library to the latest version. (10/14/2020) In which include the fix of #3982. see PRs below

|C# Code | js Code|
|-|-|
|[PR link](https://github.com/microsoft/botbuilder-dotnet/pull/4724)  |[PR link](https://github.com/microsoft/botbuilder-js/pull/2882)|

- Decouple LG ImportResolver and LU ImportResolver
- Update [ImportResolverDelegate](https://github.com/microsoft/botbuilder-js/blob/69cbd1d68d96f4e00aab255935439ae025a220f7/libraries/botbuilder-lg/src/templatesParser.ts#L33) from 
```
(sourceId: string, resourceId: string) -> { content: string; id: string }
```
 to
```
(lgResource:LGResource) => LGResource
```

## Task Item
closes #3982


boydc2014:
@zhixzhan can you help review this one?

zhixzhan:
> @zhixzhan can you help review this one?

Sure, I will

Composer/packages/lib/indexers/src/utils/lgUtil.ts
```diff
@@ -259,8 +259,8 @@ export function extractOptionByKey(nameOfKey: string, options: string[]): string
   return result;
 }
 
-export function parse(id: string, content: string, lgFiles: TextFile[]): LgFile {
-  const lgImportResolver = importResolverGenerator(lgFiles, '.lg');
+export function parse(id: string, content: string, lgResources: LGResource[]): LgFile {
```
zhixzhan, line 262: Can we create LGResource inside this methods, then for caller, needs less knowledge,  still pass lgFiles:{id, content}[]. I don't have strong opinions on this.

Composer/packages/client/src/recoilModel/parsers/workers/lgParser.worker.ts
```diff
@@ -91,8 +92,11 @@ type LgMessageEvent =
 
 type LgResources = Map<string, LgFile>;
 
-const lgFileResolver = (lgFiles) => {
-  return importResolverGenerator(lgFiles, '.lg');
+const lgFileResolver = (lgFiles: LgFile[]) => {
+  return lgImportResolverGenerator(
+    lgFiles.map((u) => new LGResource(u.id, u.id, u.content)),
```
zhixzhan, line 97: if create LGResource in lgUtil.parse , here can have no change.

zhixzhan:
I have tested few LG scenarios inline/allupview/listview/diagnostics, and run in emulator, it E2E works.

boydc2014 (dismissed):

cwhitten (approved):

PR #4384, opened 2020-10-15
feat: bf-orchestrator lib integration

### LINKED ISSUE #4132: Bf-orchestrator lib integration
Use this to track the Orchestrator lib integration.  

The deliverable is a module\plugin\extension with proper test cases that can be called by composer's authoring UX.  

We need to make sure the lib integration is not coupled with IO, which means the interface's input and output should be pure memory objects (instead of file paths) and the library itself wouldn't perform side effects IO operations.

taicchoumsft:
## Description

Integrate the Orchestrator nodejs library, which provides interfaces for working with the upcoming transformer-based [Orchestrator Recognizer](https://github.com/microsoft/BotBuilder-Samples/tree/main/experimental/orchestrator). 

This PR exposes these interfaces for upper layers in Composer to use: 
- List available NLR models 
- Download an NLR model, and
- Build a snapshot

## Task Item
fixes #4132


Composer/packages/server/jest.config.js
```diff
@@ -4,5 +4,5 @@ const { createConfig } = require('@bfc/test-utils');
 
 module.exports = createConfig('server', 'node', {
   setupFiles: [path.resolve(__dirname, 'src/__tests__/setupEnv.ts')],
-  testPathIgnorePatterns: ['src/__tests__/setupEnv.ts'],
+  testPathIgnorePatterns: ['src/__tests__/setupEnv.ts', 'orchestratorBuilder'],
```
taicchoumsft, line 7: I've added this UT to the ignore list for now since it does a model download at the start of all tests (180MB+) which is needed for other tests. I am aware is a codesmell, any suggestions appreciated. 

Orchestrator team have mentioned there is a smaller model that can be made available for UTs in the future.

Composer/packages/server/src/models/bot/__tests__/orchestratorBuilder.test.ts
```diff
@@ -0,0 +1,84 @@
...
+    // disable Orchestrator Lib console logging across the board - interferes with Jest
+    Utility.toPrintDebuggingLogToConsole = false;
+
+    const progressStatusStub = jest.fn();
+    const successStatusStub = jest.fn();
+
+    let builder = new Builder('', {} as IFileStorage, 'en-us');
+
+    // download the UT NLR model once before all tests are run - build tests don't work without a valid model
+    await builder.runOrchestratorNlrGet(nlrPath, nlrId, progressStatusStub, successStatusStub);
```
taicchoumsft, line 28: I've added this UT to the jest ignore list for now, since a 180MB+ binary file download is required to run this UT.  I'm aware this is a code-smell, any suggestions appreciated. 

Orchestrator team have mentioned making a small model available for testing purposes in the future.

boydc2014:
@lei9444  can you help check this one?

Composer/packages/server/package.json
```diff
@@ -65,6 +65,7 @@
     "@microsoft/bf-dispatcher": "^4.10.0-preview.141651",
     "@microsoft/bf-generate-library": "^4.10.0-daily.20201008.172736",
     "@microsoft/bf-lu": "^4.11.0-dev.20201005.7e5e1b8",
+    "@microsoft/bf-orchestrator": "^4.11.0-beta.20201013.20d7917",
```
tsuwandy, line 68: "@microsoft/bf-orchestrator": "beta"

This will get latest from beta when dependencies are refreshed unless you want composer to pin the orchestrator library to a specific version?

tsuwandy (approved):

Composer/packages/server/package.json
boydc2014: yes, in Composer even we specify the version here with "latest" it will always be pinned into a specific version, because we are using yarn.lock to save the version, which is considered as the best practise to make sure the product is stable.

Composer/packages/server/src/models/bot/builder.ts
```diff
@@ -98,6 +102,59 @@ export class Builder {
...
+    modelPath: string,
+    nlrId: string,
+    onProgress: IOrchestratorProgress,
+    onFinish: IOrchestratorProgress
+  ): Promise<void> {
+    await Orchestrator.nlrGetAsync(modelPath, nlrId, onProgress, onFinish);
+  }
+
+  /**
+   * Orchestrator: Build command to compile .lu files into Binary LU (.blu) snapshots.
```
feich-ms, line 132: >.lu [](start = 44, length = 3)

Tien mentioned that qna files are also supported built into blu snapshots, so I think .qna files should also be included.

Composer/packages/server/src/models/bot/builder.ts
```diff
@@ -98,6 +102,59 @@ export class Builder {
...
+   * @returns An object containing snapshot bytes and recognizer dialogs for each .lu file
+   */
+  public async runOrchestratorBuild(
+    files: FileInfo[],
+    modelPath: string,
+    isDialog = true,
+    fullEmbedding = false
+  ): Promise<IOrchestratorBuildOutput> {
+    const luObjects = files
+      .filter((fi) => fi.name.endsWith('.lu'))
```
feich-ms, line 149: >.lu [](start = 40, length = 3)

Same here. Maybe we also need to filter .qna files

Composer/packages/server/src/models/bot/builder.ts
```diff
@@ -98,6 +102,59 @@ export class Builder {
...
+    onProgress: IOrchestratorProgress,
+    onFinish: IOrchestratorProgress
+  ): Promise<void> {
+    await Orchestrator.nlrGetAsync(modelPath, nlrId, onProgress, onFinish);
+  }
+
+  /**
+   * Orchestrator: Build command to compile .lu files into Binary LU (.blu) snapshots.
+   *
+   * A snapshot (.blu file) is created per .lu supplied
```
feich-ms, line 134: >.lu [](start = 43, length = 3)

same here.

Composer/packages/server/src/models/bot/builder.ts
```diff
@@ -98,6 +102,59 @@ export class Builder {
...
+   *
+   * @param files - Array of FileInfo
+   * @param modelPath - Path to NLR model folder
+   * @param isDialog - Flag to toggle creation of Recognizer Dialogs (default: true)
+   * @param fullEmbedding - Use larger embeddings and skip size optimization (default: false)
+   * @returns An object containing snapshot bytes and recognizer dialogs for each .lu file
```
feich-ms, line 140: >.lu [](start = 82, length = 3)

same here.

Composer/packages/server/src/models/bot/builder.ts
lei9444: Do we need to filter the files with the recognizer type now? The build function parameter luFiles is referred by CrossTrainedRecognizer and LuisRecognizer. Maybe we can separate the files by recognizer type in the future.

boydc2014 (approved):

Composer/packages/server/src/models/bot/builder.ts
boydc2014: I'm not sure about qna support in orchestrator. Have you tried that @feich-ms?

Composer/packages/server/src/models/bot/builder.ts
boydc2014: I think we will filter it later maybe in the caller side of this function. Not in here. This PR's focus should be bring the lib in, and with proper test case to verify the lib is working as expected.

Composer/packages/server/src/models/bot/builder.ts
boydc2014: Get it, saw the samples, the qna support just means it can take the questions as intent "QnaMaker", it's not the way we expected to integrate with Qna service or AdaptiveQnaRecoginzer. So for now, composer should only deal with .lu.

boydc2014:
@taicchoumsft this looks good now, can you address the conflict?

taicchoumsft:
> @taicchoumsft this looks good now, can you address the conflict?

Hi @boydc2014, I've addressed the conflict in orchestrator-core, verified that the orchestrator UTs are running correctly in our Linux based CI pipeline, and then for developer convenience, disabled the orchestrator tests again until Orchestrator team provides a local model.

boydc2014:
> > @taicchoumsft this looks good now, can you address the conflict?
> 
> Hi @boydc2014, I've addressed the conflict in orchestrator-core, verified that the orchestrator UTs are running correctly in our Linux based CI pipeline, and then for developer convenience (because we d/l a large model every time), disabled the orchestrator tests again until Orchestrator team provides a local model.

Cool, it's ready to take once the CI is green, i'm running the Docker build validation now.

cdonke:
Changed the dockerfile base image to ubuntu on #4516 .

Hope it helps.

boydc2014 (approved):

PR #4389, opened 2020-10-15
feat: New adaptive forms

### LINKED ISSUE #4181: Merge ExpressionField and OneOfField
Part of improving expression inputs

### LINKED ISSUE #4182: Update OneOfField to latest designs (icon dropdown at beginning of field)
Part of improving expression inputs

### LINKED ISSUE #4183: Update individual fields (BooleanField, etc..) to latest designs
Part of improving expression inputs

### LINKED ISSUE #4184: Make expressions separate from type selection, leveraging Intellisense for discoverability
Part of improving expression inputs

LouisEugeneMSFT:
## Description

The key change of this PR is **abstracting away expressions from the basic data types**.
For example, instead of having to pick between string, number or expression, users now pick between string and number and can decide to toggle to expression within the field.
The advantages of this are multiple, among which:
- greatly decluttering the ui since most fields were a choice between one data type and expression.
- ability to now validate the return type of expressions (eg: user decides "this is a string" and types an expression, we can now validate the return type of that expression).

Additionally those changes were an opportunity to **redesign** the look and feel of the fields and field selection ("OneOf").

From: 
image

To:
image

Additionally this PR contains some "**code cleanup**" such as a deduping of OneOfField and ExpressionField (which were mostly duplicate of each other), resolving JsonField inside resolveFieldWidget and some optimizations

## Task Item

fixes #4181 
fixes #4182 
fixes #4183 
fixes #4184 

## Screenshots

New UI:
image

New type selection:
image

Leveraging Intellisense as a way to switch to expression (from TextField):
image

Leveraging Intellisense as a way to switch to expression (from Monaco editor):
image

Leveraging Intellisense as a way to switch to expression (from a dropdown):
image

Nested objects still need a dropdown for toggling to expression:
image


a-b-r-o-w-n:
@LouisEugeneMSFT does this PR close those referenced issues? If so, can you use the syntax that will automatically close them? https://docs.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords

Composer/packages/adaptive-form/src/components/ExpressionSwitchWindow.tsx
```diff
@@ -0,0 +1,51 @@
...
+      <div css={styles.fieldTypeText}>{`Start typing ${type} or`}</div>
+
+      <div
+        css={styles.switchToExpressionText}
+        onClick={() => {
+          onSwitchToExpression();
+        }}
+      >
+        <FontIcon css={styles.icon} iconName={'CalculatorEqualTo'} />
+        {`Write in expression`}
```
cwhitten, line 47: formatMessage

hatpick:
@LouisEugeneMSFT is it possible to remove the right border of the left gray div and the left border of the right element (as well as border radius) to have a more smooth look such as this:

adaptive form sug

LouisEugeneMSFT:
@hatpick the example you showed is a built in functionality of fabric TextFields:
image

In the case of this PR, the components are separate.

- On the left it can either be a dropdown (in OneOfField.tsx) or a simple icon (in withTypeIcons.tsx).
- On the right it could be any field (TextField but also dropdowns, Monaco editor, etc...)

The two solutions I had in mind to achieve the desired design were:

1. Individually overwrite all components to remove the border (either left or right depending on which component)
2. (or) Only overwrite two components (the left ones) and play with zindex and absolute position to cover the border of the right components

1 would be introducing some logic to the leaf fields that shouldn't be there. If we ever want to change the design, a dozen components would have to be modified. Any new field would have to know about the border policy.

2 would be my prefered solution but I see it as "hacky".

If you have any ideas on how to do this in a clean way, please let me know :)

hatpick:
> @hatpick the example you showed is a built in functionality of fabric TextFields:
> image
> 
> In the case of this PR, the components are separate.
> 
> * On the left it can either be a dropdown (in OneOfField.tsx) or a simple icon (in withTypeIcons.tsx).
> * On the right it could be any field (TextField but also dropdowns, Monaco editor, etc...)
> 
> The two solutions I had in mind to achieve the desired design were:
> 
> 1. Individually overwrite all components to remove the border (either left or right depending on which component)
> 2. (or) Only overwrite two components (the left ones) and play with zindex and absolute position to cover the border of the right components
> 
> 1 would be introducing some logic to the leaf fields that shouldn't be there. If we ever want to change the design, a dozen components would have to be modified. Any new field would have to know about the border policy.
> 
> 2 would be my prefered solution but I see it as "hacky".
> 
> If you have any ideas on how to do this in a clean way, please let me know :)

The snippet I included in my comment is just for visual reference, I know how you've implemented it and that's fine, all I'm saying is to remove border radius and left and right borders of the right and left elements to mimic this visual.

Composer/packages/adaptive-form/src/components/ExpressionSwitchWindow.tsx
```diff
@@ -0,0 +1,52 @@
...
+    margin-right: 5px;
+  `,
+};
+
+export const ExpressionSwitchWindow = (props: { type: string; onSwitchToExpression: () => void }) => {
+  const { type, onSwitchToExpression } = props;
+  return (
+    <React.Fragment>
+      <div css={styles.fieldTypeText}>{`Start typing ${type} or`}</div>
+
```
GeoffCoxMSFT, line 40: Nit: extra empty line

Composer/packages/adaptive-form/src/components/ExpressionSwitchWindow.tsx
```diff
@@ -0,0 +1,52 @@
...
+    display: flex;
+    align-items: center;
+    font-size: 15px;
+  `,
+  icon: css`
+    margin-right: 5px;
+  `,
+};
+
+export const ExpressionSwitchWindow = (props: { type: string; onSwitchToExpression: () => void }) => {
```
GeoffCoxMSFT, line 35: Nit: It is much cleaner to define a type Props rather than have it inline, even when only 2 properties.

Composer/packages/adaptive-form/src/components/ExpressionSwitchWindow.tsx
```diff
@@ -0,0 +1,52 @@
...
+import formatMessage from 'format-message';
+
+const styles = {
+  fieldTypeText: css`
+    height: 32px;
+    font-style: italic;
+    color: #8a8886;
+    padding: 0 4px;
+    width: 100%;
+    display: flex;
```
GeoffCoxMSFT, line 17: Why do you need flex here?

Composer/packages/adaptive-form/src/components/ExpressionSwitchWindow.tsx
```diff
@@ -0,0 +1,52 @@
...
+    align-items: center;
+    font-size: 15px;
+  `,
+  switchToExpressionText: css`
+    height: 32px;
+    cursor: pointer;
+    padding: 0 4px;
+    width: 100%;
+    display: flex;
```
GeoffCoxMSFT, line 26: Same question about needing flex here.

Composer/packages/adaptive-form/src/components/ExpressionSwitchWindow.tsx
```diff
@@ -0,0 +1,52 @@
...
+      <div css={styles.fieldTypeText}>{`Start typing ${type} or`}</div>
+
+      <div
+        css={styles.switchToExpressionText}
+        onClick={() => {
+          onSwitchToExpression();
+        }}
+      >
+        <FontIcon css={styles.icon} iconName={'CalculatorEqualTo'} />
+        {formatMessage(`Write in expression`)}
```
GeoffCoxMSFT, line 48: Write expression or Write an expression - Write in expression sounds awkward.  Usually you fill in something, not write in - except when voting.

Composer/packages/adaptive-form/src/components/SchemaField.tsx
```diff
@@ -83,6 +95,11 @@ const SchemaField: React.FC<FieldProps> = (props) => {
     schema,
     uiOptions,
     value: deserializedValue,
+    onFocus: () => setFieldFocused(true),
+    onBlur: () => {
+      setFieldFocused(false);
+      rest.onBlur && rest.onBlur(rest.id);
```
GeoffCoxMSFT, line 101: This looks strange to me.  You spread rest in the fieldProps, but then are assuming it has onBlur here.  What if rest doesn't have onBlur?  It would be nice to see if rest conformed to a HTML element type that has onBlur.

Composer/packages/adaptive-form/src/components/fields/BooleanField.tsx
```diff
@@ -28,9 +28,38 @@ const BooleanField: React.FC<FieldProps> = function CheckboxWidget(props) {
...
+  if (expression) {
+    options.push(
+      {
+        key: 'divider',
+        text: '-',
+        itemType: DropdownMenuItemType.Divider,
+      },
+      {
+        key: 'expression',
+        text: formatMessage('Write in expression'),
```
GeoffCoxMSFT, line 40: Same comment as before

Composer/packages/adaptive-form/src/components/fields/BooleanField.tsx
```diff
@@ -28,9 +28,38 @@ const BooleanField: React.FC<FieldProps> = function CheckboxWidget(props) {
...
+        case 'true':
+          optionValue = true;
+          break;
+        case 'false':
+          optionValue = false;
+          break;
+        case 'expression':
+          optionValue = '=';
+          break;
+      }
```
GeoffCoxMSFT, line 61: Because optionValue is multiple possible types, I would have a default: case that either throws and error or assigns a default value.  It isn't super clear what optionValue ends up being if no case matches: false or undefined.

Composer/packages/adaptive-form/src/components/fields/IntellisenseFields.tsx
```diff
@@ -0,0 +1,173 @@
...
+        />
+      )}
+    </Intellisense>
+  );
+};
+
+export const IntellisenseJSONField: React.FC<FieldProps<string>> = function IntellisenseJSONField(props) {
+  const { id, value = '', onChange, focused: defaultFocused, schema } = props;
+
+  const completionListOverrideResolver = (value: any) => {
```
GeoffCoxMSFT, line 126: Nit:  You are overriding the completionListResolver so it would probably be better named completionListResolverOverride or overrideCompletionListResolver.  This makes it sound like it is resolving completionListOverrides rather than overriding completionListResolvers :)

Composer/packages/adaptive-form/src/components/fields/IntellisenseFields.tsx
```diff
@@ -0,0 +1,173 @@
...
+  };
+
+  const defaultValue = schema.type === 'object' ? {} : [];
+
+  return (
+    <Intellisense
+      completionListOverrideResolver={completionListOverrideResolver}
+      focused={defaultFocused}
+      id={`intellisense-${id}`}
+      scopes={['expressions']}
```
GeoffCoxMSFT, line 156: This allocates a new array every single render which will trigger a any effects on scopes to fire within Intellisense..  Put constants into const variables outside render.

Composer/packages/adaptive-form/src/components/fields/IntellisenseFields.tsx
```diff
@@ -0,0 +1,173 @@
...
+      value={value || defaultValue}
+      onBlur={props.onBlur}
+      onChange={onChange}
+    >
+      {({ textFieldValue, onValueChanged }) => (
+        <JsonField
+          {...props}
+          style={{ height: 100 }}
+          value={textFieldValue}
+          onBlur={undefined} // onBlur managed by Intellisense
```
GeoffCoxMSFT, line 167: This is strange, why not remove specifying onBlur?

Composer/packages/adaptive-form/src/components/fields/IntellisenseFields.tsx
```diff
@@ -0,0 +1,173 @@
...
+    }
+  };
+
+  return (
+    <Intellisense
+      completionListOverrideResolver={completionListOverrideResolver}
+      focused={defaultFocused}
+      id={`intellisense-${id}`}
+      scopes={uiOptions.intellisenseScopes || []}
+      url={getIntellisenseUrl()}
```
GeoffCoxMSFT, line 32: You want to do this every render?

Composer/packages/adaptive-form/src/components/fields/IntellisenseFields.tsx
LouisEugeneMSFT: @GeoffCoxMSFT, this is because Intellisense has its own onBlur that looks at both the field and the Intellisense panel. If we don't do undefined on the children then clicking on an Intellisense result would trigger onBlur on the textfield and we would loose focus.

Composer/packages/adaptive-form/src/components/fields/IntellisenseFields.tsx
LouisEugeneMSFT: @GeoffCoxMSFT, I believe React is smart enough and do not rerender unless those values changes. Do you see a way to optimize this?

Composer/packages/adaptive-form/src/components/fields/IntellisenseFields.tsx
LouisEugeneMSFT: Okay!

Composer/packages/adaptive-form/src/components/fields/IntellisenseFields.tsx
LouisEugeneMSFT: It is actually intended. Based on the current value the _**completionListOverride**_ is constructed as such: 
image

Composer/packages/adaptive-form/src/components/fields/BooleanField.tsx
LouisEugeneMSFT: The options are listed above. The switch only go over those, there shouldn't be _no case_ match. If someone inevertently change the code to have a mismatch between options and handleChange, optionValue would be undefined which is a valid value to pass up to the parent handler.

image

Composer/packages/adaptive-form/src/components/ExpressionSwitchWindow.tsx
LouisEugeneMSFT: This is what's in the designs: 
https://www.figma.com/file/ISGuziIavuIZdwOHTx3G1u/Property-field?node-id=1%3A1726
image
@sangwoohaan

Composer/packages/adaptive-form/src/components/SchemaField.tsx
LouisEugeneMSFT: I am not 100% sure I understand what you mean here. I deconstructed onBlur on the props to leave _rest_ out of this. Was that what you had in mind?

Composer/packages/adaptive-form/src/components/ExpressionSwitchWindow.tsx
LouisEugeneMSFT: ok

Composer/packages/adaptive-form/src/components/ExpressionSwitchWindow.tsx
LouisEugeneMSFT: ok

Composer/packages/adaptive-form/src/components/ExpressionSwitchWindow.tsx
LouisEugeneMSFT: @GeoffCoxMSFT, in order to address this:
image
I could wrap the two divs inside another div but this would add unnecessary complexity to the dom

Composer/packages/adaptive-form/src/components/ExpressionSwitchWindow.tsx
```diff
@@ -0,0 +1,55 @@
...
+    display: flex;
+    align-items: center;
+    font-size: 15px;
+  `,
+  icon: css`
+    margin-right: 5px;
+  `,
+};
+
+type ExpressionSwitchWindowProps = { type: string; onSwitchToExpression: () => void };
```
hatpick, line 35: type is a reserved keyword, better not to use it: kind?

Composer/packages/adaptive-form/src/components/ExpressionSwitchWindow.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+export const ExpressionSwitchWindow = (props: ExpressionSwitchWindowProps) => {
+  const { type, onSwitchToExpression } = props;
+
+  return (
+    <React.Fragment>
+      <div css={styles.fieldTypeText}>{`Start typing ${type} or`}</div>
```
hatpick, line 42: formatMessage

Composer/packages/adaptive-form/src/components/ExpressionSwitchWindow.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+      <div
+        css={styles.switchToExpressionText}
+        onClick={() => {
+          onSwitchToExpression();
```
hatpick, line 47: no need to create another function, just use onSwitchToExpression

Composer/packages/adaptive-form/src/components/fields/BooleanField.tsx
```diff
@@ -28,9 +28,38 @@ const BooleanField: React.FC<FieldProps> = function CheckboxWidget(props) {
     },
   ];
 
+  if (expression) {
+    options.push(
+      {
+        key: 'divider',
+        text: '-',
```
hatpick, line 35: unnecessary

Composer/packages/adaptive-form/src/components/fields/BooleanField.tsx
```diff
@@ -28,9 +28,38 @@ const BooleanField: React.FC<FieldProps> = function CheckboxWidget(props) {
...
+        key: 'expression',
+        text: formatMessage('Write in expression'),
+      }
+    );
+  }
+
   const handleChange = (e, option?: IDropdownOption) => {
     if (option) {
-      const optionValue = option.key === 'none' ? undefined : option.key === 'true';
+      let optionValue: boolean | string | undefined;
```
hatpick, line 47: This looks weird, couldn't you use a union type?
what is the boolean representing?

Composer/packages/adaptive-form/src/components/fields/IntellisenseFields.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+import { JsonField } from './JsonField';
+import { NumberField } from './NumberField';
+import { StringField } from './StringField';
+
+export const IntellisenseTextField: React.FC<FieldProps<string>> = function IntellisenseTextField(props) {
+  const { id, value = '', onChange, uiOptions, focused: defaultFocused } = props;
+
+  const completionListOverrideResolver = (value: string) => {
+    if (value === '') {
```
hatpick, line 19: nit: ```value === '' ? <ExpressionSwitchWindow type={'String'} onSwitchToExpression={() => onChange('=')} />: null```

Composer/packages/adaptive-form/src/components/fields/IntellisenseFields.tsx
```diff
@@ -0,0 +1,184 @@
...
+      onBlur={props.onBlur}
+      onChange={onChange}
+    >
+      {({ textFieldValue, focused, onValueChanged, onKeyDownTextField, onKeyUpTextField, onClickTextField }) => (
+        <StringField
+          {...props}
+          focused={focused}
+          id={id}
+          value={textFieldValue}
+          onBlur={undefined} // onBlur managed by Intellisense
```
hatpick, line 46: make it optional?

Composer/packages/adaptive-form/src/components/fields/IntellisenseFields.tsx
```diff
@@ -0,0 +1,184 @@
...
+          onClick={onClickTextField}
+          onKeyDown={onKeyDownTextField}
+          onKeyUp={onKeyUpTextField}
+        />
+      )}
+    </Intellisense>
+  );
+};
+
+export const IntellisenseNumberField: React.FC<FieldProps<string>> = function IntellisenseNumberField(props) {
```
hatpick: use an arrow function

Composer/packages/adaptive-form/src/components/fields/IntellisenseFields.tsx
```diff
@@ -0,0 +1,184 @@
...
+  const { id, value = '', onChange, uiOptions, focused: defaultFocused } = props;
+
+  const completionListOverrideResolver = (value: string) => {
+    if (value === '') {
```
hatpick, line 94: same nit

Composer/packages/adaptive-form/src/components/fields/IntellisenseFields.tsx
```diff
@@ -0,0 +1,184 @@
...
+          onClick={onClickTextField}
+          onKeyDown={onKeyDownTextField}
+          onKeyUp={onKeyUpTextField}
+        />
+      )}
+    </Intellisense>
+  );
+};
+
+export const IntellisenseJSONField: React.FC<FieldProps<string>> = function IntellisenseJSONField(props) {
```
hatpick: same comment as above

Composer/packages/adaptive-form/src/components/fields/IntellisenseFields.tsx
```diff
@@ -0,0 +1,184 @@
...
+          }}
+        />
+      );
+    }
+    return null;
+  };
+
+  const defaultValue = schema.type === 'object' ? {} : [];
+  const scopes = ['expressions'];
+  const url = getIntellisenseUrl();
```
hatpick, line 160: make this a ref

Composer/packages/adaptive-form/src/components/fields/OneOfField/OneOfField.tsx
```diff
@@ -38,36 +104,89 @@ const OneOfField: React.FC<FieldProps> = (props) => {
...
+        {...props}
+        {...customProps}
+        css={{ label: 'ExpressionFieldValue' }}
+        enumOptions={enumOptions}
+        // allow object fields to render their labels
+        label={selectedSchema.type !== 'object' ? false : undefined}
+        placeholder={placeholder}
+        schema={selectedSchema}
+        transparentBorder={false}
+        uiOptions={{ ...props.uiOptions, helpLink: 'https://bing.com' }}
```
hatpick, line 137: what's up with bing.com :D

Composer/packages/adaptive-form/src/components/fields/OneOfField/OneOfField.tsx
```diff
@@ -38,36 +104,89 @@ const OneOfField: React.FC<FieldProps> = (props) => {
...
+      />
+    );
+  };
+
+  const renderDropDown = () => {
+    if (options && options.length > 1) {
```
hatpick, line 143: Please add a comment explaining this condition

Composer/packages/adaptive-form/src/components/fields/OneOfField/OneOfField.tsx
```diff
@@ -38,36 +104,89 @@ const OneOfField: React.FC<FieldProps> = (props) => {
...
+      return (
+        <Dropdown
+          ariaLabel={formatMessage('select property type')}
```
hatpick, line 146: Select ...

Composer/packages/adaptive-form/src/components/fields/OneOfField/utils.ts
```diff
@@ -23,40 +25,71 @@ function getOptionLabel(schema: JSONSchema7): string {
...
-    return options;
+    return { options, isNested };
   }
 
   if (oneOf && Array.isArray(oneOf)) {
-    return oneOf
+    const resolvedOneOf = oneOf.map((s) => (typeof s === 'object' ? resolveRef(s, definitions) : s));
+    const options = resolvedOneOf
       .map((s) => {
         if (typeof s === 'object') {
```
hatpick, line 52: type of null is also object

Composer/packages/adaptive-form/src/components/fields/OneOfField/utils.ts
```diff
@@ -23,40 +25,71 @@ function getOptionLabel(schema: JSONSchema7): string {
...
       .filter(Boolean) as IDropdownOption[];
+
+    const expression = (resolvedOneOf as JSONSchema7[]).find(({ $role }) => $role === 'expression');
+    const merged = merge({}, omit(schema, 'oneOf'), expression);
+
+    if (expression && (resolvedOneOf as JSONSchema7[]).some(({ properties, items }) => properties || items)) {
+      options.push(
+        {
+          key: 'divider',
+          text: '-',
```
hatpick, line 74: unnecessary

Composer/packages/adaptive-form/src/components/fields/SelectField.tsx
```diff
@@ -33,12 +34,28 @@ export const SelectField: React.FC<FieldProps<string | number>> = function Selec
       text: '',
     });
 
+    if (expression) {
+      opts.push(
+        {
+          key: 'divider',
+          text: '-',
```
hatpick, line 41: une...

Composer/packages/adaptive-form/src/components/fields/SelectField.tsx
```diff
@@ -33,12 +34,28 @@ export const SelectField: React.FC<FieldProps<string | number>> = function Selec
...
+        }
+      );
+    }
+
     return opts;
-  }, [enumOptions]);
+  }, [enumOptions, expression]);
 
   const handleChange = (_e: React.FormEvent<HTMLDivElement>, option?: IDropdownOption) => {
     /* istanbul ignore else */
```
hatpick, line 55: is this needed?

Composer/packages/adaptive-form/src/components/fields/SelectField.tsx
```diff
@@ -33,12 +34,28 @@ export const SelectField: React.FC<FieldProps<string | number>> = function Selec
...
-    if (option) {
+    if (option?.key === 'expression') {
```
hatpick, line 56:
```
if(option) {
   if(option.key === 'expression') {
       onChange('=');
   } else {
       onChange(option.key);
   }
} else {
   onChange(undefined);
}
```

Composer/packages/adaptive-form/src/components/sharedStyles.ts
```diff
@@ -19,3 +19,5 @@ export const focusBorder = css`
     }
   }
 `;
+
+export const sharedFieldIconStyles = { backgroundColor: '#F3F2F1', width: '44px', lineHeight: '28px', fontSize: 12 };
```
hatpick, line 23: is that color one of the NeutralColors provided by Fabric?
https://developer.microsoft.com/en-us/fluentui#/styles/web/colors/neutrals
(Grey 20)

Composer/packages/adaptive-form/src/components/ExpressionSwitchWindow.tsx
GeoffCoxMSFT: display: flex affects the items inside the element, not the element itself - it will be the equivalent of block.  If you put flex on a div that contains string content, it doesn't going to do anything.

Composer/packages/adaptive-form/src/components/fields/IntellisenseFields.tsx
GeoffCoxMSFT: Should null be returned? or would </> be better (empty fragment)?

Composer/packages/adaptive-form/src/components/fields/IntellisenseFields.tsx
GeoffCoxMSFT: Isn't that what event.preventDefault() is for?  If you must pass a value for the property, then pass a no-op function rather than undefined.  You're betting that the React call will pass a value of undefined rather than not passing the property at all.

Composer/packages/adaptive-form/src/components/fields/NumberField.tsx
```diff
@@ -1,67 +1,42 @@
...
+  const { expression, schema, onChange } = props;
   const { type } = schema;
 
-  const updateValue = (step: number) => (value: string) => {
-    if (value === '') {
-      onChange(0);
-      return;
+  const [value, setValue] = useState<string | undefined>(typeof props.value === 'number' ? props.value.toString() : '');
+
+  // if the number is a float, we need to convert to a fixed decimal place
```
GeoffCoxMSFT, line 17: Yay! I love comments that explain why the code is here.

Composer/packages/adaptive-form/src/components/fields/OneOfField/OneOfField.tsx
```diff
@@ -2,22 +2,88 @@
...
-import { resolveFieldWidget } from '../../../utils';
-import { oneOfField } from '../styles';
+import { sharedFieldIconStyles } from '../../sharedStyles';
+
+import { getOptions, getSelectedOption } from './utils';
+
+const styles = {
+  container: css`
+    width: 100%;
+
```
GeoffCoxMSFT, line 24: Nit: extra line break?

Composer/packages/adaptive-form/src/components/fields/OneOfField/OneOfField.tsx
GeoffCoxMSFT: It is critical to use the secure protocol to go to bing too!

Composer/packages/adaptive-form/src/components/fields/IntellisenseFields.tsx
hatpick: @GeoffCoxMSFT why fragment? null is like don't render anything.

Composer/packages/adaptive-form/src/components/fields/StringField.tsx
```diff
@@ -40,8 +40,17 @@ export const StringField: React.FC<FieldProps<string>> = function StringField(pr
     error,
     uiOptions,
     required,
+    focused,
   } = props;
 
+  const textFieldRef = React.createRef<ITextField>();
+
+  useEffect(() => {
+    if (focused && textFieldRef.current) {
```
GeoffCoxMSFT, line 49: Do you also need to run this when the control is mounted?

Composer/packages/adaptive-form/src/components/withTypeIcons.tsx
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) Microsoft Corporation.
```
GeoffCoxMSFT, line 1: Is this a React component?  If so, casing of the withTypeIcons.tsx should be WithTypeIcons.tsx

Composer/packages/adaptive-form/src/components/withTypeIcons.tsx
```diff
@@ -0,0 +1,45 @@
...
+    return (
+      <React.Fragment>
+        <FieldLabel
+          description={props.description}
+          helpLink={props.uiOptions?.helpLink}
+          id={props.id}
+          label={props.label}
+          required={props.required}
+        />
+        <div style={{ display: 'flex' }}>
```
GeoffCoxMSFT, line 36: Prefer component styles to inline style.

Composer/packages/intellisense/src/hooks/useLanguageServer.ts
```diff
@@ -70,7 +70,7 @@ export const useLanguageServer = (
 
   // Get completion results when selection changes
   React.useEffect(() => {
-    if (ws.current && ws.current.readyState === WebSocket.OPEN) {
+    if (ws.current && ws.current.readyState === WebSocket.OPEN && cursorPosition >= 0) {
```
GeoffCoxMSFT, line 73: This probably deserves a comment about why the cursor position could be negative..

Composer/packages/adaptive-form/src/components/fields/IntellisenseFields.tsx
LouisEugeneMSFT: This is used to override what's decomposed using {...props}.

Composer/packages/adaptive-form/src/components/withTypeIcons.tsx
hatpick: it's an hoc

Composer/packages/adaptive-form/src/components/fields/IntellisenseFields.tsx
LouisEugeneMSFT: hatpick: done. @GeoffCoxMSFT, this is an override, null means no override. </> would show an empty Intellisense panel, which is not what we want.

Composer/packages/adaptive-form/src/components/fields/BooleanField.tsx
LouisEugeneMSFT: Those are the 3 valid types for Boolean values: true and false (boolean), expression (string) and undefined for no value.
image

Composer/packages/adaptive-form/src/components/fields/StringField.tsx
LouisEugeneMSFT: Nop, we don't want fields auto focused when they appear on the screen.

GeoffCoxMSFT (approved):

PR #4411, opened 2020-10-16
fix: Move bf-generate-lib templates to asar.unpacked

### LINKED ISSUE #4381: Add hint texts to form dialogs
Add hint texts recently added by UX to form dialogs.

### LINKED ISSUE #4410: Move bf-generate-lib templates into asar.unpack
## Describe the bug

Currently, bf-generate-library uses relative paths to access its template, with the electron, it can't access them within the asar pack. These templates should be moved to asar.unpacked location and provide the path to the library for generation.

## Version

Latest on main

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [x] macOS
- [x] Windows
- [x] Ubuntu

hatpick:
## Description

Moves bf-generate-lib templates to asar.unpacked
Adds missing hint texts to form dialog UX
Updates bf-generate-lib to the latest version
Minor fixes

## Task Item

fixes #4410 
closes #4381


Composer/packages/client/src/recoilModel/dispatchers/project.ts
```diff
@@ -343,6 +344,7 @@ export const projectDispatcher = () => {
   });
 
   const reloadProject = async (callbackHelpers: CallbackInterface, response: any) => {
+    callbackHelpers.reset(filePersistenceState(response.data.id));
```
GeoffCoxMSFT, line 347: Could the loadProjectData do this.  Is there any time when we want to start trying to save stuff while we are loading?

Composer/packages/server/src/controllers/formDialog.ts
```diff
@@ -1,12 +1,36 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+import * as path from 'path';
+
```
GeoffCoxMSFT, line 5: Nit: Extra space

Composer/packages/server/src/models/bot/botProject.ts
```diff
@@ -568,7 +568,7 @@ export class BotProject implements IBotProject {
       outDir,
       metaSchema: undefined,
       allLocales: undefined,
-      templateDirs: [],
+      templateDirs: templateDirs || [],
```
GeoffCoxMSFT, line 571: Do you mean ?? - I think this will set to a true?

GeoffCoxMSFT (changes requested):
Just the || vs ?? thing and I think it is good to go.  Since your templateDirs is always either the dir or [] maybe it can't be undefined anyway.

Composer/packages/server/src/controllers/formDialog.ts
hatpick: It's prettier + eslint separating thise imports

Composer/packages/client/src/recoilModel/dispatchers/project.ts
hatpick: I think reload project should tell the filepersistance to not do anything.

Composer/packages/server/src/models/bot/botProject.ts
hatpick: it's going to be the value of templateDirs or empty array which the library expects.

GeoffCoxMSFT (approved):

GeoffCoxMSFT (approved):

PR #4413, opened 2020-10-17
feat: Added trigger grouping methods to indexers

GeoffCoxMSFT:
## Description

Added indexer method to support to grouping triggers by property.  

Supports 3 ways to determine the properties associated with a trigger:

- $designer: { "propertyGroups": ["<name1>", "<name2", ... ]
- has "property": "<name>"
- has condition : "<expresssion referencing properties>"

This association spans the trigger and its actions.

The grouping algorithm supports 
- single/multi parent of triggers in groups, choosing the first property in the case of single.
- providing a list of valid properties to exclude any discovered properties that are not part of the schema

Added indexer method to support extracting the schema property names from a dialog with a schema property.  This prefers the properties under $public, but falls back to all properties.

## Task Item

#minor
#4150 

## Screenshots

While the project tree isn't in main yet, this demonstrates that a form with a schema can be identified as a form dialog, and the triggers can be grouped by property.

image


Composer/packages/lib/indexers/src/extractSchemaProperties.ts
```diff
@@ -0,0 +1,22 @@
...
+// Licensed under the MIT License.
+
+import { DialogInfo, JsonSchemaFile } from '@bfc/shared';
+
+import { getBaseName } from './utils/help';
+
+export const extractSchemaProperties = (dialog: DialogInfo, jsonSchemaFiles: JsonSchemaFile[]): string[] => {
+  if (dialog.content?.schema) {
+    const schema = jsonSchemaFiles.find(
+      (file) => file.id === getBaseName(dialog.content.schema) || file.id === dialog.content.schema
```
hatpick, line 11: missing a cast here

GeoffCoxMSFT: The build caught that too :)  Fortunately, it helped me find another missing case to handle.

hatpick (approved):

PR #4399, opened 2020-10-15
chore: move plugins folder outside of Composer

a-b-r-o-w-n:
## Description

Also renames it to extensions.

This will help avoid yarn workspace issues with transient dependencies. Extensions should be considered external/isolated modules.

## Task Item

#minor

PR #4423, opened 2020-10-19
fix: azure publish failed

lei9444:
## Description
In #4351 we have updated the bf-cli version. But the azure publish copied the build logic from the server and client. It will fail if we update the bf-lu version. 

1. use previous bf-lu version in azure publish as a short term fix.
2. The luis settings' structure changed  (lubuild result)  in previous PR. update the get function.
3. update the azure plugin relative path. (the plugins folder path changed in #4399).
4. update the down-sampling library version

## Task Item
#minor

### DIFF
--- a/Composer/packages/server/package.json
+++ b/Composer/packages/server/package.json
@@ -62,7 +62,7 @@
     "@bfc/lg-languageserver": "*",
     "@bfc/lu-languageserver": "*",
     "@bfc/shared": "*",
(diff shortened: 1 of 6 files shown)


boydc2014:
@luhan2017 @VanyLaw  can you help review this one?

luhan2017 (dismissed):

boydc2014 (approved):

boydc2014:
@lei9444 please address the conflict

cwhitten (approved):

PR #4424, opened 2020-10-19
fix: allow extension to register multiple publish targets

### LINKED ISSUE #4394: "Publish bot to Azure" choice has no text
## Describe the bug
No text for the first entry in the dropdown

image

## Version
Version:  1.1.1-nightly.172371.8d1fc5a

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Go to 'Publish'
2. Click on 'Add new Profile'
3. See the value is missing

## Expected behavior
Should show something like "Publish bot to Azure"

## Screenshots
image

a-b-r-o-w-n:
## Description

Requires name and description for publish targets.

## Task Item

fixes #4394 

## Screenshots

image


extensions/sample-ui-plugin/src/node/index.ts
```diff
@@ -1,15 +1,26 @@
...
-  registration.addPublishMethod(plugin);
+  const plugin2 = {
+    name: 'samplePublishPlugin2',
+    description: 'Publish using custom UI (2)',
+    bundleId: 'publish2' /** we have custom UI to host */,
+    publish,
+    getStatus,
+  };
+  registration.addPublishMethod(plugin1);
+  registration.addPublishMethod(plugin2);
```
tonyanziano, line 23: sweet!

tonyanziano (approved):
Looks good to me!

Composer/packages/extension/package.json
```diff
@@ -25,6 +24,7 @@
   },
   "dependencies": {
     "@botframework-composer/types": "*",
+    "@types/passport": "^1.0.3",
```
hatpick, line 27: it's a dev dep right?

Composer/packages/extension/src/extensionRegistration.ts
```diff
@@ -58,12 +58,17 @@ export class ExtensionRegistration {
    * Publish related features
    *************************************************************************************/
   public async addPublishMethod(plugin: PublishPlugin) {
-    log('registering publish method', this.name);
-    this.context.extensions.publish[plugin.customName || this.name] = {
+    if (this.context.extensions.publish[plugin.name]) {
+      throw new Error(`Duplicate publish method. Cannot register publish method with name ${plugin.name}.`);
```
hatpick, line 62: nit: if this is surfaced to the UI, maybe use formatMessage

extensions/azurePublish/src/index.ts
```diff
@@ -6,8 +6,8 @@ import path from 'path';
 import { v4 as uuid } from 'uuid';
```
hatpick, line 6: I highly recommend avoid using arrow functions for class methods.

Composer/packages/extension/src/extensionRegistration.ts
a-b-r-o-w-n: Couple of things:

- It doesn't get to the UI. It will end up in the terminal if this is thrown.
- Our server is not yet set up to handle i18n.

beyackle: As far as I've seen, messages that get thrown out to the console (like those inside exceptions) don't need to be localized. Only strings that get displayed to a normal user do.

extensions/localPublish/src/index.ts
```diff
@@ -36,12 +38,14 @@ interface PublishConfig {
 
 const isWin = process.platform === 'win32';
 
-class LocalPublisher {
+class LocalPublisher implements PublishPlugin<PublishConfig> {
+  public name = 'localPublisher';
+  public description = 'Publish bot to local runtime';
   static runningBots: { [key: string]: RunningBot } = {};
```
hatpick, line 44: nit: Record<string, RunningBot>

extensions/azurePublish/src/index.ts
a-b-r-o-w-n: Noted. I just did minimal updates to this file, so feel free to open a PR with that change.

Composer/packages/extension/package.json
a-b-r-o-w-n: This actually needs to be in the dependencies because we reference it in an exported type.

extensions/localPublish/src/index.ts
a-b-r-o-w-n: I always forget about `Record`. However, since this PR is scoped to the publishing name and description, I am going to leave as is.

PR #4431, opened 2020-10-20
feat: Bot Project Schema Update

### LINKED ISSUE #4162: BotProject file (.botProj) Operations
*   [x] Need to support operations to read and load projects from the botproject file in state
*   [x] Add a project to the botproject file
*   [x] Delete a project from the botproject file
*   [x] Update botproject file on "Save as"
*   [x] Adding a remote skill to botproject file
*   [ ] Hosting the approved schema on Bot Framework repo

srinaath:
## Description
This PR removes workspace reference in Bot Project file. This file would always be nested directly inside the Root Bot. All skills added would be referred using a relative path from the root bot. A typo was made with the version of BotProject file. Changing it to 0.2.0 and will be upgraded to 1.0.0 at the end of R11.

Fixes #4162


Composer/packages/client/src/recoilModel/dispatchers/utils/project.ts
```diff
@@ -1,12 +1,13 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+import path from 'path';
```
hatpick, line 4: You sure you can use this module on the browser?

srinaath: Great catch. @hatpick and me had an offline conversation about this.

#4437 This would be the ticket that should do a cleanup of the path module throughout our codebase in non node context

cwhitten (approved):

PR #4433, opened 2020-10-20
feat: use Recognizer uischema `fields` to customize Recognizer instance

yeze322:

### DIFF
--- a/Composer/packages/client/config/extensions.config.js
+++ b/Composer/packages/client/config/extensions.config.js
@@ -51,5 +51,16 @@ module.exports = (webpackEnv) => {
         }),
       ],
     },
+    // Some libraries import Node modules but don't use them in the browser.
+    // Tell Webpack to provide empty mocks for them so importing them works.
+    node: {
+      module: 'empty',
+      dgram: 'empty',
+      dns: 'mock',
+      fs: 'empty',
+      net: 'empty',
+      tls: 'empty',
+      child_process: 'empty',
+    },
   };
 };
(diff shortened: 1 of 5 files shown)


yeze322:
Due to botbuilder-lg becomes a dependency of `@bfc/extension-client`, there are warnings when executing the script `plugin-host-preload.js`

image

a-b-r-o-w-n (dismissed):
This is cool. I'd like to see our usage of the expression evaluation expanded.

cwhitten:
@yeze322 conflicts

extensions/pvaPublish/webpack.config.js
```diff
@@ -38,6 +38,9 @@ function getWebpackConfigs() {
         extensions: ['.js', '.ts', '.tsx'],
       },
       plugins: [],
+      node: {
+        fs: 'empty',
```
cwhitten, line 42: what is this for?

yeze322: The `botbuilder-lg` can target both node env and browser env, to build it correctly in browser env, any project imports `botbuilder-lg` needs this line to instruct webpack where to resolve `node`. Otherwise, there will be a build error `fs not empty`. https://github.com/webpack-contrib/css-loader/issues/447#issuecomment-285598881

I add such config in this PR because `@bfc/extension-client` starts to depend on `botbuilder-lg`, this change led to a build failure in pvaPublish (it depends on `@bfc/extension-client`).

yeze322:
blocked by CI process, as a low priority feature, target it to R13
```
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
```

PR #4436, opened 2020-10-20
fix: Added missing dependencies to /types and /extensions packages.

tonyanziano:
## Description

`@bfc/extensions` and `@botframework-composer/types` have dependencies on `express-serve-static-core` but do not list it in their `package.json`

## Task Item

#minor


a-b-r-o-w-n (dismissed):

tonyanziano:
This caused some typechecks to break in `/server/` that need to be fixed.

cwhitten:
@tonyanziano can you address conflicts plz

tonyanziano:
Updated to fix typechecking errors.

a-b-r-o-w-n (dismissed):

tonyanziano:
need to fix a test that doesn't like one of the new types

a-b-r-o-w-n (approved):

PR #4453, opened 2020-10-21
fix: change copyDir implementation to copyFile function

### LINKED ISSUE #3850: Ejecting runtime to get a new copy of the runtime code can lose file attributes
## Describe the bug

The localPublish plug-in copyDir function does a content copy of files by reading the file content into memory and then writing it to the destination.  I'm not sure why it doesn't use the copyFile function imported on the IFileStorage interface.

This is the problem code:
https://github.com/microsoft/BotFramework-Composer/blob/main/Composer/plugins/localPublish/src/copyDir.ts

    // copy files
    const content = await srcStorage.readFile(srcPath);
    await dstStorage.writeFile(dstPath, content);

## Version

Main branch - fd5d15a36d7e8407f47b301b05d3af548d9da776

## Browser

- [ ] Electron distribution
- [X] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [X] Windows
- [ ] Ubuntu

## To Reproduce

1. Run dev environment (yarn start:dev)
2. Ensure that plugins/localPublish/hostedBots is empty
3. Publish a bot using the default runtime
4. Eject the runtime and choose to copy a new C# runtime
5. Inspect the file attributes on the ejected runtime.

## Expected behavior

File attributes should included with ejected runtime files

zidaneymar:
## Description

minor fix: avoid reading the content to memory and writing to assets, use copyFile instead

## Task Item
fixes #3850

### DIFF
--- a/Composer/packages/server/src/utility/storage.ts
+++ b/Composer/packages/server/src/utility/storage.ts
@@ -29,8 +29,7 @@ export async function copyDir(srcDir: string, srcStorage: IFileStorage, dstDir: 
 
     if ((await srcStorage.stat(srcPath)).isFile) {
       // copy files
-      const content = await srcStorage.readFile(srcPath);
-      await dstStorage.writeFile(dstPath, content);
+      await srcStorage.copyFile(srcPath, dstPath);
     } else {
       // recursively copy dirs
       await copyDir(srcPath, srcStorage, dstPath, dstStorage);


boydc2014 (approved):

Composer/packages/server/src/utility/storage.ts
```diff
@@ -29,8 +29,7 @@ export async function copyDir(srcDir: string, srcStorage: IFileStorage, dstDir:
 
     if ((await srcStorage.stat(srcPath)).isFile) {
       // copy files
-      const content = await srcStorage.readFile(srcPath);
-      await dstStorage.writeFile(dstPath, content);
+      await srcStorage.copyFile(srcPath, dstPath);
```
boydc2014, line 32: Sorry, this is not right, because this read and write mechanism can handle there is different storage for src and dst. But in this change, it's copying from src to src. 

Can we do a condition to check if src is same as dst, then do the copyFile, otherwise we still do read\write?

boydc2014 (changes requested):

Composer/packages/server/src/utility/storage.ts
zidaneymar: > Sorry, this is not right, because this read and write mechanism can handle there is different storage for src and dst. But in this change, it's copying from src to src.
> 
> Can we do a condition to check if src is same as dst, then do the copyFile, otherwise we still do read\write?

ok I will fix this

zidaneymar:
close this pr, no need to fix

cwhitten:
Hi @zidaneymar, does this mean the linked issue in the description can be closed?

PR #4461, opened 2020-10-21
fix: fix docker build

### LINKED ISSUE #4435: Docker not building
## Describe the bug
Using Docker to run Composer.
Several issues started happening after moving Extensions folder from Composer into root.

Most of them, during build process..

Wrong path:
image
image

Git is now required for some module, but it's not present on container.

Timeouts:
image

## Version

Latest merge - [00df46c](https://github.com/cdonke/BotFramework-Composer/commit/00df46ceaa6ef7abbd02c75fe677a190dba4ea7e)

## OS

- [ ] macOS
- [ ] Windows
- [ X ] Ubuntu

## To Reproduce

> cd ~/
> ~/BotFramework-Composer
> docker build -t bot-composer -f ./Composer/Dockerfile .

## Expected behavior

Container image is created.

a-b-r-o-w-n:
## Description

Updates docker build after moving extensions directory outside `./Composer`. Also enables docker builds in CI again.

## Task Item

fixes #4435


a-b-r-o-w-n:
@cdonke Can you take a look at this PR and let me know if there are any optimizations that can be made?

cdonke:
@a-b-r-o-w-n  Sure! right now!

cdonke:
@a-b-r-o-w-n 

Some dependencey requires the .git folder
image

Not sure if it is strictly necessary. If it is, I would COPY the .git folder as well, just after line 16
`COPY /.git ../.git`


I saw that you added a try-catch block on /Composer/scripts/compileExtensions.js because of the git command it uses. It is complaining that the git command is unknown. You have 2 options: 

1. remove the whole **getLastModified** method, because it will always fail, 
2. or to add Git to the container, adding this just after line 10:
`RUN apk add --no-cache git`

a-b-r-o-w-n:
@cdonke thanks for the review. The dependency that is complaining about the .git directory is when developing locally. It runs `eslint` before each commit. It's fine to ignore in the docker environment.

As for the `try/catch` in `compileExtensions.js`, in the case of not having git installed, I will treat that as always needing to be compiled. I don't think the docker container needs git installed because we should always build the extensions.

cdonke:
@a-b-r-o-w-n Makes sense... 

So I don't see any more changes... besides removing a redundant WORKDIR at line 35... but that is not affecting anything...

Issue #4005, opened 2020-09-02
[extension] Asset Library


a-b-r-o-w-n:
Client Requirements
- Type: page w/ project scope
- Shared UI Components
  - Toolbar
  - Dialog
  - LoadingSpinner
- Ability to re-index project after adding/removing assets

Server Requirements
- Create web routes
- Access to current bot project
  - Read/write dialog assets
  - Modify runtime by adding/removing custom actions
  - Information about current runtime (if ejected, type)
  - Ability to eject runtime
  - Update schema
- Read/write bot settings (not application.json settings, information like which assets have been installed, etc.)
- Read/write Composer settings (via dedicated extension settings file)

PR #4467, opened 2020-10-21
feat: add global styling and Fabric for extension UIs

a-b-r-o-w-n:
## Description

Makes Fabric a global so extensions do not need to bundle it as well as injects global styling into extension frames.

Also moves the extension host bundling into our main webpack config so it can be picked up by the dev server and enables loading css.

## Task Item

refs #4005


tonyanziano (approved):
Looks good!

---

Just a note for the future: we probably want to do a naming pass on `PluginHost` and `plugin-preload` and convert them to `Extension`

PR #4468, opened 2020-10-22
fix: l10n update for 2020-10-20

beyackle:
## Description

Just another l10n bubble PR with more strings needing to get added to and deleted from en-US.json.

## Task Item

#minor


beyackle:
Made redundant by the 10-26 one.

PR #4475, opened 2020-10-22
feat: Added trigger grouping for form dialogs

### LINKED ISSUE #4150: Update navigation to display form dialog assets

GeoffCoxMSFT:
## Description

Updated ProjectTree to support grouping triggers by property for form dialogs.

## Task Item

Closes #4150 

## Screenshots

When form dialogs feature is NOT enabled:
image

When form dialogs feature IS enabled:
image


Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
```diff
@@ -270,7 +283,7 @@ export const ProjectTree: React.FC<Props> = ({
       trigger: item.index,
       dialogName: dialog.id,
       isRoot: false,
-      projectId: currentProjectId,
```
GeoffCoxMSFT, line 273: @beyackle I updated the link to use the projectId parameter as the currentProjectId seemed like the incorrect thing to pass.  Let me know if this wasn't a bug.

beyackle: No, that needed to be fixed, although it didn't matter in the single-bot case. I've got my own fix for this in the draft PR I have up, too, where we need to keep track of both a projectId and a skillId for each link.

GeoffCoxMSFT: I also updated the order or params so that the first one was the primary (item, triggers, etc.) and then the later ones are the additional context (dialog, projectId)

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
```diff
@@ -306,6 +319,80 @@ export const ProjectTree: React.FC<Props> = ({
...
+      skillId: null,
+    };
+    return (
+      <span
+        css={css`
+          margin-top: -6px;
+          width: 100%;
+          label: trigger-group-header;
+        `}
+        role="grid"
```
hatpick, line 353: nit: usually used for tables or grid divs.

Composer/packages/client/src/recoilModel/selectors/project.ts
```diff
@@ -76,3 +77,17 @@ export const formDialogSchemaDialogExistsSelector = selectorFamily<boolean, { pr
     return !!dialogs.find((d) => d.id === schemaId);
   },
 });
+
+type JsonSchemaFilesByProjectId = { [projectId: string]: JsonSchemaFile[] };
```
hatpick, line 81: use Record<string, JsonSchemaFile[]>

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
GeoffCoxMSFT: I'm following the TreeItem of the dialog here.  I expect we'll pull this out to a component style in another PR.

Composer/packages/client/src/recoilModel/selectors/project.ts
GeoffCoxMSFT: good point

hatpick (approved):

PR #4483, opened 2020-10-22
fix: correctly read project paths on Windows

beyackle:
## Description

Loading multi-skill bots was broken on Windows due to strange behavior in Node's own `path.resolve` function, where it returns paths like `/C:/Users`. The leading slash breaks the server's ability to load files, so now we detect it. This also adds some null-checking for safety to the ProjectTree so you can load in a multi-bot project even on Windows and see it displayed (although the links won't work yet; that PR is still in progress).

## Task Item

#minor

## Screenshots

image


Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
```diff
@@ -212,11 +213,11 @@ export const ProjectTree: React.FC<Props> = ({
 
   const renderDialogHeader = (skillId: string, dialog: DialogInfo) => {
     const warningContent = notificationMap[currentProjectId][dialog.id]
```
srinaath, line 215: is it better to put both error and warning content inside if(notificationMap\[currentProjectId\]\[dialog.id\]) instead of typescript ?

srinaath (approved):

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
beyackle: I think it's about equivalent. I've gotten used to using `?.` for cases like this.

PR #4484, opened 2020-10-22
feat: Add notification panel

### LINKED ISSUE #4462: Add notification panel
https://www.figma.com/file/LlTlh5vXwq91zjGnFtrUUR6l/Composer-main-design-spec-%2F-UI-library?node-id=5204%3A3

tdurnford:
## Description
Adds the notification button to the header and the notification panel. 

## Task Item
Closes #4462

## Screenshots
image
image


Composer/packages/client/src/components/Header.tsx
```diff
@@ -50,17 +52,22 @@ const divider = css`
   margin: 0px 0px 0px 20px;
 `;
 
-const updateAvailableIcon = {
+const controls = css`
+  display: flex;
+  position: absolute;
```
hatpick, line 57: nit: this can be achieved without absolute positioning and has the advantage of vertically centering it via css instead of calculations. Please consider that.

Composer/packages/client/src/components/Notifications/NotificationButton.tsx
```diff
@@ -0,0 +1,69 @@
...
+    }
+    setIsOpen(!isOpen);
+  };
+
+  return (
+    <div aria-label={formatMessage('Open notification panel')}>
+      <IconButton iconProps={{ iconName: 'Ringer' }} styles={buttonStyles} onClick={toggleIsOpen}>
+        <div css={styles.container}>
+          <div aria-hidden css={styles.count(!isOpen && !!unreadNotification.length)}>
+            {unreadNotification.length}
```
hatpick, line 60: fun fact: chrome on android shows :) when there are more than 99 (3 digits) tabs open

Composer/packages/client/src/components/Notifications/NotificationContainer.tsx
```diff
@@ -24,12 +24,20 @@ const container = css`
 
 export const NotificationContainer = React.memo(() => {
```
hatpick, line 25: this component has no props, React.memo has no effect

Composer/packages/client/src/components/Notifications/NotificationPanel.tsx
```diff
@@ -0,0 +1,99 @@
...
+    text-align: center;
+    width: 100%;
+  `,
+  panelButtons: css`
+    display: flex;
+    justify-content: center;
+  `,
+};
+
+interface NotificationPanelProps {
```
hatpick, line 38: please use type

Composer/packages/client/src/components/Notifications/index.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Corporation.
```
hatpick, line 1: nit: I recommend avoid using barrel files. It has complication with webpack tree shaking.

Composer/packages/client/src/components/Notifications/NotificationButton.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+const NotificationButton: React.FC<NotificationButtonProps> = ({ buttonStyles }) => {
+  const [isOpen, setIsOpen] = useState(false);
+  const { deleteNotification, markNotificationAsRead } = useRecoilValue(dispatcherState);
+  const notifications = useRecoilValue(notificationsSelector);
+  const unreadNotification = notifications.filter(({ read }) => !read);
+
+  const toggleIsOpen = () => {
+    if (!isOpen) {
+      notifications.map(({ id }) => markNotificationAsRead(id));
```
hatpick, line 50: Question: as soon as one clicks the icon, notifications are set to read and hidden?

Also, given that the addNotification adds to the end of the array of notifications, users may miss unread notifications that are located below the viewport (scrolled).

Composer/packages/client/src/components/Notifications/NotificationPanel.tsx
```diff
@@ -0,0 +1,99 @@
...
+  isOpen: boolean;
+  onDismiss: (event?: React.SyntheticEvent<HTMLElement, Event>) => void;
+  onDeleteNotification: (id: string) => void;
+}
+
+const NotificationPanel: React.FC<NotificationPanelProps> = ({ isOpen, onDeleteNotification, onDismiss }) => {
+  const notifications = useRecoilValue(notificationsSelector);
```
hatpick, line 45: NotificationPanel is half-smart, it knows how to get a list of notifications but don't know how to delete them. I think it should be fully smart, meaning remove the callback for delete and get the dispatcher directly.

OR

go fully "dumb component" and pass a list of notifications + delete callback.

Composer/packages/client/src/recoilModel/dispatchers/notification.ts
```diff
@@ -6,10 +6,10 @@ import { CallbackInterface, useRecoilCallback } from 'recoil';
 import { v4 as uuid } from 'uuid';
```
hatpick, line 6: I think currently a new notification is added to the end of the array, shouldn't it be added to the front of the array?

hatpick (changes requested):

Composer/packages/client/src/recoilModel/dispatchers/notification.ts
hatpick: @sangwoohaan

Composer/packages/client/src/components/Notifications/NotificationButton.tsx
tdurnford: When there is a new notification, it renders as a toast. When the user opens the panel, the notifications should be cleared from the clear. Setting the notification to hidden removes it from the toaster - not the panel. A more descriptive name or a comment would be helpful here. Also, when the panel is opened, the notification is considered read.

hatpick (approved):

cwhitten (approved):

PR #4485, opened 2020-10-22
fix: Optimization of dialog validation

### LINKED ISSUE #4203: Optimize unnecessary extra validations (eg arrays of items that cause validation on every children)

LouisEugeneMSFT:
## Description

Most recoil atoms in the Composer state are tracking data through arrays. In the case of dialogs, this was problematic since all dialogs were being validated any time any of them had any change (say a new character in a field). 

This PR addresses this issue by introducing an atom family tracking the individual dialogs. Any creation, update, validation, deletion can now target an individual dialog instead of modifying the entire array of dialogs. To be able to still list all the dialogs, an array of dialogIds is introduced. All the dialogs data can still be accessed, but it is now through a selector rather than an atom.

image

## Task Item

fixes #4203


Composer/packages/client/src/recoilModel/selectors/dialogs.ts
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { DialogInfo } from '@bfc/shared';
+import { selectorFamily } from 'recoil';
+
+import { dialogIdsState, dialogState } from '../atoms';
+
+export const dialogsState = selectorFamily<DialogInfo[], string>({
```
hatpick, line 9: the name better to indicate that it's a selectorFamily

hatpick (approved):

Composer/packages/client/src/pages/knowledge-base/QnAPage.tsx
```diff
@@ -14,7 +14,7 @@ import { navigateTo } from '../../utils/navigation';
 import { TestController } from '../../components/TestController/TestController';
 import { INavTreeItem } from '../../components/NavTree';
 import { Page } from '../../components/Page';
-import { dialogsState, qnaFilesState } from '../../recoilModel/atoms/botState';
```
GeoffCoxMSFT, line 17: I don't think we should have an index barrel file for the recoilModel.  We found in LUIS how much is slows down build and messes with hot module reload.

Composer/packages/client/src/recoilModel/DispatcherWrapper.tsx
```diff
@@ -10,11 +10,11 @@ import { BotAssets } from '@bfc/shared';
 import { useRecoilValue } from 'recoil';
 import isEmpty from 'lodash/isEmpty';
 
+import { dialogsState } from './selectors';
```
GeoffCoxMSFT: I know you're trying to have minimal change impact, but if it is a selector it should probably be called a selector.

Composer/packages/client/src/recoilModel/selectors/dialogs.ts
```diff
@@ -0,0 +1,27 @@
...
+  key: 'dialogs',
+  get: (projectId: string) => ({ get }) => {
+    const dialogIds = get(dialogIdsState(projectId));
+
+    return dialogIds.map((dialogId) => {
+      return get(dialogState({ projectId, dialogId }));
+    });
+  },
+  set: (projectId: string) => ({ set }, newDialogs) => {
```
GeoffCoxMSFT, line 18: Nit: I'm not a fan of the set on a selector, but I can't claim it won't work.

GeoffCoxMSFT, line 18: Isn't newDialogs expected to be an array (according to Typescript)?  Do you need to defend / silently fail if someone passes a non-array?

Composer/packages/client/src/recoilModel/selectors/dialogs.ts
LouisEugeneMSFT: done

GeoffCoxMSFT (approved):

GeoffCoxMSFT (approved):

PR #4491, opened 2020-10-23
fix: Adjust cross-train to use id as key-value instead of path in config

lei9444:
## Description
use file names without any path or extension in config.

need to wait for the @microsoft/bf-lu nightly build

## Task Item
refs https://github.com/microsoft/botframework-cli/pull/1030
#minor


lei9444:
Please don't merge this PR

lei9444:
Close this PR, fixed in #4549

Issue #4448, opened 2020-10-21
Suggested refinement of the validation error messages


zxyanliu:
### 1. Error message in the **Knowledge base name** field.
* Spaces are not allowed. This should be mentioned in the error message.
* Letters, numbers, -, and _ are allowed. This should be mentioned in the error message.

  ![image](https://user-images.githubusercontent.com/32497439/96655656-e1866900-12f2-11eb-8a6b-dd4e325279ae.png)
   
 #### Suggested change

   Spaces and special characters are not allowed. Use letters, numbers, -, or _.

### 2. Error message in the **Knowledge source** field. 

   ![image](https://user-images.githubusercontent.com/32497439/96658683-04684b80-12fa-11eb-8692-57e9d2ca651e.png)

   ![image](https://user-images.githubusercontent.com/32497439/96658811-5d37e400-12fa-11eb-9224-a5a6039cfb05.png)

  #### Suggested change

   A valid URL should start with http:// or https://.
   A valid URL is required.

### 3. Watermark message in the **Knowledge source** field. 

   ![image](https://user-images.githubusercontent.com/32497439/96659469-dd127e00-12fb-11eb-9670-0441d62e4460.png)
    
  #### Suggested change (if "browse to upload a file" is not supported)
  Enter a valid URL

zxyanliu:
@mareekuh I need your feedback on this so that I can proceed to update. 
cc: @hibrenda

mareekuh:
1. Error message in the Knowledge base name field.
Watermark text : QnA-Maker_KB_1
Error text : A Knowledge base name cannot contain spaces or special characters. Use letters, numbers, -, or _.

2. Error message in the Knowledge source field.
Watermark text : Upload a .pdf file or type a URL starting with http:// or https://
or
Watermark text : Upload a .pdf file or type a URL
Error text : A knowledge base URL should start with http:// or https://.

PR #4497, opened 2020-10-23
fix: error message in QnA KB creation

zxyanliu:
## Description

* Updated the error messages to be more user-friendly. See details in https://github.com/microsoft/BotFramework-Composer/issues/4448.
* Updated two target URLs to be aka links and point to the right places.
* Run `yarn l10n` script and bring resulting changes to en-US.json committed alongside it.

### Before

image
image

## Task Item

https://github.com/microsoft/BotFramework-Composer/issues/4448
#minor

### After
qna-1


beyackle (changes requested):
This PR should have the `yarn l10n` script run and the resulting changes to `en-US.json` committed alongside it.

zxyanliu:
> beyackle
This PR should have the yarn l10n script run and the resulting changes to en-US.json committed alongside it.

That's a great point, Ben! Thanks for bringing it up! 
I will update it.

zxyanliu:
@beyackle FYI.
As suggested, I ran the `yarn l10n` script and committed the resulting changes to en-US.json. 
But why am I seeing more changes to the en-US file than what I updated? Please take a look.

beyackle:
> @beyackle FYI.
> As you suggested, I ran the `yarn l10n` script and committed the resulting changes to en-US.json.
> But why am I seeing more changes to the en-US file than what I updated? Please take a look.

That's the result of other earlier changes that didn't run the script. I'm okay with merging these just so the file stays up-to-date without needing another PR just to take care of those.

zxyanliu:
> > @beyackle FYI.
> > As you suggested, I ran the `yarn l10n` script and committed the resulting changes to en-US.json.
> > But why am I seeing more changes to the en-US file than what I updated? Please take a look.
> 
> That's the result of other earlier changes that didn't run the script. I'm okay with merging these just so the file stays up-to-date without needing another PR just to take care of those.

Thank you! That makes sense to me.

beyackle (approved):

PR #4499, opened 2020-10-23
feat: allow extensions to read/write data

### LINKED ISSUE #4004: Extensions should be able to read/write data to their own database

a-b-r-o-w-n:
## Description

Adds ability for extensions to access their own store to read/write arbitrary data and persist across updates / reloads.

## Task Item

closes #4004


a-b-r-o-w-n:
I have tested the electron distro locally, including starting a bot.

Composer/packages/extension/src/storage/store.ts
```diff
@@ -0,0 +1,71 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import path from 'path';
+
+import { existsSync, writeJsonSync, readJsonSync, unlinkSync } from 'fs-extra';
+
+export type IStore<T> = {
+  read: () => T;
+  write: (data: T) => void;
```
a-b-r-o-w-n, line 10: @benbrown would you prefer this interface take a key when reading and writing?

```ts
type IStore {
  read: (key: string) => unkown;
  write<T>: (key: string, value: T) => void
}
```

benbrown: yes
read write delete? or write null to delete?

benbrown (dismissed):

benbrown (approved):

PR #4506, opened 2020-10-23
feat: Links the generated dialog to its schema

hatpick:
## Description

Currently one can navigate from a form dialog schema to its generated dialog.
This PR adds the capability to do the reverse, from generated dialog navigate to its the form dialog schema. (if it's still there)

## Task Item

#minor

## Screenshots
edit schema


Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
```diff
@@ -167,6 +167,13 @@ export const ProjectTree: React.FC<Props> = ({
     return process.env.COMPOSER_ENABLE_FORMS && dialog.content?.schema !== undefined;
   };
 
+  const formDialogSchemaExists = (projectId: string, dialog: DialogInfo) => {
+    return (
+      dialogIsFormDialog(dialog) &&
```
GeoffCoxMSFT, line 172: I coded dialogIsFormDialog to return false if the feature switch is false.  If the user deletes the dialog with the feature switch off, will it leave files on disk?

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
```diff
@@ -238,10 +245,14 @@ export const ProjectTree: React.FC<Props> = ({
       displayName: dialog.displayName,
       isRoot: dialog.isRoot,
       projectId: currentProjectId,
-      skillId: null,
+      skillId,
       errorContent,
       warningContent,
     };
+
+    const isFormDialog = dialogIsFormDialog(dialog);
```
GeoffCoxMSFT, line 253: Nit: Why separate this out to a variable when it is only used once?

GeoffCoxMSFT (approved):

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
hatpick: If the feature flag is not set, user should not see the menu item.

PR #4507, opened 2020-10-23
fix: Expressions not starting with "=" can resolve to a regular TextField

### LINKED ISSUE #4479: Composer boolean expression field no longer supports the expressions type (only equalsExpressions)
## Describe the bug

Cannot see or edit condition expressions when they don't start with equals (=).

## Version

main branch
10/22/2020 9:38AM

## Browser

- [ ] Electron distribution
- [ X] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [ X] Windows
- [ ] Ubuntu

## To Reproduce

1. Create a bot project
2. Add any activity that has a condition
3. Use the drop down to write an expression with the = prefix  (=!$Name for example)
4. Close the bot project
5. Edit the dialog file directly to remove the = prefix
6. Open the bot project
7. Select the action and note the state of the condition field

## Expected behavior

The expression to continue to appear and be editable for the boolean/expression field because the value is neither true nor false so it must be an expression.

## Screenshots

image
image

image
image

LouisEugeneMSFT:
## Description

**resolveFieldWidget** assumed that all expressions start with "=". But there are two types of expressions:

- **equalsExpression**
- **expression**

As such, schema types that were **oneOf** between **expression** (not **equalsExpression**) and a list of non string types were enable to resolve when the input was a string. Such an example is **condition**:

```

"condition": {
      "$role": "expression",
      "title": "Boolean condition",
      "description": "Boolean constant or expression to evaluate.",
      "oneOf": [
        {
          "$ref": "#/definitions/expression"
        },
        {
          "type": "boolean",
         "title": "Boolean",
          "description": "Boolean value.",
          "default": true,
          "examples": [
            false
          ]
        }
      ]
    },

```

This PR addresses this issue by resolving to a regular input field when the type is an **expression**.

## Task Item

fixes #4479 

## Screenshots

image

### DIFF
--- a/Composer/packages/adaptive-form/src/utils/resolveFieldWidget.ts
+++ b/Composer/packages/adaptive-form/src/utils/resolveFieldWidget.ts
@@ -58,8 +58,24 @@ export function resolveFieldWidget(params: {
       return { field: DefaultFields.OneOfField };
     }
 
-    if (expression && typeof value === 'string' && value.startsWith('=')) {
-      return { field: isOneOf ? DefaultFields.IntellisenseExpressionField : IntellisenseExpressionFieldWithIcon };
+    if (expression && typeof value === 'string') {
+      // The schema has two types of expressions: "equalsExpression" and "expression".
+      // "equalsExpression" inputs start with "=". For those we want to have access to the adaptive expressions built-in functions and have intellisense surface it, thus using IntellisenseExpressionField.
+      // "expression" inputs don't leverage the built-in functions. For those, we only want to show IntellisenseTextField.
+      if (value.startsWith('=')) {
+        return { field: isOneOf ? DefaultFields.IntellisenseExpressionField : IntellisenseExpressionFieldWithIcon };
+      } else {
+        if (showIntellisense && isOneOf) {
+          return { field: DefaultFields.IntellisenseTextField };
+        } else if (showIntellisense && !isOneOf) {
+          return { field: IntellisenseTextFieldWithIcon };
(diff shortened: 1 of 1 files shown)


Composer/packages/adaptive-form/src/utils/resolveFieldWidget.ts
```diff
@@ -58,8 +58,24 @@ export function resolveFieldWidget(params: {
...
 
-    if (expression && typeof value === 'string' && value.startsWith('=')) {
-      return { field: isOneOf ? DefaultFields.IntellisenseExpressionField : IntellisenseExpressionFieldWithIcon };
+    if (expression && typeof value === 'string') {
+      // The schema has two types of expressions: "equalsExpression" and "expression".
+      // "equalsExpression" inputs start with "=". For those we want to have access to the adaptive expressions built-in functions and have intellisense surface it, thus using IntellisenseExpressionField.
+      // "expression" inputs don't leverage the built-in functions. For those, we only want to show a regular text field (that could potentially leverage Intellisense for results other than built-in expression functions).
+      if (value.startsWith('=')) {
+        return { field: isOneOf ? DefaultFields.IntellisenseExpressionField : IntellisenseExpressionFieldWithIcon };
+      } else {
```
GeoffCoxMSFT, line 67: Nit: Could merge else and else if to just else ifs

GeoffCoxMSFT (approved):

PR #4514, opened 2020-10-23
feat: Change Container image to Ubuntu

cdonke:
## Description
Changed the base image to ubuntu, instead of node with Debian.

Instead of running a Node+Debian image and installing the .NET Core, I used a .NET Core image and installed the NodeJS package.

The base image size is 400Mb smaller.
image

## Task Item
#minor


cdonke:
Docker build is failing at when it tries to start. Cannot find a module.
image

I'll try to merge with the orchestrator branch,

cdonke:
Created a new PR #4516 based on Main Branch, because of too many changes from tachau/orchestrator branch. 
I wasn't able to identify from where was the issues comming.

PR #4515, opened 2020-10-24
feat: Selector to get all lg and lu files imported by a form dialog

hatpick:
## Description

Adds two selectors to get all lg and lu files imported by a form dialog

## Task Item

#minor #4360


Composer/packages/client/src/recoilModel/selectors/formDialogs.ts
```diff
@@ -0,0 +1,80 @@
+// Copyright (c) Microsoft Corporation.
```
GeoffCoxMSFT, line 1: I would name the file dialogImports (it isn't completely form dialogs related)

Composer/packages/types/src/indexers.ts
```diff
@@ -150,6 +150,11 @@ export type LgParsed = {
   templates: LgTemplate[];
 };
 
+export type ImportExpression = {
```
GeoffCoxMSFT, line 153: Avoid the work expression here as it has other meaning in bot framework.  How about LuLgImport, LanguageImport, or even MarkdownImport?

Composer/packages/client/src/recoilModel/selectors/formDialogs.ts
hatpick: It can be soon, so I'd keep it this way.

GeoffCoxMSFT: My point is that the LU/LG imports won't be behind the form dialog feature switch - they are a general improvement to support users editing imports.

GeoffCoxMSFT (approved):

PR #4520, opened 2020-10-26
chore: update server api to prevent overwrite the root dialog

lei9444:
## Description
Add preserveRoot to inform create project to not overwrite the root dialog

## Task Item
#minor


boydc2014 (approved):

a-b-r-o-w-n:
What is the reason for this change? I don't see a linked issue.

cwhitten:
@a-b-r-o-w-n this is something I worked with Dong and Leilei on for a PVA requirement

a-b-r-o-w-n:
This probably warrants some commenting in the code because our code base isn't directly using this parameter.

cwhitten:
I'm going to implement the client-side of this and will be sure to include some comments.

a-b-r-o-w-n:
We should also consider typing the request body on our controller actions like in https://github.com/microsoft/BotFramework-Composer/blob/main/Composer/packages/server/src/controllers/extensions.ts#L55

We can then add comments on the TS properties.

PR #4523, opened 2020-10-26
ci: Disable docker cache temporarily

boydc2014:
After taking #4516,  our docker pipeline fails in last post caching step because of no enough space. Seems like a normal issue and no easy way to increase space unless switching to a custom runner (non-trivial work). 
image
image

After digging into more, i found the cache step is taking about 4GB but seems not benefiting the build step. 

So, disable the docker cache temporarily will save 4G space for now, before we switch a custom runner or any other ideas.

#minor


cdonke:
@boydc2014 the default action for docker build uses the [buildx ](https://docs.docker.com/buildx/working-with-buildx/) as template... but it still marked as Experimental Preview...

Since building the image is not so current (many times a day, I mean), wouldn't it be better to use plain docker commands and skip the buildx template?

boydc2014:
> @boydc2014 the default action for docker build uses the [buildx ](https://docs.docker.com/buildx/working-with-buildx/) as template... but it still marked as Experimental Preview...
> 
> Since building the image is not so current (many times a day, I mean), wouldn't it be better to use plain docker commands and skip the buildx template?

Will buildx taking lots of space? The error i see is no space left, and i searched around, many people encountered this also and looks like the default disk size is 16GB.   And from the log of this cache restore step, i saw ~4GB cache restored + when i looked into the build step, i didn't any step says "from cache" if you look at recent PR, so i thought it might be a good idea of not using this cache. What do you think?

cdonke:
> > @boydc2014 the default action for docker build uses the [buildx ](https://docs.docker.com/buildx/working-with-buildx/) as template... but it still marked as Experimental Preview...
> > Since building the image is not so current (many times a day, I mean), wouldn't it be better to use plain docker commands and skip the buildx template?
> 
> Will buildx taking lots of space and using plain docker command will save significant disk space for us? The error i see is no space left, and i searched around, many people encountered this also and looks like the default disk size is 16GB. And from the log of this cache restore step, i saw ~4GB cache restored + when i looked into the build step, i didn't any step says "from cache" if you look at recent PR, so i thought it might be a good idea of not using this cache. What do you think?

Cache is load on lines 37/38 of Dockerfile.

each stage on dockerfile (4 stages) will be new image.. 
With plain docker cmds we can add the _--rm_ that will remove the intermediate images when done...
I could not find this option on buildx.

buildx runs a new container (on the _--use_ cmd)... This image has ~85Mb.. And the build is ran in it...
Without it, we would be using all the capabilities of the host.. I could not find any size limitations, besides the 14Gb SSD.

The benefits of using buildx is for scoping (dev and testing, for instance)... Also, for building for multiple environments (Windows, Linux and Mac)... Both I don't see it being used...

IMHO, buildx and Caching is overhead.

boydc2014:
> > > @boydc2014 the default action for docker build uses the [buildx ](https://docs.docker.com/buildx/working-with-buildx/) as template... but it still marked as Experimental Preview...
> > > Since building the image is not so current (many times a day, I mean), wouldn't it be better to use plain docker commands and skip the buildx template?
> > 
> > 
> > Will buildx taking lots of space and using plain docker command will save significant disk space for us? The error i see is no space left, and i searched around, many people encountered this also and looks like the default disk size is 16GB. And from the log of this cache restore step, i saw ~4GB cache restored + when i looked into the build step, i didn't any step says "from cache" if you look at recent PR, so i thought it might be a good idea of not using this cache. What do you think?
> 
> Cache is load on lines 37/38 of Dockerfile.
> 
> each stage on dockerfile (4 stages) will be new image..
> With plain docker cmds we can add the _--rm_ that will remove the intermediate images when done...
> I could not find this option on buildx.
> 
> buildx runs a new container (on the _--use_ cmd)... This image has ~85Mb.. And the build is ran in it...
> Without it, we would be using all the capabilities of the host.. I could not find any size limitations, besides the 14Gb SSD.
> 
> The benefits of using buildx is for scoping (dev and testing, for instance)... Also, for building for multiple environments (Windows, Linux and Mac)... Both I don't see it being used...
> 
> IMHO, buildx and Caching is overhead.

Thanks, buildx and caching is overhead to me as well. 

The 37/38 line in docker.yml of cache loading replying the cache downloaded in line 25, right? That's the reason i ignore the downloading in this PR. I saw from the log, in this step,~4GB cache is downloaded into /tmp/.buildx-cache.  By commenting out this, it should save us those spaces for sure.  So that's the first thing i'd try. 

And yes, it's a multi-stage build which we can clean timely if using plain docker command instead of buildx. But actually buildx don't block us to clean it later, we can also clean in last step with docker prune or something else right?

The previous log shows that the out of disk space issue is **not in the middle of build**, it happens in the post build steps. So i guess we probably do need to clean the image right after one stage.

So i will propose disable the cache first at least to unblock other PRs at least for now, does this make sense to you? While at the same time, we can experiment with other optimization like using plain docker command instead of buildx to do finer-grain control.

a-b-r-o-w-n (approved):

PR #4524, opened 2020-10-26
fix: Action title not displayed

yeze322:
## Description

#minor

**Problem**
Title of an Action node can be empty when Form plugin overrides the schema title with a function.

**Fixes**
1. This PR fixes this issue by handling the case when `label` is a function.
2. Besides that, also fixes the wrong `aria-label` issue which is not reported yet.

## Screenshots

image
image


Composer/packages/client/src/pages/design/VisualEditor.tsx
```diff
@@ -81,8 +81,15 @@ const VisualEditor: React.FC<VisualEditorProps> = (props) => {
...
-        sdkDefinitions[$kind].title = formOptions?.label;
+      const formLabel = formOptions?.label;
+      const definition = sdkDefinitions[$kind];
+      if (formLabel && definition) {
+        if (typeof formLabel === 'string') {
+          definition.title = formLabel;
+        } else if (typeof formLabel === 'function') {
+          const computedTitle = formLabel({});
+          if (computedTitle) definition.title = computedTitle;
+        }
```
boydc2014, line 92: Could look better if we use switch instead of if else

boydc2014: We may also catch the "default" case or the "else" case with a value saying "unknown"

boydc2014 (approved):

Composer/packages/client/src/pages/design/VisualEditor.tsx
yeze322: I moved the type judgement logic inside title resolver considering that the label fn needs to be aware of data context.

yeze322: it can resolve the 'else' case

Composer/packages/lib/shared/src/viewUtils.ts
```diff
@@ -183,19 +183,26 @@ export function getDialogGroupByType(type) {
 
 const truncateSDKType = ($kind) => (typeof $kind === 'string' ? $kind.replace('Microsoft.', '') : '');
 
+const getSchemaTitle = (sdkschema: JSONSchema7, data) => {
+  const rawTitleValue = sdkschema.title as string | ((d) => string) | undefined;
+  // Handle the case that FormUIOption label function overrides sdk.schema.
+  return typeof rawTitleValue === 'function' ? rawTitleValue(data) : rawTitleValue;
```
a-b-r-o-w-n, line 189: Why would the sdk schema contain a function? Where does this get overwritten?

Composer/packages/lib/shared/src/viewUtils.ts
```diff
@@ -183,19 +183,26 @@ export function getDialogGroupByType(type) {
 
 const truncateSDKType = ($kind) => (typeof $kind === 'string' ? $kind.replace('Microsoft.', '') : '');
 
+const getSchemaTitle = (sdkschema: JSONSchema7, data) => {
+  const rawTitleValue = sdkschema.title as string | ((d) => string) | undefined;
```
hatpick, line 187: did you mean to use || instead of |

Composer/packages/lib/shared/src/viewUtils.ts
yeze322: It do bring confusion, I'll make it clear

Composer/packages/lib/shared/src/viewUtils.ts
yeze322: They are type declarations, I'll clarify the definition.

Composer/packages/lib/shared/src/viewUtils.ts
yeze322: To make the Action title be consistent with Form's title, the FormConfig is merged into sdkSchema object in client/VisualEdtitor.tsx, the `label` option overrides the `title` option.

It can but I didn't make it consume FormConfig directly because I think it's improper to couple with Form uischema.

cwhitten (approved):

PR #4542, opened 2020-10-27
chore: add `onOpenDialog` api to support dialog jump

### LINKED ISSUE #3942: Create a new navigation API to handle jumping between dialogs
## Describe the bug

When fixing breadcrumb and navigation behavior, jumping between dialogs by clicikng on BeginDialog was broken silently.
We need a standalone dialog jump API to treat this scenario specially and as an anti-corruption layer.

## Browser

- [ ] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

yeze322:
## Description

closes #3942 

Previously, we use `navTo()` to support both breadcrumb jump and BeginDialog jump. It has several problems
1. Updating `navTo` api broke the dialog jump behavior, we fixed this bug.
2. The `navTo()` API belongs to `ApplicationContextApi` while Flow Editor prefers to consume `DialogEditingContextApi`. Using `navTo` breaks the context boundary.

For providing a more descriptive interface, this PR implements the new api `onOpenDialog` used only in the dialog jump scenario. In the future, we can freely update the `navTo()` api without bothering breaking the dialog jump behavior.


boydc2014:
This looks reasonable to me for creating a dedicated api in this BeginDialog jump.

My question is should we just update the logic to use this new API, so that we will how much changes is needed or what the impact is. That will help us to determine whether take this or not, right?

yeze322:
@boydc2014 The new API is consumed in `useEditorEventApi.tsx` where handles the click event on dialog link. That's the only usage for now, it seems not noticeable.

boydc2014:
> @boydc2014 The new API is consumed in `useEditorEventApi.tsx` where handles the click event on dialog link. That's the only usage for now, it seems not noticeable.

I didn't see useEditorEventApi.tsx is updated in this PR, is that expected? I mean, i saw a new API add, but don't see any caller in this PR

boydc2014 (approved):

cwhitten:
@yeze322 can you address conflicts

cwhitten:
@yeze322 still have conflicts

cwhitten (approved):

PR #4550, opened 2020-10-27
Provide proper index to trigger links

GeoffCoxMSFT:
## Description

The index for triggers is used to build the link to the trigger. This needs to be relative to the list of triggers for the dialog.  The current filter and grouping miscalculates the trigger.  This updates to ensure the index is always relative to the dialog trigger array.

## Task Item

#minor
#4150 

## Screenshots

N/A

### DIFF
--- a/Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
+++ b/Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
@@ -72,6 +72,8 @@ const SUMMARY_ARROW_SPACE = 28; // the rough pixel size of the dropdown arrow to
 
 // -------------------- Helper functions -------------------- //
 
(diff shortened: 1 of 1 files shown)


Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
```diff
@@ -72,6 +72,8 @@ const SUMMARY_ARROW_SPACE = 28; // the rough pixel size of the dropdown arrow to
 
 // -------------------- Helper functions -------------------- //
 
+type GetTriggerIndex = (trigger: ITrigger) => number;
```
hatpick, line 75: how about a util function that takes a dialog and a trigger and it return the trigger's index.

hatpick (changes requested):

hatpick (approved):

Issue #4505, opened 2020-10-23
Remove generated dialog (forms) is broken


hatpick:
## Describe the bug
When I remove a generated dialog from form dialog, it only removes files related to the top level dialog. All the lu and lg files for the properties remain on the disk which will cause a merging problem if I re-generate the dialog from the schema.

## Version

Latest on main.

## Browser

All browsers + electron.

## OS

All platforms.

## To Reproduce

0. Build composer with forms feature flag.
1. Create a form dialog schema.
2. Generate dialog.
3. Go to the generated dialog.
4. Select remove this dialog from the project tree.
5. See the error as well as check the bot project file to see they are not removed.

## Expected behavior

The entire files/dirs related to this dialog should be removed.

Note: router.delete('/formDialogs/:projectId/:dialogId', FormDialogController.deleteDialog);

PR #4553, opened 2020-10-27
fix: Add delete form dialog to Composer server API

GeoffCoxMSFT:
## Description

Deleting a form dialog requires deleting form dialog assets which are all contained within a folder under dialogs.

## Task Item

#minor
#4505

## Screenshots

N/A

### DIFF
--- a/Composer/packages/server/src/controllers/formDialog.ts
+++ b/Composer/packages/server/src/controllers/formDialog.ts
@@ -77,8 +77,27 @@ const generate = async (req: Request, res: Response) => {
   }
 };
 
+const deleteDialog = async (req: Request, res: Response) => {
+  const projectId = req.params.projectId;
+  const dialogId = req.params.dialogId;
+
+  const user = await ExtensionContext.getUserFromRequest(req);
+
+  const currentProject = await BotProjectService.getProjectById(projectId, user);
+  if (currentProject !== undefined) {
+    await currentProject.deleteFormDialog(dialogId);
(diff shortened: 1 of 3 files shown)


Composer/packages/server/src/controllers/formDialog.ts
```diff
@@ -77,8 +77,27 @@ const generate = async (req: Request, res: Response) => {
...
 
+const deleteDialog = async (req: Request, res: Response) => {
+  const projectId = req.params.projectId;
+  const dialogId = req.params.dialogId;
+
+  const user = await ExtensionContext.getUserFromRequest(req);
+
+  const currentProject = await BotProjectService.getProjectById(projectId, user);
+  if (currentProject !== undefined) {
+    await currentProject.deleteFormDialog(dialogId);
```
hatpick, line 88: do we want a try catch here with 500 response?

Composer/packages/server/src/controllers/formDialog.ts
```diff
@@ -77,8 +77,27 @@ const generate = async (req: Request, res: Response) => {
...
+    const updatedProject = await BotProjectService.getProjectById(projectId, user);
+    res.status(200).json({ id: projectId, ...updatedProject.getProject() });
+  } else {
+    res.status(404).json({
```
hatpick, line 92: nit: the error for when the caller doesn't provide project id should be different

hatpick (approved):

PR #4556, opened 2020-10-28
fix: Calls form dialog dedicated delete api when dialog is removed

hatpick:
## Description

Calls form dialog dedicated delete api when dialog is removed

## Task Item

#minor
#4505

### DIFF
--- a/Composer/packages/client/src/recoilModel/dispatchers/dialogs.ts
+++ b/Composer/packages/client/src/recoilModel/dispatchers/dialogs.ts
@@ -15,6 +15,7 @@ import {
   showCreateDialogModalState,
   dialogState,
 } from '../atoms/botState';
+import { dispatcherState } from '../DispatcherWrapper';
 
 import { createLgFileState, removeLgFileState } from './lg';
 import { createLuFileState, removeLuFileState } from './lu';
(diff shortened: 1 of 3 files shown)


Composer/packages/client/src/recoilModel/dispatchers/dialogs.ts
```diff
@@ -24,7 +25,16 @@ import { removeDialogSchema } from './dialogSchema';
 export const dialogsDispatcher = () => {
   const removeDialog = useRecoilCallback(
     (callbackHelpers: CallbackInterface) => async (id: string, projectId: string) => {
-      const { set, reset } = callbackHelpers;
+      const { set, reset, snapshot } = callbackHelpers;
+
+      const dialog = await snapshot.getPromise(dialogState({ projectId, dialogId: id }));
+
+      // If the dialog is a generated form dialog, delete using form dialog dispatcher
+      if (dialog.content?.$schema) {
```
GeoffCoxMSFT, line 33: Do you think we should have a shared helper function for isDialogFormDialog?

Composer/packages/client/src/recoilModel/dispatchers/formDialogs.ts
```diff
@@ -79,14 +79,37 @@ export const formDialogsDispatcher = () => {
...
+      const dialog = await snapshot.getPromise(dialogState({ projectId, dialogId }));
+      const { reloadProject } = await snapshot.getPromise(dispatcherState);
+
+      try {
+        if (!dialog) {
+          return;
+        }
+
+        const response = await httpClient.delete(`/formDialogs/${projectId}/${dialogId}`);
+        await reloadProject(callbackHelpers, response);
```
GeoffCoxMSFT, line 103: Do we need to do anything with the file state to prevent the delete from causing that issue with reloading (that thing you previously saw during generation)?

GeoffCoxMSFT (approved):

Composer/packages/client/src/recoilModel/dispatchers/dialogs.ts
hatpick: probably yes.

PR #4562, opened 2020-10-28
feat: sync skillbot's locale with root bot

### LINKED ISSUE #4121: Locale handling for BotProjects
The Root Bot's locale holds precedence over the locale of all skills in the Bot Project. When the active locale for the root bot is changed all the skills would also need to swap to the root bots' changed locale. In the edge case of a skill not supporting the active locale of the root bot we create a new locale inside the skill which would be a copy of the contents of the default locale of the skill. In this way, the Root bots' locale would always be present in the skill.

https://www.figma.com/file/LlTlh5vXwq91zjGnFtrUUR6l/Composer-main-design-spec-UI-library?node-id=5690%3A2180

### LINKED ISSUE #4859: Settings page should list only local bots with no errors
The settings page should only list local bots without errors. Use the `localBotsWithoutErrorsSelector`. The settings page is currently using the botprojectSpace selector. As a result bots with errors are shown in the project tree of settings.

Screen Shot 2020-11-17 at 2 38 48 PM
Screen Shot 2020-11-17 at 2 38 36 PM

liweitian:
## Task Item
Closes #4121 
Closes #4859


Composer/packages/client/src/recoilModel/dispatchers/multilang.ts
```diff
@@ -59,7 +60,26 @@ export const multilangDispatcher = () => {
...
-
+      const botProjectsMetaData = await snapshot.getPromise(botProjectSpaceSelector);
+      const rootBotProjectId = await snapshot.getPromise(rootBotProjectIdSelector);
+      if (projectId === rootBotProjectId) {
+        for (let i = 0; i < botProjectsMetaData.length; i++) {
+          if (!botProjectsMetaData[i].isRootBot && !botProjectsMetaData[i].isRemote) {
+            const skillBotProjectId = botProjectsMetaData[i].projectId;
+            const settings = await snapshot.getPromise(settingsState(skillBotProjectId));
+            const languages = settings.languages;
+            const defaultLang = settings.defaultLang;
```
srinaath, line 71: Should we pick the active locale of the root bot instead of default locale?

srinaath (changes requested):

Composer/packages/client/src/recoilModel/dispatchers/multilang.ts
liweitian: From current code logic, I think a new language is copied from the default language.

srinaath (approved):

cwhitten:
@liweitian please address conflicts

liweitian:
pls merge https://github.com/microsoft/BotFramework-Composer/pull/4870 first

srinaath:
@liweitian  #4870 is merged now

Composer/packages/client/src/router.tsx
```diff
@@ -119,11 +120,12 @@ const projectStyle = css`
 const ProjectRouter: React.FC<RouteComponentProps<{ projectId: string; skillId: string }>> = (props) => {
   const { projectId = '' } = props;
   const schemas = useRecoilValue(schemasState(projectId));
-  const { fetchProjectById, setSettings } = useRecoilValue(dispatcherState);
+  const { fetchProjectById, setSettings, setLocale } = useRecoilValue(dispatcherState);
   const botProjects = useRecoilValue(botProjectIdsState);
   const botProjectsMetaData = useRecoilValue(botProjectSpaceSelector);
   const botProjectSpaceLoaded = useRecoilValue(botProjectSpaceLoadedState);
   const rootBotProjectId = useRecoilValue(rootBotProjectIdSelector);
+  const botName = useRecoilValue(botDisplayNameState(rootBotProjectId || ''));
```
srinaath, line 128: the default for this state is an empty string.

Composer/packages/client/src/router.tsx
```diff
@@ -138,6 +140,14 @@ const ProjectRouter: React.FC<RouteComponentProps<{ projectId: string; skillId:
     }
   }, [botProjectSpaceLoaded, rootBotProjectId]);
 
+  useEffect(() => {
+    if (botProjectSpaceLoadedState && rootBotProjectId && botName) {
+      const storedLocale = languageStorage.get(botName)?.locale;
+      setLocale(storedLocale, rootBotProjectId);
+    }
+  }, [botProjectSpaceLoaded, rootBotProjectId, botName]);
```
srinaath, line 148: The bot display name would be set when botProjectSpaceLoaded is set to true. Do we need another useEffect just for the additional botName. Feels like it could be combined with the above useEffect

srinaath (changes requested):

Composer/packages/client/src/router.tsx
liweitian: Argument of type 'string | undefined' is not assignable to parameter of type 'string'.

srinaath (dismissed):

boydc2014 (approved):

PR #4574, opened 2020-10-29
fix: Adds validation for form dialogs propery name

hatpick:
## Description

As per Melanie's request, this PR adds property name validation to the form dialog property names.

## Task Item

#minor
Screen Shot 2020-10-28 at 5 00 50 PM


Composer/packages/form-dialogs/src/utils/constants.ts
```diff
@@ -2,3 +2,5 @@
 // Licensed under the MIT License.
 
 export const jsPropertyListClassName = 'jsPropertyListClassName';
+
+export const nameRegex = /^[a-zA-Z0-9-_]+$/;
```
GeoffCoxMSFT, line 6: I see this in the SDK schema for AgeEntityRecognizer.  I wonder if the sdk schema should define a pattern for a valid property name rather than string?  Or maybe this is too meta for a JSON schema.

"$kind": {
          "title": "Kind of dialog object",
          "description": "Defines the valid properties for the component you are configuring (from a dialog .schema file)",
          "type": "string",
          "pattern": "^[a-zA-Z][a-zA-Z0-9.]*$",
          "const": "Microsoft.ActivityTemplate"
        },

GeoffCoxMSFT (approved):

PR #4612, opened 2020-11-02
fix: dialog resources recognizer/qna incorrect creation 

### LINKED ISSUE #4378: QnA files are sometimes added to project in the wrong place
## Describe the bug

When a project is opened, for each dialog that DOES NOT ALREADY include a .qna file, one is added. However, these files are always added at the /dialogs/nameOfDialog folder.

If the dialogs in question are not in the /dialogs folder, or are in a subfolder of the dialogs folder, the related .qna files should be placed in the existing folder, not in a new one.

For example, if /dialogs/imported/getEmail is added to the project, the .qna file should be in /dialogs/imported/getEmail, not in /dialogs/getEmail as currently happens, resulting in 2 folders.

Whatever code is creating these .qna files should NOT be opinionated about the location of the dialogs, as they can technically exist in any folder or subfolder of the project.

## Version

1.1.1

# Steps to reproduce the behavior:

1. Create a project
2. Open it in the file system and copy an existing dialog into /dialogs/imported/someFolderName
3. Open the project in Composer
4. Look at the file system again - you will see /dialogs/someFolderName has been erroneously created with an empty qna file.

## Expected behavior

The new qna file will be created inside the existing folder.

### LINKED ISSUE #4617: Empty QnA files generated for form dialog generated dialogs upon opening a project
## Describe the bug

Composer creates empty QnA files for generated dialogs (form dialogs) when a project containing one is loaded (opened).

## Version

Latest on main.

## Browser

- [x] Electron distribution
- [x] Chrome
- [x] Safari
- [x] Firefox
- [x] Edge

## OS

- [x] macOS
- [x] Windows
- [x] Ubuntu

## To Reproduce

1. Turn on form dialog feature flag from settings.
2. Navigate to forms (table icon on the main left nabvigation)
3. Create an schema and add one property of type "any string"
4. Click generate dialog.
5. Once the generation is complete, navigate to design page to see the generated dialog.
6. Open explorer to your bot, look at under dialogs, you should see a folder with the same name as the schema from step 3.
7. Now refresh the Composer app and check your explorer again.
8. You can now see the generated files next to the directory from step 6.

## Expected behavior

Don't generated empty QnA files for generated form dialogs. This can break the ProjectTree as you can see in the attached screenshot.

## Screenshots

MicrosoftTeams-image (4)
MicrosoftTeams-image (3)
MicrosoftTeams-image (2)

## Additional context

Check this PR:
https://github.com/microsoft/BotFramework-Composer/commit/72e5700746c29831ce3401020905285d0bde2473
separateDialogsAndRecognizers detects recognizers from generated dialogs and causes Composer to create those files.

zhixzhan:
## Description

1. Customize imported dialog's resource should be created inside it's child folder.
```
/imported/Welcome
  /language-generation/...
  /language-understanding/...
  /knowledge-base/...
  /recognizers/...
```

2. Form dialog generated dialog should not be touched, avoid create 'recognizers' / 'knowledge-base' folder on it.

## Task Item
close #4378 
close #4617


Composer/packages/lib/indexers/src/dialogIndexer.ts
```diff
@@ -180,6 +180,7 @@ function parse(id: string, content: any) {
   const luFile = typeof content.recognizer === 'string' ? content.recognizer : '';
   const qnaFile = typeof content.recognizer === 'string' ? content.recognizer : '';
   const lgFile = typeof content.generator === 'string' ? content.generator : '';
+  const isFormDialog = has(content, '$Generator'); // mark as form generated dialog;
```
boydc2014, line 183: We should leverage a more systematic to tell isFormDialog.

The correct way to tell isFormDialog is schema property is not null dialog.content?.schema !== undefined

boydc2014 (changes requested):

Composer/packages/lib/indexers/src/dialogIndexer.ts
zhixzhan: @boydc2014 updated to `const isFormDialog = has(content, 'schema');`

Composer/packages/server/src/models/bot/botProject.ts
```diff
@@ -790,11 +798,22 @@ export class BotProject implements IBotProject {
...
+    this.files = new Map<string, FileInfo>([...files]);
     const dialogFiles: FileInfo[] = [];
     const qnaFiles: FileInfo[] = [];
     files.forEach((file) => {
       if (file.name.endsWith('.dialog')) {
-        dialogFiles.push(file);
+        try {
+          // filter form dialog generated file.
+          const dialogJson = JSON.parse(file.content);
+          const isFormDialog = has(dialogJson, '$Generator');
```
boydc2014, line 810: this should be also updated.

boydc2014 (approved):

PR #4614, opened 2020-11-02
fix: generate correct URLs for root bots

beyackle:
## Description

This fixes a small glitch in generating URLs for the links; dialogs and triggers that belong to the root of a botProject now generate links that look like the older non-botProject ones, which should fix some issues with undo/redo.

## Task Item

#minor

## Screenshots

image

### DIFF
--- a/Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
+++ b/Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
@@ -279,7 +279,7 @@ export const ProjectTree: React.FC<Props> = ({
       displayName: dialog.displayName,
       isRoot: dialog.isRoot,
       projectId: rootProjectId,
-      skillId: skillId,
+      skillId: skillId === rootProjectId ? undefined : skillId,
       errorContent,
       warningContent,
     };
@@ -332,7 +332,7 @@ export const ProjectTree: React.FC<Props> = ({
   const renderTrigger = (item: any, dialog: DialogInfo, projectId: string): React.ReactNode => {
     const link: TreeLink = {
       projectId: rootProjectId,
-      skillId: projectId,
+      skillId: projectId === rootProjectId ? undefined : projectId,
       dialogId: dialog.id,
       trigger: item.index,
       displayName: item.displayName,


cwhitten (approved):

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
```diff
@@ -279,7 +279,7 @@ export const ProjectTree: React.FC<Props> = ({
       displayName: dialog.displayName,
       isRoot: dialog.isRoot,
       projectId: rootProjectId,
-      skillId: skillId,
+      skillId: skillId === rootProjectId ? undefined : skillId,
```
srinaath, line 282: @beyackle  I see a bug where I click on a skill's dialog file url changes correctly to `http://localhost:3000/bot/25141.842060596977/skill/5927.301138798824/dialogs/googlekeepsync?selected=triggers[%22376720%22]` . Navigate to a different page and come back to the DesignPage. The url is an invalid URL. It needs to set the url obtained from the previous designPageLocationState right?

srinaath (changes requested):

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
beyackle: This looks to me like a problem in the sidebar - when I have a skill open (so, the URL is /bot/<bot Id>/skill/<skill Id>) and click on the sidebar button, it takes me to a URL with /bot/<skill Id>/ in it, and then clicking back into Design keeps that around. Single-bot projects don't have this problem. The reason this change is going into main and not the feature branch is that undo/redo was breaking even in the single-bot case, right?

srinaath: This fix has no relation to undo/redo. The URL was broken irrespective of whether we were doing Undo/redo.

srinaath: This is also not an issue with the sidebar. DesignPageLocation state stores the current focussed actions etc that are particular to the design page. On rendering the designPage it should always use the currentProjectId state and parse the designPagelocation state to set the URL correctly?

srinaath (approved):

Issue #2560, opened 2020-04-08
Bring other SDK features in Composer


luhan2017:
This issue is to track what are the SDK actions/inputs/conditions/recognizers are not yet enabled in Composer.

#Actions 
- [ ] ContinueConversationLater
- [ ] GotoAction
- [ ] DebugBreak
- [ ] CodeAction
- [ ] EditActions
- [x] GetActivityMembers
- [x] GetConversationMembers
- [x] ThrowException

#EntityRecognizers 
- [ ] ChannelMentionEntityRecognizer
- [ ] ConfirmationEntityRecognizer
- [ ] CurrencyEntityRecognizer
- [ ] DateTimeEntityRecognizer
- [ ] DimensionEntityRecognizer
- [ ] EmailEntityRecognizer
- [ ] GuidEntityRecognizer
- [ ] HashtagEntityRecognizer
- [ ] IpEntityRecognizer
- [ ] MentionEntityRecognizer
- [ ] NumberEntityRecognizer
- [ ] NumberRangeEntityRecognizer
- [ ] OrdinalEntityRecognizer
- [ ] PhoneNumberEntityRecognizer
- [ ] RegexEntityRecognizer
- [ ] TemperatureEntityRecognizer
- [ ] TextEntityRecognizer
- [ ] UrlEntityRecognizer

#OthersRecognizers 
- [ ] OnContinueConversation
- [ ] OnInstallationUpdateActivity

#TeamsChannelRelated
- [ ] Teams.OnAppBasedLinkQuery
- [ ] Teams.OnCardAction
- [ ] Teams.OnChannelCreated
- [ ] Teams.OnChannelDeleted
- [ ] Teams.OnChannelRenamed
- [ ] Teams.OnFileConsent
- [ ] Teams.OnMessagingExtensionCardButtonClicked
- [ ] Teams.OnMessagingExtensionConfigurationQuerySettingUrl
- [ ] Teams.OnMessagingExtensionConfigurationSetting
- [ ] Teams.OnMessagingExtensionFetchTask
- [ ] Teams.OnMessagingExtensionQuery
- [ ] Teams.OnMessagingExtensionSelectItem
- [ ] Teams.OnMessagingExtensionSubmitAction
- [ ] Teams.OnO365ConnectorCardAction
- [ ] Teams.OnTaskModuleFetch
- [ ] Teams.OnTaskModuleSubmit
- [ ] Teams.OnTeamRenamed

[...]

PR #4625, opened 2020-11-03
feat: update sdk package to 4.11.0-rc2

luhan2017:
## Description

Updated schemas in this PR:

Added:
- Microsoft.ChannelMentionEntityRecognizer (only a entity recognizer, we didn't expose all of them in composer yet)
- Microsoft.ContinueConversationLater (not exposed)
- Microsoft.OnInstallationUpdateActivity (not exposed)
- Microsoft.ThrowException (should expose as a normal action)

## Task Item
#minor


benbrown (approved):

benbrown:
How is the `ChannelMentionEntityRecognizer` going to manifest in Composer?

srinaath (approved):

luhan2017:
> How is the `ChannelMentionEntityRecognizer` going to manifest in Composer?

We didn't expose all the EntityRecognizer in composer yet. The EntityRecognizer binds with a RegexRecognizer in most of the cases. Previously we decide to move the RegexRecognizer editing in the trigger modal from FormEditor part. At that time we removed them all and we think we should have a better way to edit RegexRecognizer for example: edit in lu file. 
We may need to revisit this topic in R12.

This issue lists all the SDK and Composer gaps: 
https://github.com/microsoft/BotFramework-Composer/issues/2560

PR #4667, opened 2020-11-04
fix: let newly created Action be focused automatically

### LINKED ISSUE #4616: When an action node is newly created, it is supposed to be focused immediately

Currently, when you create a new action node in the authoring canvas, the focus goes to the action node below the newly created. 

Screen Shot 2020-11-02 at 12 56 23 PM
1. Creating a new Send a response node

Screen Shot 2020-11-02 at 12 56 37 PM
2. I would expect the Send a response node to be in the focus state, not the node below.

The newly created action node is supposed to be in the focus state immediately after it has been created so that the user can seamlessly continue authoring that node without having to click on it.

yeze322:
## Description
closes #4616 

**Problem**:
New action not focused automatically

**Root cause**:
Related to $designer.id encoded URL.
The focusPath is calculated based on old dialog data where $designerId is for wrong Action.

**Solution**:
Let `saveData()` return a Promise, postpone the calculation of next focused Action to after dialog updated.
It will correct the behavior of:
1. Create new Action via flyout menu
2. Paste Action from clipboard

### DIFF
--- a/Composer/packages/adaptive-flow/__tests__/adaptive-flow-editor/stubs/ShellApiStub.ts
+++ b/Composer/packages/adaptive-flow/__tests__/adaptive-flow-editor/stubs/ShellApiStub.ts
@@ -10,7 +10,7 @@ const fnPromise = () => Promise.resolve({} as any);
 export const ShellApiStub: ShellApi = {
   getDialog: fn,
   saveDialog: fn,
-  saveData: fn,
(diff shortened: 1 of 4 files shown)


yeze322:
@lei9444 please review the change on ShellApi interface

cwhitten (approved):

PR #4673, opened 2020-11-04
fix: Sorting OneOfField options / better default type

### LINKED ISSUE #4654: Change the default type of the value field from object to text in Set a property, Set properties action.
Screen Shot 2020-11-02 at 3 21 28 PM

The default type of the value field in both Set a property, and Set properties is object. We want to change this to text as default.

LouisEugeneMSFT:
## Description

"object" was the default type for fields accepting **valueExpression**, such as **Microsoft.SetProperty**.
This was due to the fact that the default type is simply the first one mentioned in the schema:

image

As such the dropdown for fields using **valueExpression** was listing in that order: object -> array -> string -> boolean -> number, with object being the default.

One solution to this problem could be to modify the ordering in the schema.

The approach taken in this PR instead is to have a mappings of type to weights to give more priority to certain types (such as string) and order based of that.

## Task Item

closes #4654

## Screenshots

image

### DIFF
--- a/Composer/packages/adaptive-form/src/components/fields/OneOfField/utils.ts
+++ b/Composer/packages/adaptive-form/src/components/fields/OneOfField/utils.ts
@@ -25,6 +25,18 @@ function getOptionLabel(schema: JSONSchema7): string {
   return type || 'unknown';
 }
 
+const typePriorityWeights = {
+  string: 5,
+  number: 4,
+  boolean: 3,
+  array: 2,
+  object: 1,
+};
+
+const sortOptionsByTypeWeights = ({ key: type1 }, { key: type2 }): number => {
+  return (typePriorityWeights[type1] || 0) < (typePriorityWeights[type2] || 0) ? 1 : -1;
+};
+
 export function getOptions(
   schema: JSONSchema7,
   definitions?: SchemaDefinitions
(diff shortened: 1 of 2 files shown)


Composer/packages/adaptive-form/src/components/fields/OneOfField/utils.ts
```diff
@@ -25,6 +25,18 @@ function getOptionLabel(schema: JSONSchema7): string {
   return type || 'unknown';
 }
 
+const typePriorityWeights = {
```
GeoffCoxMSFT, line 28: It would be cool to weight the types based on their order in the sdk schema.  Then the schema has control and maybe uischema could provide weight/ordering overrides.

GeoffCoxMSFT (approved):

Composer/packages/adaptive-form/src/components/fields/OneOfField/utils.ts
LouisEugeneMSFT: @GeoffCoxMSFT Technically speaking it was taking the order of the schema before this PR. So changing the order there would have solved the issue. This PR is just a way to fix the issue client side, which is quicker.

PR #4675, opened 2020-11-04
fix: Form dialog hidden fixed feature flags

### LINKED ISSUE #4621: Hide form dialog preview feature switch for 1.2 release
As form dialog won't be fully operational for 1.2, we should hide the feature switch in main and then restore it once 1.3 work starts in main.

GeoffCoxMSFT:
## Description

Hiding form dialogs feature switch for composer 1.2 (hidden:true in the defaults)
There were issues with the feature switch service that caused hidden:true to not hide the switch.

I refactored to fix the following:
- filter statements were returning values and not booleans
- hidden switches needed to follow the process.env for enabled
- hidden switches that were updated weren't properly causing the storage to save again
- the display name and description shouldn't be saved as user data
- the defaults were only applied if the key was not present, needed to merge defaults and currents per key

## Task Item

Fixes #4621 

## Screenshots

image

image

### DIFF
--- a/Composer/packages/lib/shared/src/featureFlagUtils/index.ts
+++ b/Composer/packages/lib/shared/src/featureFlagUtils/index.ts
@@ -28,7 +28,7 @@ export const getDefaultFeatureFlags = (): FeatureFlagMap => ({
   FORM_DIALOG: {
     displayName: formatMessage('Show Form Dialog'),
     description: formatMessage('Show form dialog editor in the canvas'),
-    isHidden: false,
+    isHidden: true,
     enabled: false,
   },
 });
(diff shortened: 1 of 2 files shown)


GeoffCoxMSFT:
@cwhitten Appreciate if you review & approve as this is for 1.2.

pavolum:
So the users data,json will never contain the display name and description text and those will always be pulled from getDefaultFeatureFlags? That makes a lot of sense since its not user specific. 

I will pull my bug fix from my PR since this handles what I had fixed in a better way. Thanks!

pavolum (approved):

hatpick (approved):

PR #4677, opened 2020-11-04
feat: Integrate OneAuth library

### LINKED ISSUE #4131: Download bot assets from PVA and use them to create a new bot project in Composer
Given some PVA context -- tenant ID, environment ID, and bot ID -- download bot assets from PVA and use them to create a Composer bot project.

- [x] Download bot assets in .zip from PVA API
  - Dialog assets, runtime schema, UI schema, `appsettings.json` with serialized publishing profiles
- [x] Convert bot assets into a Composer bot project and save to disk
- [x] Open the bot project within Composer

### LINKED ISSUE #4385: Pull from publish target (PVA scenario)
As part of the PVA integration, there will be a scenario in which the user tries to publish to PVA, and the bot content is out of sync with what is stored in PVA.

In this case, the user will need a way to sync their bot content by "pulling" from the publish target.

For now, we want to enable this only for the PVA scenario, and then later we want to add this functionality in a general sense to any publishing plugin that wants to / can implement it.

Requirements:

- [ ] [P0] Add notification UX when content is out of sync (412) to launch into pull UX
- [ ] [P0] Backup old bot content
- [ ] [P0] Copy new, pulled bot content into bot project directory
- [ ] [P1] Make UX more user-friendly
  - allow user to select where to backup
  - let the user know where they just backed up

tonyanziano:
## Description

This PR updates the Electron-side authentication code to use the new OneAuth library that allows the app to authenticate against Microsoft resources in a compliant way.

This is a requirement for the Electron application due to a recent tenant-level security policy change.

Fixes #4131
Fixes #4385

### Highlights:

- Auth flow is initiated at the server-level `/api/auth/getAccesssToken` route
  - added CSRF protection to this route that should prevent external client code from being able to call this route and get a token via Composer

- This route then delegates the auth flow to the `ElectronAuthProvider` or the `WebAuthProvider` depending on the platform

  - this supports both Electron & Web (to be implemented) auth flows

- The `ElectronAuthProvider` uses the `oneAuthService.ts` to call the OneAuth library's APIs and do OS-level authentication

- New build steps for the OneAuth library


Composer/packages/server/src/utility/electronContext.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export type ElectronContext = {
+  getAccessToken: (
+    params: ElectronAuthParameters
+  ) => Promise<{ accessToken: string; acquiredAt: number; expiryTime: number }>;
+};
+
+let context;
+
+export const useElectronContext = (): ElectronContext => context;
```
srinaath, line 14: Can we just use property accessors for setting and using?

let obj = { _get propName()_ { // getter, the code executed on getting obj.propName }, _set propName(value)_ { // setter, the code executed on setting obj.propName = value } };?

Composer/packages/electron-server/scripts/copy-artifacts.js
```diff
@@ -0,0 +1,92 @@
...
+    log.info('Mac detected. Copying artifacts to: ', destinationDir);
+    break;
+  }
+
+  case 'linux':
+    destinationDir = path.resolve(__dirname, '../dist/linux-unpacked/resources/app.asar.unpacked');
+    log.info('Linux detected. Copying artifacts to: ', destinationDir);
+    break;
+
+  case 'win32':
```
cwhitten, line 49: is `win64` a valid case?

tonyanziano: win64 is not a possible value of `process.platform`:

https://nodejs.org/api/process.html#process_process_platform

cwhitten (approved):

PR #4682, opened 2020-11-05
fix: Update disabled property to booleanExpression across all the kinds + two uischema update

### LINKED ISSUE #4651: Disabling a node causes a validation error and the bot cannot be started.
## Describe the bug

Disabling a node causes a validation error and the bot cannot be started.

http://g.recordit.co/5bcl5DM0iN.gif

5bcl5DM0iN

## Version

nightly build

## Browser

- [x] Chrome

## OS

- [x] macOS
## To Reproduce

1. Create a dialog
2. insert a send activity as the first node
3. disable the node
4. See a validation error appear in the upper right, " end a response: disabled the return type does not match"

## Expected behavior

This should not show a validation error.

luhan2017:
## Description

- Update disable property of Microsoft.SendActivity and Microsoft.Ask from stringExpression to booleanExpression. 
- add uischema, OnChooseIntent
- update helplink, LogAction and TraceActivity

Note, this schema is not actually generated from 4.11.0-rc2, I generated from C# daily package for include the latest change. 

## Task Item

fix #4651


boydc2014:
Can we verify will this fix the normal case

boydc2014 (approved):

Issue #4681, opened 2020-11-05
[Custom action] Misordered using directive


zxyanliu:
## Describe the bug

![image](https://user-images.githubusercontent.com/32497439/98184316-f2bca180-1ebe-11eb-9db1-9c588c87951e.png)

## Version

Composer version: [v1.1.1-nightly.181539.5e9576e](https://github.com/microsoft/BotFramework-Composer-Nightlies/releases/tag/v1.1.1-nightly.181539.5e9576e)

## Browser

- [x] Electron distribution

## OS

- [x] Windows

PR #4684, opened 2020-11-05
fix: misordered using directives in runtime

zxyanliu:
## Description

Fix the error of using directives not ordered alphabetically by the namespaces. 
This will cause an error after uncommenting the line to use custom action.

## Task Item
https://github.com/microsoft/BotFramework-Composer/issues/4681
#minor

## Screenshots

Before:

image

After:

image


runtime/dotnet/azurewebapp/Startup.cs
```diff
@@ -26,9 +26,8 @@
 using Microsoft.Bot.Connector.Authentication;
 using Microsoft.BotFramework.Composer.Core;
 using Microsoft.BotFramework.Composer.Core.Settings;
-using Microsoft.BotFramework.Composer.WebAppTemplates.Authorization;
-
 //using Microsoft.BotFramework.Composer.CustomAction;
```
zxyanliu, line 29:
```suggestion

//using Microsoft.BotFramework.Composer.CustomAction;
```

zxyanliu:
@luhan2017 please review this PR. Thanks!

luhan2017 (approved):

PR #4707, opened 2020-11-06
fix: multi publish still loading in the same profile

VanyLaw:
## Description
fix multi publish still loading in the same profile. 
**Reason**: publish with the same profile will used the same project name and profile name to save the history and status. if user click publish again, those value will be messed up.
**Solution**: disable publish button if already publishing with the same profile.

## Task Item

#minor


Composer/packages/client/src/pages/publish/Publish.tsx
```diff
@@ -86,6 +86,19 @@ const Publish: React.FC<RouteComponentProps<{ projectId: string; targetName?: st
...
+    () => () => {
+      // if current profile is already publishing, set disable to true
+      if (selectedTargetName && selectedTargetName !== 'all') {
+        if (thisPublishHistory?.length >= 0) {
+          return thisPublishHistory[0]?.status === 202;
+        }
+      } else {
+        return selectedTargetName === 'all';
+      }
+    },
```
a-b-r-o-w-n, line 99: Why does this need to return a function?

VanyLaw: fixed

Composer/packages/client/src/pages/publish/Publish.tsx
```diff
@@ -86,6 +86,17 @@ const Publish: React.FC<RouteComponentProps<{ projectId: string; targetName?: st
     [projectId, publishTypes]
   );
 
+  const isPublishDisable = useMemo(() => {
+    // if current profile is already publishing, set disable to true
+    if (selectedTargetName && selectedTargetName !== 'all') {
+      if (thisPublishHistory?.length >= 0) {
```
hatpick, line 92: when this ```if``` is not true, nothing is returned, do you want it to be undefined?

hatpick:
@VanyLaw would you please address the comments?

Composer/packages/client/src/pages/publish/Publish.tsx
VanyLaw: fixed

VanyLaw:
> @VanyLaw would you please address the comments?
> Also please update the branch to resolve the conflicts.
@hatpick Thanks for the comment, I addressed it.

hatpick (approved):

VanyLaw:
Due to publish page was change to a large extent, I need to confirm with @alanlong9278 how to disable those profiles which were used by current publish, make sure it can not select during current publish.
So I close it now. Will send another PR later.

PR #4712, opened 2020-11-06
fix: validate bot name when passed via URL params

### LINKED ISSUE #4441: CreationFlow does not validate name passed in via URL params
## Describe the bug

Navigating to the creation wizard with a url such as:

`http://localhost:3000/projects/create/<template>?name=somename`

will cause the `name` field to be prepopulated with `somename` from the URL params.

However, the filename validation does not run in this scenario until you start typing something, so this initial value never gets validated and can result in trying to create a duplicate bot project.

## Version

`main` as of 10/20/2020

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Open Dev Tools
2. enter into console: `location.href ="http://localhost:3000/projects/create/<template-name>?name=<duplicate-bot-name>"`
3. Observer no duplicate error

## Expected behavior

The initial value from the URL params should be validated when the field is populated.

## Screenshots

image

a-b-r-o-w-n:
## Description

Validates name when passed via URL query string.

Also adds some QOL improvements to electron development.

## Task Item

fixes #4441


Composer/package.json
```diff
@@ -47,7 +47,9 @@
     "start": "cross-env NODE_ENV=production PORT=3000 yarn start:server",
     "startall": "yarn start",
     "start:dev": "concurrently  \"npm:start:client\" \"npm:start:server:dev\"",
+    "start:dev:electron": "concurrently  \"npm:start:client\" \"npm:start:electron\"",
```
tonyanziano, line 50: This might not work on Windows machines due to how long the client takes to start up.

Normally when I try to start the electron server without the dev server running, it throws an error and never starts. Usually the client dev server takes longer to start than the electron process, so I would assume if they are started concurrently that it would error out most of the time, but I would have to test it to make sure.

Composer/packages/electron-server/package.json
```diff
@@ -9,7 +9,7 @@
     "node": ">=12"
   },
   "scripts": {
-    "build": "tsc -p tsconfig.build.json && ncp src/preload.js build/preload.js",
+    "build": "tsc -b tsconfig.json && ncp src/preload.js build/preload.js",
```
tonyanziano, line 12: cool!!!

tonyanziano (approved):
Looks good to me

Composer/package.json
a-b-r-o-w-n: Ok. I can update to wait for dev server to start before starting electron

a-b-r-o-w-n: @tonyanziano Can you check out using `wait-on`?

Composer/packages/server/tsconfig.json
```diff
@@ -1,9 +1,10 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
+    "rootDir": "./src",
     "outDir": "./build/",
     "sourceMap": true,
     "target": "es6"
   },
-  "include": ["src/**/*.ts"]
+  "include": ["src"]
```
hatpick, line 9: Shouldn't it be "src/**/*"?

hatpick (dismissed):
Changes look fine to me, but here's my overall "nit" suggestion to consider :)
I suggest breaking this PR into two PRs, one that has electron-related changes, and one have the validation changes. This helps with reverting changes easier and keeping track of them if things go south.

Issue #4620, opened 2020-11-03
Undo/Redo


srinaath:
## Describe the bug

Undo/ Redo is broken

1. Undo/Redo needs to handle the new encoded paths in url
2. Undo/Redo commits changes before the selector value is updated
3. Undo/Redo needs to be supported for skills

PR #4718, opened 2020-11-09
fix: undo/redo can't navigate to the correct position

lei9444:
## Description
Editors use the designpage location to calculate some important values.
The data flow now is saveData -> commitChanges -> navigate. So undoHistory can't get the correct locations.

in this PR add a commit version for every stack, and update the stack after the designpage location updated

## Task Item
#minor
refs #4620 

## Screenshots
undo


lei9444:
This is not solution to fix the issue, I will have another PR to fix it

lei9444:
close this one useEffect for detect the location update from where
In #4722 fix the issue

PR #4720, opened 2020-11-09
fix: Flow Editor 'onResize' handler not working

yeze322:
## Description
#minor 
Follows #4672 

In #4672, we use `scroll` rather than `bounds` to measure node size. This change broke the layouter of Flow Editor.
This PR fixes the issue by reading `scroll` instead of `bounds` in the `onResize` handler.

**Wrong layout**:
image
**Correct layout**:
image


yeze322:
@alanlong9278 FYI

boydc2014 (approved):

PR #4736, opened 2020-11-09
fix: correct URL when switching page modes

### LINKED ISSUE #4615: In DesignPage set URL by reading the designPageLocationState or to the root bot's main dialog
###  Steps to recreate

In the context of a botproject with a root bot and skill, click on a dialog inside a skill the URL is set correctly `(bot/{​​​​​​​rootBotId}​​​​​​​/skill/{​​​​​​​skillId}​`​​​​​​). Navigate using the sidebar to a different page and come back to the design page. The url is set at `bot/{​​​​​​​skillId}​​​​​​​/dialogs/*`. This URL is wrong as a skill needs to be nested inside the RootBotId in the url always.

### Solution

  
Read the designPageLocationState and set the URL from that state or at a minimum navigate to the RootBot's main dialog whenever the user selects the DesignPage icon on the left nav

beyackle:
## Description

This adds awareness of the project and skill structure to the LG, LU, and QnA pages. They will now display the dialogs of whichever the current selected bot or skill is, keeping that around in the URL and using it to correctly point to the right place back in Design View.

## Task Item

closes #4615


Composer/packages/client/src/pages/language-generation/LGPage.tsx
```diff
@@ -22,19 +22,25 @@ const CodeEditor = React.lazy(() => import('./code-editor'));
...
+  skillId?: string;
 }
 
 const LGPage: React.FC<RouteComponentProps<LGPageProps>> = (props: RouteComponentProps<LGPageProps>) => {
-  const { dialogId = '', projectId = '' } = props;
-  const dialogs = useRecoilValue(validateDialogsSelectorFamily(projectId));
+  const { dialogId = '', projectId = '', skillId = null } = props;
+  const dialogs = useRecoilValue(validateDialogsSelectorFamily(skillId ?? projectId));
+
+  console.log(skillId, projectId, skillId ?? projectId, dialogs);
```
srinaath, line 32: remove logs

Composer/packages/client/src/pages/knowledge-base/QnAPage.tsx
```diff
@@ -25,14 +25,18 @@ const CodeEditor = React.lazy(() => import('./code-editor'));
...
   dialogId?: string;
+  skillId?: string;
 }
 
 const QnAPage: React.FC<QnAPageProps> = (props) => {
-  const { dialogId = '', projectId = '' } = props;
+  const { dialogId = '', projectId = '', skillId = null } = props;
+
+  const baseURL =
+    projectId === skillId || skillId == null ? `/bot/${projectId}/` : `/bot/${projectId}/skill/${skillId}/`;
```
srinaath, line 35: projectId === skillId should never happen correct?

srinaath (changes requested):

Composer/packages/client/src/pages/language-generation/LGPage.tsx
```diff
@@ -22,19 +22,25 @@ const CodeEditor = React.lazy(() => import('./code-editor'));
...
 
   const path = props.location?.pathname ?? '';
 
   const edit = /\/edit(\/)?$/.test(path);
 
+  const baseURL =
+    projectId === skillId || skillId == null ? `/bot/${projectId}/` : `/bot/${projectId}/skill/${skillId}/`;
```
srinaath, line 39: same issue here projectId===skillId is a scenario that should never occur

Composer/packages/client/src/pages/language-understanding/LUPage.tsx
```diff
@@ -21,17 +21,20 @@ const CodeEditor = React.lazy(() => import('./code-editor'));
 const LUPage: React.FC<RouteComponentProps<{
   dialogId?: string;
   projectId: string;
+  skillId: string;
 }>> = (props) => {
-  const { dialogId = '', projectId = '' } = props;
-  const dialogs = useRecoilValue(validateDialogsSelectorFamily(projectId));
+  const { dialogId = '', projectId = '', skillId = null } = props;
```
srinaath, line 26: is there a reason why skillId is instanitated to null and other props are'nt? Can we make the instantiation consistent?

beyackle: I originally had this as `''`, but that broke the `??` melding I was doing. If anything, I think they should all default to `null`, so I'll make that change too.

Composer/packages/client/src/pages/language-generation/LGPage.tsx
beyackle: Oops. Thanks for catching this.

Composer/packages/client/src/pages/language-understanding/LUPage.tsx
beyackle: Update on this - changing them all to `null` caused more typechecking issues downstream, so I found another solution that looks less strange than this. We need a projectId and dialogId in all cases, but the skillId is optional and gets marked as such.

Composer/packages/client/src/pages/knowledge-base/QnAPage.tsx
```diff
@@ -21,17 +21,18 @@ import TableView from './table-view';
 
 const CodeEditor = React.lazy(() => import('./code-editor'));
 
-interface QnAPageProps extends RouteComponentProps<{}> {
-  projectId?: string;
+const QnAPage: React.FC<RouteComponentProps<{
```
hatpick, line 24: RouteComponentProps<TParams> makes the passed TParams partial, so all of them will be optional.

hatpick (commented):
Please use:
```
`${baseUrl}restOfUrl`
```
instead of ```+``` concat.

srinaath (approved):

PR #4744, opened 2020-11-10
[IGNORE ME] 

tonyanziano:


.vscode/launch.json
```diff
@@ -88,7 +88,13 @@
         "DEBUG": "composer*",
         "COMPOSER_ENABLE_ONEAUTH": "false"
       },
-      "outputCapture": "std"
+      "outputCapture": "std",
+      "outFiles": [
+        "${workspaceRoot}/Composer/packages/electron-server/build/**/*.js",
+        "${workspaceRoot}/extensions/**/*.js",
+        "${workspaceRoot}/Composer/packages/server/build/**/*.js"
```
a-b-r-o-w-n, line 95: Does this allow us to put breakpoints in those packages as well?

tonyanziano: it should -- when I was testing it was working

PR #4745, opened 2020-11-10
chore: bundle extensions

a-b-r-o-w-n:
## Description

Use webpack to bundle node extensions to reduce overall app bundle size.

## Task Item

#minor


tonyanziano (approved):
🚀

PR #4751, opened 2020-11-10
feat: Allow importing bot content from an external source

### LINKED ISSUE #4125: Implement "import" protocol handler
Implement a new protocol route `bfcomposer://import` that can accept a `source` and `payload` parameter and use these to 
download bot assets from a remote location outside of Composer.

`source` - where the bot is being imported from
`payload` - metadata about the assets to be imported (can specify a URL to request the assets from)

**Ex:**

`bfcomposer://import?source=pva&payload=%7B%22url%22%3A%22https%3A%2F%2Fwww.importme.com%2Fimport%22%2C%22username%22%3A%22johndoe1%22%7D` 

- [x] [P0] Pull bot content down from some service, launch creation wizard with populated fields, create a new bot from content
- [ ] [P0] Navigate to additional deep link after creation
- [ ] [P1] Implement new UX leveraging Notification component
- [ ] [P1] Recognize that the bot content being imported has already been opened as a project -- prompt to re-open

tonyanziano:
## Description

This PR allows Composer to accept a deep link with the following form:

`bfcomposer://import?source={source}&payload={urlEncodedJSONWithMetadata}`

This will then display UI that allows the user to sign in to the external service -- if necessary -- and download the bot content to either be saved as a new bot project within Composer, or imported into an existing project with a matching identifier.

**NOTES**:
- tests on the way
- ~~small refactors on the way~~

## Task Item

Fixes #4125, #4131

## Screenshots

**Connecting to service**

import5
import1
import2
import3

**Update existing project**

import4

**Import as new project**

import6
import7


Composer/packages/client/src/components/ImportModal/ImportModal.tsx
```diff
@@ -0,0 +1,338 @@
...
+  const [importPayload, setImportPayload] = useState<ImportPayload>({ name: '', description: '' });
+  const [importedProjectInfo, setImportedProjectInfo] = useState<ImportedProjectInfo | undefined>(undefined);
+  const [modalState, setModalState] = useState<ImportModalState>('connecting');
+  const [existingProject, setExistingProject] = useState<ExistingProjectInfo | undefined>(undefined);
+  const [error, setError] = useState<any>(undefined);
+  const [backupLocation, setBackupLocation] = useState<string>('');
+  const { addNotification } = useRecoilValue(dispatcherState);
+
+  const importAsNewProject = useCallback((info: ImportedProjectInfo) => {
+    // navigate to creation flow with template selected
```
hatpick, line 66: Have you considered passing this data as state to ```navigate``` function instead of the query params?
The reason I'm asking is that this url pattern is being only used to pass the data (e.g., url is not going to be used to share a deep linked content)

Composer/packages/client/src/components/ImportModal/ImportModal.tsx
```diff
@@ -0,0 +1,338 @@
...
+
+      const backupAndSave = async () => {
+        try {
+          // call server to do backup and then save to existing project
+          let res = await fetch(`/api/projects/${existingProject.id}/backup`, { method: 'POST' });
+          const { path } = await res.json();
+          setBackupLocation(path);
+
+          const { eTag, templateDir } = importedProjectInfo;
+          res = await fetch(`/api/projects/${existingProject.id}/copyTemplateToExisting`, {
```
hatpick, line 100: is there a particular reason that you are using the vanilla fetch api instead of axios (httpClient)?

Composer/packages/client/src/components/ImportModal/ImportModal.tsx
```diff
@@ -0,0 +1,338 @@
...
+              }),
+            });
+            addNotification(notification);
+            navigate(`/bot/${existingProject?.id}`);
+          } else {
+            const err = await res.json();
+            throw err;
+          }
+        } catch (e) {
+          console.error('Something went wrong while saving bot to existing project: ', e);
```
hatpick, line 123: Do you want to keep this log?

Composer/packages/client/src/components/ImportModal/ImportModal.tsx
```diff
@@ -0,0 +1,338 @@
...
+            const { description, name } = importPayload;
+
+            let res = await fetch(
+              `/api/import/${importSource}?payload=${encodeURIComponent(JSON.stringify(importPayload))}`,
+              {
+                method: 'POST',
+              }
+            );
+            if (res.status !== 200) {
+              throw `Something went wrong during import: ${res.status} ${res.statusText}`;
```
hatpick, line 146: ```throw new Error('string')```

Composer/packages/client/src/components/ImportModal/ImportModal.tsx
```diff
@@ -0,0 +1,338 @@
...
+            const err = await res.json();
+            throw err;
+          }
+        } catch (e) {
+          console.error('Something went wrong while saving bot to existing project: ', e);
+          setError(e);
+          setModalState('failed');
+        }
+      };
+      backupAndSave();
```
hatpick, line 128: await this?

Composer/packages/client/src/components/ImportModal/ImportModal.tsx
```diff
@@ -0,0 +1,338 @@
...
+            }
+            importAsNewProject(projectInfo);
+          } catch (e) {
+            // something went wrong, abort and navigate to the home page
+            console.error(`Aborting import: ${e}`);
+            navigate('/home');
+          }
+        }
+      };
+      importBotContent();
```
hatpick, line 172: await this?

Composer/packages/client/src/components/ImportModal/ImportModal.tsx
```diff
@@ -0,0 +1,338 @@
...
+            throw `Something went wrong during authenticating import: ${res.status} ${res.statusText}`;
+          }
+          setModalState('downloadingContent');
+        } catch (e) {
+          // something went wrong, abort and navigate to the home page
+          console.error(`Aborting import: ${e}`);
+          navigate('/home');
+        }
+      };
+      signIn();
```
hatpick, line 196: await this?

Composer/packages/client/src/components/ImportModal/ImportModal.tsx
```diff
@@ -0,0 +1,338 @@
...
+      const signIn = async () => {
+        try {
+          const res = await fetch(
+            `/api/import/${importSource}/authenticate?payload=${encodeURIComponent(JSON.stringify(importPayload))}`,
+            {
+              method: 'POST',
+            }
+          );
+          if (res.status !== 200) {
+            throw `Something went wrong during authenticating import: ${res.status} ${res.statusText}`;
```
hatpick, line 187: same comment as above

Composer/packages/client/src/components/ImportModal/ImportModal.tsx
```diff
@@ -0,0 +1,338 @@
...
+          }
+          setModalState('downloadingContent');
+        } catch (e) {
+          // something went wrong, abort and navigate to the home page
+          console.error(`Aborting import: ${e}`);
```
hatpick, line 192: same comment

Composer/packages/client/src/components/ImportModal/ImportModal.tsx
```diff
@@ -0,0 +1,338 @@
...
+
+  useEffect(() => {
+    if (location && location.href) {
+      try {
+        // parse data from url and store in state
+        const url = new URL(location.href);
+        const source = url.searchParams.get('source');
+        const payload = url.searchParams.get('payload');
+        if (!source || !payload) {
+          throw 'Missing source or payload.';
```
hatpick, line 208: new Error

Composer/packages/client/src/components/ImportModal/ImportModal.tsx
```diff
@@ -0,0 +1,338 @@
...
+        }, CONNECTING_STATUS_DISPLAY_TIME);
+      } catch (e) {
+        console.error('Aborting import: ', e);
+        navigate('/home');
+      }
+    }
+  }, []);
+
+  const cancel = useCallback(() => {
+    console.log('Cancelled import.');
```
hatpick, line 223: remove?

Composer/packages/client/src/components/ImportModal/ImportModal.tsx
```diff
@@ -0,0 +1,338 @@
...
+    navigate('/home');
+  }, []);
+
+  const openExistingProject = useCallback(() => {
+    navigate(`/bot/${existingProject?.id}`);
```
hatpick, line 228: this may end up being ```/bot/undefined``` url

Composer/packages/client/src/components/ImportModal/ImportModal.tsx
```diff
@@ -0,0 +1,338 @@
...
+  }, [existingProject]);
+
+  const createNewProxy = useCallback(() => {
+    if (importedProjectInfo) {
+      importAsNewProject(importedProjectInfo);
+    }
+  }, [importedProjectInfo]);
```
hatpick, line 235: add importAsNewProject to the list

Composer/packages/client/src/components/ImportModal/ImportModal.tsx
```diff
@@ -0,0 +1,338 @@
...
+          <Dialog
+            hidden={false}
+            dialogContentProps={{ type: DialogType.normal }}
+            modalProps={{ isBlocking: true }}
+            styles={{ main: { display: 'none' } }}
+          />
+        );
+
+      default:
+        return <div style={{ display: 'none' }} />;
```
hatpick, line 324: maybe throw here to catch a run time mistake

Composer/packages/client/src/components/ImportModal/ImportModal.tsx
```diff
@@ -0,0 +1,338 @@
+/** @jsx jsx */
+import { jsx, css } from '@emotion/core';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
+import { navigate, RouteComponentProps } from '@reach/router';
+import { Dialog, DialogType, DialogFooter } from 'office-ui-fabric-react/lib/Dialog';
```
hatpick, line 5: Have you considered using DialogWrapper from shared ui lib? It basically uses the same Dialog from Fabric but allows us to do app wide customization+styling in the future.

Composer/packages/client/src/components/ImportModal/ImportModal.tsx
```diff
@@ -0,0 +1,338 @@
...
+  }, [existingProject]);
+
+  const createNewProxy = useCallback(() => {
+    if (importedProjectInfo) {
+      importAsNewProject(importedProjectInfo);
+    }
+  }, [importedProjectInfo]);
+
+  const modalContent = useMemo(() => {
+    switch (modalState) {
```
hatpick, line 238: I highly recommend to extract the larger Dialog components below to their own components, it makes code more readable, debugging in dev tools much easier and allows for better testing of each individual (dialog) component.

Composer/packages/client/src/components/ImportModal/ImportStatus.tsx
```diff
@@ -0,0 +1,167 @@
...
+          <ProgressIndicator label={label} styles={{ itemName: { textAlign: 'center' } }} />
+        </Dialog>
+      );
+    }
+
+    case 'downloading': {
+      const sourceName = getUserFriendlySource(source);
+      const label = (
+        <p style={{ fontSize: 16, whiteSpace: 'normal' }}>
+          {formatMessage('Importing ')}
```
hatpick, line 99: please consider doing this using formatMessage.rich
https://github.com/format-message/format-message/tree/master/packages/format-message#formatmessagerichpattern-args-locales

Composer/packages/client/src/components/ImportModal/ImportStatus.tsx
```diff
@@ -0,0 +1,167 @@
...
+        aria-label={formatMessage('Data transferring between services')}
+        src={dataTransferLine}
+        style={{ margin: '0 16px', width: '78px' }}
+      />
+    </React.Fragment>
+  );
+}
+
+// TODO: create a type for possible publish sources
+export function getUserFriendlySource(source?: string): string {
```
hatpick, line 159: Why is this exported? Also please do define then use.

Composer/packages/server/src/controllers/import.ts
```diff
@@ -0,0 +1,83 @@
...
+type StartImportRequest = Request & {
+  params: {
+    source: ExternalContentProviderType;
+  };
+  query: {
+    payload: string;
+  };
+};
+
+type AuthenticateRequest = StartImportRequest;
```
hatpick, line 20: why the alias?

Composer/packages/server/src/controllers/import.ts
```diff
@@ -0,0 +1,83 @@
...
+  const metadata = JSON.parse(payload);
+
+  const contentProvider = contentProviderFactory.getProvider({ type: source, metadata });
+  if (contentProvider) {
+    if (contentProvider.authenticate) {
+      const accessToken = await contentProvider.authenticate();
+      return res.status(200).json({ accessToken });
+    }
+    // no-op
+    res.sendStatus(200);
```
hatpick, line 35: Shouldn't ```next``` be called?

Composer/packages/server/src/controllers/import.ts
```diff
@@ -0,0 +1,83 @@
...
+      await extractZip(zipPath, { dir: templateDir });
+      log('Done extracting.');
+      await contentProvider.cleanUp();
+
+      let alias;
+      if (contentProvider.getAlias) {
+        alias = await contentProvider.getAlias();
+      }
+
+      setTimeout(() => {
```
hatpick, line 66: why the timeout?

Composer/packages/server/src/controllers/import.ts
```diff
@@ -0,0 +1,83 @@
...
+        res.json({ alias, eTag, templateDir, urlSuffix });
```
hatpick, line 67: res.status(200).json(data)?

Composer/packages/server/src/externalContentProvider/contentProviderFactory.ts
```diff
@@ -0,0 +1,26 @@
+import { ExternalContentProvider, ContentProviderMetadata } from './externalContentProvider';
+import { PowerVirtualAgentsProvider, PowerVirtualAgentsMetadata } from './powerVirtualAgentsProvider';
+import { ExternalContentProviderType } from './providerType';
+
+type ProviderInfo = PowerVirtualAgentsProviderInfo; // union with additional provider info types (ex. PowerVirtualAgentsProviderInfo | ABSProviderInfo | etc.)
+
+type ProviderInfoBase<T extends ContentProviderMetadata> = {
+  type: ExternalContentProviderType;
```
hatpick, line 8: nit: type is a reserved keyword, use kind?

Composer/packages/server/src/externalContentProvider/externalContentProvider.ts
```diff
@@ -0,0 +1,53 @@
+export interface ContentProviderMetadata {
+  [key: string]: any;
```
hatpick, line 2: Record<string, any>

Composer/packages/server/src/externalContentProvider/externalContentProvider.ts
```diff
@@ -0,0 +1,53 @@
+export interface ContentProviderMetadata {
```
hatpick, line 1: since it's an interface maybe name it ```IContentProviderMetadata```

Composer/packages/server/src/externalContentProvider/powerVirtualAgentsProvider.ts
```diff
@@ -0,0 +1,173 @@
...
+import { join } from 'path';
+
+import logger from '../logger';
+import { authService } from '../services/auth/auth';
+
+import { BotContentInfo, ContentProviderMetadata, ExternalContentProvider } from './externalContentProvider';
+
+const log = logger.extend('pva-provider');
+
+const COMPOSER_1P_APP_ID = 'ce48853e-0605-4f77-8746-d70ac63cc6bc';
```
hatpick, line 13: Just checking if it's fine to have this checked in with the code

Composer/packages/server/src/externalContentProvider/powerVirtualAgentsProvider.ts
```diff
@@ -0,0 +1,173 @@
...
+        method: 'GET',
+        headers: await this.getRequestHeaders(),
+      };
+      const result = await fetch(url, options);
+
+      const eTag = result.headers.get('etag') || '';
+      const contentType = result.headers.get('content-type');
+      if (!contentType || contentType !== 'application/zip') {
+        const json = await result.json();
+        throw `Did not receive zip back from PVA: ${prettyPrintError(json)}`;
```
hatpick, line 91: new Error(msg)

Composer/packages/server/src/externalContentProvider/powerVirtualAgentsProvider.ts
```diff
@@ -0,0 +1,173 @@
...
+      }
+
+      // write the zip to disk
+      if (result && result.body) {
+        ensureDirSync(this.tempBotAssetsDir);
+        const zipPath = join(this.tempBotAssetsDir, 'bot-assets.zip');
+        const writeStream = createWriteStream(zipPath);
+        await new Promise((resolve, reject) => {
```
hatpick, line 99: Is it fine to overwrite an existing one?

Composer/packages/server/src/externalContentProvider/powerVirtualAgentsProvider.ts
```diff
@@ -0,0 +1,173 @@
...
+          writeStream.once('finish', resolve);
+          writeStream.once('error', reject);
+          result.body.pipe(writeStream);
+        });
+        return { eTag, zipPath, urlSuffix: this.getDeepLink() };
+      } else {
+        throw 'Response containing zip does not have a body';
```
hatpick, line 106: new Error

Composer/packages/server/src/externalContentProvider/powerVirtualAgentsProvider.ts
```diff
@@ -0,0 +1,173 @@
...
+      }
+    } catch (e) {
+      throw `Error while trying to download the bot content: ${prettyPrintError(e)}`;
```
hatpick, line 109: new Error

Composer/packages/server/src/server.ts
```diff
@@ -57,6 +57,8 @@ export async function start(electronContext?: ElectronContext): Promise<number |
   setEnvDefault('COMPOSER_BUILTIN_EXTENSIONS_DIR', path.resolve(__dirname, '../../../../extensions'));
   // Composer/.composer/extensions
   setEnvDefault('COMPOSER_REMOTE_EXTENSIONS_DIR', path.resolve(__dirname, '../../../.composer/extensions'));
+  setEnvDefault('COMPOSER_TEMP_DIR', path.resolve(__dirname, '../../../.composer/.temp'));
```
hatpick, line 60: nit: I'd do this similar to .git structure, .composer dir and then normal dirs under (non dot directories):
.composer/backup
.composer/temp

Composer/packages/server/src/services/project.ts
```diff
@@ -278,4 +362,30 @@ export class BotProjectService {
       return '';
     }
   };
+
+  public static backupProject = async (project: BotProject): Promise<string> => {
```
hatpick, line 366: What's the difference between this and backupProject defined under controller

hatpick (changes requested):

Composer/packages/client/src/components/CreationFlow/DefineConversation.tsx
```diff
@@ -146,6 +157,15 @@ const DefineConversation: React.FC<DefineConversationProps> = (props) => {
...
   const { formData, formErrors, hasErrors, updateField, updateForm } = useForm(formConfig);
+  const [isImported, setIsImported] = useState<boolean>(false);
+
+  useEffect(() => {
+    if (props.location?.search) {
+      const decoded = decodeURIComponent(props.location.search);
+      const { imported } = querystring.parse(decoded);
+      setIsImported(imported === 'true');
+    }
+  }, []);
```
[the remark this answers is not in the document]
tonyanziano: Yeah, I've tested it out and it only needs to be checked when it's mounted.

Composer/packages/client/src/components/ImportModal/ImportModal.tsx
tonyanziano: I had no idea this was a thing 😮 I think that's a way better approach.

Composer/packages/client/src/components/ImportModal/ImportModal.tsx
tonyanziano: I was under the impression you couldn't pass an async function to `useCallback()`. I'll change this

Composer/packages/client/src/components/ImportModal/ImportModal.tsx
tonyanziano: I was under the impression you couldn't pass an async function to `useCallback()`. I'll change this

Composer/packages/client/src/components/ImportModal/ImportModal.tsx
tonyanziano: I plan to add better error handling to this flow in the case of errors during sign-in (probably a notification). So I'd prefer to keep some sort of log until then.

Composer/packages/client/src/components/ImportModal/ImportModal.tsx
tonyanziano: Nope, I can switch over to Axios

Composer/packages/client/src/components/ImportModal/ImportModal.tsx
tonyanziano: Is there a pattern for awaiting inside of `useEffect()`? The research I did online pointed to this pattern in which I can't await signIn because it's not inside of an async function.

Composer/packages/client/src/components/ImportModal/ImportModal.tsx
tonyanziano: Same question as below

Composer/packages/client/src/components/ImportModal/ImportStatus.tsx
tonyanziano: This PR is a chunk of a larger feature branch where this function is reused. I'll remove the export for now

Composer/packages/client/src/components/ImportModal/ImportModal.tsx
tonyanziano: I don't really see the benefit in defining a new Dialog type over in the DialogWrapper for the styles used in the Import flow.

I would have to extend the DialogWrapper props to allow me to control the `isBlocking` and whether or not the close button shows on the dialog.

The dialog styles I'm using in this flow are only shared between different states of `ImportStatus`, so if I need to change the styles, I can just change the styles object in that file.

If this becomes a problem in the future, I can change it as needed.

Composer/packages/server/src/controllers/import.ts
tonyanziano: I actually have no idea -- good catch

Composer/packages/server/src/externalContentProvider/powerVirtualAgentsProvider.ts
tonyanziano: Yup, it's not a secret 👍

Composer/packages/server/src/externalContentProvider/powerVirtualAgentsProvider.ts
tonyanziano: Yes, but I'll timestamp it just to be careful

Composer/packages/server/src/server.ts
tonyanziano: works for me

Composer/packages/server/src/services/project.ts
tonyanziano: This is another artifact from the larger feature branch. This same logic is used by the "Pull" feature which will be broken out into a separate PR following this one. I will delete this one for now until I open that PR.

Composer/packages/client/src/components/ImportModal/ImportModal.tsx
hatpick: What you are doing is perfectly fine, I was wondering if this component relies on the results of those two async functions since I'm not super familiar with the business logic here.

Another way of achieving the same is using an IIFE:
```
(async () => {
   await awaitable();
})();
```

PR #4756, opened 2020-11-11
fix: Remove unnecessary code

peterbozso:
#minor

It doesn't make sense to `Hard code the settings path to 'ComposerDialogs' while deployment` on this line, since the `bot` configuration will always have a value at this point, thanks to the ` builder.UseBotPathConverter` call on line 26. And inside that method it's actually being set to the value `ComposerDialogs` in case of a deployment.


a-b-r-o-w-n:
@luhan2017 Can you take a look at this one?

luhan2017 (approved):

peterbozso:
@luhan2017 is there anything keeping this from being merged?

PR #4760, opened 2020-11-11
fix: zip file can not be deleted

### LINKED ISSUE #4704: publishing issue, temporary folder can't be deleted if any exception happens
## Describe the bug
If exception happens in publishing thread, the temporary archived code.zip can't be deleted any more.
If try to execute another publishing instance, since the code.zip couldn't be deleted, composer will throw an exception, but this exception will not be caught by azurePublish plugin, and the publishing status won't be updated, the publishing page would get stuck at loading. 

## Version
main branch and electron build

## Browser

- [x] Electron distribution
- [x] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Add a publishing profile and publish the bot
2. Try to throw an exception, for example close the vpn while sending the zip folder to azure (the sending process would be shut down by composer)
3. The code.zip temporary folder would never be deleted.
4. If try to execute another publishing instance, since the code.zip couldn't be deleted, composer will throw an exception, but this exception will not be caught by azurePublish plugin, and the publishing status won't be updated, the publishing page would get stuck at loading. 

## Expected behavior
1. The temporary folder could be deleted successfully.
2. Exception could be caught by azure publish plugin.

VanyLaw:
## Description
Reason: when zip upload failed, the read file stream of zip did not close.

## Task Item
close #4704


extensions/azurePublish/src/deploy.ts
```diff
@@ -149,19 +149,30 @@ export class BotProjectDeploy {
     const publishEndpoint = `https://${
       hostname ? hostname : name + '-' + env
     }.scm.azurewebsites.net/zipdeploy/?isAsync=true`;
+    const fileReadStream = fs.createReadStream(zipPath, {autoClose:true});
+    fileReadStream.on( "close", function () {
```
boydc2014, line 153: Can we remove this debugging prints?

VanyLaw: done.

boydc2014 (approved):

extensions/azurePublish/src/deploy.ts
```diff
@@ -149,19 +149,27 @@ export class BotProjectDeploy {
     const publishEndpoint = `https://${
       hostname ? hostname : name + '-' + env
     }.scm.azurewebsites.net/zipdeploy/?isAsync=true`;
+    const fileReadStream = fs.createReadStream(zipPath, {autoClose:true});
+    fileReadStream.on( "error", function (err) {
+      console.log(err);
```
a-b-r-o-w-n, line 154: Let's use the composer's logger for this instead.

extensions/azurePublish/src/index.ts
```diff
@@ -166,9 +166,14 @@ export default async (composer: ExtensionRegistration): Promise<void> => {
...
     private async cleanup(resourcekey: string) {
-      const projFolder = this.getRuntimeFolder(resourcekey);
-      await emptyDir(projFolder);
-      await rmdir(projFolder);
+      try {
+        const projFolder = this.getRuntimeFolder(resourcekey);
+        await emptyDir(projFolder);
+        await rmdir(projFolder);
+      } catch (error) {
+        console.error(error);
```
a-b-r-o-w-n, line 174: Use the composer logger here as well.

extensions/azurePublish/src/index.ts
```diff
@@ -334,34 +339,53 @@ export default async (composer: ExtensionRegistration): Promise<void> => {
...
+          mergedSettings,
+          runtime,
+          project.id,
+          profileName,
+          jobId,
+          resourcekey,
+          customizeConfiguration
+        );
+      } catch (err) {
+        console.log(err);
```
a-b-r-o-w-n, line 372: Composer logger.

a-b-r-o-w-n (approved):

PR #4762, opened 2020-11-11
chore: Updated strings from building

GeoffCoxMSFT:
## Description

Updated strings from building main

## Task Item

#minor

## Screenshots

N/A


Composer/packages/server/src/locales/en-US.json
```diff
@@ -2369,12 +2357,12 @@
...
   },
+  "spaces_and_special_characters_are_not_allowed_use__2a61c454": {
+    "message": "Spaces and special characters are not allowed. Use letters, numbers, -, or _ and don''t use number at the beginning."
+  },
   "spaces_and_special_characters_are_not_allowed_use__48acec3c": {
     "message": "Spaces and special characters are not allowed. Use letters, numbers, -, or _."
   },
-  "spaces_and_special_characters_are_not_allowed_use__d24a8636": {
-    "message": "Spaces and special characters are not allowed. Use letters, numbers, -, or _, and begin the name with a letter."
-  },
```
beyackle, line 2377: Why did this one revert? I changed it on purpose because it was misleading (you can't begin a name with `-` or `_` either).

GeoffCoxMSFT: Not sure - it was the result of the build.

hatpick (approved):

PR #4764, opened 2020-11-11
ci: increase unit test timeout to 30 min

a-b-r-o-w-n:
## Description

Our unit test workflow has been bumping up against the 20 min timeout.

## Task Item

#minor


tonyanziano (approved):
🕓

PR #4771, opened 2020-11-11
feat: Property Editor Header when botname is selected in Project tree

srinaath:
This PR adds a header to the property  editor when the botname is clicked in the project tree and no dialogs are selected. Default routing to main dialog is stopped.
nav

### DIFF
--- a/Composer/packages/client/src/pages/design/PropertyEditorHeader.tsx
+++ b/Composer/packages/client/src/pages/design/PropertyEditorHeader.tsx
@@ -41,10 +41,10 @@ const styles = {
     margin-bottom: 10px;
     white-space: pre-line;
     font-size: ${FontSizes.size12};
+    margin-bottom: 10px;
   `,
 
   helplink: css`
-    margin-top: 15px;
     font-size: ${FontSizes.size12};
   `,
 };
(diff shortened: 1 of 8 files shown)


Composer/packages/client/src/pages/design/PropertyEditorHeader.tsx
```diff
@@ -0,0 +1,111 @@
...
+  helplink: css`
+    margin-top: 15px;
+    font-size: ${FontSizes.size12};
+  `,
+};
+
+export type PropertyEditorHeaderProps = {
+  projectData: { isRootBot: boolean; isRemote: boolean };
+  botName: string;
+  description?: string;
```
hatpick, line 55: is this gonna be used anywhere?

srinaath: There is actually no use of the property. I was hoping it could be something that @zhixzhan  passes to the component initially. But the static text is just inside the propertyeditorHeader component now

srinaath:
@zhixzhan tests dont pass on ur branch. I think its safe to take this PR into ur branch when u feel its comfortable as i made sure no new tests fail because of this branch.

hatpick (approved):

PR #4778, opened 2020-11-12
chore: update l10n files (2020-11-11)

beyackle:
## Description

Another set of changes just got merged into main, so we should update the l10n stuff accordingly.

## Task Item

#minor


beyackle:
Stale; will do a new update in a separate PR.

PR #4799, opened 2020-11-13
fix: Move doesn't work

yeze322:
## Description
#minor
follows #4740  #4792


yeze322:
open another work item to avoid having this kind of bugs again #4800

cwhitten (approved):

PR #4803, opened 2020-11-13
fix: remove special case for single-bot view in tree

### LINKED ISSUE #4797: Context menu does not show up in a single bot (only rootbot) context
## Describe the bug
The Add toolbar will no longer exist after the BotProjects work. Essentially, the context menus to add dialog, trigger etc are existing only when a root bot has skills. In the context of a single rootbot with no skills the Project tree view is incorrect.
Screen Shot 2020-11-12 at 5 41 15 PM

In the above image we just have a root bot and no context menu.

Screen Shot 2020-11-12 at 5 42 23 PM

The above image is how it should look like even in a single project context in order to add dialogs or triggers to the root bot

beyackle:
## Description

This removes the check to see if a bot is alone in a project; it was needed so the tree would look right before the BotProject branch was merged in. Now that we need to see the bot headers all the time, though, we shouldn't do that check.

## Task Item

closes #4797

## Screenshots

image


srinaath:
Nice!!

@zhixzhan this is the fix for the single bot context.

srinaath (approved):

PR #4809, opened 2020-11-13
fix: misc creation fixes

### LINKED ISSUE #4601: Minor UI features to support R11 remote template creation
[ ] add a banner above the template list in the creation flow when REMOTE_TEMPLATE feature flag is turned on. Banner contains text letting the user know that remote/conversational core templates are in the list. 
[ ] When a project is created with a remote template the create load time is much longer. Need to add loading UI to the spinner to indicate creation progress. Currently its a Spinner for ~30 seconds which may convince some users that the app is hung up.

pavolum:
## Description

- Adding informational banner to creation flow when conversational core template feature is turned on
- Adding documentationLink to feature flag type and added a doc link for conversational core feature flag
- Altering feature flag text

## Task Item

fixes #4601 

## Screenshots
Feature flag is hidden by default but if it werent it would look like this (Strings and terminology are not final)
image

image


Composer/packages/client/src/pages/setting/app-settings/FeatureFlagCheckBox.tsx
```diff
@@ -14,13 +16,19 @@ type FeatureFlagCheckBoxProps = {
...
 
-const renderLabel = (featureName: string, description: string) => () => (
+const renderLabel = (featureName: string, description: string, documentationLink?: string) => () => (
   <span>
     <span css={styles.featureFlagTitle}>{`${featureName}.`}</span>
-    {` ${description}`}
+    {` ${description}.`}
+    {documentationLink && (
+      <Link href={documentationLink} target="_blank">
+        {formatMessage(' Learn More.')}
```
hatpick, line 29: instead of the space in the text, use css to add the margin you want

Composer/packages/client/src/pages/setting/app-settings/FeatureFlagCheckBox.tsx
```diff
@@ -14,13 +16,19 @@ type FeatureFlagCheckBoxProps = {
...
+  documentationLink?: string;
   toggleFeatureFlag: (FeatureFlagKey: FeatureFlagKey, enabled: boolean) => void;
 };
 
-const renderLabel = (featureName: string, description: string) => () => (
+const renderLabel = (featureName: string, description: string, documentationLink?: string) => () => (
   <span>
     <span css={styles.featureFlagTitle}>{`${featureName}.`}</span>
-    {` ${description}`}
+    {` ${description}.`}
```
hatpick, line 26: same here

Composer/packages/lib/shared/src/constant.ts
```diff
@@ -173,3 +173,5 @@ export const LgTemplateSamples = {
     };
   },
 };
+
+export const creationFeatureFlagReadMe = 'https://aka.ms/AAabzf9';
```
hatpick, line 177: nit: I think resources related to a feature flagged feature should come from the feature code.

pavolum: What do you mean by this? Like the documentation should live near the code in GH? The aka.ms link redirects to a GH file that Andrew is working on for this feature.

mewa1024:
@emivers8

Composer/packages/lib/shared/src/constant.ts
hatpick: creationFeatureFlagReadMe variable is defined under lib/shared right now, I think it should be near the creation flow code itself.

pavolum: Right but this constant is used in both the shared and client workspace. Since client references shared and shared does not reference client, this is the only place it can be

benbrown (dismissed):
With the caveat that before we ship, we need to update the actual text that is displayed when this flag is enabled, this is otherwise good to go.

Composer/packages/client/src/components/CreationFlow/CreateOptions.tsx
```diff
@@ -20,14 +20,18 @@ import {
   DetailsRow,
 } from 'office-ui-fabric-react/lib/DetailsList';
 import { Sticky, StickyPositionType } from 'office-ui-fabric-react/lib/Sticky';
-import { BotTemplate } from '@bfc/shared';
+import { BotTemplate, creationFeatureFlagReadMe } from '@bfc/shared';
```
benbrown, line 23: Since this URL is temporary, lets just put it inline in here instead of shared.  fewer places to clean up!

Composer/packages/client/src/components/CreationFlow/CreateOptions.tsx
```diff
@@ -265,6 +274,14 @@ export function CreateOptions(props: CreateOptionsProps) {
       <DialogWrapper isOpen {...dialogWrapperProps} dialogType={DialogTypes.CreateFlow} onDismiss={onDismiss}>
         <ChoiceGroup label={choiceGroupTitle} options={choiceOptions} selectedKey={option} onChange={handleChange} />
         <h3 css={listHeader}>{formatMessage('Examples')}</h3>
+        {featureFlags?.REMOTE_TEMPLATE_CREATION_EXPERIENCE?.enabled && (
```
benbrown, line 277: Add a TODO comment here about removing this code once the feature flag is removed.

benbrown (changes requested):

PR #4829, opened 2020-11-16
fix: prevent minification of AbortSignal

boydc2014:
Update webpackconfig of azure publish plugin to prevent the minification of AbortSignal type

#minor


luhan2017 (approved):
Verified for azure publish scenario

image

a-b-r-o-w-n (approved):

PR #4831, opened 2020-11-16
fix: Added a build:dev yarn task to PVA Publish to create source maps

tonyanziano:
## Description

Webpack-generated source maps for the node-side code of the PVA publish extension do not work inside of VS Code and are not recognized by the debugger.

I have added a `yarn build:dev` task that will skip webpack when compiling the node code and will instead pass it to `tsc` to be compiled. The `tsc`-generated source maps do work with the debugger.

This should have no effect on the current production environment, and the developer will have to explicitly opt-in to this mode by calling `yarn build:dev`.

#minor


extensions/pvaPublish/webpack.config.js
```diff
@@ -1,54 +1,66 @@
...
+      },
+      plugins: [],
     },
-    resolve: {
-      extensions: ['.js', '.ts', '.tsx'],
+    {
+      entry: './src/node/index.ts',
+      mode: isDevelopment ? 'development' : 'production',
+      target: 'node',
+      devtool: isDevelopment ? 'source-map' : undefined,
```
a-b-r-o-w-n, line 46: shouldn't this be the other way around?

tonyanziano: I want it to behave like this:

```
if (isDevelopment)
  // generate source maps
else 
  // don't generate source maps
```

a-b-r-o-w-n: interesting. it's usually the other way around. until we start reporting errors, I guess it doesn't matter too much.

tonyanziano: I would think you would want source maps when developing so you can step through your code and work out any bugs, and then you would not want them in production to reduce size and because you aren't often debugging the production code.

We also don't copy the source files of the extensions into the production app so the source maps wouldn't even work would they?

tonyanziano:
The source mapping might actually be a little wonky... I need to investigate and see if the breakpoints are being hit properly

extensions/pvaPublish/webpack.config.js
a-b-r-o-w-n: Normally, you would use something like `cheap-eval-source-map` for dev because of build speed. In this case, we aren't concerned as much with build speed.

tonyanziano:
I tested the branch with a clean build and the source maps are working as expected 👍

extensions/pvaPublish/tsconfig.json
```diff
@@ -1,7 +1,8 @@
 {
```
GeoffCoxMSFT, line 1: Nit: Should you be importing a base tsconfig that has the stricter settings?

GeoffCoxMSFT (approved):

extensions/pvaPublish/tsconfig.json
tonyanziano: I have another PR incoming for this package that deals with types, I will add some of the strict settings in that PR.

PR #4835, opened 2020-11-16
fix: keep highlight on item with the menu open

### LINKED ISSUE #4810: Project tree focus not locking when context menu is opened
Screen Shot 2020-11-13 at 10 46 53 AM

When the context menu is opened on a particular tree item the focus needs to lock on the tree until the context menu is closed. The current focus state is just on hover over the tree item which leaves the context menu orphaned.

beyackle:
## Description

This adds some logic to the ProjectTree and TreeItem components to let them keep track of which, if any, tree item is selected. This lets us visually "lock" the highlight in place when we have a menu open; if any menu is open, that one will be shown as highlighted. If not, the highlight will be on whichever item is actively shown.

## Task Item

closes #4810 

## Screenshots
(note: my cursor is over on another item during this screenshot)

image


zhixzhan:
I have a try on this, seems on context menu open, current item turn out to be 'selected' style (blue) , not 'hover' style (gray) ,

image
image

Is it by design ?

beyackle:
It's possible I misunderstood the issue, and on reading it again, that would make more sense (to keep the gray highlight in place, not the blue one which shows which one is active). I'll change this to do that instead.

beyackle:
A screenshot of the new behavior: 
image

The gray highlight now stays in place while the menu is out, even when the cursor moves.

srinaath (approved):

PR #4837, opened 2020-11-16
fix: Added better error handling in pva publish extension's publish method

tonyanziano:
## Description

Previously, in the case that PVA would send back a response other than what was expect (202), we would discard any info in the response and simply display a 500 and say something went wrong.

With this change, we look for the 202, and if we do not receive a 202, we try to surface the response from PVA in the Composer UI.

#minor

### DIFF
--- a/extensions/pvaPublish/src/node/publish.ts
+++ b/extensions/pvaPublish/src/node/publish.ts
@@ -92,23 +92,37 @@ export const publish = async (
         'If-Match': project.eTag,
       },
     });
-    const job: PVAPublishJob = await res.json();
-    logger.log('Publish job started: %O', job);
+    if (res.status === 202) {
+      const job: PVAPublishJob = await res.json();
+      logger.log('Publish job started: %O', job);
 
-    // transform the PVA job to a publish response
-    const result = xformJobToResult(job);
+      // transform the PVA job to a publish response
+      const result = xformJobToResult(job);
 
-    // add to publish history
-    const botProjectId = project.id;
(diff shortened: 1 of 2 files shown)


extensions/pvaPublish/src/node/publish.ts
```diff
@@ -92,23 +92,37 @@ export const publish = async (
...
 
-    return {
-      status: result.status,
-      result,
-    };
+      return {
+        status: result.status,
+        result,
+      };
+    } else {
```
GeoffCoxMSFT, line 113: The if clause only checks for 202.  Would this else clause report an error for a 200 if there was text in the result?  Should this be an else if clause looking for error codes?

tonyanziano: Correct. We don't expect any other status codes back, so if we get a 200 back then something went wrong.

Here is the [actual route on PVA's side for this endpoint](https://msazure.visualstudio.com/CCI/_git/BotDesigner?path=%2Fsrc%2FBotDesigner.Management.Server%2FBotDesigner.Citizen.Server%2FBotDesigner.Management.Api%2FControllers%2FComposerContentManagementController.cs&version=GC708b173f4bab91833fa7e97cd1fca7876cb4e4c4&line=63&lineEnd=64&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents). You can see that it is registered to return a 202 Accepted.

GeoffCoxMSFT (approved):

cwhitten (approved):

PR #4864, opened 2020-11-18
feat: Add data collection dialog and settings

### LINKED ISSUE #4766: [Telemetry] Add data collection dialog and the ability to change selection in settings
## Description
Composer should prompt the user with a dialog to opt-in or opt-out of data collection when the user first opens the app. The dialog should not reappear upon subsequent sessions, and it should not conflict with any of the PVA logic or onboarding flows. The user should also have the ability to change their selection in the application settings.

tdurnford:
## Description
Adds a prompt when the user opens the app to opt-in or opt-out of data collection. It also adds the ability for them to change their data collection preferences in the Application Settings. 

## Task Item
Closes #4766

### Data Collection Dialog
image

### Data Collection Application Settings
image

### DIFF
--- a/Composer/packages/client/src/recoilModel/dispatchers/serverSettings.tsx
+++ b/Composer/packages/client/src/recoilModel/dispatchers/serverSettings.tsx
@@ -1,0 +1,51 @@
+/* eslint-disable react-hooks/rules-of-hooks */
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { CallbackInterface, useRecoilCallback } from 'recoil';
+import { produce } from 'immer';
+
+import httpClient from '../../utils/httpUtil';
+import { ServerSettingsState } from '../atoms/appState';
+
+import { logMessage } from './shared';
+
+export const serverSettingsDispatcher = () => {
+  const setAllowDataCollection = useRecoilCallback(
+    (callbackHelpers: CallbackInterface) => async (allowDataCollection: boolean) => {
+      const { set, snapshot } = callbackHelpers;
+      try {
+        const currentSettings = await snapshot.getPromise(ServerSettingsState);
+
+        const settings = produce(currentSettings, (current) => ({
+          ...current,
+          telemetry: {
+            ...current.telemetry,
+            allowDataCollection,
+          },
+        }));
+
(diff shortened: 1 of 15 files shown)


Composer/packages/client/src/recoilModel/dispatchers/serverSettings.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+      set(ServerSettingsState, settings);
+    } catch (error) {
+      logMessage(callbackHelpers, `Error fetching data collection settings: ${error}`);
+    }
+  });
+
+  return {
+    fetchServerSettings,
+    setAllowDataCollection,
```
hatpick, line 49: per server setting in the future a new dispatcher callback is required, have you considered having one callback to update settings like:
```
const updateServerSettings= useRecoilCallback(
    (callbackHelpers: CallbackInterface) => async (settings: Partial<ServerSettings>) => {
        // merge with current settings and update app and server state
    }
  )
```

Composer/packages/types/src/telemetry.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+export type TelemetrySettings = {
+  allowDataCollection?: boolean | null;
```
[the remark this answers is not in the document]
tdurnford: It's not necessary - I just need to change the default value to `false` in the app state.

Composer/packages/client/src/recoilModel/dispatchers/serverSettings.tsx
tdurnford: Sounds good to me

Composer/packages/types/src/telemetry.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+export type TelemetrySettings = {
+  allowDataCollection: boolean | null;
```
hatpick, line 5: I still think this should be:
{
   allowDataCollection ?: boolean;
}

hatpick (changes requested):

hatpick (approved):

PR #4868, opened 2020-11-18
fix: show bots without compile error

### LINKED ISSUE #4859: Settings page should list only local bots with no errors

liweitian:
## Description
only show bots without compile error
## Task Item
Closes #4859


srinaath:
@liweitian  **#4562 seems to have overlapping code with this. Would it make sense to just move these changes to #4562 itself and you can have 1 PR close the two tickets. In that way u dont have to wait for all the checks to pass for a small piece of code long**

liweitian:
all changes copied to #4562

PR #4869, opened 2020-11-18
fix: let uischema take priority over custom action logic

yeze322:
## Description

#minor
Previously, if an action is detected as 'custom Action' (not implemented in base SDK but implemented in ejected SDK), it will be always displayed as a dark grey card even uischema tries to override it.

image

This PR fixes this issue by adjusting the override order. Now UI Schema takes priority over custom Action logic, so that user can customize their non-builtin Actions.


yeze322:
@cwhitten I think the 'Custom Action' mechanism can be retired progressively since we already have the plan on versioning bot runtime. Custom Action simply compares project sdk.schema with server's sdk.schema and it indeed brings extra maintainance cost to us. UI Schema can entirely replace it.

a-b-r-o-w-n (approved):

PR #4889, opened 2020-11-19
feat: Updated statuses on bot runtime start

### LINKED ISSUE #4880: Change status message to queued when start all bots is clicked
When a botproject contains a root bot and multiple skills and "Start all bots" is clicked. The root bot is started after all skills are started. The status message needs to be changed to "Queued" instead of inactive on the root bot.

srinaath:
## Description

Screen Shot 2020-11-18 at 2 30 08 PM

Screen Shot 2020-11-18 at 2 29 22 PM

Fixes #4880


Composer/packages/client/src/utils/botRuntimeUtils.ts
```diff
@@ -12,14 +12,17 @@ export const getBotStatusText = (currentBotStatus) => {
...
     }
     case BotStatus.publishing:
       return BotStatusesCopy.publishing;
 
+    case BotStatus.queued:
+      return BotStatusesCopy.queued;
+
     default:
     case BotStatus.unConnected:
       return BotStatusesCopy.unConnected;
```
beyackle, line 28: Wouldn't `BotStatusesCopy[currentBotStatus] ?? BotStatusesCopy.unConnected` handle all of these cases?

beyackle (approved):
Looks good - I have one question about it, but I don't want to block getting this PR out on that. If there's an issue there, it can be addressed in a future PR.

Composer/packages/client/src/utils/botRuntimeUtils.ts
srinaath: I think so. I'll do the change.

beyackle (approved):

PR #4893, opened 2020-11-19
feat: Enable add a knowledge base to a skill in context menu.

zhixzhan:
## Description

Enable add a knowledge base to a skill in context menu.
the create buttons would only show in dialog header, cause our current logic is createQnA must in project's dialog

## Task Item
close #4892 4855

## Screenshots
Screen Shot 2020-11-19 at 11 35 05 AM


zhixzhan:
Now create QnA is work on DesignPage & QnAPage.
But QnAPage still have issue to display right file in multiple bots scenarios, it may related to @beyackle work. I just exclude it from this PR.

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
```diff
@@ -358,6 +366,20 @@ export const ProjectTree: React.FC<Props> = ({
           onDialogCreateTrigger(skillId, dialog.id);
         },
       },
+      {
+        label: formatMessage('Create KB from scratch'),
```
hatpick, line 370: are these labels passed by UX?

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
```diff
@@ -358,6 +366,20 @@ export const ProjectTree: React.FC<Props> = ({
...
+        onClick: () => {
+          createQnAFromScratchDialogBegin({ projectId: skillId });
+          onDialogCreateQnAKBFromUrl(skillId, dialog.id);
```
hatpick, line 373: I think you have these callbacks in reverse with line 380

Composer/packages/client/src/pages/knowledge-base/QnAPage.tsx
```diff
@@ -26,13 +26,13 @@ const QnAPage: React.FC<RouteComponentProps<{
   projectId: string;
   skillId: string;
 }>> = (props) => {
-  const { dialogId = '', projectId = '', skillId } = props;
-
+  const { dialogId = '', projectId = '', skillId = '' } = props;
+  const currentBotProjectId = skillId ?? projectId;
```
hatpick, line 30: when skillId is undefined, it gets the empty string '' default and then your currentBotProjectId becomes empty string as well, is that what you want?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

srinaath: I think skillID needs to be instantiated to null

beyackle: Careful here. The coalescing operator does _not_ treat `''` as a null value, so if there were no skillId given, the current it would end up with an empty string. That's why I had it without a default (because `undefined` would let the projectId pass through). @hatpick's comment about the projectId possibly also being empty is never going to happen, though; the QnAPage component gets values from the Router, and projectId is going to be read from the URL. The only reason it has a default value here is because TypeScript isn't aware of that fact.

srinaath: skillId is now left without instantiation

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
srinaath: Addressed it in the PR now

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
srinaath: UI has been updated with the latest copy now

srinaath:
@zhixzhan  I think this PR is ready. Except a unit test is failing. Once you can get the test passing we can merge the PR

beyackle (dismissed):
Changes look good.

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
```diff
@@ -359,6 +365,14 @@ export const ProjectTree: React.FC<Props> = ({
           onDialogCreateTrigger(skillId, dialog.id);
         },
       },
+      {
+        label: formatMessage('Add new knowledge base'),
```
srinaath, line 369: This should be only if the schema is a non PVA

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
```diff
@@ -359,6 +365,14 @@ export const ProjectTree: React.FC<Props> = ({
           onDialogCreateTrigger(skillId, dialog.id);
         },
       },
+      {
```
srinaath, line 368: This should be shown only if non PVA schema

srinaath (changes requested):

boydc2014 (approved):

PR #4898, opened 2020-11-19
fix: e2e test failed

lei9444:
## Description
image
The data collection dialog should be closed when run e2e test

## Task Item
#minor


Composer/cypress/support/commands.ts
```diff
@@ -4,6 +4,12 @@
 import '@testing-library/cypress/add-commands';
 
 Cypress.Commands.add('createBot', (botId: string, botName?: string) => {
+  cy.get('body').then((body) => {
+    if (body.find('[data-testid="data collection dialog"]').length > 0) {
+      cy.findByText('Not now').click();
+    }
+  });
```
a-b-r-o-w-n, line 11: nit:

```ts
if (cy.findByTestId('data collection dialog')) {
  // ...
}
```

Composer/packages/client/src/components/DataCollectionDialog.tsx
```diff
@@ -23,6 +23,7 @@ const DataCollectionDialog: React.FC = () => {
 
   return (
     <Dialog
+      data-testid="data collection dialog"
```
a-b-r-o-w-n, line 26: nit: camel case this.

a-b-r-o-w-n (changes requested):
This is a fine thing to do, but here are a couple of questions:

1. Can we just disable this altogether when not in production?
2. If not, is there an api we can interact with during set up that allows us to bypass interacting with the ui?
3. I don't think this is the right place to put this code since it probably only happens on a new user session. It doesn't have anything to do with creating a bot. It should go in a set up function.

@tdurnford do you have any thoughts about the first 2 questions?

hatpick:
I'm with @a-b-r-o-w-n on disabling it when not in production.

tdurnford:
@sorgh @a-b-r-o-w-n I think disabling it when it's not in production is probably the best approach.

a-b-r-o-w-n:
@tdurnford Something to consider is if we run the e2e tests with a prod build or not. I don't recall.

tdurnford:
@a-b-r-o-w-n We run it with a production build. I'm trying to set `allowDataCollection` to false in the `Composer/cypress/__e2e_data.json`, but I'm running into issues with that at the moment.

lei9444:
this will be fixed in #4907 .
close this

PR #4924, opened 2020-11-20
fix: get history always return empty array

VanyLaw:
## Description

fix getHistory always return empty array.
reason: the function defined in publishplugin interface is getHistory, not history.
image

## Task Item

#minor


boydc2014 (approved):

boydc2014:
@VanyLaw looks good to me, please cherry-pick this to 1.2.2 branch as well

boydc2014 (approved):

PR #4927, opened 2020-11-20
chore: cherry pick get history always return empty array

alanlong9278:
## Description

cherry pick get history always return empty array

## Task Item

#minor

### DIFF
--- a/Composer/packages/server/src/controllers/publisher.ts
+++ b/Composer/packages/server/src/controllers/publisher.ts
@@ -172,7 +172,7 @@ export const PublishController = {
 
     if (profile && extensionImplementsMethod(extensionName, 'history')) {
       // get the externally defined method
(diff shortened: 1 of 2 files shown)


boydc2014:
Looks good to me, but let's leave this until @tonyanziano get a chance to confirm this.

boydc2014 (dismissed):

Composer/packages/server/src/controllers/publisher.ts
```diff
@@ -172,7 +172,7 @@ export const PublishController = {
 
     if (profile && extensionImplementsMethod(extensionName, 'history')) {
```
tonyanziano, line 173: shouldn't this be `extensionImplementsMethod(extensionName, 'getHistory')` ?

tonyanziano: Fixed it and tested -- looks like it's working

PR #4930, opened 2020-11-20
fix: Toggle interaction on start bots

srinaath:
## Description
Change start bots controller to a toggle.

mov

@liweitian I had to make a change to setLocale as it gets an undefined locale and messes the entire LG view like below

MicrosoftTeams-image

#minor

PR #4931, opened 2020-11-21
fix: update E2E tests to use new menu names

beyackle:
## Description

We need to make sure our E2E tests get updated when interface names do.

## Task Item

#minor


beyackle:
Incorporated into #4930 .

PR #4934, opened 2020-11-22
fix: Dialog creation, trigger creation error resolved

### LINKED ISSUE #4821: Create dialog will show alert
## Describe the bug
alert

## Version
main

- [ ] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [x] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

Create a new locale. Set it as the default. Try adding a dialog now.

srinaath:
## Description
Dialog creation uses the currently selected dialog to seed the data for the new dialog. With the new bot project work we can have urls to the bot name and not redirected to the main dialog always. Essentially, this PR makes sure if no dialog is selected we get the seed data from root dialog.

Fixes #4821

### DIFF
--- a/Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
+++ b/Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
@@ -224,6 +224,10 @@ export const ProjectTree: React.FC<Props> = ({
 
   const jsonSchemaFilesByProjectId = useRecoilValue(jsonSchemaFilesByProjectIdSelector);
 
+  const createSubtree = useCallback(() => {
+    return projectCollection.map(createBotSubtree);
+  }, [projectCollection]);
+
   if (rootProjectId == null) {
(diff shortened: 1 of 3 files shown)


Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
```diff
@@ -224,6 +224,10 @@ export const ProjectTree: React.FC<Props> = ({
 
   const jsonSchemaFilesByProjectId = useRecoilValue(jsonSchemaFilesByProjectIdSelector);
 
+  const createSubtree = useCallback(() => {
```
srinaath, line 227: @cwhitten we are now memoizing the project tree creation. This has removed the delay we have when we click on a tree item

cwhitten (approved):

PR #4968, opened 2020-11-24
feat: telemetry api

### LINKED ISSUE #4765: [Telemetry] Add Application Insights telemetry routes, controllers, and service to the server
## Description
We need to add two routes to the server's api: `/telemetry/trackEvent` and `/telemetry/pageView`. The payloads sent to these routes will be forwarded the to Application Insights through the telemetry controllers and service.

tdurnford:
## Description
Adds support for telemetry api in the client and the server. Also adds the hashed mac address to the electron context.

## Task Item
Closes #4765


Composer/packages/electron-server/src/utility/machineId.ts
```diff
@@ -0,0 +1,21 @@
...
+// Licensed under the MIT License.
+
+import crypto from 'crypto';
+
+import { v4 as uuid } from 'uuid';
+
+import { getMacAddress } from './getMacAddress';
+
+export async function getMachineId(): Promise<string> {
+  return (await getMacMachineId()) || uuid.generateUuid(); // fallback, generate a UUID
```
hatpick, line 11: I'll review more thoroughly soon but here's a quick comment:
I think the very first time this id should be generated and persisted and later on just use the persisted id.
(Check persisted file for ```telemetryId```, if it doesn't exist, generate one: it's gonna be either the mac address or a uuid. Later on, use the persisted ```telemetryId```)

a-b-r-o-w-n (changes requested):
I think I'd like to see the server API support event batching instead of processing singular events since the payloads are small.

The client should then implement a pooling mechanism that processes the events in batches of 10 or 20. This will allow the client-side telemetry to scale better.

Some things to consider with this approach:
- Set an interval to process all events in queue (like you have now, but call `flush` every n seconds).
- Be sure to drain the queue when navigating away from the page.

Also, I might have missed this, but is `logPageView` being called anywhere?

tdurnford:
@a-b-r-o-w-n No, `logPageView` is not used in this PR, but it will be used a follow up PR that wires up all the events.

Composer/packages/server/src/services/telemetry.ts
```diff
@@ -0,0 +1,80 @@
...
+        switch (type) {
+          case TelemetryEventTypes.TrackEvent:
+            client?.trackEvent({ name, properties });
+            break;
+          case TelemetryEventTypes.PageView:
+            client?.trackPageView({ name, url, properties });
+            break;
+        }
+      } catch (error) {
+        // swallow the exception on a failed attempt to collect usage data
```
a-b-r-o-w-n, line 63: consider adding in retry logic here and in the client.

Composer/packages/client/src/telemetry/hooks.ts
```diff
@@ -0,0 +1,44 @@
...
+      projectId,
+    }));
+
+    forceRender({});
+  }, [telemetry, page, projectId]);
+
+  const handleBeforeUnload = useCallback(() => {
+    const eventLogger = getEventLogger();
+    eventLogger.flush();
+    window.removeEventListener('beforeunload', handleBeforeUnload);
```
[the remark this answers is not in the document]
tdurnford: I noticed `handleBeforeUnload` was called multiple times if the listener wasn't removed in the function call.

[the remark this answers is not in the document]
a-b-r-o-w-n: That might be a dev-only thing. 

Try removing the listener in the useEffect below instead.

```ts
useEffect(() => {
  window.addEventListener('beforeunload', handleBeforeUnload);
  () => {
    window.removeEventListener('beforeunload', handleBeforeUnload);
  }
}, []);
``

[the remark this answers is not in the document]
Comment (tdurnford) [reply Composer/packages/client/src/telemetry/hooks.ts:34]: That's what I had before when I noticed it being called multiple times, but if it's just a dev behavior, I'll change it back.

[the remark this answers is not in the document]
Comment (a-b-r-o-w-n) [reply Composer/packages/client/src/telemetry/hooks.ts:34]: I'm not saying that it is that for sure, but could be due to hot reloading in the client. It doesn't hurt to have this though, so I would leave.

File: Composer/packages/client/src/telemetry/useInitializeLogger.ts [hunk 4]
```diff
@@ -0,0 +1,36 @@
...
+export const useInitializeLogger = () => {
+  const projectId = useRecoilValue(currentProjectIdState);
+  const page = useRecoilValue(currentModeState);
+  const { telemetry } = useRecoilValue(userSettingsState);
+
+  useEffect(() => {
+    TelemetryClient.setup(telemetry, () => ({
+      timestamp: new Date().toUTCString(),
+      composerVersion: process.env.COMPOSER_VERSION || 'unknown',
+      sdkPackageVersion: process.env.SDK_PACKAGE_VERSION || 'unknown',
```
a-b-r-o-w-n, line 20: I think these 3 properties could be added in the client's methods since they are "static".

Composer/packages/client/src/telemetry/useInitializeLogger.ts
```diff
@@ -0,0 +1,36 @@
...
+
+import TelemetryClient from './TelemetryClient';
+
+export const useInitializeLogger = () => {
+  const projectId = useRecoilValue(currentProjectIdState);
+  const page = useRecoilValue(currentModeState);
+  const { telemetry } = useRecoilValue(userSettingsState);
+
+  useEffect(() => {
+    TelemetryClient.setup(telemetry, () => ({
```
a-b-r-o-w-n, line 17: This can probably just move outside of the useEffect since it is not costly and will ensure the client is setup before the next render phase.

Composer/packages/client/src/telemetry/TelemetryClient.ts
```diff
@@ -0,0 +1,53 @@
...
+  private static _getProperties?: () => LogData;
+  private static _telemetrySettings?: TelemetrySettings;
+
+  public static setup(telemetrySettings: TelemetrySettings, additionalProperties: LogData | (() => LogData)) {
+    if (this._telemetrySettings?.allowDataCollection !== telemetrySettings.allowDataCollection) {
+      this.client?.drain();
+    }
+
+    this._getProperties =
+      typeof additionalProperties === 'function' ? additionalProperties : () => additionalProperties;
```
a-b-r-o-w-n, line 19: What I would do instead is store `additionalProperties` as a private property and define a getter like this:

```ts
private static additionalProperties?: () => LogData;

private static get sharedProperties() {
  return {
    ...this.additionalProperties?.(),
    timestamp: Date.now(),
    composerVersion: process.env.COMPOSER_VERSION || 'unknown',
    sdkPackageVersion: process.env.SDK_PACKAGE_VERSION || 'unknown',
  }
}
```

Composer/packages/client/src/telemetry/TelemetryClient.ts
```diff
@@ -0,0 +1,53 @@
...
+    this._telemetrySettings = telemetrySettings;
+  }
+
+  public static log<TN extends TelemetryEventName>(
+    eventName: TN,
+    ...args: TelemetryEvents[TN] extends undefined ? [never?] : [TelemetryEvents[TN]]
```
a-b-r-o-w-n, line 25: This typing is a little confusing to me because of `args` usage below. It looks like there is only 1 argument expected ever which is an object. Is that correct?

While I think this is some neat Typescript, what is it buying us here?

Composer/packages/client/src/telemetry/ConsoleClient.ts
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { LogData } from '../../../types/src';
```
a-b-r-o-w-n, line 4:
```suggestion
import { LogData } from '@botframework-composer/types';
```

Composer/packages/client/src/telemetry/AppInsightsClient.ts
```diff
@@ -0,0 +1,56 @@
...
+    this.startInterval();
+    this._eventPool.push({ type: TelemetryEventTypes.TrackEvent, name, properties });
+  }
+
+  public static logPageView(name: string, url: string, properties: LogData) {
+    this.startInterval();
+    this._eventPool.push({ type: TelemetryEventTypes.PageView, name, properties, url });
+  }
+
+  public static drain() {
```
a-b-r-o-w-n, line 24: This should return the `Promise.all` so that it can be awaited elsewhere.

Composer/packages/client/src/telemetry/AppInsightsClient.ts
```diff
@@ -0,0 +1,56 @@
...
+      const events = this._eventPool.splice(0, BATCH_SIZE);
+      await this.postEvents(events);
+    }
+  }
+
+  private static async postEvents(events: TelemetryEvent[]) {
+    try {
+      await httpClient.post('/telemetry/events', { events });
+    } catch (error) {
+      // Swallow error to avoid crashing the app while sending telemetry
```
a-b-r-o-w-n, line 45: We should add some retry logic here.

a-b-r-o-w-n (changes requested):

Composer/packages/client/src/telemetry/TelemetryClient.ts
tdurnford: This allows us to strongly type the `eventName` and the required properties for that event, so we don't inadvertently track an event without the required data.

a-b-r-o-w-n: Is the spread arg required or can it be something like

```ts
log<TN extends TelemetryEventName>(
    eventName: TN,
    properties?: TelemetryEvents[TN] extends undefined ? never : TelemetryEvents[TN]
)
```

a-b-r-o-w-n (dismissed):

Composer/packages/client/src/telemetry/TelemetryClient.ts
tdurnford: Unfortunately, it doesn't work without the spread operator.

tdurnford: Actually.. that does work..

a-b-r-o-w-n (approved):

PR #4978, opened 2020-11-25
fix: Memory leak & UI adjust for publish page

alanlong9278:
## Description

- fix memory leak 
- delete backgroud color for publish dialog

## Task Item

#minor 

## Screenshots

image


yeze322 (dismissed):
It's better to paste a screenshot of the error message

PR #4984, opened 2020-11-25
fix: publish page url

### LINKED ISSUE #4982: root id and skill id are switched in url
## Describe the bug
Create two todoluisbots, click skill bot on design page. (gif shows it clicks the warning, but the result is the same. Because it changes currentProjectIdState) Go to publish page. The two ids are switched. Then if we click the manage profiles button, it may overwrite publish Targets.

## Browser

- [ ] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

## To Reproduce
bug

liweitian:
## Description
publish page link on left navigation list is not correct. 

## Task Item
Closes #4982 

## Screenshots


liweitian:
weird commits history. open another PR

PR #4991, opened 2020-11-26
fix: a11y in all-up view

### LINKED ISSUE #4825: editing issue (accessibility) in LU/LG table view
## Describe the bug
In LG table view, I can not hit "enter" if the focus is in the response for the LG template
image

In LU table view, when I hit the  "up arrow" inside the table view for one luis intent, it will go to other component instead of the upper line in Manaco editor.  
image

## Version

Composer 1.2

## Browser

- [X] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [X] Windows
- [ ] Ubuntu

zhixzhan:
#### 1. Refine Enter submit in list-view
  - For single line input, press Enter do submit
  - For multiple line input, press Enter do submit, Shift+Enter get a new line,

## Task Item
close #4825 

## Screenshots
image


a-b-r-o-w-n (approved):

beyackle (approved):
Good use of `?:` to make the logic more apparent.

PR #5010, opened 2020-11-30
fix: only support en-us on qna for tmp use

### LINKED ISSUE #5004: [MCS-DE-GO] Dialog creation error in qna-filename when language is not "en-us"
When selecting a Bot with a default language (set as only language) which is not "en-us", the .qna-Files are still created with the "en-us"-extension which causes Composer to produce an error that it could not find the .qna-File.

@joaollq @goergenj @sgellock @benbrown @Bindsi @cwhitten

liweitian:
## Description
QnA only support en-us at this moment. It causes some bugs when we have multi-locale.

## Task Item
Closes #5004


liweitian:
@luhan2017 Could you help checking this PR?

luhan2017 (approved):

Composer/packages/client/src/recoilModel/dispatchers/qna.ts
```diff
@@ -50,7 +50,9 @@ export const createQnAFileState = async (
 ) => {
   const { set, snapshot } = callbackHelpers;
   const qnaFiles = await snapshot.getPromise(qnaFilesState(projectId));
-  const locale = await snapshot.getPromise(localeState(projectId));
+  //const locale = await snapshot.getPromise(localeState(projectId));
+  //To do: support other languages
+  const locale = 'en-us';
```
zhixzhan, line 55: For other methods in this dispatcher should also use fixed locale en-us.
`removeQnAFileState`
`renameKBFileState`
`updateQnAFileState`

zhixzhan (changes requested):

Composer/packages/client/src/recoilModel/dispatchers/qna.ts
liweitian: solved. And it seems createQnAKBFromScratch and createQnAKBFromUrl should use fixed locale en-us

PR #5011, opened 2020-11-30
feat: azure publish with orchestrator

### LINKED ISSUE #4980: Azure publish plugin should not copy code to re-build bot
### Context
Today, azure publish is totally decoupled from local publish (local start), which means azure publish will assume the bot is not built yet. It will ignore the bot build result of local run (ie, the generated folder) and build the bot from scatch by doing the cross-train, down-sampling, luis-build, qna-build, all that. 

### Issue
The assumption may make sense, but the issue is today the build process is done by **copying code**, which is anti-pattern and definitely not encouraged. Every time there is a update in the build process (when qna came in, when orchestrator came in), we have to duplicated the code again in azure publish. This is not a consistent approach and very hard to maintain. 

### Solution
The proposed solution of this, is to decouple the local start bot to 2 modules 
* build
* run

Then share the build module to azure publish plugin by making it a shared lib or make it an extension, to avoid copying code and  all the consequences brought by it.

lei9444:
## Description
In azure publish we use the old build flow and copy some code from local build and generate the crosstrain config by itself. It is difficult to upgrade this part code of build and add new feature for azure publish.

In this PR:

- remove the duplicate code for build and use the builder instance from current project to build the bot.

- add orchestrator build for azure publish

> 1. update the build result in orchestrator.setting.json.(use relative path)

> 2. copy the orchestrator build model to bot folder

## Task Item
closes #4980

## Screenshots
image
image
image


boydc2014:
@taicchoumsft @luhan2017 can you help verify this?

boydc2014:
@VanyLaw can you also help take a look at this one, there is some refactor along with the new feature, we basically removed a lot of duplicated code, can you confirm if it make sense?

Composer/packages/server/src/models/bot/builder.ts
```diff
@@ -26,6 +27,7 @@ const SETTINGS = 'settings';
 const INTERRUPTION = 'interruption';
 const SAMPLE_SIZE_CONFIGURATION = 2;
 const CrossTrainConfigName = 'cross-train.config.json';
+const MODLE = 'model';
```
taicchoumsft, line 30: >MODLE [](start = 6, length = 5)

minor typo

extensions/azurePublish/src/luisAndQnA.ts
```diff
@@ -6,190 +6,76 @@ import { promisify } from 'util';
...
+  qnaResources: string[];
+}
 
-  private notEmptyModel(file: string) {
-    return fs.readFileSync(file).length > 0;
-  }
+type BuildSettingType = {
+  luis: ILuisConfig,
+  qna: QnaConfigType
+} & Resources
```
taicchoumsft, line 30: >& Resources [](start = 2, length = 11)

minor: for perf reasons listed [here:](https://github.com/microsoft/TypeScript/wiki/Performance#preferring-interfaces-over-intersections), is it easy to switch to interfaces?

taicchoumsft (dismissed):
:shipit:

luhan2017:
The qna build error is not properly handled:
image
image
The publish is success, but I can't query the remote bot

lei9444:
> The qna build error is not properly handled:
> image
> image
> The publish is success, but I can't query the remote bot

fixed it
image

extensions/azurePublish/src/deploy.ts
```diff
@@ -58,18 +58,28 @@ export class BotProjectDeploy {
...
+
+      this.logger({
+        status: BotProjectDeployLoggerType.DEPLOY_INFO,
+        message: 'Build Success!',
+      });
 
       // this function returns an object that contains the luis APP ids mapping
       // each dialog to its matching app.
-      const { luisAppIds, qnaConfig } = await publisher.publishLuisAndQna(
+      const { luisAppIds, qnaConfig } = await publishLuisToPrediction(
```
VanyLaw, line 75: publishLuisToPrediction method only return luisAppIds, why we add the qnaConfig parameter here? Doesn't it affect the qna build or publish?

lei9444: Thanks @VanyLaw

luhan2017 (approved):

boydc2014 (approved):

PR #5012, opened 2020-11-30
fix: some bug fix about publish page

alanlong9278:
### bugfix
- can not get all bot publish histories. It just get the first bot publish history when initial the page
-  if the previous publish failed or success, start a new publish process, the notification modal will pop up the old ones.  The notification card display wrong when publish bot again. It will show the old notification card when publishing again

## Task Item

#minor

## Screenshots

notification
image


Composer/packages/client/src/pages/publish/Publish.tsx
```diff
@@ -210,24 +219,28 @@ const Publish: React.FC<RouteComponentProps<{ projectId: string; targetName?: st
     }
   }, [projectId]);
 
+  const usePrevious = (value) => {
+    const ref = useRef();
```
a-b-r-o-w-n, line 223: I think this violates react rules on using hooks. They need to be used deterministically in the render of the component.

a-b-r-o-w-n (changes requested):

Composer/packages/client/src/pages/publish/Publish.tsx
alanlong9278: Move usePrevious out of component

Composer/packages/client/src/pages/publish/Publish.tsx
```diff
@@ -210,24 +226,20 @@ const Publish: React.FC<RouteComponentProps<{ projectId: string; targetName?: st
     }
   }, [projectId]);
 
+  const prevBotStatusList: IBotStatus[] | undefined = usePrevious(botStatusList);
   useEffect(() => {
-    // init publishHistoryList
-    if (Object.keys(publishHistoryList).length > 0) {
+    // init bot status list for the botProjectData is empty array when first mounted
+    if (!isEqual(prevBotStatusList, statusList)) {
```
a-b-r-o-w-n, line 232: I am having a hard time understanding why this is necessary.

This is only used to compare the status list and then update if changed. Can that not be done using the current state value?

```ts
if (!isEqual(currentStatusList, statusList) {
  setCurrentStatusList(statusList)
}
```

The `usePrevious` hook updates the ref in an effect, which is async anyways so this value could be stale.

yeze322 (changes requested):
Several advices to make sure readability:
- Avoid nested if else clauses, return early.
- Extract complicated state computation logic as pure functions.
- Define necessary local variables to replace appearances of `botPublishHistory[0].status`. Avoid context switching when reading

Composer/packages/client/src/pages/publish/Publish.tsx
alanlong9278: Remove `useRef` hook.

boydc2014:
I agree we can merge it if this is functional working. But a few more comments for refactoring. See below

Composer/packages/client/src/pages/publish/Publish.tsx
```diff
@@ -159,49 +159,85 @@ const Publish: React.FC<RouteComponentProps<{ projectId: string; targetName?: st
   };
 
   const [pendingNotification, setPendingNotification] = useState<Notification>();
+  const [previousBotPublishHistoryList, setPreviousBotPublishHistoryList] = useState(botPublishHistoryList);
```
boydc2014, line 162: can we have a better naming for this what previousHistory means? do we have a current history?

and there is no good to add "list" as postfix, botPublishHistory is good enough

Composer/packages/client/src/pages/publish/Publish.tsx
```diff
@@ -159,49 +159,85 @@ const Publish: React.FC<RouteComponentProps<{ projectId: string; targetName?: st
...
-                if (item.id === botProjectId) {
-                  item.status = botPublishHistory[0].status;
-                  item.comment = botPublishHistory[0].comment;
-                  item.message = botPublishHistory[0].message;
-                  item.time = botPublishHistory[0].time;
-                }
-                return item;
-              })
-            );
+      if (!(bot.publishTarget && bot.publishTargets)) {
```
boydc2014, line 167: First, the comments "// most recent item is a 202, which means we should poll for updates..." is not related to this line.  there is even no 202 in many lines. 

For such a long use effect and the many other long use effects below, we should give a top-level summary.

boydc2014 (approved):

Issue #483, opened 2019-07-11
can't set the app id and app password in config settings to null once they are set with any value


zidaneymar:
![image](https://user-images.githubusercontent.com/9328933/61044428-35758300-a40b-11e9-9e16-fd9bb6861d9a.png)

known issue of botbuilder sdk

luhan2017:
This is not fix in that PR, this needs the BotManager and BotRuntime decouple to fix.

PR #5029, opened 2020-12-01
feat: npm based 'new bot' flow behind feature flag

### LINKED ISSUE #5025: creation: introduce npm-feed backed template selection
This will begin the work on the new "enhanced bot creation experience" that will bring a new set of templates and a new process for scaffolding projects during creation.

Preliminary work:
- [x] Create a clone of the current create flow component
- [x] Introduce a new feature flag to gate access to the new version of the create flow
- [x] If the new create flag is enabled, when a user clicks the "new" button, they should see the NEW create component
- [x] In the new version of the create component, clone the current API calls (client and server) using a /v2/ in the API URL. For now, these can just be exact copies.

Once we have the new component gated behind a feature flag, we can start to update the functionality:

- [x] Create an updated version of fetchTemplates that takes a feed URL as a parameter.  This new endpoint should accept NPM feeds as well as NUGET feeds, see below for example.  It should then fetch the feed content, normalize the fields into a format that matches our needs, and return it to the client. For now, we will have 2 feeds - 1 for JS templates, 1 for C# templates.    It is ok to just default this to a static URL for now...
- [x] Ensure that the new templates sourced from the feed appear in the template picker.
- [x] Ensure that a user can select the template and move to the next screen of the flow
- [x] Ensure that when the user submits the second page, the *NEW* version of the create bot endpoint is called and all the appropriate information (template name, project name, folder) are passed correctly.

Sample feed URL:
https://registry.npmjs.org/-/v1/search?text=bot&keywords:botframework-composer&size=100&from=0&quality=0.65&popularity=0.98&maintenance=0.5

Spec items:
1.0, 1.2, 1.4

### LINKED ISSUE #5026: creation: introduce ability to create from npm-hosted package
Building on #5025, we will introduce the ability to create a new project based on a template sourced from NPM:

- [x] Update the createprojectv2 endpoint to fetch the package contents from npm 
- [x] We should cache this content somewhere. Like in the composer data folder I think. Only download if not present in the cache.
- [x] Create a new project using the package content as the source. This should be achieved using the `yeoman-environment` package.

Spec items:
1.0, 1.1, 1.2, 1.3, 1.4

### LINKED ISSUE #5488: build/publish from Composer works with new runtime
End-to-end build/publish works inside Composer using the new runtime and packages.

### LINKED ISSUE #5489: Add telemetry for new create flow

### LINKED ISSUE #5498: Ensure PVA scenarios work with new creation flow in Composer
- [x] Ensure PVA scenarios work accordingly with the new creation endpoint

### LINKED ISSUE #5502: Handle versioning of templates
Ensure that versions passed by template feed are the versions of the selected template that is instantiated for the user

### LINKED ISSUE #5566: Create UT for new creation flow
Create unit & integration tests for the new component model based creation flow.

- [x] Template feed population unit test
- [x] Yeomen generator installation unit test
- [x] Yeomen generator instantiation unit test
- [ ] Read me gathering unit test

pavolum:
## Description

This PR introduces the new creation experience work behind a hidden feature flag. It was implemented in such a way that the existing production creation experience should not be impacted whatsoever. 

This PR merges the new creation experience so far, there are various edge cases it does not yet account for, hence why it is behind a hidden feature flag. Below is a high level list of what was done

- Front and back end creation flow cloned and the copy was altered to support the new creation requirements
    - The cloned creation flow follows a 'v2' pattern where server and front end routing calls are prefaces with some 'v2' in the path
- Implementation of getReadMe server endpoint which calls npm with a given package name and returns the readMe html for that package
- Implementation of getProjTemplatesV2 server endpoint which is used to grab the npm template generators that are available in a given feed(s)
- Implementation of hidden 'New Creation Experience' feature flag which routes to the new 'new bot' flow if checked
- Added scaffolding of new runtime template for the new C# adaptive runtime, only slightly edited to ensure friendly local publish flow and build works.
- CreateProjectAsync refactored to handle npm driven template installation, building and package resolution
    - As a result the 'Conversational Core' feature will no longer work, this feature flag has been reverted to be hidden
        - Done this way to not have to support three different creation flows at once

## Task Item

fixes #5025
fixes #5026 
fixes #5488 
fixes #5502  
fixes #5498 
fixes #5489 
fixes #5566 

## Known limitations
- PVA scenarios in the new creation flow accounted for as part of this PR #5498 
- The templates instantiated in the new creation experience use the new adaptive runtime. New runtime integration is early and with this PR edge cases have not been validated, particularly around publishing. #5488  
- Detection of old runtime and bot file structure and routing to older runtime template not implemented. #5500
- Remote package acquisition does not follow agreed upon naming conventions which are TBD (component repo issue #483)
- ConversationalCore feature flag will no longer work


pavolum:
@benbrown pushing out draft PR for tracking. Didnt get as much time to tackle this today as id have liked but have set up the new flow behind a feature flag with cloned versions of the existing flow. Will start configuring the npm driven template feed tomorrow.

Composer/packages/server/src/models/asset/assetManager.ts
```diff
@@ -92,58 +100,79 @@ export class AssetManager {
...
+    locale?: string
+  ): Promise<LocationRef> {
+    // user storage maybe diff from template storage
+    const dstStorage = StorageService.getStorageClient(ref.storageId, user);
+    const dstDir = Path.resolve(ref.path);
+    if (await dstStorage.exists(dstDir)) {
+      log('Failed copying template to %s', dstDir);
+      throw new Error('already have this folder, please give another name');
+    }
+    await mkdirs(dstDir, (err) => {
```
joshgummersall, line 117: Generally, async functions work in either Promise mode or callback mode, not both. [The docs](https://github.com/jprichardson/node-fs-extra/blob/HEAD/docs/ensureDir.md) indicate that if you pass a callback, a promise is not returned. Testing with the CLI seems to confirm this:

```bash
> let fsx = require('fs-extra')
undefined
> fsx.mkdirs
[Function: mkdirs]
> fsx.mkdirs('foo')
Promise { <pending> }
> fsx.mkdirs('foo', () => console.log('done'))
undefined
```

pavolum: Ah good catch I changed it to use mkdirSync and added a try catch around it as seems to be the standard in other parts of the code base.

joshgummersall: `await mkdirs()`, without the callback, wrapped in a `try`-`catch` would also work!

Composer/packages/client/src/components/CreationFlow/v2/CreateOptions.tsx
```diff
@@ -0,0 +1,252 @@
...
+      const emptyBotTemplate = find(templates, ['id', EmptyBotTemplateId]);
+      if (emptyBotTemplate) {
+        setCurrentTemplate(emptyBotTemplate.id);
+        setEmptyBotKey(emptyBotTemplate.id);
+      }
+    }
+  }, [templates]);
+
+  useEffect(() => {
+    if (selectedFeed?.props?.headerText?.toLowerCase() === 'typescript') {
```
benbrown, line 189: There must be a better way to access the value of the selected feed.

Also, we should pull these feeds out into a dictionary rather than define them inline here.  

`props.fetchTemplates(feeds[selectedFeed])`

benbrown: maybe at least use the `itemKey` field instead of the actual label text.

Eventually these feeds will be user modifiable, so they will have a key, title, and url value.

Composer/packages/client/src/components/CreationFlow/v2/CreateOptions.tsx
```diff
@@ -0,0 +1,252 @@
...
+  return (
+    <Fragment>
+      <DialogWrapper
+        isOpen
+        {...DialogCreationCopy.CREATE_NEW_BOT_V2}
+        dialogType={DialogTypes.CreateFlow}
+        onDismiss={onDismiss}
+      >
+        <Pivot onLinkClick={setSelectedFeed}>
+          <PivotItem headerText="C#"></PivotItem>
```
benbrown, line 213: add `itemKey` to these pivotitems

Composer/packages/client/src/components/CreationFlow/v2/CreateOptions.tsx
```diff
@@ -0,0 +1,252 @@
...
+    });
+  }, []);
+
+  const handleJumpToNext = () => {
+    let routeToTemplate = emptyBotKey;
+    if (option === optionKeys.createFromTemplate) {
+      routeToTemplate = currentTemplate;
+    }
+
+    if (option === optionKeys.createFromQnA) {
```
benbrown, line 138: We need to figure out how this feature is surfaced.  This is what allows a user to start in the QNA maker mode.

benbrown: This should appear in the template list even though it is not in the feed.

Composer/packages/client/src/components/CreationFlow/v2/CreationFlow.tsx
```diff
@@ -0,0 +1,192 @@
...
+  const projectId = useRecoilValue(currentProjectIdState);
+  const storages = useRecoilValue(storagesState);
+  const templateReadMe = useRecoilValue(templateReadMeHtml);
+  const focusedStorageFolder = useRecoilValue(focusedStorageFolderState);
+  const { appLocale } = useRecoilValue(userSettingsState);
+  const cachedProjectId = useProjectIdCache();
+  const currentStorageIndex = useRef(0);
+  const storage = storages[currentStorageIndex.current];
+  const currentStorageId = storage ? storage.id : 'default';
+  // const [formData, setFormData] = useState({ name: '', description: '', location: '' });
```
benbrown, line 61: remove?

Composer/packages/client/src/components/CreationFlow/v2/DefineConversation.tsx
```diff
@@ -0,0 +1,329 @@
...
+            <StackItem grow={0} styles={halfstack}>
+              <TextField
+                multiline
+                label={formatMessage('Description')}
+                resizable={false}
+                styles={description}
+                value={formData.description}
+                onChange={(_e, val) => updateField('description', val)}
+              />
+            </StackItem>
```
benbrown, line 311: There are some fields missing from this --

we need to also collect the webapp/functions runtime choice here, as well as the primary language.  I believe there is a design for the language field in the figma files

Composer/packages/client/src/pages/home/Home.tsx
```diff
@@ -163,9 +173,7 @@ const Home: React.FC<RouteComponentProps> = () => {
                 styles={home.newBotItem}
                 title={addButton}
                 onClick={() => {
-                  TelemetryClient.track('CreateNewBotProject', { method: 'newCallToAction' });
-                  setCreationFlowStatus(CreationFlowStatus.NEW);
-                  navigate('projects/create');
+                  onClickNewBot();
```
benbrown, line 176: this change will cause the telemetry to incorrectly register clicks on the call to action as clicks on the toolbar.
Maybe pass in a parameter? `onClickNewBot(method)`

Composer/packages/client/src/recoilModel/atoms/appState.ts
```diff
@@ -271,3 +271,8 @@ export const displaySkillManifestState = atom<undefined | string>({
   key: getFullyQualifiedKey('displaySkillManifest'),
   default: undefined,
 });
+
+export const templateReadMeHtml = atom<string>({
```
benbrown, line 275: why is this in the app state? seems like this should stay in the component.

Composer/packages/client/src/recoilModel/dispatchers/storage.ts
```diff
@@ -155,6 +156,43 @@ export const storageDispatcher = () => {
     }
   });
 
+  const fetchTemplatesV2 = useRecoilCallback(({ set }: CallbackInterface) => async (feedUrls?: string[]) => {
```
benbrown, line 159: prefer all of this to be in the component itself - it doesn't need to be in app-wide state, right? It is only used in the one part of the create new wizard flow.

Composer/packages/server/src/controllers/project.ts
```diff
@@ -518,51 +518,50 @@ function createProjectV2(req: Request, res: Response) {
...
-    schemaUrl,
-    locale,
-    preserveRoot,
-    templateDir,
-    eTag,
-    alias,
-  } = req.body;
+  const { templateId } = req.body;
+
+  // todo: add back templateDir, eTag, alias from req extraction for PVA scenarios
```
benbrown, line 525: lets do this before merging

Composer/packages/server/src/controllers/project.ts
```diff
@@ -518,51 +518,50 @@ function createProjectV2(req: Request, res: Response) {
...
+    await AssetService.manager.copyBoilerplate(currentProject.dataDir, currentProject.fileStorage);
 
     if (currentProject !== undefined) {
       await ejectAndMerge(currentProject, jobId);
       BackgroundProcessManager.updateProcess(jobId, 202, formatMessage('Initializing bot project'));
       await currentProject.updateBotInfo(name, description, preserveRoot);
-      if (schemaUrl && !createFromRemoteTemplate) {
-        await currentProject.saveSchemaToProject(schemaUrl, locationRef.path);
-      }
+
```
benbrown, line 560: is this part of the PVA related stuff?

Composer/packages/server/src/controllers/project.ts
```diff
@@ -518,51 +518,50 @@ function createProjectV2(req: Request, res: Response) {
...
     BackgroundProcessManager.updateProcess(jobId, 202, formatMessage('Getting template'));
-    const newProjRef = await getNewProjRef(templateDir, templateId, locationRef, user, locale);
+
+    const newProjRef = await AssetService.manager.copyRemoteProjectTemplateToV2(templateId, name, locationRef, user);
+    BackgroundProcessManager.updateProcess(jobId, 202, formatMessage('Bot files created'));
 
     const id = await BotProjectService.openProject(newProjRef, user);
     // in the case of a remote template, we need to update the eTag and alias used by the import mechanism
-    createFromRemoteTemplate && BotProjectService.setProjectLocationData(id, { alias, eTag });
+    // BotProjectService.setProjectLocationData(id, { alias, eTag });
```
benbrown, line 550: restore pva related stuff?

Composer/packages/server/src/models/asset/assetManager.ts
```diff
@@ -1,15 +1,18 @@
...
+/* eslint-disable @typescript-eslint/no-var-requires */
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
 import fs from 'fs';
-import path, { join } from 'path';
-import { exec } from 'child_process';
-import { promisify } from 'util';
+import path from 'path';
+// import { exec } from 'child_process';
```
benbrown, line 8: remove

Composer/packages/server/src/models/asset/assetManager.ts
```diff
@@ -1,15 +1,18 @@
...
+// import { promisify } from 'util';
```
benbrown, line 9: remove

Composer/packages/server/src/models/asset/assetManager.ts
```diff
@@ -1,15 +1,18 @@
...
 
 import find from 'lodash/find';
-import { UserIdentity, BotTemplate, FileExtensions } from '@bfc/extension';
-import { mkdirs, readFile } from 'fs-extra';
-import rimraf from 'rimraf';
+import { UserIdentity, FileExtensions } from '@bfc/extension';
+import { mkdirSync, readFile } from 'fs-extra';
+// import rimraf from 'rimraf';
```
benbrown, line 14: remove

Composer/packages/server/src/models/asset/assetManager.ts
```diff
@@ -21,8 +24,9 @@ import StorageService from '../../services/storage';
 import { IFileStorage } from '../storage/interface';
 import { BotProject } from '../bot/botProject';
 
-const execAsync = promisify(exec);
-const removeDirAndFiles = promisify(rimraf);
+// TODO: pass in working directory param to createEnv for desired location of local Yeomen Repo
```
benbrown, line 27: do we still need to do this?

Composer/packages/server/src/models/asset/assetManager.ts
```diff
@@ -92,58 +96,78 @@ export class AssetManager {
...
+  }
+
+  private async installRemoteTemplate(generatorName: string, npmPackageName: string, dstDir: string): Promise<boolean> {
+    const registeredGenerators: string[] = await yeomanEnv.getGeneratorNames();
+
+    if (registeredGenerators.indexOf(generatorName) !== -1) {
+      return true;
     } else {
-      throw new Error('selected template has no local or external address');
+      await yeomanEnv.installLocalGenerators({ [npmPackageName]: '*' });
```
benbrown, line 141: I don't see how this is ensuring that the locally installed generator is going into our composer data folder.

Perhaps setting the yeoman.cwd will do it?

Composer/packages/server/src/models/bot/botProject.ts
```diff
@@ -205,12 +205,22 @@ export class BotProject implements IBotProject {
...
     if (settings?.runtime?.customRuntime && settings.runtime.path && !Path.isAbsolute(settings.runtime.path)) {
       const absolutePath = Path.resolve(this.dir, 'settings', settings.runtime.path);
+
       if (fs.existsSync(absolutePath)) {
         settings.runtime.path = absolutePath;
         await this.updateEnvSettings(settings);
       }
     }
 
+    if (
```
benbrown, line 215: is this not something that can be done inside the yeoman template?

Composer/packages/server/src/services/project.ts
```diff
@@ -182,8 +182,8 @@ export class BotProjectService {
       BotProjectService.deleteRecentProject(locationRef.path);
       throw new Error(`file ${locationRef.path} does not exist`);
     }
-
-    if (!(await StorageService.checkIsBotFolder(locationRef.storageId, locationRef.path, user))) {
+    const isBotFolder = await StorageService.checkIsBotFolder(locationRef.storageId, locationRef.path, user);
```
benbrown, line 185: why this change?

benbrown (changes requested):

Composer/packages/server/src/models/asset/assetManager.ts
pavolum: I have now initialized yeomen env with the cwd set to the appData directory, it seems to be working in both web and electron context. I want to verify and see the dir itself in an electron setting. Where would that live when you run yarn start:electron:dev?

Composer/packages/server/src/models/asset/assetManager.ts
pavolum: Resolved

Composer/packages/client/src/pages/home/Home.tsx
pavolum: Good catch, I changed it to call  the telem method from the respective onClicks for the UI components, now OnClickNewBot only calls shared functionality.

Composer/packages/client/src/recoilModel/dispatchers/storage.ts
pavolum: If we are getting rid of the template view in the home.tsx view then yeah it only needs to be in the createOptions component.

Composer/packages/client/src/components/CreationFlow/v2/CreateOptions.tsx
pavolum: Ive created task #5506 to track this. I say we leave it for now and revisit it once we have an empty bot template which presumably will be used under the hood when this template is selected (instantiate empy bot template -> Route to import from qna modal and apply the uploaded KB to empty bot)

Composer/packages/server/src/controllers/project.ts
pavolum: I was originally planning on leaving PVA out in this PR (I am tracking its reintegration in #5498 ) but can look into it. It should be simple to put back in this v2 flow and test since it has its own isolated flow.

benbrown: Would this then break PVA in the nightly builds until fixed?

Composer/packages/client/src/recoilModel/dispatchers/storage.ts
benbrown: OK good point. We can leave that in place for now, but when/if the homepage gets revamped, we should move this.

Composer/packages/server/src/models/asset/assetManager.ts
benbrown: in your `~/BotFrameworkComposer` folder

Composer/packages/server/src/controllers/project.ts
pavolum: It should not, there are no PVA unit tests running through this code. This PR should introduce a separate code path for creation that no existing production code paths or build blocking UTs run through unless you have the hidden feature flag checked.

Composer/packages/server/src/models/bot/botProject.ts
pavolum: Good point, I just validated it and it works nicely

pavolum: Kinda makes you think how much of the flow should/could be handle by the template itself. Things like the build command could be templated as well but that may be overkill
image
In a larger sense the template itself could probably run things like dialog merge once it completes file generation but may be best to keep that logic in Composer if we are not needing to optimize on non Composer use of this

pavolum: Yeah I guess im just realizing the power of Yeomen, cool stuff!

pavolum:
> make itemkey `node` to match dictionary

Okay I have resolved this and made it a bit more strongly typed as well.

Composer/packages/server/src/controllers/project.ts
pavolum: @benbrown Ive restored it and done initial testing with favorable results except there are still some quirks. Getting an env to test this e2e with PVA is blocking me to iron out these quirks efficiently. I have created a task to validate the PVA flows and Tony has offered to help which should expedite the PVA integration (I believe its just one more small change needed I just cant test the damn thing effectively).

Since this will not affect anything in prod, PVA included, is it okay to iron out the kinks in its own PR so we can unblock this PR? I should I work with Tony now to get this in during this PR?

Composer/packages/server/src/controllers/project.ts
pavolum: Resolving as PVA discussion tracked elsewhere

Composer/packages/client/src/components/CreationFlow/v2/DefineConversation.tsx
pavolum: Okay I have added those fields and nothing is done with the selection as of this PR. We can take the selections and apply them to the runtime once the runtime is ready for that, I have created a task to track that.

Composer/packages/client/src/components/ImportModal/ImportModal.tsx
```diff
@@ -70,7 +70,8 @@ export const ImportModal: React.FC<RouteComponentProps> = (props) => {
       templateDir,
       urlSuffix,
     };
-    let creationUrl = `/projects/create/${encodeURIComponent(source)}`;
+    // TODO: remove after testing
```
benbrown, line 73: remove?

Composer/packages/server/src/models/asset/assetManager.ts
```diff
@@ -92,58 +94,78 @@ export class AssetManager {
...
+
+  private async installRemoteTemplate(generatorName: string, npmPackageName: string): Promise<boolean> {
+    this.yeomanEnv.cwd = templateGeneratorPath;
+    const registeredGenerators: string[] = await this.yeomanEnv.getGeneratorNames();
+
+    if (registeredGenerators.indexOf(generatorName) !== -1) {
+      return true;
     } else {
-      throw new Error('selected template has no local or external address');
+      await this.yeomanEnv.installLocalGenerators({ [npmPackageName]: '*' });
```
benbrown, line 139: we should do some error catching around this. not a blocker for now, but yeoman could fail.

benbrown: since this is inside a try catch its probably ok but .. maybe some error correction

extensions/runtimes/src/index.ts
```diff
@@ -308,4 +308,174 @@ export default async (composer: any): Promise<void> => {
...
+        composer.log('OUTPUT FROM BUILD', stdout);
+        if (stderr) {
+          composer.log('ERR FROM BUILD: ', stderr);
+        }
+      } catch (err) {
+        composer.log('Error doing dotnet publish', err);
+        throw err;
+        return;
+      }
+      // Then, copy the declarative assets into the build artifacts folder.
```
benbrown, line 415: this is probably not necessary anymore with the new runtime structure

pavolum: It seems you have already commented this out, I will leave as is with commented out code

Composer/packages/client/src/components/CreationFlow/v2/CreateOptions.tsx
```diff
@@ -0,0 +1,268 @@
...
+          <div css={content} tabIndex={-1}>
+            {item.name}
+          </div>
+        </div>
+      ),
+    },
+  ];
+
+  const onRenderRow = (props) => {
+    if (props) {
```
cwhitten, line 171: nit: invert this guard

```typescript
if (!props) return null;
return (<div />)
```

Composer/packages/client/src/components/CreationFlow/v2/CreateOptions.tsx
```diff
@@ -0,0 +1,268 @@
...
+import { IPivotItemProps, Pivot, PivotItem } from 'office-ui-fabric-react/lib/Pivot';
+import { csharpFeedKey, nodeFeedKey } from '@botframework-composer/types';
+
+import { DialogCreationCopy, EmptyBotTemplateId, feedDictionary, QnABotTemplateId } from '../../../constants';
+import httpClient from '../../../utils/httpUtil';
+
+import { TemplateDetailView } from './TemplateDetailView';
+
+// -------------------- Styles -------------------- //
+
```
cwhitten, line 33: `detailsListContainer`, `templateDetailContainer`, `pickerContainer` don't appear to be responsive. Is this by design?

Composer/packages/client/src/components/CreationFlow/v2/CreateOptions.tsx
```diff
@@ -0,0 +1,268 @@
...
+// -------------------- CreateOptions -------------------- //
+type CreateOptionsProps = {
+  templates: BotTemplate[];
+  onDismiss: () => void;
+  onNext: (data: string) => void;
+  fetchTemplates: (feedUrls?: string[]) => Promise<void>;
+} & RouteComponentProps<{}>;
+
+export function CreateOptionsV2(props: CreateOptionsProps) {
```
cwhitten, line 113: I don't like this V2 convention in the class name. Can you think of alternative ways to represent a different component path than hardcoding v2 in both the directories and the classes?

Composer/packages/client/src/components/CreationFlow/v2/CreateOptions.tsx
```diff
@@ -0,0 +1,268 @@
...
+  const [option] = useState(optionKeys.createFromTemplate);
+  const [disabled] = useState(false);
+  const { templates, onDismiss, onNext } = props;
+  const [templateReadMe, setTemplateReadMe] = useState('');
+  const [currentTemplate, setCurrentTemplate] = useState('');
+  const [emptyBotKey, setEmptyBotKey] = useState('');
+  const [selectedFeed, setSelectedFeed] = useState<{ props: IPivotItemProps }>({ props: { itemKey: csharpFeedKey } });
+
+  const selection = useMemo(() => {
```
cwhitten, line 122: `selection` of what? It's opaque. 

How about `selectedTemplate` or something more descriptive.

Composer/packages/client/src/components/CreationFlow/v2/CreateOptions.tsx
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    if (props.location && props.location.search) {
+      routeToTemplate += props.location.search;
+    }
+
+    onNext(routeToTemplate);
+  };
+
+  const tableColums = [
```
cwhitten, line 150: s/tableColums/tableColumns

Composer/packages/client/src/components/CreationFlow/v2/CreateOptions.tsx
```diff
@@ -0,0 +1,268 @@
...
+      return (
+        <DetailsRow {...props} data-testid={props.item.id} styles={rowDetails(disabled)} tabIndex={props.itemIndex} />
+      );
+    }
+    return null;
+  };
+
+  const fetchReadMe = async (moduleName: string) => {
+    try {
+      const response = await httpClient.post(`assets/templateReadme`, {
```
cwhitten, line 181: We're breaking the layering here by fetching data directly in the view.

Also, why is this HTTP method a post and not a get?

Composer/packages/client/src/components/CreationFlow/v2/DefineConversation.tsx
```diff
@@ -0,0 +1,369 @@
...
+      marginBottom: '1rem',
+    },
+  ],
+};
+
+// -------------------- DefineConversation -------------------- //
+
+const MAXTRYTIMES = 10000;
+
+interface DefineConversationFormData {
```
cwhitten, line 77: this shape should be better typed, and our convention is to use types over interfaces

Composer/packages/client/src/components/CreationFlow/v2/DefineConversation.tsx
```diff
@@ -0,0 +1,369 @@
...
+  location?: string;
+
+  templateDir?: string; // location of the imported template
+  eTag?: string; // e tag used for content sync between composer and imported bot content
+  urlSuffix?: string; // url to deep link to after creation
+  alias?: string; // identifier that is used to track bots between imports
+  preserveRoot?: boolean; // identifier that is used to determine ay project file renames upon creation
+}
+
+interface DefineConversationProps
```
cwhitten, line 92: our convention is to use types over interfaces

Composer/packages/client/src/components/CreationFlow/v2/DefineConversation.tsx
```diff
@@ -0,0 +1,369 @@
...
+  const writable = focusedStorageFolder.writable;
+
+  // template ID is populated by npm package name which needs to be formatted
+  const normalizeTemplateId = (templateId?: string) => {
+    if (templateId) {
+      return templateId.replace(invalidNameCharRegex, '_');
+    }
+  };
+
+  const getDefaultName = () => {
```
cwhitten, line 126: was this function copied from somewhere? it seems out of place

Composer/packages/client/src/components/CreationFlow/v2/DefineConversation.tsx
```diff
@@ -0,0 +1,369 @@
...
+            onRenderCardContent: ImportSuccessNotificationWrapper({
+              importedToExisting: false,
+            }),
+          });
+          addNotification(notification);
+        }
+      }
+
+      onSubmit(
+        {
```
cwhitten, line 269: does prettier format this object parameter like it's checked-in? why not `{ ...dataToSubmit },`

Composer/packages/client/src/components/CreationFlow/v2/TemplateDetailView.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { jsx } from '@emotion/core';
+import React from 'react';
+
+type TemplateDetailViewProps = {
+  templateId: string;
+  readMe: string;
+};
+
+export const TemplateDetailView: React.FC<TemplateDetailViewProps> = (props) => {
+  return <div className="content" dangerouslySetInnerHTML={{ __html: props.readMe }}></div>;
```
cwhitten, line 15: We really should not `dangerouslySetInnerHTML`

Composer/packages/server/src/controllers/asset.ts
```diff
@@ -2,8 +2,11 @@
 // Licensed under the MIT License.
```
cwhitten, line 2: A general ask for this file -- is all the data fetching in the controller layer the right thing?

Composer/packages/server/src/models/asset/assetManager.ts
```diff
@@ -92,58 +93,86 @@ export class AssetManager {
...
+      const npmPackageName = templateId;
+      const generatorName = npmPackageName.toLowerCase().replace('generator-', '');
+
+      const remoteTemplateAvailable = await this.installRemoteTemplate(generatorName, npmPackageName);
+
+      if (remoteTemplateAvailable) {
+        await this.instantiateRemoteTemplate(generatorName, dstDir, projectName);
       }
+
+      ref.path = ref.path + `/${projectName}`;
```
cwhitten, line 123: nit: interpolation over concatenation

extensions/azurePublish/src/node/luisAndQnA.ts
```diff
@@ -12,8 +12,8 @@ import { AzurePublishErrors } from './utils/errorHandler';
 import {BotProjectDeployLoggerType} from './types';
 
 const readdir: any = promisify(fs.readdir);
-
-const botPath = (projPath: string) => path.join(projPath, 'ComposerDialogs')
+// const botPath = (projPath: string) => path.join(projPath, 'ComposerDialogs')
```
cwhitten, line 15: kill this if unneeded?

cwhitten (dismissed):
general comments: there is random formatting changes across a host of files adding things like whitespace. why?

there are no new tests are part of this. it seems like there should be.

it would be beneficial to improve the typing across the board

why are we checking in package-lock.json's?

Composer/packages/client/src/components/CreationFlow/v2/CreateOptions.tsx
benbrown: I thought it would be appropriate to have this data being fetched at the component level since it is not used anywhere else. Would you have it live in the app state?

Composer/packages/client/src/components/CreationFlow/v2/CreateOptions.tsx
pavolum: Our plan is to remove the v2 convention when we make the hard shift to this creation experience prior to the build release.  That way this creation code path has the naming and file structure as the current path. 

Would there be an issue with that? I could change the 'v2' verbiage but pre build we will likely change it again to i.e. 'CreateOptions.ts' (once that file is no longer needed to support the current creation flow).

Composer/packages/client/src/components/CreationFlow/v2/CreateOptions.tsx
pavolum: I agree that it should be in app state, ill add it back there

Composer/packages/client/src/components/CreationFlow/v2/CreateOptions.tsx
pavolum: Handled

Composer/packages/client/src/components/CreationFlow/v2/CreateOptions.tsx
pavolum: Handled

Composer/packages/client/src/components/CreationFlow/v2/CreateOptions.tsx
benbrown: ¯\_(ツ)_/¯

Composer/packages/client/src/components/CreationFlow/v2/DefineConversation.tsx
pavolum: Handled

Composer/packages/client/src/components/CreationFlow/v2/DefineConversation.tsx
pavolum: Handled

Composer/packages/client/src/components/CreationFlow/v2/DefineConversation.tsx
pavolum: Yes this is from the original DefineConversation.tsx and slightly altered to provide a valid default project name to prepopulate the name field in the UI. There is a good bit of redundant code between the two flows which is intended since they will only live side by side for a short period of time.

Composer/packages/client/src/components/CreationFlow/v2/DefineConversation.tsx
pavolum: Handled

Composer/packages/client/src/components/CreationFlow/v2/TemplateDetailView.tsx
pavolum: Yeah Im not a big fan of it for the xss attack vulnerability although we control the src it pulls the html from. 

We would have to rethink how we grab the template detail view. I like the idea of grabbing the readme as it makes template development easier and it sticks with the npm standard. 

Alternatively, the feed sends metadata that could pass the core info of the template (i.e. supported languages, description, etc) to Composer for rendering. It would make visual styling easier and more consistent then pulling readMe markdown.

@benbrown what do you think the meta data config would be in package.json of the template. So instead of populating a readme they would populate a handful of fields in the package.json of their template package.

Composer/packages/server/src/models/asset/assetManager.ts
pavolum: Handled

extensions/azurePublish/src/node/luisAndQnA.ts
pavolum: I didnt make this change, @benbrown is it okay to remove this? Note changes to this file will affect the existing creation flow

Composer/packages/client/src/components/CreationFlow/v2/CreateOptions.tsx
pavolum: Its like the whole division of business logic and presentational logic.

Ive also been on teams that ban local state in favor of central. Like even if a component is the only one that needs that data it may not be the case down the line so always place state in a central repo like recoil or redux. 

Thats a little overkill imo but I do like the division of business logic and presentational logic aspect.

Composer/packages/server/src/controllers/asset.ts
pavolum: Okay I have exported most of these additions to AssetManager. I have left the readme function here since your other comment may lead to the removal of this technique.

extensions/azurePublish/src/node/luisAndQnA.ts
pavolum: I have removed it but will leave this comment to confirm with @benbrown

Composer/packages/client/src/components/CreationFlow/v2/CreateOptions.tsx
pavolum: Actually I am going to leave this as is. We need to settle the ideal read me/template detail acquisition story. If we are no longer going down the readme grabbing route then it does not make sense to move this to app state since that will potentially be throw away code.

pavolum:
> general comments: there is random formatting changes across a host of files adding things like whitespace. why?
**Must be an auto formatter mismatch that altered those files, I have reverted these changes.** 

> there are no new tests are part of this. it seems like there should be.
**Small aspects of the flow may be subject to change and this is behind a hidden feature flag which is why we are waiting to develop UTs. That being said if its a hard block for merges to main even if the code is hidden from end users, then I can tackle an initial set now and alter them as/if the code changes. @benbrown what do you think, should UTs be prioritized before pushing this PR?**

> it would be beneficial to improve the typing across the board
**I have improved some of the typing you explicitly called out. I largely stuck with the patterns implemented in the existing creation experience and the standards their seem to be outdated (interfaces are used over types, etc) so I have updated those as well. Are there any other specific typing issues that you would want to see resolved before this is pushed?**

> why are we checking in package-lock.json's?
**I have removed this**

cwhitten:
> I can tackle an initial set now and alter them as/if the code changes.

Sounds good! 

> I largely stuck with the patterns implemented in the existing creation experience and the standards their seem to be outdated

I understand. The downside of copying a bunch of code and checking it in is you now own the copy. I'll take a closer look at the types soon.

Composer/packages/client/src/components/CreationFlow/v2/TemplateDetailView.tsx
pavolum: Resolving as the new readMe approach has been added that takes call logic out of the component layer, adds readme to app state and does not dangerouslyinjecthtml

Composer/packages/client/src/components/CreationFlow/v2/CreateOptions.tsx
pavolum: Its parent up the div chain allows for decent responsiveness at current state with these existing rules. That being said there is a task item for design iterations on this modal (responsiveness details, boilerplate readme structure, general pixel corrections). I am going to leave this as is and revisit when we do the smaller PR design iterations.

pavolum:
Some things are being brought up here that are good example of the need to clear up the following devops question that is introduced with the hidden feature flag concept.

**Should we be working on a separate feature branch until this is at its first e2e acceptable state and then PR or should we push incrementally to main and hide the incomplete (working but no UT, end designs, etc) code behind a hidden feature flag?**

This PR assumes the latter which is a new approach then what we are used to. 

Two options for feature work:
- Large features are worked on in their own feature branch and not pushed to main until they are in shipping condition (UTs developed, final designs implemented, etc.)
  - Cons:
    - Need to switch to and build feature branches to try new features
    - Potential for larger rifts from main if you do not update feature branches regularly
  - Pros:
    - Everything in main is in its final state and code is guaranteed to not go unreviewed
- Large feature work incrementally pushed to main behind a hidden feature flag
  - Pros:
    - Regular integration with main leads to less branch rifting 
    - Other devs can easily turn on hidden feature flag work in main to test in development work.
  - Cons:
    - Potential for unimplemented code (i.e first incremental push to main has not UTs or final designs and its on the word of the dev to actually implement those at a later PR prior to shipping)

pavolum:
Oops accidentally deleted a quote reply comment that was outdated and it seems to have deleted the original, will try to recover.

Composer/packages/client/src/components/CreationFlow/v2/CreateOptions.tsx
benbrown: We can use the <Stack> element, it does a nice job in these cases.

Composer/packages/client/src/components/CreationFlow/v2/CreateOptions.tsx
```diff
@@ -0,0 +1,259 @@
...
+        <Pivot
+          defaultSelectedKey={csharpFeedKey}
+          onLinkClick={(item) => {
+            if (item) {
+              setSelectedFeed(item);
+            }
+          }}
+        >
+          <PivotItem headerText="C#" itemKey={csharpFeedKey}></PivotItem>
+          <PivotItem headerText="Typescript" itemKey={nodeFeedKey}></PivotItem>
```
benbrown, line 221: Before we ship, we can kill this pivot since we will not have Typescript templates. Til then let's leave it to test underyling features.

Composer/packages/client/src/components/CreationFlow/v2/CreationFlow.tsx
```diff
@@ -0,0 +1,198 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// TODO: Remove path module
```
benbrown, line 4: why this todo?

Composer/packages/client/src/constants.ts
```diff
@@ -379,3 +390,25 @@ export const triggerNotSupportedWarning = () =>
   formatMessage(
     'This trigger type is not supported by the RegEx recognizer. To ensure this trigger is fired, change the recognizer type.'
   );
+
+export const feedDictionary: { [key in FeedName]: string } = {
+  firstPartyCsharp:
+    'https://registry.npmjs.org/-/v1/search?text=conversationalcore&size=100&from=0&quality=0.65&popularity=0.98&maintenance=0.5',
+  firstPartyNode:
```
benbrown, line 397: we'll want to remove/disable the JS feed before shipping

Composer/packages/server/src/models/asset/assetManager.ts
```diff
@@ -92,58 +97,86 @@ export class AssetManager {
...
+
+      // find selected template
+      const npmPackageName = templateId;
+      const generatorName = npmPackageName.toLowerCase().replace('generator-', '');
+
+      const remoteTemplateAvailable = await this.installRemoteTemplate(generatorName, npmPackageName, templateVersion);
+
+      if (remoteTemplateAvailable) {
+        await this.instantiateRemoteTemplate(generatorName, dstDir, projectName);
       }
```
benbrown, line 126: do we need an else here?

benbrown: maybe we should return true / false from installRemoteTemplate instead of using throw/catch?

lets pick one instead of both.

Composer/packages/server/src/models/asset/assetManager.ts
```diff
@@ -92,58 +97,86 @@ export class AssetManager {
...
   }
 
+  private async instantiateRemoteTemplate(
+    generatorName: string,
+    dstDir: string,
+    projectName: string
+  ): Promise<boolean> {
+    this.yeomanEnv.cwd = dstDir;
+
+    log('About to instantiate a template!', dstDir, generatorName, projectName);
```
benbrown, line 162: does awaiting this work?


normally when there is a callback, it doesn't work as a promise.

We should double check on this - otherwise, wrap this in a Promise and resolve inside the callback.

Composer/packages/server/src/models/asset/assetManager.ts
```diff
@@ -234,4 +267,56 @@ export class AssetManager {
       }
     }
   }
+
+  private getFeedType(): FeedType {
+    // TODO: parse through data to detect for npm or nuget package schema and return respecive result
```
benbrown, line 272: I have code that does this in package manager.

Composer/packages/client/src/components/CreationFlow/v2/CreateOptions.tsx
pavolum: Okay I will disable it as an option so users wont be able to click it

Composer/packages/client/src/components/CreationFlow/v2/CreationFlow.tsx
pavolum: Removed

Composer/packages/client/src/constants.ts
pavolum: Handled

Composer/packages/client/src/components/CreationFlow/v2/CreateOptions.tsx
pavolum: Hmm it seems that UI element is not disableable. I have removed it for now so there is only the C# tab. Will iterate with design for a better short term solution if needed. (iterating with design task is created and will be handled in its own PR)

Composer/packages/server/src/models/asset/assetManager.ts
pavolum: Handled

Composer/packages/server/src/models/asset/assetManager.ts
pavolum: Just ran a test, awaiting it does work :D. It runs the callback prior to continuing the function and returning 'true'

Composer/packages/server/src/models/asset/assetManager.ts
pavolum: Ah thats right, I remember you mentioning that a while back. How shall we proceed? Want to send me the deets to that implementation and I can make the change to use that function?

Composer/packages/server/package-lock.json
```diff
@@ -0,0 +1,1780 @@
+{
```
cwhitten, line 1: we should also remove this

Composer/packages/server/src/controllers/__tests__/project.test.ts
```diff
@@ -172,6 +172,49 @@ describe('should get recent projects', () => {
...
+
+  it('should start to create a new project', async () => {
+    BotProjectService.createProjectAsync = jest.fn();
+
+    ProjectController.createProjectV2(mockReq, mockRes);
+    expect(mockRes.status).toHaveBeenCalledWith(202);
+  });
+
+  it('should create a new project', async () => {
+    // await ProjectController.createProjectAsync(mockReq, '123');
```
cwhitten, line 214: is this intended to be commented out?

Composer/packages/server/src/controllers/asset.ts
```diff
@@ -16,6 +18,57 @@ async function getProjTemplates(req: Request, res: Response) {
...
+export async function getProjTemplatesV2(req: any, res: any) {
+  try {
+    let templates: BotTemplateV2[] = [];
+
+    // Get FeedUrl
+    const { feedUrls, getFirstPartyNpm } = req.body;
+
+    // Grab templates from FeedURls
+    if (feedUrls) {
+      // TODO: Current assumption is that the feed is an npm feed with given result structure (https://registry.npmjs.org/-/v1/search?text=docker&size=100&from=0&quality=0.65&popularity=0.98&maintenance=0.5)
```
cwhitten, line 30: what's the future plan here with this TODO?

Composer/packages/server/src/controllers/asset.ts
```diff
@@ -16,6 +18,57 @@ async function getProjTemplates(req: Request, res: Response) {
...
+export async function getTemplateReadMe(req: any, res: any) {
+  try {
+    const moduleName = req.query?.moduleName;
+
+    if (!moduleName) {
+      res.status(400).json({
+        message: 'missing module name on request',
+      });
+    } else {
+      const moduleURL = 'https://registry.npmjs.org/' + moduleName;
```
cwhitten, line 58: nit: interpolation over concatenation

Composer/packages/server/src/models/asset/assetManager.ts
```diff
@@ -92,58 +97,88 @@ export class AssetManager {
...
+  private async instantiateRemoteTemplate(
+    generatorName: string,
+    dstDir: string,
+    projectName: string
+  ): Promise<boolean> {
+    this.yeomanEnv.cwd = dstDir;
+
+    log('About to instantiate a template!', dstDir, generatorName, projectName);
+    await this.yeomanEnv.run([generatorName, projectName], {}, () => {
+      console.log('DONE');
```
cwhitten, line 166: Is this necessary?

extensions/packageManager/src/client/Library.tsx
```diff
@@ -294,16 +295,20 @@ const Library: React.FC = () => {
...
-    } catch (err) {
-      setApplicationLevelError({
-        status: err.response.status,
-        message: err.response && err.response.data.message ? err.response.data.message : err,
-        summary: strings.libraryError,
-      });
+    if (!isLoadingInstalled) {
+      setIsLoadingInstalled(true);
+      try {
+        updateInstalledComponents([]);
```
cwhitten, line 301: why does this function invocation need to accept an empty array?

benbrown: That is resetting the state to an empty list while it loads...

Composer/packages/server/package-lock.json
pavolum: Man, I accidentally run one 'npm' command instead of 'yarn' and these files sneak in there. Ill remove now.

Composer/packages/server/src/controllers/__tests__/project.test.ts
pavolum: Ah thats remnants of an old test, removing now.

Composer/packages/server/src/controllers/asset.ts
pavolum: Ive removed the TODO, the work needed for that revolved around documentation tasks that have been created and are being tracked. Essentially we just need to let folks know through the docs the standard expected structure of the template feeds once we allow custom feeds for templates (post R12). 
The expected structure is set by the respective package manager (npm) so we are not creating our own feed schema or anything, we are sticking to the format that  npm dictates to not uneccisarily reinvent the wheel. 

I have removed the TODO.

Composer/packages/server/src/controllers/asset.ts
pavolum: Handled

Composer/packages/server/src/models/asset/assetManager.ts
pavolum: Console.log is a bit invasive. I have changed to to use the log() function.

cwhitten:
> ConversationalCore feature flag will no longer work

Other work streams are relying on this feature flag. Composer Adapters for one. Can you make sure to connect with Ben Y on the appropriate way to guard UX associated with the VA scenarios?

pavolum:
> > ConversationalCore feature flag will no longer work
> 
> Other work streams are relying on this feature flag. Composer Adapters for one. Can you make sure to connect with Ben Y on the appropriate way to guard UX associated with the VA scenarios?

Good to know, Ill ping him now!

pavolum:
> > > ConversationalCore feature flag will no longer work
> > 
> > 
> > Other work streams are relying on this feature flag. Composer Adapters for one. Can you make sure to connect with Ben Y on the appropriate way to guard UX associated with the VA scenarios?
> 
> Good to know, Ill ping him now!

I spoke with Ben Y and we have a plan in place. Once this gets into main he will pull latest and make it such that the adapter feature is only shown if the NEW_CREATION_FLOW feature flag is checked. That way when I do the smaller PR that gets rid of the REMOTE_TEMPLATE_CREATION feature flag then his branch wont break when he pulls latest.

cwhitten (approved):

PR #5032, opened 2020-12-01
fix: Lg content missing when switch the page

### LINKED ISSUE #4988: LG Editor keeps dropping changes
## Describe the bug

I don't have a consistant repro per say but the LG Editor in the latest composer build is dropping changes **A LOT**. I did about 5 minutes worth of changes to an LG file clicked "Restart Bot" and it was if none of my changes took.  I went back into composer and they were gone.  A second ago I add a new template to a file, clicked away to another file, and when I clicked back my new template was gone :(

Earlier I was typing and when I'd run I got an error saying "turn..value" is invalid. I looked at that expression in composer and it said "turn.dialogEvent.value".  I reloaded the project and sure enough it changed to "turn..value".

## Version

1.2.1

lei9444:
## Description
**root cause**

1. the text in editor is updated, the lg content in state is blocked by the parser and will be updated delay. If we change the page before the state get the new content, the old state content will be rendered in the editor

2. the worker for parse is single thread,  so when we do a full parse for all files, it will take a long time to update the state.

**In this PR**

1. add a new worker to generate the diagnostics, drop the expired parse result and attach the latest diagnostics to the lg state.
2. update the content directly when we call the updatelgfile dispatcher and attach the parse result after the parse is over.

**Also in this PR**
LG server did some parsing in main thread before, now it's all move to worker thread, so now Electron build will have better performance.

## Task Item
closes #4988


boydc2014:
Thanks, I and many others confirmed this do fix the data loss in Steven's big bot case. 

But we are still facing the lagging for big bot in electron (web app is good), let's take this chance to fix that as well, put a temporary on-hold for this.

zhixzhan:
> Thanks, I and many others confirmed this do fix the data loss in Steven's big bot case.
> 
> But we are still facing the lagging for big bot in electron (web app is good), let's take this chance to fix that as well, put a temporary on-hold for this.

Updates: The lagging in electron is caused by a main thread LG parsing, now it's all moved to worker thread. Web app is good because LSP is running on server, won't block client UI thread.

boydc2014 (approved):

PR #5037, opened 2020-12-01
doc: updates to Package manager readme

zxyanliu:
## Description

- Updated language, format, and style.
- Updated the help link in the Package manager page pointing to [this readme](https://github.com/microsoft/BotFramework-Composer/blob/main/extensions/packageManager/readme.md) file.
#minor


extensions/packageManager/src/client/Library.tsx
```diff
@@ -17,7 +17,7 @@ import { WorkingModal } from './workingModal';
 
 const DEFAULT_CATEGORY = formatMessage('Available');
 
-const docsUrl = `https://aka.ms/composer-use-package-library`;
+const docsUrl = `https://github.com/microsoft/BotFramework-Composer/blob/main/extensions/packageManager/readme.md`;
```
benbrown, line 20: Why this change?  Can we just redirect the aka.ms link instead?

benbrown (dismissed):
Aside from my comment about the aka.ms link, this looks great!

extensions/packageManager/readme.md
```diff
@@ -1,144 +1,155 @@
...
+3. Inside this new folder, create a subfolder called `exported`.
+4. Copy the dialogs you want to share from the original `dialogs` folder into the new `exported` folder.
+5. Follow the instructions ([NuGet](#for-nuget) |[NPM](#for-npm)) below to publish the package to NuGet and/or NPM.
 
-To share dialogs you've created in Composer:
+### Package LG files
 
-1. Open your bot project and find the `dialogs` folder. The files for each dialog are contained in a named sub-folder here.  Identify the dialogs you want to share.
-2. Create a new folder with the name of your new package. For example, `cool-dialogs`
+1. Open your bot project and find the `language-generation` folder that contains the LG files you want to share.
```
benbrown, line 136: There is a little bit more complexity here.

Inside language-generation, there will be locale folders like `en-us` -- and then inside these, there will be the actual LG files.

To support multiple languages, the structure should retain the locale subfolders -

so you might end up where you have 2 LG files in 2 languages, for example:

exported/en-us/fancy.en-us.lg
exported/en-us/cool.en-us.lg
exported/de-de/fancy.de-de.lg
exported/de-de/cool.de-de.lg

extensions/packageManager/src/client/Library.tsx
zxyanliu: Sure. I updated this link to: 
https://aka.ms/composer-package-manager-readme

extensions/packageManager/readme.md
zxyanliu: I updated the wording to clarify this.

PR #5041, opened 2020-12-02
fix: l10n files special handling for apostrophes and "other" numbers

### LINKED ISSUE #5040: Using Composer in other languages can lead to crashing
## Describe the bug

Changing the Application Settings to some other languages will cause Composer to fail to load (e.g. French makes Design View crash).

## Expected behavior

The app should load without crashing in all supported locales.

beyackle:
## Description

This fixes the formatting errors in our l10n files. Some of them relate to using single quotes (`'`) instead of doubling them (`''`), which makes strings in those languages that use lots of single quotes (e.g. French and Turkish) crash on load. Another is that, in forms that use the ICU format for numbers (e.g. `=0 {No bots} =1 {One bot} other {# bots}`), the word "other" had been translated into the target language instead of being left alone as part of ICU syntax.

## Task Item

fixes #5040 

## Screenshots

French loads now without crashing the app (and I've verified this for all our supported languages)
image


cwhitten:
how we will avoid new translations drops introducing these problems?

liweitian:
Double checked. No crashed page found.

boydc2014 (approved):

beyackle:
> how we will avoid new translation drops introducing these problems?

I'll get in touch with the localization team with more guidelines.

PR #5046, opened 2020-12-02
fix: change runtime identifier default to win-x64

zidaneymar:
## Description
minor fix: change runtime identifier default to win-x64

## Task Item
#minor


cwhitten:
@zidaneymar what is driving this change? the orchestrator requirement?

zidaneymar:
> @zidaneymar what is driving this change? the orchestrator requirement?

yes, this is actually an issue because if user provide an old version of publishing profile which doesn't have the configuration for runtime identifier, currently composer would not build the runtime as a self-contained app, which means some dependencies would not be loaded on azure. So I change this config to win-x64 as default.

luhan2017 (approved):

PR #5051, opened 2020-12-02
fix: use currentProjectId to filter the updates in file persistence

### LINKED ISSUE #5043: File persistence currently happening for all bots
## Describe the bug
https://github.com/microsoft/BotFramework-Composer/blob/main/Composer/packages/client/src/recoilModel/DispatcherWrapper.tsx#L107

Currently we are checking for updates on all projects in the BotProject. We need to be using the currentProjectIdState which needs to be set based on the URL. This could improve our performance slightly

lei9444:
## Description
This PR's intention is to only use currentProjectId to filter the persistence layer to make it more efficient.  But we can't simply do that because of the follow issues found. 

### Issue 1: 
When navigating in project tree, only the design page will update the current project id based on the selection. The lu\lg\qna\settings's project tree won't update the project id, resulting in changes across different projects. In this case, simply filtering by current project id will cause data loss.  

### Issue 2:
Even in design page, if we add a dialog using the context menu to another un-selected bot, the current project id hasn't switched to the new selection, causing the the adding dialog behavior is not added to the current project id. In this case, simply filtering by current project id will cause data loss.  

**Fixes**
1. add a location listener to update the current project Id in different pages. 
2. If we click the context menu icon, do a navigation first before adding dialogs (or any other actions)
3. use current project id to filter the updates in file persistence

## Task Item
closes #5043


boydc2014:
Looks like we spotted several cases that the changes are not only just in the current project, maybe due to some data flow order, or whatever reasons, so i wonder should we even do this optimization, given the risk that, an incorrect filter will cause data loss. And updating all projects seems to be pretty cheap, means we have very little gain here.

What do you think @srinaath @cwhitten

PR #5060, opened 2020-12-02
fix: l10n fix by moving constant strings to thunks

### LINKED ISSUE #5042: Lots of console warnings about missing translations on load
## Describe the bug

When Composer starts, we get dozens of warnings in the browser console about translations being missing. We shouldn't see these warnings for a locale ("en") that we don't technically support - there should be a workaround so it gets rerouted to "en-US". Our format-message library defaults to "en" before anything gets initialized, so that might be where this comes from.

## Screenshots

image

beyackle:
## Description

Some calls to formatMessage() have been getting run before the first call to formatMessage.setup (which happens when the app first loads, once it knows what locale to load). Most of these are in lists of constants, and another was in the Recoil initial values. The constants were moved to thunks that get run when a component is initialized, which lets them delay until after formatMessage.setup is called. Meanwhile, setting that Recoil value to initially undefined fixes the problem there; we don't display that value until after setup happens, so it's safe to leave that blank.

There are still some lingering warnings with strings that aren't in the en-US.json file, but this is part of a different problem where the l10n script is removing a lot of strings that rightfully should still be there, and this is something I haven't figured out yet. But the immediate issue is taken care of.

## Task Item

closes #5042


a-b-r-o-w-n (changes requested):
Since most of the constants are exported as objects already, let's make the properties getters instead so that the callers do not have to change.

beyackle:
> Since most of the constants are exported as objects already, let's make the properties getters instead so that the callers do not have to change.

Good idea. I'll  make that change.

beyackle:
The E2E failure is in TriggerCreation, which is currently flaky. This is safe to merge.

PR #5068, opened 2020-12-03
fix: Install 64-bit vcredist package

### LINKED ISSUE #5067: Orchestrator:  Composer hangs on Splash Screen (Windows)
## Describe the bug
Orchestrator requires several native dependencies.  On Windows, we require Visual Studio with the C++ workload installed, or the Visual Studio 2015-2019 redistributable (found [here](https://support.microsoft.com/en-us/help/2977003/the-latest-supported-visual-c-downloads)

Without this dependency, Composer hangs with a cryptic message that is neither descriptive or helpful:
image

We will partner with the Orchestrator team to fix this together: On Composer, we will install the Visual Studio 2015-2019 redistributable, and Orchestrator will fix their error handling to be more robust for 1.3

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

taicchoumsft:
## Description
On Windows, Orchestrator has a dependency on `vcruntime140.dll`, which is provided by either installing Visual Studio with the C++ Workload, or installing the Visual Studio 2015-2019 redistributable package.  

We will install the Visual Studio 2015-2019 redistributable package if the user hasn't installed it yet.  

The user will receive a UAC prompt asking if they want VC redist to modify their system folders.  This is the only user intervention required in the installer.

We fix this for the Electron 64-bit builds only - Orchestrator will not work on 32-bit. 

## Task Item
closes #5067 

## Screenshots
There will be an additional stage in the installer where the latest vcredist packages will be downloaded (~15MB) and installed if it hasn't been installed yet. 
image


boydc2014:
@feich-ms can you help verify this fixed the issue?

hatpick:
This built in plugin has some limitations listed here:
https://nsis.sourceforge.io/Builtin_NSISdl_plug-in

why not use https://nsis.sourceforge.io/Inetc_plug-in

taicchoumsft:
> This built in plugin has some limitations listed here:
> https://nsis.sourceforge.io/Builtin_NSISdl_plug-in
> 
> why not use https://nsis.sourceforge.io/Inetc_plug-in

Thanks @hatpick.  I stuck with the built-in NSISdl plugin since I didn't run into any issues with http with the vcredist download link.  If you feel we should make the switch to inetc I'll do it, I haven't tested this configuration out within electron-builder.

feich-ms:
Verified the latest Electron composer built from this PR [Pipelines](https://fuselabs.visualstudio.com/Composer/_build/results?buildId=190904&view=results). The installation successed without any issues now. Thanks for the fix.

boydc2014:
> > This built in plugin has some limitations listed here:
> > https://nsis.sourceforge.io/Builtin_NSISdl_plug-in
> > why not use https://nsis.sourceforge.io/Inetc_plug-in
> 
> Thanks @hatpick. I stuck with the built-in NSISdl plugin since I didn't run into any issues with http with the vcredist download link. If you feel we should make the switch to inetc I'll do it, I haven't tested this configuration out within electron-builder.

Thanks @hatpick, @taicchoumsft if it's a one line change that can gave us https support, i think we should just go head and do that, what do you think?

taicchoumsft:
> > > This built in plugin has some limitations listed here:
> > > https://nsis.sourceforge.io/Builtin_NSISdl_plug-in
> > > why not use https://nsis.sourceforge.io/Inetc_plug-in
> > 
> > 
> > Thanks @hatpick. I stuck with the built-in NSISdl plugin since I didn't run into any issues with http with the vcredist download link. If you feel we should make the switch to inetc I'll do it, I haven't tested this configuration out within electron-builder.
> 
> Thanks @hatpick, @taicchoumsft if it's a one line change that can give us https support, i think we should just go head and do that, what do you think?

Thanks @boydc2014 . Unfortunately this may not be a one-liner - it looks like for inetc we have to take that plugin dependency into our build since it's not a built-in plugin.  It's offered as a zip from the link above. There is a fork of the project, but it looks like we have to build that one from source.

boydc2014:
> > > > This built in plugin has some limitations listed here:
> > > > https://nsis.sourceforge.io/Builtin_NSISdl_plug-in
> > > > why not use https://nsis.sourceforge.io/Inetc_plug-in
> > > 
> > > 
> > > Thanks @hatpick. I stuck with the built-in NSISdl plugin since I didn't run into any issues with http with the vcredist download link. If you feel we should make the switch to inetc I'll do it, I haven't tested this configuration out within electron-builder.
> > 
> > 
> > Thanks @hatpick, @taicchoumsft if it's a one line change that can give us https support, i think we should just go head and do that, what do you think?
> 
> Thanks @boydc2014 . Unfortunately this may not be a one-liner - it looks like for inetc we have to take that plugin dependency into our build since it's not a built-in plugin. It's offered as a zip from the link above. There is a fork of the project, but it looks like we have to build that one from source.

Thanks for the helpful details. In that case, i would lean to keep at it as, i don't think we want to pay that price to get this extra again. Let's confirm this in tomorrow's platform chat, i will suggest we keep it is for now. And @hatpick Do you feel strong about that we should use intec?

boydc2014 (approved):

hatpick:
@boydc2014 unless the following is a show stopper I don't have any preference :)
_Even if you specify a HTTPS link, this plugin will download via HTTP. If your link is redirected from HTTP to HTTPS via a server-sent HTTP header, you will encounter a timeout error. You might want to use the Inetc plug-in instead._

PR #5078, opened 2020-12-03
fix: put package manager behind a feature flag

### LINKED ISSUE #4904: Place package manager behind a feature flag or settings toggle

For the 1.3 release, we want the package manager to be OFF by default, and require a user to toggle it ON.

The design plans call for this to be part of the feature flag UI on the settings page.

However, because package manager is an extension, it cannot see the feature flag setting. Also, it may require a restart to activate or deactivate the extension.  We need to hook the UI up to this capability so that the feature can be easily enabled.

benbrown:
## Description

Introduce a feature flag to hide the package manager feature unless enabled.

## Task Item

Fixes #4904 

## Screenshots

Screen Shot 2020-12-03 at 11 40 04 AM


a-b-r-o-w-n (approved):

hatpick:
NIT: @benbrown It's up to you and UX design but I think there should be an indicator in the UX telling user this is in "preview". Something as simple as "Package manager (preview)" on the left navigation.

PR #5090, opened 2020-12-03
fix: Cause Recognizers component to update after project reload

### LINKED ISSUE #5034: After installing a component, Composer has to reload before the imported dialogs will work in the runtime
## Describe the bug

It appears that extra steps are required to fully import and utilize the imported dialogs -- and that these steps currently only happen during the project opening phase.

## Version

current `main`

## To Reproduce

1. install a component
2. connect it to your root dialog
3. start the bot and trigger the newly imported dialog
4. you will get an error in emulator about an invalid object

5. Reload the browser window
6. Restart the bot and test in emulator -- it should work this time.

Crucially: I notice that upon reloading the project, additional files are added to the imported dialog folder under the `recognizers/` subfolder.  These files only get created when the project is initially opened.

## Expected behavior

The component should work without having to reload the bot.

benbrown:
## Description

Adds a lastReloadedFromServer member so that the Recognizers onEffect will fire to generate the new recognizer files for newly added dialogs.

HOWEVER, while the recognizer onEffect does fire, it does not generate the recogniezr files as expected.

## Task Item

Fixes #5034

### DIFF
--- a/Composer/packages/client/src/recoilModel/Recognizers.tsx
+++ b/Composer/packages/client/src/recoilModel/Recognizers.tsx
@@ -11,7 +11,7 @@ import { getBaseName, getExtension } from '../utils/fileUtil';
 
 import * as luUtil from './../utils/luUtil';
 import * as buildUtil from './../utils/buildUtil';
-import { crossTrainConfigState, luFilesState, qnaFilesState, settingsState } from './atoms';
+import { crossTrainConfigState, luFilesState, qnaFilesState, settingsState, projectMetaDataState } from './atoms';
 import { dialogsSelectorFamily } from './selectors';
 import { recognizersSelectorFamily } from './selectors/recognizers';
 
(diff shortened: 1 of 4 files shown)


lei9444:
Hi @benbrown  @srinaath I have a PR to fix this issue #5109 

**Root cause**
When we install a new component from the package manager page, the data flow in the state is:

download component -> reset the filepersistenceState to empty -> init bot state (dialog, lu, lg and other bot resources)-> set filepersistenceState.

the file persistence is used to update the file on the disk. when the filepersistence is empty, we will not notify composer to update the changes to server. 

The useEffect for recognizer generate function will be fired during the init bot state stage when the dialog or lu files updating, but during this stage the filepersistenceState is empty . So the recognizer state is updated by recognizer generate function but the filepersistence is empty and don't know the change has happened. After the bot reload, the recognizer updates in state is also over.  

**fix**
add filepersistenceState to the recognizer generate function listener. If the filepersistence is empty, we do nothing. This fix will protect the init bot stage. new updates can be fired after the bot reload.

PR #5092, opened 2020-12-04
chore: Upgrade Orchestrator Lib packages

### LINKED ISSUE #5095: Orchestrator: Composer hangs on Splash Screen (Mac)
This is the same issue as #5067 , Composer will hang on the splash screen if the `icu4c` dependency is not installed.  

A fix for this is in progress on the Orchestrator side. 

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [x] macOS
- [ ] Windows
- [ ] Ubuntu

taicchoumsft:
## Description
closes #5095

Update Orchestrator Lib to `4.11.0-beta.20201203.cd29e25`.  

This change allows Composer on OS X to start without having to manually install `icu4c` first, Orchestrator now bundles this dependency in the lib. 

Disable `npmRebuild` feature in electron-builder as well - the native rebuild has no benefit since Orchestrator's node packages run properly on our Electron version, and additionally causes a linking issue for Orchestrator on Mac.  Only Orchestrator is affected by this flag - **no other libraries in Composer are affected by `npmRebuild`**

Tested for regressions by running electron build on Windows 10, Mac OSX 10.15 and Ubuntu 20.04.  On Mac, the app is now bootable without icu4c being installed first.


tsuwandy (approved):

boydc2014:
@lei9444  can you confirm this with your mac?

lei9444:
I have tried the installer in mac, I can open it and update the bot, but when I start bot, composer crashed. But I think it is the m1 cpu cause the crashed issue

boydc2014:
> I have tried the installer in mac, I can open it and update the bot, but when I start bot, composer crashed. But I think it is the m1 cpu cause the crashed issue

I got a regular Mac and tried the same issue, app is opening fine, and editing is good, but the moment i click start, my app crashed.  @taicchoumsft i think we need to onhold on this. 

Also, i found that, orchestrator seems to not allowing me to train an empty lu, some weird errors happens if i just switch everything from ToDoLuisSample to orchestrator.

zhixzhan:
I tested on my intel iMac, without pre-install icu4c, it works.

boydc2014:
OK, we confirmed this would fix the issue using Zhixiang's env, which represents the most common case. The reason this fix wasn't working for me is my icu4c version is too low and in Leilei's case, he has the M1 cpu.

boydc2014 (approved):

PR #5096, opened 2020-12-04
fix: publish button and botStatus item should be disabled to click when some bot is publishing

alanlong9278:
## Description

Bugfix: publish button and botStatus item should be disabled to click when some bot is publishing status

## Task Item

#minor

## Screenshots

image


Composer/packages/client/src/pages/publish/Publish.tsx
```diff
@@ -162,7 +163,7 @@ const Publish: React.FC<RouteComponentProps<{ projectId: string; targetName?: st
   const [previousBotPublishHistoryList, setPreviousBotPublishHistoryList] = useState(botPublishHistoryList);
   // check history to see if a 202 is found
   useEffect(() => {
-    // most recent item is a 202, which means we should poll for updates...
+    let publishFinished = true;
```
yeze322, line 166: It's better to caculate this boolean value independently via `selectedBots.some(...)` for maintainability, even though it causes an extra iteration.

boydc2014: Agree, use built-in primitives whenever you can.

alanlong9278: move it independently

boydc2014 (approved):

PR #5184, opened 2020-12-07
fix: Allow choose Orchestrator recognizer type when lu file is empty

### LINKED ISSUE #5126: Orchestrator Recognizer - LU FILE IS EMPTY
On a fresh bot, I switch to the Orchestrator Recognizer and see:

Screen Shot 2020-12-04 at 12 48 43 PM

lei9444:
## Description
**Why we can't choose Orchestrator type in the previous design**

When the dialog recognizer type is CrossTrainedRecognizerSet, the value of the recognize field in dialog is '{fileBasename}.{locale}.lu.qna'. So we don't know the the dialog recognizer type from the recognize field in the dialog directly. 

When we choose the Orchestrator type, we will update the {fileBasename}.{locale}.lu.dialog content. We will write the Orchestrator type to this file. Composer then will use the recognize field from the dialog and the type from it's recognizer file to determine if it is Orchestrator type.

But in the previous design, we will not create a recognizer file for empty lu file and in order to avoid users to choose the Orchestrator type when the lu file is empty, we add a warning here.

**fix**

Generate the luis recognizer for empty file. But the file will not been add to the crosstrained recognizer file

## Task Item
closes #5126

### DIFF
--- a/Composer/packages/client/src/recoilModel/Recognizers.tsx
+++ b/Composer/packages/client/src/recoilModel/Recognizers.tsx
@@ -76,7 +76,7 @@ export const getMultiLanguagueRecognizerDialog = (
 
 export const getLuisRecognizerDialogs = (target: string, luFiles: LuFile[]) => {
   return luFiles
-    .filter((item) => !item.empty && getBaseName(item.id) === target)
+    .filter((item) => getBaseName(item.id) === target)
     .map((item) => ({ id: `${item.id}.lu.dialog`, content: LuisRecognizerTemplate(target, item.id) }));
 };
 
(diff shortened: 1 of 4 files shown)


boydc2014 (changes requested):
image

I saw this when i open a echo bot and select orchestrator and click start. 

I think you need to filter the empty lu, seems like orchestrator doesn't support empty input.

lei9444:
> image
> 
> I saw this when i open a echo bot and select orchestrator and click start.
> 
> I think you need to filter the empty lu, seems like orchestrator doesn't support empty input.

add a filter to fix this issue now

boydc2014 (approved):

PR #5196, opened 2020-12-07
feat: add extension settings

### LINKED ISSUE #3995: Users can configure extension settings
From the extensions page, a user can edit extension settings via a json editor.

The UX will be modeled by the experience in VS Code. 

Extensions can provide a JSON schema describing their settings interface and Composer will add a key into the settings object with the extension name.

Example:
Extension name: 
my-cool-extension

Settings Schema:
```json
{
  "type": "object",
  "properties": {
    "someSetting": {
      "type": "string",
      "default": "foo"
    }
  }
}
```

Will produce settings object like:
```json
{
  // other settings
  "my-cool-extension": {
    "someSetting": "foo"
  }
}
```

When installing a new extension, the default settings should be automatically added to the settings file.

a-b-r-o-w-n:
## Description

Adds api for reading and writing to extension settings file as well as ability for extensions to configure their settings via a JSON schema definition.

## Task Item

closes #3995

## Screenshots

Extension package.json:
```json
{
  "composer": {
    "configuration": {
      "sources": {
        "description": "Array of library sources.",
        "type": "array",
        "items": {
          "type": "string"
        },
        "default": [
          "https://raw.githubusercontent.com/microsoft/botframework-components/main/experimental/feeds/components.json"
        ]
      }
    }
  }
}
```

Usage:
```ts
export default function initialize(composer) {
  const packageSources = (composer.settings.sources as string[]) || [
    `https://raw.githubusercontent.com/microsoft/botframework-components/main/experimental/feeds/components.json`,
  ];
}
```

Settings editor:
Settings editor gif


Composer/packages/types/src/extension.ts
```diff
@@ -38,6 +38,14 @@ export type ExtensionBundle = {
   path: string;
 };
 
+export type ExtensionConfigurationSchema = {
```
hatpick, line 41: nit: replace all [key]: value with Record

hatpick (approved):

hatpick (approved):

PR #5213, opened 2020-12-08
fix: Allowed Callers list instantiated to an empty string array.

### LINKED ISSUE #5165: Update allowedCallers behavior when working locally
We currently expose all skills when built locally with allowedCallers ['*'], which is fine for local development, but places the security burden on users to manage this when deployed.

We can still avoid requiring Channel Registration & AppId/Password via some recent undocumented behavior of the SDK that will bypass allowedCallers entirely when running locally.

This change is to:

P0 - update allowedCallers to not be ['*'], but [], or some other model that will allow skills to communicate locally but still require appropriate security measures when deployed
P1 - take a UX item to cross-reference any channel-registered Skill Appid's of the current bot project into the allowedCallers array.

### LINKED ISSUE #5193: Use relative paths for bot projects

Currently, adding new or existing skills to a bot project uses absolute paths rather than relative paths. This will be problematic in a team environment where developers organize their files differently.

image

I'd like to have the child projects be referenced using a relative path to the root project, for example

image

I manually edited the botproj file based on the screenshot above and was able to get this to work but it would be good if this would work out of the box.

## Additional context

From the UI perspective, it is not clear which one is the "root" bot and which ones are the child ones. It would be good to use a different icon or some UI hint that would let me know which one is the one that has the references. 

This is should also be clear in the open bot window so I can clearly see which one is the root and which ones is are the child ones:

In the screenshot below, ComplexComposerHostBot is the one with the external references and the ones above are the skills but it is not obvious from looking at it.

image

Consider introducing a botsln file (sort of like Visual Studio files) that allows you to organize projects in a solution without having to modify the botproj files.

srinaath:
## Description

1. Fixes an edge case where absolute paths are saved in the botproject file
2. allowedCallersList is now set to an empty array instead of *.

## Task Item
Fixes #5193 
Fixes #5165


hibrenda:
Verified following scenarios. all passed.
- local skill with empty array[] (call success)
- remote skill with empty array[] (expect 401 error) 
- remote skill with botid setting in array["botid"] (call success)

hibrenda:
by the way, old bot still keep old format. only new created bot with this updated skill settings.

luhan2017 (approved):

boydc2014 (approved):

srinaath:
@hibrenda Thanks for testing the scenarios.

srinaath:
@clearab @gabog The allowedCallers list is now set to an empty array by default

PR #5215, opened 2020-12-08
fix: highlight selected item in LGPage and LUPage

### LINKED ISSUE #5205: Form dialogs: in LG all up page, you can select a property under the tree, but the property does not show as selected
## Version

v1.2.1-nightly.192397.d88f6af

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Create a form, generate the dialog.
2. Go to the Bot Responses page.
3. Select the dialog you just generated:
image

4. Then select an property. The content to the right changes, but the tree does not: 
image

## Expected behavior

The example above, "Cheese" should appear selected instead of "sandwich".

hatpick:
## Description

This PR fixes the highlight selection in ProjectTree within Lg and Lu pages.

## Task Item
closes #5205


Composer/packages/client/src/components/Page.tsx
```diff
@@ -101,6 +101,7 @@ type IPageProps = {
   projectId?: string;
   skillId?: string;
   dialogId?: string;
+  fileId?: string;
```
GeoffCoxMSFT, line 104: I debated if it shouldn't just be fileId all the way down rather than having luFileId and lgFileId.  What do you think?

GeoffCoxMSFT (approved):

Composer/packages/client/src/components/Page.tsx
hatpick: it's already overloaded by props, kept it at minimum, it's called "Page" (reusable) so shouldn't have any knowledge about projectId, skillId, dialogId, lg/lu ids imho.

PR #5223, opened 2020-12-08
chore: set USER_AGENT to environment variables before luis and qna build

lei9444:
## Description
This PR is based on the https://github.com/microsoft/botframework-cli/pull/1070
We will use nightly-build to replace the @bfcomposer/bu-lu.

## Task Item
#minor

## Screenshots
image
image


Composer/packages/server/src/constants.ts
```diff
@@ -18,3 +18,5 @@ export enum ClaimNames {
 export const APPINSIGHTS_INSTRUMENTATIONKEY = process.env.APPINSIGHTS_INSTRUMENTATIONKEY;
 
 export const piiProperties = ['sessionId', 'userId', 'projectId'];
+
+export const COMPOSER_VERSION = '1.2.0';
```
boydc2014, line 22: let's take a look as a best effort to unify the version in one place if possible.

lei9444: I can create a new PR to fix this

boydc2014 (approved):

PR #5243, opened 2020-12-08
fix: add LUIS endpoint to setting page

### LINKED ISSUE #5190: Need to include authoring key and region under external services
## Describe the bug
In the bot settings editor, it only shows LUIS Key and LUIS region. However, for any LUIS app used in production, the authoring key does not provide enough quota, so there should always be an endpoint key and region, as well as an authoring key and region, to avoid users hitting 403 errors and not being able to identify the source:

image

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

beyackle:
## Description

This adds a new field to the Bot Settings pages for both root and skill. 

## Task Item

closes #5190

## Screenshots

image
image


Composer/packages/client/src/pages/botProject/SkillBotExternalService.tsx
```diff
@@ -197,19 +207,34 @@ export const SkillBotExternalService: React.FC<SkillBotExternalServiceProps> = (
...
+        <div ref={luisKeyFieldRef}>
+          <TextFieldWithCustomButton
+            ariaLabel={formatMessage('LUIS endpoint key')}
+            buttonText={formatMessage('Use custom LUIS endpoint key')}
+            errorMessage={!rootLuisEndpointKey ? formatMessage('Root Bot LUIS endpoint key is empty') : ''}
+            id={'luisEndpointKey'}
+            label={formatMessage('LUIS endpoint key')}
+            placeholder={formatMessage('Enter LUIS endpoint key')}
+            placeholderOnDisable={rootLuisEndpointKey}
+            required={isLUISKeyNeeded}
```
srinaath, line 232: This key should be not be a required key. Authoring keys are the key we usually have. I think this key is only to increase LUIS deployment limits

beyackle: Thanks for catching that.

cwhitten (approved):

PR #5251, opened 2020-12-09
fix: DetailsList hijacks EditableField navigation keys such as arrows, home/end, page up/down and Mac combos

### LINKED ISSUE #5207: Bot Response page, default view is difficult to edit template with multiple lines
## Version

v1.2.1-nightly.192397.d88f6af

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Go to Bot Responses.
2. Find a template that has multiple lines (either multiple response variations or a multi-line response), and select the responses:
image

3. If you try to use the arrow keys to move up or down a line, you select the previous or next template instead.

Conversely, if you try to add newlines to a single-line template, it deselects the text input. 
To get around this, you can edit in the "Show code" view:
image

but then if you switch back to the default view, the responses all appear on one line:
image

## Expected behavior

In the default view, when a response template is selected, I'd expect to be able to use arrow keys to navigate within the field, like any text input. I'd also expect to be able use the Enter key to add new lines.

hatpick:
## Description

DetailsList internal FocusZone hijacks the key events from TextBoxes within to change the focus and that's making text editing cumbersome.

## Task Item

fixes #5207

## Screenshots

EditableField

### DIFF
--- a/Composer/packages/client/src/components/EditableField.tsx
+++ b/Composer/packages/client/src/components/EditableField.tsx
@@ -10,6 +10,9 @@ import { IconButton } from 'office-ui-fabric-react/lib/Button';
 import { IIconProps } from 'office-ui-fabric-react/lib/Icon';
 
 import { FieldConfig, useForm } from '../hooks/useForm';
+
+const allowedNavigationKeys = ['ArrowDown', 'ArrowUp', 'ArrowLeft', 'ArrowRight', 'PageDown', 'PageUp', 'Home', 'End'];
+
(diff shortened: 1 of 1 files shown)


hatpick:
@zhixzhan FYI

hatpick:
Closing in favor of #5220 (added commits there)

PR #5256, opened 2020-12-09
fix: Only do down sampling for interruption intent

### LINKED ISSUE #5265: Downsampling adjustments
This ticket represents turning off downsampling for all intents, and instead only runs it on the interruption intents when they exist.

This ticket also addresses a regression where LU models were published when there has been no change to them.

lei9444:
### issue
Today we will do down sampling with two config(maxImbalanceRatio, maxUtteranceAllowed) after cross train.
The maxImbalanceRatio is used to balance the ratio of utterances' number from different intent.
The maxUtteranceAllowed is used to downsize all utterances‘ number.

This design will check all intents and may down size some important intents created by user. 
And our initial goal is to balance the interruption intent generated by the cross train. 

### In this PR
1. Only do down sampling for interruption intent
2. set ratio to -1 by default, We don't do down sampling when ratio is -1.
3. remove maxUtteranceAllowed, the luis service will do this by itself.

## Task Item
closes #5265


boydc2014 (approved):

boydc2014 (changes requested):
Let's hold on to take this until we figure out the how to avoid the republishing.

boydc2014 (approved):

PR #5259, opened 2020-12-09
fix: add multi-bot nav to the package library

benbrown:
## Description

This adds local (simplified) implementation of the multi-bot nav to the package library.

#minor

## Screenshots

Screen Shot 2020-12-09 at 12 25 09 PM


beyackle (approved):
LGTM, and it's passing unit tests now, so it's building properly.

PR #5268, opened 2020-12-09

GeoffCoxMSFT:


beyackle (approved):
LGTM. I had to verify that these aren't ICU strings (just so we don't run into apostrophe problems again) and they're not; they just get put into a new bot on creation as they are. So I think this is good to do.

PR #5290, opened 2020-12-10
fix: dismiss open bot loading screen

tdurnford:
## Description
Moved `set(botOpeningState, false);` to the `finally` block in the `try/catch` so the loading screen will always be dismissed.

## Task Item
#minor

### DIFF
--- a/Composer/packages/server/src/services/telemetry.ts
+++ b/Composer/packages/server/src/services/telemetry.ts
@@ -69,7 +69,7 @@ if (instrumentationKey) {
 
       // remove PII
       for (const property of piiProperties) {
-        if (data.baseData.properties[property]) {
+        if (typeof data.baseData.properties[property] !== 'undefined') {
           delete data.baseData.properties[property];
         }
       }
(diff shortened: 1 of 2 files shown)


a-b-r-o-w-n (approved):

Composer/packages/server/src/services/telemetry.ts
```diff
@@ -69,7 +69,7 @@ if (instrumentationKey) {
 
       // remove PII
       for (const property of piiProperties) {
-        if (data.baseData.properties[property]) {
+        if (typeof data.baseData.properties[property] !== 'undefined') {
```
hatpick, line 72: general comment: you can just compare with ```undefined``` itself instead of comparing the typeof with a string

GeoffCoxMSFT, line 72: This could be if (data.baseData.properties[property] !== undefined)

The only typeof that will equal 'undefined' is an object set to undefined.

hatpick (approved):

PR #5291, opened 2020-12-10
chore: Remove onnxruntime.pdb from Publishing

taicchoumsft:
## Description
As of today, onnxruntime still does not properly separate out their large debug symbols [(see here)] (https://github.com/microsoft/onnxruntime/issues/2742).   Orchestrator depends on this lib.

Using flags like `CopyOutputSymbolsToPublishDirectory` and `debug=None` [(see here)](https://github.com/dotnet/sdk/issues/5575) don't work here - they work to strip out generated pdbs from the csprojs that use them, but do not strip out dependency-of-dependencies' pdbs. 

We perform a workaround here by zeroing out the `onnxruntime.pdb`, instead of deleting it. Deleting it will cause issues when the `self-contained`  deploy option is selected like it is in Composer, since the deps will be written into a json file and checked for existence before the app is booted. Credit to @hanabi1224 for this idea in the link above.

Saves ~20MB on upload size after zipping.  Tested that bot deploys and boots to Azure WebApp.

Affects only publishing, local debugging symbols are still present in the build `OutDir`. 

## Task Item
#minor

### DIFF
--- a/Composer/packages/client/config/env.js
+++ b/Composer/packages/client/config/env.js
@@ -88,7 +88,7 @@ function getClientEnvironment(publicUrl) {
         PUBLIC_URL: publicUrl,
         GIT_SHA: getGitSha().toString().replace('\n', ''),
         SDK_PACKAGE_VERSION: '4.11.0', // TODO: change this when Composer supports custom schema/custom runtime
-        COMPOSER_VERSION: '1.3.0-rc2',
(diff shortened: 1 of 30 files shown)


tsuwandy (approved):
Great job Tai, simplest fix!

boydc2014:
@luhan2017 can you help verify this?

boydc2014 (approved):

PR #5299, opened 2020-12-10
fix: generate responses

### LINKED ISSUE #5295: Composer server should return error information for form dialog generation
## Describe the bug

The generation call is always returning success even when there are problems reported by bf-generate.

## Version

main branch, 2020-12-10

## Browser

- [ ] Electron distribution
- [ X] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [ X] Windows
- [ ] Ubuntu

## To Reproduce

N/A

## Expected behavior

Return errors from the server.

GeoffCoxMSFT:
## Description

- updated to track errors during and success of generate call
- return reduces payload for generate and deleteDialog (project ID rather than entire project)
- update to add logs to form dialog error state

## Task Item

Closes #5295 

## Screenshots

### DIFF
--- a/Composer/packages/server/src/controllers/formDialog.ts
+++ b/Composer/packages/server/src/controllers/formDialog.ts
@@ -71,7 +71,9 @@ const generate = async (req: Request, res: Response) => {
 
     if (!result.success) {
       res.status(500).json({ id: projectId, errors: result.errors });
+      return;
     }
+
     res.status(200).json({ id: projectId });
   } else {
     res.status(404).json({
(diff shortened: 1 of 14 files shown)


Composer/packages/server/src/controllers/formDialog.ts
```diff
@@ -67,9 +67,12 @@ const generate = async (req: Request, res: Response) => {
 
     const templatesDirs = await getTemplateDirs();
 
-    await currentProject.generateDialog(name, templatesDirs);
-    const updatedProject = await BotProjectService.getProjectById(projectId, user);
-    res.status(200).json({ id: projectId, ...updatedProject.getProject() });
+    const result = await currentProject.generateDialog(name, templatesDirs);
+
+    if (!result.success) {
+      res.status(500).json({ id: projectId, errors: result.errors });
```
hatpick, line 73: I think you need to return here.

hatpick (changes requested):

hatpick (approved):

PR #5305, opened 2020-12-11
fix: prevent create root dialog (PVA scenario)

### LINKED ISSUE #5304: When add a new dialog with the same name as root dialog, publish PVA bot error
## Describe the bug

I am pulling a PVA bot which is called test.
Then, I add a dialog in the bot also called test. 

I will get an error for publishing, because now there are two dialogs placed in the root folder:
PowerVirtualAgentRoot.dialog
test.dialog

---
{
  "code": "InvalidDialogLocation",
  "range": null,
  "severity": "Error",
  "source": "PowerVirtualAgentRoot.dialog",
  "message": "The dialog location 'PowerVirtualAgentRoot.dialog' is invalid. All dialogs except the main dialog should be in 'dialogs' folder."
}
---

Though this is PVA specific. We should not allow to create a dialog with the same name as root dialog. 

## Browser

- [ ] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

zhixzhan:
## Description

PVA bot's root dialog has different name with bot name, when create a botName dialog, current check will pass because there is no duplicate dialog name. that mess up PVA bot structure.

The solution add a check prevent root dialog (bot name dialog). that both work for normal composer bot and PVA bot.

## Task Item
close #5304 

## Screenshots
image

### DIFF
--- a/Composer/packages/client/src/pages/design/createDialogModal.tsx
+++ b/Composer/packages/client/src/pages/design/createDialogModal.tsx
@@ -14,7 +14,7 @@ import { DialogWrapper, DialogTypes } from '@bfc/ui-shared';
 import { DialogCreationCopy } from '../../constants';
 import { StorageFolder } from '../../recoilModel/types';
 import { FieldConfig, useForm } from '../../hooks/useForm';
-import { actionsSeedState, schemasState, validateDialogsSelectorFamily } from '../../recoilModel';
+import { actionsSeedState, botDisplayNameState, schemasState, validateDialogsSelectorFamily } from '../../recoilModel';
 import TelemetryClient from '../../telemetry/TelemetryClient';
 
 import { name, description, styles as wizardStyles } from './styles';
@@ -38,8 +38,8 @@ export const CreateDialogModal: React.FC<CreateDialogModalProps> = (props) => {
 
   const schemas = useRecoilValue(schemasState(projectId));
   const dialogs = useRecoilValue(validateDialogsSelectorFamily(projectId));
+  const botName = useRecoilValue(botDisplayNameState(projectId));
   const actionsSeed = useRecoilValue(actionsSeedState(projectId));
-
   const { shellApi, ...shellData } = useShellApi();
   const { defaultRecognizer } = useRecognizerConfig();
 
@@ -55,6 +55,10 @@ export const CreateDialogModal: React.FC<CreateDialogModalProps> = (props) => {
 
         if (dialogs.some((dialog) => dialog.id.toLowerCase() === value.toLowerCase())) {
           return formatMessage('Duplicate dialog name');
+        }
+
+        if (botName.toLowerCase() === value.toLowerCase()) {
+          return formatMessage('Duplicate root dialog name');
         }
       },
       defaultValue: '',


boydc2014:
@zhixzhan this check make sense to me. But why this dialog is somehow created in top level previously, is that another issue?

zhixzhan:
> @zhixzhan this check make sense to me. But why this dialog is somehow created in top level previously, is that another issue?

The file path of dialogs is defined in `botStructrure.ts`. 
```
const BotStructureTemplate = {
  entry: '${BOTNAME}.dialog',
  lg: 'language-generation/${LOCALE}/${BOTNAME}.${LOCALE}.lg',
  lu: 'language-understanding/${LOCALE}/${BOTNAME}.${LOCALE}.lu',
  qna: 'knowledge-base/en-us/${BOTNAME}.en-us.qna',
  sourceQnA: 'knowledge-base/source/${FILENAME}.source.qna',
  dialogSchema: '${BOTNAME}.dialog.schema',
  schema: '${FILENAME}',
  settings: 'settings/${FILENAME}',
  common: {
    lg: 'language-generation/${LOCALE}/common.${LOCALE}.lg',
  },
  dialogs: {
    entry: 'dialogs/${DIALOGNAME}/${DIALOGNAME}.dialog',
    ...
```
When create a dialog or reformat a bot, all file path is calculated by `defaultFilePath` method:
https://github.com/microsoft/BotFramework-Composer/blob/8446514076d7eb17d05739cfd2945d6bada2ad10/Composer/packages/server/src/models/bot/botStructure.ts#L188-L191

In this issue, the created dialog name  just hit root dialog. so it was placed in top level (root path).

boydc2014:
> > @zhixzhan this check make sense to me. But why this dialog is somehow created in top level previously, is that another issue?
> 
> @boydc2014 The file path of dialogs is defined in `botStructrure.ts`.
> 
> ```
> const BotStructureTemplate = {
>   entry: '${BOTNAME}.dialog',
>   lg: 'language-generation/${LOCALE}/${BOTNAME}.${LOCALE}.lg',
>   lu: 'language-understanding/${LOCALE}/${BOTNAME}.${LOCALE}.lu',
>   qna: 'knowledge-base/en-us/${BOTNAME}.en-us.qna',
>   sourceQnA: 'knowledge-base/source/${FILENAME}.source.qna',
>   dialogSchema: '${BOTNAME}.dialog.schema',
>   schema: '${FILENAME}',
>   settings: 'settings/${FILENAME}',
>   common: {
>     lg: 'language-generation/${LOCALE}/common.${LOCALE}.lg',
>   },
>   dialogs: {
>     entry: 'dialogs/${DIALOGNAME}/${DIALOGNAME}.dialog',
>     ...
> ```
> 
> When create a dialog or reformat a bot, all file path is calculated by `defaultFilePath` method:
> https://github.com/microsoft/BotFramework-Composer/blob/8446514076d7eb17d05739cfd2945d6bada2ad10/Composer/packages/server/src/models/bot/botStructure.ts#L188-L191
> 
> In this issue, the created dialog name just hit root dialog. so it was placed in top level (root path).

thanks, get it

boydc2014 (approved):

PR #5307, opened 2020-12-11
fix: no need to update script if script folder is missing in bot project

liweitian:
## Description
In PVA scenario, if scripts is missing in botproject folder, we do not need to write assets scripts into bot project. So update script is hidden in this scenario.
## Task Item
#minor


boydc2014:
This looks good to me. But i'm not sure we need this in 1.3. @tonyanziano can you help confirm if for PVA, we don't have to show the update scripts button, right?

a-b-r-o-w-n (approved):

PR #5311, opened 2020-12-11
fix: js runtime publish to azure functions

### LINKED ISSUE #5300: publishing JS runtime to azure functions results in a misconfiguration
## Describe the bug

When I publish a Javascript runtime to the Azure Functions target, I get an error in Azure:

Screen Shot 2020-12-10 at 5 08 37 PM

This may be related to the `FUNCTIONS_WORKER_RUNTIME` setting, which is always set to `dotnet`

Screen Shot 2020-12-10 at 5 21 40 PM

## Version

`wenyluo/provision` branch

## To Reproduce

1. Create an azure functions environment
2. Eject runtime and select JS
3. Publish to azure functions
4. Navigate to the azure portal and look at the "overview" tab - you will see an error.

## Expected Behavior

The publishing tool should set the necessary configuration to run the language I am using.

VanyLaw:
## Description
Fix the bug about js runtime publish to azure functions.

Solution: 
1. remove lock file, because some packages link from myget were deleted.
2. nodejs runtime folder structure is different from dotnet, have no need to split into azurewebapp folder and functions folder.
3. some environment value need to be set in azure service after publishing, so that service will do the `npm install` automatically.

## Task Item

close #5300


luhan2017:
I can make the JS runtime deploy to azure function by the following steps, we should fix 2 and 3 in this PR, but we are still trying to investigate how to do 1 in the deployment process. 

1. go to Azure Function Configuration Page:
Set: FUNCTIONS_WORKER_RUNTIME=node
Set: WEBSITE_NODE_DEFAULT_VERSION=~14 (I think both ~12 ~13 should be fine, but we need to set it, otherwise it will use a very low version)

2. add local.settings.json in node runtime folder
```
{
  "IsEncrypted": false,
  "Values": {
    "FUNCTIONS_WORKER_RUNTIME": "node"
  }
}
```

3. Fix the folder issue and process.env.runtime_environment issue
somwhow the process.env.runtime_environment is not set correctly. 
```
export const getProjectRoot = (): string => {
  // get the root folder according to environment
  if (process.env.runtime_environment === 'production') {
    return path.join(__dirname, '../../azurefunctions/ComposerDialogs');
  } else {
    return path.join(__dirname, '../../../');
  }
```
};

PR #5313, opened 2020-12-11
fix: runtime throw exception when there is no cosmosdb config

luhan2017:
## Description
This PR include 3 fixes:
1. This is a fix to backward compatible issue with 1.0 created bot, which don't have cosmosdb config but our latest runtime code assume there is always one cosmosdb section and throw exception if there isn't. Fixing by loosing the constraints here.
2. remove the myget package source in Nuget.config
3. use logger instead of this.logger to fix the exception when publishing multiple lu files which triggers the retry logic. 

#minor

### DIFF
--- a/extensions/azurePublish/src/luisAndQnA.ts
+++ b/extensions/azurePublish/src/luisAndQnA.ts
@@ -112,7 +112,7 @@ export async function publishLuisToPrediction(
         break;
       } catch (err) {
         if (retryCount < 1) {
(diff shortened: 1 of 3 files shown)


boydc2014 (dismissed):

boydc2014 (approved):

boydc2014:
CI is unstable because of myget, local run of this PR looks good to me
image

PR #5315, opened 2020-12-11
fix: make all resource names unique

### LINKED ISSUE #5479: Deployment script is not able to create multiple environments with the same name
## Describe the bug

If you'd like to deploy two resource groups with the script using the same name, but different environments, the second deployment would fail with and error.

This happens because both runs of the script would try to create a `Microsoft.BotService/botServices` resource with the same name, which of course would fail for the second try.

#### 1st deployment

```
node .\provisionComposer.js `
   --subscriptionId ***** `
   --appId ***** `
   --appPassword ***** `
   --location westeurope `
   --name my-composer `
   --environment dev
```

#### 2nd deployment

```
node .\provisionComposer.js `
   --subscriptionId ***** `
   --appId ***** `
   --appPassword ***** `
   --location westeurope `
   --name my-composer `
   --environment test
```

(Note the only difference is the `--environment` parameter: `dev` first, `test` second.)

The error you would get is:

```
×
{
  "error": {
    "code": "InvalidBotData",
    "message": "Bot is not valid. Errors: The bot name is already registered to another bot application..  See https://aka.ms/bot-requirements for detailed requirements."
  }
}
** Provision failed **
```

## Expected behavior

The second run should succeed. The resource names in the two environments shouldn't clash.

peterbozso:
## Description

Without this fix, if you'd like to deploy two resource groups with the script using the same name, but different environments, the second deployment would fail with and error.

With my fix, the error produced by the example in #5479 won't happen again, because the `Microsoft.BotService/botServices` resource's name would be equal to the resource group's name, so `my-bot-dev` for the first resource group and `my-bot-test` for the second. This way, you can group your resource groups by name, but still distinguish between them by environment. I believe this was the initial intent when creating this deployment script.

## Task Item

Fixes #5479


extensions/samples/assets/shared/scripts/provisionComposer.js
```diff
@@ -465,7 +465,7 @@ const create = async (
     message: `> Create App Id Success! ID: ${appId}`,
   });
 
-  const resourceGroupName = `${name}-${environment}`;
+  const resourceGroupName = (name = `${name}-${environment}`);
```
GeoffCoxMSFT, line 468: This re-assignment could lead to bugs because it is easy to miss and it changes a parameter mid-function.
Consider passing resourceGroupName rather than name to getDeploymentTemplateParam and getQnaTemplateParam instead.

peterbozso: Good point! How about now?

extensions/samples/assets/shared/scripts/provisionComposer.js
```diff
@@ -451,7 +453,7 @@ const create = async (
     });
 
     // create the app registration
-    const appCreated = await createApp(graphClient, name, appPassword);
+    const appCreated = await createApp(graphClient, botName, appPassword);
```
peterbozso, line 456: @GeoffCoxMSFT I caught a possible bug here, where we are creating an app registration if no app ID is provided. I haven't noticed that in my previous commit, because as you can see in the issue description, I was providing already existing app ID-s to the script when testing this fix. Now that is fixed as well and taking this into account, I think `botName` as the variable name is more descriptive for the whole function.

peterbozso:
The pipeline currently fails for this PR because it expects the PR's description to either include a reference to an issue or a #minor tag. I don't think this is something minor, because without this fix the usability of the deployment script is rather limited, so I wouldn't be comfortable with tagging this as #minor.  So should I create an issue for this and reference that here?

zidaneymar (changes requested):
I think it's not a good idea to use "botName" everywhere, just keep resourceGroupName and name variable, and just replace the botName in creating bot channel registration process.

mrivera-ms:
Thank you for this contribution, @peterbozso. Are you able to merge main to eliminate the conflicts and reduce the number of files affected?

mrivera-ms:
@peterbozso, thank you. We are unable to code review at this time.

PR #5323, opened 2020-12-14
fix: LG diagnostics mismatch by update reference in cache 

zhixzhan:
## Description

For performance consideration, LG file parseResults are cached in client. the next time do CRUD, will directly use previous parseResult do increment parse. 
1. initially we have two file (`common.lg`, `a.lg`) and it's parsed resource ( `CommonParseResource:T0`, `AParseResource:T0` )
2. when modify common.lg file, will update common.lg to `CommonParseResource:T1`.
3. when modify a.lg (referred common.lg), a.lg will use  `AParseResource:T0 ` update itself. but in `AParseResource `, it referred
common.lg is expired `CommonParseResource:T0`.  which cause mismatch error.

### Solution
At step2, not only update itself, but also update all rest lg files 's  resource reference.
 

## Task Item
#minor


lei9444:
have tested it and LGTM

boydc2014 (approved):

PR #5346, opened 2020-12-15
fix: Updated publish extensions to use updated APIs

tonyanziano:
## Description

Upon merging the provisioning work, there were some changes made to the publishing extension API that were not translated into all the publishing extensions, breaking some of them.

This PR rewrites the `pvaPublish` and `sample-ui-plugin` extensions to use those updated APIs.

Gets rid of the `setPublishConfig()` and `setConfigIsValid()` APIs.

#minor

## Screenshots

**`pvaPublish`**

pva1
pva2
pva3

**`sample-ui-plugin`**

sample1
sample2

PR #5348, opened 2020-12-16
fix: Fixed issue where an env object was being stored in state instead of …

### LINKED ISSUE #5270: Add a Publish Profile > Publish Bot to PVA > No Bots Found
## Describe the bug

In our default environment (and only environment) in our test tenant, I am trying out the Add a Publish Profile in Bot Composer and attempting to Publish my bot to PVA. We have 4 published bots and yet the message I get is "no bots found." Does anyone know why the Publish can't find the bots (it did find the default environment)?
Add-Publish-Profile

## Version

Latest: 1.21

## Browser

- [ ] Chrome

## OS

- [ ] Windows

## To Reproduce

In Bot Composer, go to publish and choose "Add new profile"
Give profile a name, choose destination type as "Publish Bot to PVA" > Login with my test tenant account > message back after it finds the default environment is "no bots found." There are 4 published PVA bots here. 

## Expected behavior

I expected it would publish to PVA and find the bots.

tonyanziano:
…an env ID string.

## Description

There was a [line in the PVA publish extension code](https://github.com/microsoft/BotFramework-Composer/blob/main/extensions/pvaPublish/src/ui/pvaDialog.tsx#L90) that was storing the entire PVA environment object in state instead of just the environment ID.

This was then causing the following HTTP request to PVA to target the wrong environment endpoint because it was encoding the stringified environment object instead of just the environment ID in the URL:

**Before fix:**

`https://powerva.microsoft.com/api/botmanagement/v1/environments/%5Bobject%20Object%5D/bots`

**After fix:**

`https://powerva.microsoft.com/api/botmanagement/v1/environments/839eace6-59ab-4243-97ec-a5b8fcc104e4/bots`

This was preventing users with only one selectable environment from showing their list of available bots within the extension.

## Task Item

closes #5270

### DIFF
--- a/extensions/pvaPublish/src/ui/pvaDialog.tsx
+++ b/extensions/pvaPublish/src/ui/pvaDialog.tsx
@@ -87,7 +87,7 @@ export const PVADialog: FC = () => {
         setFetchingEnvironments(false);
         setEnvs(envs);
         if (envs && envs.length) {
-          setEnv(envs[0]);
+          setEnv(envs[0].id);
         }
       };
       fetchEnvs();


luhan2017:
@tonyanziano, in composer v1.3.0, we can't add a new PVA publish profile, it seems the PVA UX extension is not hooked up in  the new extension structure. Could you please check if it is a bug or by-design? I am not familiar with the experience. 

image

tonyanziano:
@luhan2017 this was due to some changes in the provision branch that were not transferred over. That was fixed in another PR I submitted yesterday: #5346

srinaath (approved):

PR #5353, opened 2020-12-16
chore: Concurrency safe state manipulation for lg/lu/qna/multilang

### LINKED ISSUE #5339: Concurrency safe state manipulation for lgFiles/luFiles/QnAFiles/multilang
### The issue 

In current dispatchers methods, some of them, state manipulations are not concurrency safe. For example :
https://github.com/microsoft/BotFramework-Composer/blob/dc4bb80d9ee6a92634a5d227d58d864cb12de05a/Composer/packages/client/src/recoilModel/dispatchers/lu.ts#L143-L150

if a concurrency call incoming ( and also mutated 'luFilesState' ) during line-148, line-149 executing, The the line-150  set will overwrite the concurrency mutation. 

It caused bug issues 

- LG Editor keeps dropping changes https://github.com/microsoft/BotFramework-Composer/issues/4988

we opened some PR fix this type bug

- fix: LG Editor dropping changes #4997
- fix: Lg content missing when switch the page #5032

### Solution

This is doable by use setter function provide by recoil.  just like bellow
https://github.com/microsoft/BotFramework-Composer/blob/dc4bb80d9ee6a92634a5d227d58d864cb12de05a/Composer/packages/client/src/recoilModel/dispatchers/lg.ts#L268-L270

In setter function, the state 'lgFiles' is latest, and only touched file is the updated file..

### Todos
Just like the safe state operation demonstrate above, all state needs to check and review see it is concurrency safe, especially for <List>[] type state. make sure the mutation would not overwrite others.

zhixzhan:
## Description
As describe in #5339. In current dispatchers methods, some of them, state manipulations are not concurrency safe.
If a concurrency call incoming also mutate same state variable, last time long time parsing will overwrite the concurrency mutation.

This PR reviewed those code again, change to use updater callback function make fine-gained operations.

## Task Item
close #5339

### DIFF
--- a/Composer/packages/client/src/recoilModel/dispatchers/multilang.ts
+++ b/Composer/packages/client/src/recoilModel/dispatchers/multilang.ts
@@ -20,7 +20,7 @@ import {
   onDelLanguageDialogCompleteState,
   showDelLanguageModalState,
   botDisplayNameState,
-  qnaFilesState,
+  // qnaFilesState,
 } from './../atoms/botState';
 
 const copyLanguageResources = (files: any[], fromLanguage: string, toLanguages: string[]): any[] => {
@@ -58,6 +58,7 @@ const deleteLanguageResources = (
 };
 
 export const multilangDispatcher = () => {
+  // When skill bot do not have the locale to be set in root bot, create it for skill bot.
   const setLocale = useRecoilCallback(
     ({ set, snapshot }: CallbackInterface) => async (locale: string, projectId: string) => {
       const botName = await snapshot.getPromise(botDisplayNameState(projectId));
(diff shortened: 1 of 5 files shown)


Composer/packages/client/src/recoilModel/dispatchers/multilang.ts
```diff
@@ -131,27 +133,31 @@ export const multilangDispatcher = () => {
...
-      }
+      set(lgFilesState(projectId), (prevlgFiles) => {
+        const { left: leftLgFiles } = deleteLanguageResources(prevlgFiles, languages);
+        return leftLgFiles;
+      });
+      set(luFilesState(projectId), (prevluFiles) => {
+        const { left: leftLuFiles } = deleteLanguageResources(prevluFiles, languages);
+        return leftLuFiles;
+      });
+      // set(qnaFilesState(projectId), (prevQnAFiles) => {
```
srinaath, line 147: Remove comment

srinaath (approved):

srinaath:
Looks good @zhixzhan using the updater function as a callback for atom updates.

@zhixzhan with this version of Recoil we could use the updater functions for filePersistence too.

PR #5361, opened 2020-12-17
fix: project tree show incorrect imports

### LINKED ISSUE #5347: LG navigation pane shows all markdown links as imported LG files
## Describe the bug
When an LG file has a markdown link using the `"[]()"` syntax, it is shown in the left navigation pane. 
## Version
1.3

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## Expected behavior
It should only show the imported LG files, not all the links.

## Screenshots
Here is the link in the LG template:
image

Here is what shows up in the nav:
image

zhixzhan:
## Description
Previous imports info is get from content by RegExp match, that is not accurate when LG body include markdown style text.

Fix it by get imports from LG parser side, same to LU/QnA.

## Task Item
close #5347 

## Screenshots
image


Composer/packages/lib/indexers/src/utils/luUtil.ts
```diff
@@ -104,13 +105,24 @@ export function convertLuParseResultToLuFile(id: string, resource, luFeatures: I
     convertLuDiagnostic(e, id)
   ) as Diagnostic[];
 
+  const imports = Sections.filter(({ SectionType }) => SectionType === SectionTypes.ImportSection).map(
+    ({ Path, Description }) => {
```
beyackle, line 109: Sections is just defined as `any[]`, so I can't tell if these fields actually exist. Would you mind tracking down the definition for what a "section" actually is and changing it accordingly in `packages/types/src/indexers.ts`? That's where the LuFile and LuParseResource types are defined.

beyackle (changes requested):
This looks like it should work, but I think the type structure needs more clarity.

hatpick (dismissed):
Looks good to me, thank you.

hatpick (approved):

hatpick (approved):

PR #5375, opened 2020-12-17
refactor: Refactor publish page.

alanlong9278:
## Description

Refactor publish page.
### Take out the polling logic into pollingUpdater class. So polling logic will become clearer.
- Start polling scenario:
1.  initial render, 
2. change publish target 
3. publish begin.
- Stop polling scenario：
1. Auto stop when response code is 200 || 404 || 500
2. Manual stop when unmount the page

After the class used, we can reduce `useEffect` to calculate state data. 

### Rewrite data flow
- Split recoil value botProjectData into botProjectData and botPublishHistory. This is because botProjectData is stable than publishHistory. So the split action will reduce recalculate data from recoil
- reduce state data use.  Add const data, ref data to replace state data. 
- split generatedComputedData function into `publish` Page & `botStatusList` component. `botStatusList` component can calculate owner render data from props instead of using the original props.

After that, the data flow is clearer to track data change.

## Task Item

#minor


Composer/packages/client/src/utils/publishStatusPollingUpdater.ts
```diff
@@ -0,0 +1,67 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import httpClient from './httpUtil';
+
+enum CircuitBreakerStateEnum {
```
srinaath, line 6: What is this CircuitBreakerState? Can this be a more descriptive name?

srinaath (changes requested):

Composer/packages/client/src/utils/publishStatusPollingUpdater.ts
```diff
@@ -0,0 +1,67 @@
...
+      this.status = CircuitBreakerStateEnum.Open;
+
+      const firstResponse = await httpClient.get(`/publish/${this.botProjectId}/status/${this.targetName}`);
+      onData && onData({ botProjectId: this.botProjectId, targetName: this.targetName, apiResponse: firstResponse });
+      this.timerId = window.setInterval(async () => {
+        try {
+          const response = await httpClient.get(`/publish/${this.botProjectId}/status/${this.targetName}`);
+          onData && onData({ botProjectId: this.botProjectId, targetName: this.targetName, apiResponse: response });
+        } catch (err) {
+          console.log(this.timerId);
```
srinaath, line 43: Remove logs

Composer/packages/client/src/utils/publishStatusPollingUpdater.ts
```diff
@@ -0,0 +1,67 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import httpClient from './httpUtil';
```
srinaath, line 4: It feels like you could leverage the useIntervalHook to accomplish the polling instead of building a class component for polling. Can you check the component BotRuntimeStatus.tsx?

Composer/packages/client/src/utils/publishStatusPollingUpdater.ts
```diff
@@ -0,0 +1,67 @@
...
+        try {
+          const response = await httpClient.get(`/publish/${this.botProjectId}/status/${this.targetName}`);
+          onData && onData({ botProjectId: this.botProjectId, targetName: this.targetName, apiResponse: response });
+        } catch (err) {
+          console.log(this.timerId);
+          onData && onData({ botProjectId: this.botProjectId, targetName: this.targetName, apiResponse: err.response });
+        }
+      }, this.pollingInterval);
+    } catch (err) {
+      console.log(this.timerId);
```
srinaath, line 48: remove logs

Composer/packages/client/src/utils/publishStatusPollingUpdater.ts
```diff
@@ -0,0 +1,67 @@
...
+    }
+  }
+  stop() {
+    if (this.timerId) {
+      window.clearInterval(this.timerId);
+    }
+    this.timerId = 0;
+    this.status = CircuitBreakerStateEnum.Closed;
+  }
+  restart(onData) {
```
srinaath, line 59: space between functions

Composer/packages/client/src/utils/publishStatusPollingUpdater.ts
```diff
@@ -0,0 +1,67 @@
...
+    this.stop();
+    this.start(onData);
+  }
+  beEqual(botProjectId, targetName) {
```
srinaath, line 63: more descriptive name than beEqual and also space between methods

Composer/packages/client/src/recoilModel/dispatchers/publisher.ts
```diff
@@ -220,6 +220,13 @@ export const publisherDispatcher = () => {
     }
   );
 
+  // get bot status from target publisher
+  const getPublishStatusV2 = useRecoilCallback(
```
srinaath, line 224: If its a refactor why use V2? Can we just update the original functions for getPublishStatus?

Composer/packages/client/src/pages/publish/type.ts
```diff
@@ -0,0 +1,42 @@
...
+  message?: string;
+  comment?: string;
+};
+export type IBot = {
+  id: string;
+  name: string;
+  publishTarget: string;
+};
+
+export type IBotSetting = {
```
srinaath, line 24: we are avoiding the prefix 'I'. The type can just be BotSettings.

Composer/packages/client/src/pages/publish/type.ts
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { DialogSetting, PublishResult, PublishTarget } from '@bfc/shared';
+
+import { PublishType } from '../../recoilModel/types';
+
+export type IBotStatus = {
```
srinaath, line 8: remove interface prefix

Composer/packages/client/src/pages/publish/type.ts
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) Microsoft Corporation.
```
srinaath, line 1: Remove interace 'I' prefix on all types. Its not our naming pattern.

Composer/packages/client/src/pages/publish/type.ts
```diff
@@ -0,0 +1,42 @@
...
+};
+
+export type IBotPublishType = {
+  projectId: string;
+  publishTypes: PublishType[];
+};
+
+export type IBotPublishHistory = {
+  projectId: string;
+  publishHistory: { [key: string]: PublishResult[] };
```
srinaath, line 41: use Record

Composer/packages/client/src/pages/publish/type.ts
```diff
@@ -0,0 +1,42 @@
...
+import { DialogSetting, PublishResult, PublishTarget } from '@bfc/shared';
+
+import { PublishType } from '../../recoilModel/types';
+
+export type IBotStatus = {
+  id: string;
+  name: string;
+  publishTargets?: PublishTarget[];
+  publishTarget?: string;
+  time?: string;
```
srinaath, line 13: timestamp?

Composer/packages/client/src/pages/publish/PublishToolbar.tsx
```diff
@@ -0,0 +1,63 @@
...
+export const PublishToolbar: React.FC<PublishToolbarProps> = ({ canPublish, canPull, onPublish, onPull }) => {
+  const toolbarItems: IToolbarItem[] = [
+    {
+      type: 'element',
+      align: 'left',
+      element: (
+        <ActionButton
+          data-testid="publishPage-Toolbar-Publish"
+          disabled={!canPublish}
+          styles={{ root: { fontSize: '16px' } }}
```
srinaath, line 28: FontSize.16?

Composer/packages/client/src/pages/publish/PublishStatusList.tsx
```diff
@@ -160,7 +162,7 @@ export const PublishStatusList: React.FC<IStatusListProps> = (props) => {
             allowDisabledFocus
             styles={{ root: { color: '#0078D4' } }}
```
srinaath, line 163: Check matching color in SharedColors or NeutralColors. It will be useful when we add themes to Composer

Composer/packages/client/src/pages/publish/Publish.tsx
```diff
@@ -3,357 +3,222 @@
...
-  const statusList: IBotStatus[] = [];
-  const botPublishTypesList: { projectId: string; publishTypes: PublishType[] }[] = [];
-  const publishTargetsList: { projectId: string; publishTargets: PublishTarget[] }[] = [];
+// CR: Data model. What is stable data? What is frequently changed data? Manage them separately rather than repeating them.
+const generateComputedData = (botProjectData) => {
+  const botSettingList: IBotSetting[] = [];
+  const botPublishTargetsList: IBotPublishTargets[] = [];
+  const botPublishTypesList: IBotPublishType[] = [];
+  const botList: IBot[] = [];
   botProjectData.forEach((bot) => {
```
srinaath, line 39: it looks like you can use

`botList= botProjectData.map(() =>` instead of forEach which makes u create a temporary variable

Composer/packages/client/src/pages/publish/Publish.tsx
```diff
@@ -3,357 +3,222 @@
...
-
-  // items to show in the list
-  const [selectedVersion, setSelectedVersion] = useState<PublishResult | null>(null);
+  useEffect(() => {
+    botList
+      .filter((bot) => !!bot.publishTarget && !pollingUpdaterList.some((u) => u.beEqual(bot.id, bot.publishTarget)))
+      .map((bot) => {
+        const updater = new PublishStatusPollingUpdater({ botProjectId: bot.id, targetName: bot.publishTarget });
+        updater.start(updateData);
+        pollingUpdaterList.push(updater);
```
srinaath, line 89: Could you explain this section. I'm slighltly lost. On every useEffect you create a new pollingUpdater object. Why do we do this?

Composer/packages/client/src/pages/publish/BotStatusList.tsx
```diff
@@ -7,63 +7,81 @@ import moment from 'moment';
...
   onRollbackClick: (selectedVersion: PublishResult, item: IBotStatus) => void;
 };
 
+const generateBotList = (
+  botList: IBot[],
+  botPublishTargetsList: IBotPublishTargets[],
+  botPublishHistoryList: IBotPublishHistory[]
+): IBotStatus[] => {
+  const bots: IBotStatus[] = [];
+  botList.forEach((bot) => {
```
srinaath, line 45: use bots = botList.map

yeze322:
@srinaath Alan was in vacation, he will be back next week (Dec 28)

Composer/packages/client/src/utils/publishStatusPollingUpdater.ts
alanlong9278: rename `PollingStateEnum`

Composer/packages/client/src/recoilModel/dispatchers/publisher.ts
alanlong9278: Because `BotRunTimeStatus` use the original function. The refactor move the http call to the polling, but `BotRunTimeStatus` still need the http call in the function. So I create a V2 function for refactor

Composer/packages/client/src/pages/publish/Publish.tsx
alanlong9278: Using `forEach` is because we need assign values to `botList`, `botSettingList`, `botPublishTargetsList` and `botPublishTypesList` in one loop

Composer/packages/client/src/pages/publish/Publish.tsx
alanlong9278: There is a `pollingUpdater` List, so we create a new updater after filtering the existing updater. 
`
      .filter(
        (bot) => !!bot.publishTarget && !pollingUpdaterList.some((u) => u.isSameUpdater(bot.id, bot.publishTarget))
      )
`

Composer/packages/client/src/pages/publish/BotStatusList.tsx
alanlong9278: use map

Composer/packages/client/src/pages/publish/type.ts
alanlong9278: Using `time` is for sync with `publishResult` type

Composer/packages/client/src/utils/publishStatusPollingUpdater.ts
alanlong9278: use `isSameUpdater`

Composer/packages/client/src/utils/publishStatusPollingUpdater.ts
alanlong9278: The reason for using polling class is that we can not only clean up actions when a component destroyed, but also can stop the interval manually.  And the polling updater is a list, we can manage the updater object easier.

srinaath: I understand the reason to use polling but why not just use the useInterval hook to accomplish it

Composer/packages/client/src/pages/publish/Publish.tsx
srinaath: You can do that inside map too.

Composer/packages/client/src/pages/publish/Publish.tsx
srinaath: Can we connect offline on this. Adding, unit test cases testing the scenario would also be helpful in understanding this flow.

Composer/packages/client/src/utils/publishStatusPollingUpdater.ts
```diff
@@ -0,0 +1,72 @@
...
+enum PollingStateEnum {
+  Closed,
+  Open,
+}
+
+export const ApiStatus = {
+  Publishing: 202,
+  Success: 200,
+  Failed: 500,
+  Unknow: 404,
```
srinaath, line 15: Unkown

yeze322 (approved):

Composer/packages/client/src/pages/publish/BotStatusList.tsx
```diff
@@ -7,82 +7,55 @@ import moment from 'moment';
...
-  botPublishTypesList: { projectId: string; publishTypes: PublishType[] }[];
-  publishDisabled: boolean;
-  updateItems: (items: IBotStatus[]) => void;
-  updatePublishHistory: (items: PublishResult[], item: IBotStatus) => void;
-  updateSelectedBots: (items: IBotStatus[]) => void;
-  changePublishTarget: (PublishTarget: string, item: IBotStatus) => void;
-  onLogClick: (item: PublishResult) => void;
-  onRollbackClick: (selectedVersion: PublishResult, item: IBotStatus) => void;
-};
+export interface BotStatusListProps {
```
hatpick, line 26: use type instead of interface

Composer/packages/client/src/pages/publish/BotStatusList.tsx
```diff
@@ -7,82 +7,55 @@ import moment from 'moment';
...
-    onRollbackClick,
-  } = props;
-  const [selectedBots, setSelectedBots] = useState<IBotStatus[]>([]);
-  const [showHistoryBots, setShowHistoryBots] = useState<string[]>([]);
+  /** When set to true, disable the checkbox. */
+  disabled: boolean;
+  managePublishProfile: (skillId: string) => void;
+  checkedIds: string[];
+  onCheck: (skillIds: string[]) => void;
+  changePublishTarget: (PublishTarget: string, item: BotStatus) => void;
```
hatpick, line 35: please name this as if it's a callback. onChange....

Composer/packages/client/src/pages/publish/BotStatusList.tsx
```diff
@@ -7,82 +7,55 @@ import moment from 'moment';
...
-    changePublishTarget,
-    updateSelectedBots,
-    onLogClick,
-    onRollbackClick,
-  } = props;
-  const [selectedBots, setSelectedBots] = useState<IBotStatus[]>([]);
-  const [showHistoryBots, setShowHistoryBots] = useState<string[]>([]);
+  /** When set to true, disable the checkbox. */
+  disabled: boolean;
+  managePublishProfile: (skillId: string) => void;
```
hatpick, line 32: onManage....

Composer/packages/client/src/pages/publish/BotStatusList.tsx
```diff
@@ -7,82 +7,55 @@ import moment from 'moment';
...
-    updatePublishHistory,
-    changePublishTarget,
-    updateSelectedBots,
-    onLogClick,
-    onRollbackClick,
-  } = props;
-  const [selectedBots, setSelectedBots] = useState<IBotStatus[]>([]);
-  const [showHistoryBots, setShowHistoryBots] = useState<string[]>([]);
+  /** When set to true, disable the checkbox. */
+  disabled: boolean;
```
hatpick, line 31: disableCheckbox

Composer/packages/client/src/pages/publish/BotStatusList.tsx
```diff
@@ -7,82 +7,55 @@ import moment from 'moment';
...
+  botStatusList,
+  botPublishHistoryList,
+  disabled,
+  checkedIds,
+  onCheck,
+  managePublishProfile,
+  changePublishTarget,
+  onRollbackClick,
+}) => {
+  const [expandedBotIds, setExpandedBotIds] = useState<{ [botId: string]: boolean }>({});
```
hatpick, line 49: use type Record<string, boolean>

Composer/packages/client/src/pages/publish/BotStatusList.tsx
```diff
@@ -7,82 +7,55 @@ import moment from 'moment';
...
-  };
 
-  const publishTargetOptions = (item: IBotStatus): IDropdownOption[] => {
+  const displayedItems: BotStatus[] = useMemo(() => {
+    if (currentSort.key !== 'Bot') return botStatusList;
+    if (currentSort.descending) return botStatusList;
+    return botStatusList.slice().reverse();
+  }, [botStatusList, currentSort]);
+
+  const publishTargetOptions = (item: BotStatus): IDropdownOption[] => {
```
hatpick, line 58: either make this a useMemo or name it appropriately for a function: getPublishTargetOptions

Composer/packages/client/src/pages/publish/LogDialog.tsx
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import React from 'react';
+import { Dialog } from 'office-ui-fabric-react/lib/Dialog';
+import { TextField } from 'office-ui-fabric-react/lib/TextField';
+
+export interface LogDialogProps {
```
hatpick, line 8: use type

Composer/packages/client/src/pages/publish/LogDialog.tsx
```diff
@@ -0,0 +1,28 @@
...
+  };
+  return (
+    <Dialog
+      dialogContentProps={logDialogProps}
+      hidden={false}
+      minWidth={700}
+      modalProps={{ isBlocking: true }}
+      onDismiss={onDismiss}
+    >
+      <TextField multiline readOnly placeholder="Log Output" style={{ minHeight: 300 }} value={value} />
```
hatpick, line 25: use formatMessage in placeholder and message should be: "Log output"

Composer/packages/client/src/pages/publish/LogDialog.tsx
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import React from 'react';
+import { Dialog } from 'office-ui-fabric-react/lib/Dialog';
+import { TextField } from 'office-ui-fabric-react/lib/TextField';
+
+export interface LogDialogProps {
+  value?: string;
```
hatpick, line 9: It seems like the caller always provides a value, in fact it only shows this dialog when the value is not null, why is this optional then?

Composer/packages/client/src/pages/publish/PublishToolbar.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { jsx } from '@emotion/core';
+import React from 'react';
+import formatMessage from 'format-message';
+import { ActionButton } from 'office-ui-fabric-react/lib/Button';
+import { SharedColors } from '@uifabric/fluent-theme/lib/fluent/FluentColors';
+import { FontSizes } from '@uifabric/fluent-theme';
+
+import { Toolbar, IToolbarItem } from '../../components/Toolbar';
+
+export interface PublishToolbarProps {
```
hatpick, line 14: use type

Composer/packages/client/src/pages/publish/publishPageUtils.tsx
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { ApiStatus } from '../../utils/publishStatusPollingUpdater';
+
+import { Bot, BotStatus, BotPublishHistory, BotPropertyType } from './type';
+
+export const deleteNotificationInterval = 5000;
+
+export const generateComputedData = (botProjectData) => {
```
hatpick, line 10: what's the type of botProjectData? please add.

hatpick, line 10: what does this mean "generatedComputedData" what does this computed data represent? (use the answer to this question to name the function more appropriately)

Composer/packages/client/src/pages/publish/publishPageUtils.tsx
```diff
@@ -0,0 +1,69 @@
...
+  // fill Settings, status, publishType, publish target for bot from botProjectMeta
+  const botPropertyData: BotPropertyType = {};
+  const botList: Bot[] = [];
+  botProjectData.forEach((bot) => {
+    const botProjectId = bot.projectId;
+    const publishTargets = bot.setting ? bot.setting.publishTargets || [] : [];
```
hatpick, line 16: ```const publishTargets = bot.setting?.publishTargets || []; ```

Composer/packages/client/src/pages/publish/type.ts
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) Microsoft Corporation.
```
hatpick, line 1: Where is there a mixture of types and interfaces? please use types all across.

Composer/packages/client/src/utils/publishStatusPollingUpdater.ts
```diff
@@ -0,0 +1,88 @@
...
+  private pollingInterval = 10000;
+
+  constructor(botProjectId: string, publishTargetId: string) {
+    this.botProjectId = botProjectId;
+    this.publishTargetName = publishTargetId;
+  }
+
+  private async fetchPublishStatusData(botProjectId: string, publishTargetName: string, onData: OnDataHandler) {
+    if (typeof onData !== 'function') {
+      console.error(new Error('onData should be a function.'));
```
hatpick, line 41: why not throw this Error here? instead of logging it

Composer/packages/client/src/utils/publishStatusPollingUpdater.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    onData({
+      botProjectId,
+      targetName: publishTargetName,
+      apiResponse: response,
+    });
+  }
+
+  async start(onData: OnDataHandler) {
+    if (!this.botProjectId || !this.publishTargetName) return;
```
hatpick, line 58: I think you need to throw here and let the caller know what's missing

Composer/packages/client/src/utils/publishStatusPollingUpdater.ts
```diff
@@ -0,0 +1,88 @@
...
+    if (this.status === PollingStateEnum.Running) return;
```
hatpick, line 59: same here

Composer/packages/client/src/utils/publishStatusPollingUpdater.ts
```diff
@@ -0,0 +1,88 @@
...
+  apiResponse: any;
+}
+type OnDataHandler = (payload: OnDataPayload) => void;
+
+export class PublishStatusPollingUpdater {
+  private status: PollingStateEnum = PollingStateEnum.Stopped;
+  private timerId = 0;
+  private publishTargetName: string;
+  private botProjectId: string;
+  private pollingInterval = 10000;
```
hatpick, line 32: I think this should be configurable (add a setter)

Composer/packages/client/src/utils/publishStatusPollingUpdater.ts
```diff
@@ -0,0 +1,88 @@
...
+    this.timerId = 0;
+    this.status = PollingStateEnum.Stopped;
+  }
+
+  restart(onData: OnDataHandler) {
+    this.stop();
+    this.start(onData);
+  }
+
+  isSameUpdater(botProjectId, targetName) {
```
hatpick, line 83: add type to args

Composer/packages/client/src/utils/publishStatusPollingUpdater.ts
```diff
@@ -0,0 +1,88 @@
...
+    return this.botProjectId === botProjectId && this.publishTargetName === targetName;
+  }
+}
+
+export const pollingUpdaterList: PublishStatusPollingUpdater[] = [];
```
hatpick, line 88: what's this? and where is it used?

hatpick (changes requested):

Composer/packages/client/src/utils/publishStatusPollingUpdater.ts
alanlong9278: This is the polling updater list for store updaters. Using global value insteads of component's value is because component will re-render when getting publish status, so the component's value like `ref` or `state` will reset too.

Composer/packages/client/src/pages/publish/publishPageUtils.tsx
alanlong9278: use detial name instead of `computedData`

Composer/packages/client/src/pages/publish/BotStatusList.tsx
yeze322: `disableCheckbox` is accurate under current logic (only disable the checkbox). However, the dropdown buttons need to be disabled as well (but not implemeneted yet), so I used `disabled` here. Either is ok to me.

Composer/packages/client/src/pages/publish/BotStatusList.tsx
yeze322: In most of the practices I can find, props of a JSX is declared as an interface. Do we have any reference of interface vs type?

yeze322: One of the reason I prefer to use `interface` over `type` is it can trigger VSCode's intellisense:
image

yeze322:
About Type v.s. Interface, I'm also confused with that. Previously, I didn't think too much about that, just follow the convention that using `interface` everywhere for defining object shapes while `type` for declaring aliases.

Today I queried this topic on web, found some differences. For saving your time, summarize them here. @srinaath @alanlong9278 

### Language feature differences
1. **Type** supports
    - Mapped Types: https://www.typescriptlang.org/docs/handbook/advanced-types.html#mapped-types
    - Conditional Types: https://www.typescriptlang.org/docs/handbook/advanced-types.html#conditional-types
    - Can be indexed: https://github.com/microsoft/TypeScript/issues/15300

2. **Interface** supports
    - Declaration Merging: https://www.typescriptlang.org/docs/handbook/declaration-merging.html
    - Polymorphism via `implements`

### Official Doc comparasion
Link: http://typescriptlang.org/docs/handbook/advanced-types.html#interfaces-vs-type-aliases (they named it `type-aliases`?)
It says that:
> Because an interface more closely maps how JavaScript objects work by being open to extension, we recommend using an interface over a type alias when possible.

I think that's also the reason why many opensource boilerpolates use `interface` everywhere.

On the opposite, in several blog posts, there are some people prefer type over interface. But from my perspective, their explaination is not quite solid.

### Personal preference

I usually prefer to use `interface` with a very straightforward but stong reason- it can trigger IDE's intellisense while `type` doesn't 💯! Here is an example

image

Composer/packages/client/src/pages/publish/BotStatusList.tsx
alanlong9278: rename

Composer/packages/client/src/pages/publish/BotStatusList.tsx
alanlong9278: rename

Composer/packages/client/src/pages/publish/BotStatusList.tsx
alanlong9278: rename

Composer/packages/client/src/pages/publish/BotStatusList.tsx
alanlong9278: fixed

Composer/packages/client/src/pages/publish/BotStatusList.tsx
alanlong9278: use `getPublishTargetOptions`

Composer/packages/client/src/pages/publish/LogDialog.tsx
alanlong9278: use formatMessage

Composer/packages/client/src/pages/publish/LogDialog.tsx
alanlong9278: cancel `?`

Composer/packages/client/src/pages/publish/publishPageUtils.tsx
alanlong9278: Add type

Composer/packages/client/src/pages/publish/publishPageUtils.tsx
alanlong9278: refactor it

Composer/packages/client/src/pages/publish/type.ts
alanlong9278: rename

Composer/packages/client/src/utils/publishStatusPollingUpdater.ts
alanlong9278: move the log in start function

Composer/packages/client/src/utils/publishStatusPollingUpdater.ts
alanlong9278: Add log

Composer/packages/client/src/utils/publishStatusPollingUpdater.ts
alanlong9278: Add args in construtor function

Composer/packages/client/src/utils/publishStatusPollingUpdater.ts
alanlong9278: Add types

Composer/packages/client/src/utils/publishStatusPollingUpdater.ts
hatpick: I dont see it being used anywhere

srinaath:
@yeze322  great read up on Types and interfaces. I was always preferring types over interfaces for things I never imagined to be extended or dont want to be extended. I do not see the intellisense issue you have pointed out. Maybe im missing something . I did try the cases of renaming a property from a different file and the refactor propogated the change into the Type and I see our regular intellisense kick in on Type.

I do agree interface depicts a more Prototypical inheritance pattern Javascript is basing itself. But, for things that we dont want to be extended it makes sense to stick to Type. I do not see any memory advantages of using Type vs Interface. So, it maybe fine to go either ways but to be consistent throughout the codebase.

hatpick:
Adding to what @srinaath has said, I think Interfaces and their exclusive features (you mentioned some above) belong to OOP, while in react/typescript land a big portion of the code is following a functional programming paradigm with composition. My vote is also for being consistent across the codebase.

yeze322:
@srinaath @hatpick Yes, I agree with you that consistency is of most importantance.
Out of the range of this PR, I think we add an eslint rule to limit usages? Either interface or type will be fine

Composer/packages/client/src/utils/publishStatusPollingUpdater.ts
alanlong9278: It is used in `publish.tsx`.
```
      // Start updaters
      botList
        .filter(
          (bot) => !!bot.publishTarget && !pollingUpdaterList.some((u) => u.isSameUpdater(bot.id, bot.publishTarget))
        )
        .forEach((bot) => {
          if (pollingUpdaterList.some((updater) => updater.isSameUpdater(bot.id, bot.publishTarget))) return;
          const updater = new PublishStatusPollingUpdater(bot.id, bot.publishTarget);
          pollingUpdaterList.push(updater);
          updater.start(onReceiveUpdaterPayload);
        });
```
and
```
    // Add new updater
    if (!pollingUpdaterList.some((u) => u.isSameUpdater(currentBotStatus.id, publishTarget))) {
      const newUpdater = new PublishStatusPollingUpdater(currentBotStatus.id, publishTarget);
      newUpdater.start(onReceiveUpdaterPayload);
      pollingUpdaterList.push(newUpdater);
    }
```

alanlong9278:
@srinaath  @hatpick  Have resolved all comments. Do u guys want to have any other comments

hatpick (approved):

srinaath (approved):

yeze322:
We can merge this PR after patch release

yeze322 (approved):

PR #5387, opened 2020-12-18
fix: luis deploy e2e test case failed due to unnecessary clicking on the start bots panel.

liweitian:
## Description
Current: after clicking start bots button, the "start bots panel" will open automatically if it is failed to start bots. 
In our e2e test case, there is one more clicking on the panel which will actually close it, so the See Details text cannot be seen.

Fix: delete the clicking on the "start bots panel" operation.

## Task Item
#minor


boydc2014:
Thanks for looking at this, but @liweitian can you describe more about what the issue is?

boydc2014 (approved):

boydc2014:
Great, really appreciate it

PR #5401, opened 2020-12-22
chore: refactor client code for cleanliness

beyackle:
## Description

This cleans up some type-safety and style things in the front-end code:
* Replaces a lot of `foo && foo.method()` calls with `foo?.method()` (and similarly for arrays).
* Replaces "reduce" calls with slightly more efficient maps and flatMaps when possible (note: Node 11 and recent browsers have flatMap as a built-in array method, but the code wouldn't compile with it that way, leading to me just using the Lodash one)
* Replaces a lot of 'any' type annotations with more specific ones.

There should be no changes to any functionality; this is meant to be purely a refactoring change for cleanup purposes.

## Task Item

#minor


Composer/packages/client/package.json
```diff
@@ -14,7 +14,8 @@
     "test": "jest",
     "lint": "eslint --quiet --ext .js,.jsx,.ts,.tsx ./src ./__tests__",
     "lint:fix": "yarn lint --fix",
-    "typecheck": "tsc --noEmit"
+    "typecheck": "tsc --noEmit",
+    "typecheck:watch": "tsc --noEmit --watch --noImplicitAny"
```
a-b-r-o-w-n, line 18: I think this was left in unintentionally. I don't mind leaving it, but I would remove the `--noImplicitAny` because that differs from our tsconfig.

Composer/packages/client/src/components/AppUpdater.tsx
```diff
@@ -60,8 +61,15 @@ const dialogContent: Partial<IDialogContentStyles> = {
...
 
-function SelectOption(props) {
-  const { checked, text, key } = props;
+type SelectOptionProps = {
+  checked?: boolean;
+  text: string;
+  key: string;
+};
+
+function SelectOption(props: SelectOptionProps | undefined) {
```
a-b-r-o-w-n, line 70: This could be desired, but this enforces the caller to pass an argument.

Composer/packages/client/src/pages/design/DesignPage.tsx
```diff
@@ -599,7 +599,7 @@ const DesignPage: React.FC<RouteComponentProps<{ dialogId: string; projectId: st
 
   async function handleDeleteDialog(projectId: string, dialogId: string) {
     const refs = getAllRef(dialogId, dialogs);
-    let setting: any = {
+    let setting: {} = {
```
a-b-r-o-w-n, line 602:
```suggestion
    let setting: Record<string, string> = {
```

beyackle: Good call. I'll do this.

Composer/packages/client/package.json
beyackle: It was intentional, but I'll remove the implicit-any thing.

Composer/packages/client/src/components/AppUpdater.tsx
beyackle: This was done to match the type signature of the `onRenderField` prop on a `ChoiceGroup`'s `options` attribute. It's expecting us to handle the "undefined" case even though this function is only used internally to this component and in cases we control completely.

Composer/packages/client/src/types/window.d.ts
```diff
@@ -36,3 +36,11 @@ declare global {
     __csrf__?: string;
   }
 }
+
+export type PageLink = {
```
tonyanziano, line 40: Why is this type included in this file? It is supposed to represent the properties on the window object

tonyanziano (commented):
See my one comment above. Other than that I think this PR looks good

Composer/packages/client/src/types/window.d.ts
beyackle: Oops - I added this to the wrong `types` file. I'll fix this.

tonyanziano (approved):

PR #5417, opened 2020-12-29
fix: make ClientStorage generic for type-safety

beyackle:
## Description

This is a minor and self-contained bit of cleanup that adds a generic to our ClientStorage class so we have a little more assurance of what we're getting out of it and can get rid of a few "any" type annotations.

In draft because it needs to be rebased on main once my other client-side type thing is merged; there really are only about 5-6 files being changed here, but that one needs to get in first.

## Task Item
#minor


beyackle:
Rebasing turned this into a mess; I'll close and redo it (it definitely is not 71 files worth of change).

PR #5420, opened 2020-12-30
fix: Remove Unnecessary Resource Group Name Validation

### LINKED ISSUE #5414: Resource Group Name Validation Not Needed When Creating Publish Profiles.
The deployment wizard warns the user that the resource group exists. However, the existence of it is not a problem at all.

image

When the user fills in the `HostName` first and selects the subscription next, the wizard validates the input, and the deployment can proceed with no issues. 

I suggest the validation be removed.

image

This will help with the confusion of having three different field descriptions as well. From what I can tell, **Resource Group Name** is the correct one.

image

This is on Bot Framework Composer v1.3.0. I hope it helps. Thanks.

daronyondem:
## Description

The deployment wizard warns the user that the resource group exists. However, the existence of it does not cause any issues during deployment. Removing the resource group check accordingly. 

Considering the `HostName` is used for both the resource group and web site/function app names ([See](https://github.com/microsoft/BotFramework-Composer/blob/d79ef62f1c36242260c1db29f607c87482cadce6/extensions/azurePublish/src/ui/azureProvisionDialog.tsx#L248)), changing the `placeholder` text to better reflect the use of the input.

## Task Item

Closes #5414


luhan2017 (approved):

VanyLaw:
I think it's ok to merge

daronyondem:
Apparently, the team decided to implement the [same fix under their name](https://github.com/microsoft/BotFramework-Composer/pull/5446/files). Closing this one.

PR #5428, opened 2020-12-31
fix: use generic-typed ClientStorage

beyackle:
## Description

This adds a generic to our ClientStorage class so we have more type-safety about what we put into and get out of there (which lets us use fewer "any" type annotations).

## Task Item
#minor


beyackle:
Somehow this picked up a commit I didn't even write.

PR #5429, opened 2020-12-31
refactor: add generics to ClientStorage

beyackle:
## Description

This adds a generic to our ClientStorage class so we have more type-safety about what we put into and get out of there (which lets us use fewer "any" type annotations).

## Task Item

#minor

### DIFF
--- a/Composer/packages/client/src/utils/luFileStatusStorage.ts
+++ b/Composer/packages/client/src/utils/luFileStatusStorage.ts
@@ -6,14 +6,16 @@ import storage, { ClientStorage } from './storage';
 
 const KEY = 'LuFileStatus';
 
-interface ILuFileStatus {
+type LuFileStatus = {
   [fileId: string]: boolean;
-}
+};
+
+type LuFileRecord = { [projectId: string]: LuFileStatus };
 
 // add luis publish status to local storage
 class LuFileStatusStorage {
-  private storage: ClientStorage;
-  private _all: { [projectId: string]: ILuFileStatus };
+  private storage: ClientStorage<LuFileRecord>;
+  private _all: LuFileRecord;
 
   constructor() {
     this.storage = storage;
--- a/Composer/packages/client/src/utils/qnaFileStatusStorage.ts
+++ b/Composer/packages/client/src/utils/qnaFileStatusStorage.ts
@@ -6,14 +6,16 @@ import storage, { ClientStorage } from './storage';
 
 const KEY = 'QnaFileStatus';
 
-interface IQnaFileStatus {
+type QnaFileStatus = {
   [fileId: string]: boolean;
-}
+};
+
+type QnaFileRecord = { [projectId: string]: QnaFileStatus };
 
 // add qna publish status to local storage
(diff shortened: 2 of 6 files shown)


beyackle:
There's a bug here; a `ClientStorage<string>` will try to deserialize a string as if it were JSON and crash instead of just spitting the string out. The older version handled this using a slightly clumsy mechanism, so I want to figure out a better way to do this.

beyackle:
No, actually, string storage will work. JSON.stringify sticks an extra pair of quotes around strings when they serialize (so `abc` becomes `"abc"`), and then JSON.parse will turn that back into a regular string as intended.

Composer/packages/client/src/utils/luFileStatusStorage.ts
```diff
@@ -6,14 +6,16 @@ import storage, { ClientStorage } from './storage';
 
 const KEY = 'LuFileStatus';
 
-interface ILuFileStatus {
+type LuFileStatus = {
   [fileId: string]: boolean;
-}
+};
+
+type LuFileRecord = { [projectId: string]: LuFileStatus };
```
a-b-r-o-w-n, line 13: These can be converted to `Record<string, *>`

Composer/packages/client/src/utils/qnaFileStatusStorage.ts
```diff
@@ -6,14 +6,16 @@ import storage, { ClientStorage } from './storage';
 
 const KEY = 'QnaFileStatus';
 
-interface IQnaFileStatus {
+type QnaFileStatus = {
   [fileId: string]: boolean;
-}
+};
+
+type QnaFileRecord = { [projectId: string]: QnaFileStatus };
```
a-b-r-o-w-n, line 13: Use `Record` here as well.

Composer/packages/client/src/utils/storage.ts
```diff
@@ -12,15 +12,17 @@ export class ClientStorage {
...
       return this.remove(this.prefix(key));
     }
     this.storage.setItem(this.prefix(key), this.serialize(val));
     return val;
   }
 
-  public get<T = any>(key: string, def?: T): T {
+  public get(key: string): T | undefined;
+  public get(key: string, def: T): T;
+  public get(key: string, def?: T): T | undefined {
```
a-b-r-o-w-n, line 25: Can we take this opportunity to rename this argument. Something like `defaultValue` would be better than `def` which is ambiguous.

Composer/packages/client/src/utils/storage.ts
```diff
@@ -37,36 +39,37 @@ export class ClientStorage {
     this.storage.clear();
   }
 
-  public getAll(): { [key: string]: any } {
+  public getAll(): { [key: string]: TextDecoderOptions } {
```
a-b-r-o-w-n, line 42: Consider the `Record` type here.

a-b-r-o-w-n (changes requested):

a-b-r-o-w-n:
Stale - closing. cc @beyackle

PR #5440, opened 2021-01-05
fix: fix australia region not work when publishing luis

### LINKED ISSUE #5407: Publish LUIS to "australia" failed and to "australiaeast" succeeded.
## Describe the bug

1. **Select `australia` as LUIS region.** 

   image

2. **Publish and get error**

   image

cc: @hibrenda

## Solution

Change `australia` to `australiaeast` worked.

image

Consider changing the LUIS region menu to: `australiaeast` instead of `australia`.

luis-region

## Version

v1.3

## Browser

- [x] Electron distribution

## OS

- [x] Windows

## To Reproduce

https://docs.microsoft.com/en-us/composer/how-to-add-luis

## Expected behavior

Should not fail.

## Additional text
https://docs.microsoft.com/en-us/azure/cognitive-services/luis/luis-reference-regions#luis-authoring-regions

VanyLaw:
## Description

change australia to australiaeast 

## Task Item

close #5407


hatpick:
@beyackle what do you think about the changes to the localization file?

beyackle (approved):
Normally I'd want the `en-US.json` changes to be part of a separate PR, but for the sake of getting that bubble ironed out with some speed, I'll approve it being done here. It does look like a lot of removals, but I've verified a sample of them and those strings really are absent.

PR #5446, opened 2021-01-05
fix: remove resoruce group checking in provision UI

### LINKED ISSUE #5414: Resource Group Name Validation Not Needed When Creating Publish Profiles.

VanyLaw:
## Description
support provision in existed resource group if webapp or other resources name not conflict.

## Task Item

close #5414


VanyLaw:
close it, because repeat with https://github.com/microsoft/BotFramework-Composer/pull/5420

luhan2017 (approved):

Issue #5396, opened 2020-12-21
Composer nearly unusably slow for editing dialogs after updating to 1.3.0 


ttur:
## Describe the bug

After updating the composer from 1.1.1 to 1.3.0 editing a complicated bot is so slow that actions (such as editing a sendresponse) take 5-10 seconds, and occasionally fail. 

Simpler dialogs can be edited okay, more complicated ones not at all. 

We have been building an advisory bot since composer early betas. Our bot has 30+ dialogs. Simple dialogs have a few activities. Complex dialogs have a few hundred. We support three languages. 

Most recently we were using version 1.1.1 and that worked quite well for this. 

## Version

Release: 1.3.0

## Browser

- [X] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [X] macOS
- [X] Windows
- [ ] Ubuntu

## Screenshots

Here is an example of the edit operation failing (editing Finnish version of the dialog that hasn't yet been properly translated from English): 

![image](https://user-images.githubusercontent.com/996293/102791359-93075080-43af-11eb-8843-740fbe0d2f77.png)

## Additional context

The process is taking just under 1G of memory on my laptop. I would be happy to give it more, if that helps, if I knew how.

It does not seem to matter whether the local bot is running or not. Editing is very slow and occasionally fails regardless.

ttur:
Some testing results:

- Release 1.2.1 is a bit faster than 1.3.0, but still way slower than 1.1.1 (tested on Windows and MacOS, same bot)
- Removing our extra languages (Finnish and Swedish) – leaving only English – does not make 1.3.0 much faster; maybe a little

gabog:
Thank you @ttur, we will address this issue in an upcoming release.

cwhitten:
This will be addressed in an upcoming patch (1.3.1)

PR #5455, opened 2021-01-06
fix: LG worker cache expires occasionally

zhixzhan:
## Description

# The issue
Inside LG worker, we maintained a cache which holds the parsed lg index, to avoid a full parse every time. We also designed a cache expiring mechanism that once we found the current state of lg file is not equaling the cached lg file, we consider lg file is changed outside, and thus the cache is invalid, we will do a full parse and resyncing with the master state. 
But we found that this caching is expiring more often than we think, and causing more full parsing than the expected incremental parsing.

Cache is expiring too fast and too often is because our store is updating slower than the caching itself. When user is typing fast, the store must be updated after the parsed result is obtained, because today, our raw lg content is tightened up with the parsed lg result and be updated together (This is actually the same case with dialog and lu).

For example, typing from '' to '12345' like:

image

If get-from-store and get-from-cache content not match, will do a full re-parse.

# The Solution
Instead of compare content, just use cache as single truth of fact, because all updates are supposed to be happen in worker, and worker will always update cache.

## Task Item
 
refs #5396

### DIFF
--- a/Composer/packages/client/src/recoilModel/parsers/workers/lgParser.worker.ts
+++ b/Composer/packages/client/src/recoilModel/parsers/workers/lgParser.worker.ts
@@ -151,7 +151,8 @@ const filterParseResult = (lgFile: LgFile) => {
 const getTargetFile = (projectId: string, lgFile: LgFile) => {
   const cachedFile = cache.get(projectId, lgFile.id);
 
-  return !cachedFile || cachedFile.content !== lgFile.content ? lgFile : cachedFile;
+  // Instead of compare content, just use cachedFile as single truth of fact, because all updates are supposed to be happen in worker, and worker will always update cache.
+  return cachedFile ?? lgFile;
 };
 
 export const handleMessage = (msg: LgMessageEvent) => {
--- a/Composer/packages/lib/indexers/src/utils/lgUtil.ts
+++ b/Composer/packages/lib/indexers/src/utils/lgUtil.ts
@@ -44,6 +44,7 @@ function templateToLgTemplate(templates: Template[]): LgTemplate[] {
   });
 }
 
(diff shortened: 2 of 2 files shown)


a-b-r-o-w-n:
@boydc2014 @zhixzhan @lei9444 

I think it would be helpful for us to draw up a system diagram of this. I'm not sure how to articulate this, but our usage of this cache is a bit of a code smell to me. It confuses what the single source of state is in our system. One general question I have is "is it even necessary to keep the file content in the recoil store"? It seems that operations on this are costly and we end up needing to circumvent the store in order to prioritize rendering performance. 

It would be interesting to explore keeping a reference to the files in the store but not the contents. We could use VSCode's Document model as inspiration perhaps.

This change is probably ok for now, but I'd like us to start thinking this.

boydc2014:
> @boydc2014 @zhixzhan @lei9444
> 
> I think it would be helpful for us to draw up a system diagram of this. I'm not sure how to articulate this, but our usage of this cache is a bit of a code smell to me. It confuses what the single source of state is in our system. One general question I have is "is it even necessary to keep the file content in the recoil store"? It seems that operations on this are costly and we end up needing to circumvent the store in order to prioritize rendering performance.
> 
> It would be interesting to explore keeping a reference to the files in the store but not the contents. We could use VSCode's Document model as inspiration perhaps.
> 
> This change is probably ok for now, but I'd like us to start thinking this.
> 
> cc @cwhitten

Sure, Andy, thanks for raising the concern. This PR particularly is the short-term fix to the cache invalding meachnasim. Our plan next will be putting down a design spec to document what's happening today with detailed diagram, what's the issue and what's long-term solution proposed (seperating cold\hot data). We will be able to share out something out next week.

a-b-r-o-w-n (approved):

PR #5457, opened 2021-01-06
fix: mutli re-render when editing in form editor

lei9444:
## Description
**Root cause**

1. when updating the dialog data, the state of dialogs in recoil will trigger one re-render. But in the <DesignPage/> the local state currentDialog will listen to the dialogs in useEffect and then trigger another re-render. The <ProjectTree/> has the same issue.
image

2. The debouncedUpdateLgTemplate API used by inline editor will invoke on the leading edge and trailing edge. This may trigger twice re-render.
image

## Task Item
refs #minor

## Screenshots
image


boydc2014 (dismissed):

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
```diff
@@ -239,7 +239,9 @@ export const ProjectTree: React.FC<Props> = ({
   }));
 
   useEffect(() => {
-    setSelectedLink(defaultSelected);
+    if (!isEqual(selectedLink, defaultSelected)) {
+      setSelectedLink(defaultSelected);
+    }
```
a-b-r-o-w-n, line 244: What scenario causes this effect to run? When does `defaultSelected` change any why?

Composer/packages/client/src/shell/lgApi.ts
```diff
@@ -99,7 +99,7 @@ function createLgApi(
     addLgTemplate: updateLgTemplate,
     getLgTemplates,
     updateLgTemplate,
-    debouncedUpdateLgTemplate: debounce(updateLgTemplate, 250, { leading: true, trailing: true }),
+    debouncedUpdateLgTemplate: debounce(updateLgTemplate, 250),
```
a-b-r-o-w-n, line 102: This (and the lu api) should be thoroughly tested. When changed in the past it caused some bugs with the editing experience.

a-b-r-o-w-n (commented):
@lei9444 make sure to test the changes the debounce on the lg and lu api thoroughly.

Composer/packages/client/src/shell/lgApi.ts
boydc2014: Make sense, i can help do more testing on this one. My expectation was it should be OK, since it's not removing denounce (which I remember we did in certain places and caused issues), it's just do the debounce in the default config. I will double check is the {leading: true, trailing: true} config was added for a specific reason, and will test whether it can be removed. Thanks.

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
srinaath: @beyackle could you assist here

lei9444: HI, @a-b-r-o-w-n Thanks to review the code. The trigger flow is: update dialog data -> update the dialogs in the recoil -> update the current dialog in the <DesignPage/> -> update the trigger item -> update the defaultSelected.

And I will sync with @beyackle if we can remove the useEffect directly, or why we need it here.

Composer/packages/client/src/shell/lgApi.ts
lei9444: I will do more test here

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
hatpick: it's probably from when it parses the url and knows what should be selected in the tree based on url params.

a-b-r-o-w-n:
@lei9444 ping me when you feel confident with the LG/LU api update -- I'll approve it.

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
lei9444: Thanks @hatpick , I have sync with @beyackle and @srinaath . We need this for changes like add a new dialog, a new trigger at the design page level. And we can make it more efficient by breaking the object into fields.

lei9444:
> @lei9444 ping me when you feel confident with the LG/LU api update -- I'll approve it.

Thanks @a-b-r-o-w-n ,I have repro it. when we use the  { leading: true, trailing: false }, we may drop some characters. I found you update trailing to true in #4790. And I have tested my PR, it's OK

boydc2014 (approved):

PR #5466, opened 2021-01-07
chore: Avoid re-render from states value change

zhixzhan:
## Description

Some components consume too much unnecessary states which can cause itself re-render. By reducing them can avoid re-render.

1. Reduce states in ProjectRouter components
2. Reduce states in LGPage/LUPage/QnAPage components, and avoid submit unchanged value. 

## Task Item
refs #5396 

## Screenshots
Inline editing on LG all up view table-list:
image

Before (several re-render):
image

After (one render): 
image


boydc2014:
Looks good, can you list how much time is saved?

Composer/packages/client/src/pages/language-generation/LGPage.tsx
```diff
@@ -25,16 +25,15 @@ const LGPage: React.FC<RouteComponentProps<{
   lgFileId: string;
 }>> = (props) => {
   const { dialogId = '', projectId = '', skillId, lgFileId = '' } = props;
-  const dialogs = useRecoilValue(validateDialogsSelectorFamily(skillId ?? projectId ?? ''));
-
+  const dialogs = useRecoilValue(dialogIdsState(skillId ?? projectId ?? ''));
```
srinaath, line 28: Is there any reason why projectId passed to table-view would be emoty string?

skillId ?? projectId would do right?

zhixzhan:
> Looks good, can you list how much time is saved?

Number of render times has been reduced from 2 to 1, it should save 50% render UI time.

boydc2014 (approved):

PR #5471, opened 2021-01-07
fix: Remove native Orchestrator DLLs from Runtime

### LINKED ISSUE #5447: New bot created with 1.3 fails to build - BadImageFormatException (Orchestrator not found)
## Describe the bug

A colleague in Cloud Advocacy is experiencing an issue that I think could likely affect multiple users.

Upgraded to 1.3 via automatic upgrade prompt.  Created a new bot (no ejected runtime) and made a simple change to the welcome message LG.  

Tried to run the bot from Composer and got the below error.

MicrosoftTeams-image

## Version

MicrosoftTeams-image (1)

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

taicchoumsft:
## Description
Orchestrator requires a 64-bit environment, which in turn requires a bump in Azure tier to S1 for the published bot, and might cause issues for users publishing an ejected runtime from scripts or Visual Studio (they will see a `BadImageFormatException`, since they are currently unaware of this 64-bit requirement). 

This PR removes the Orchestrator native DLLs so that the runtime can be compiled with `AnyCPU` and hosted on Azure WebApp Free Tier.  

A separate PR will document how to get Orchestrator running again for those who want to experiment with it.

## Task Item
closes #5447


boydc2014:
Is the way to get Orchestrator running again is through package manager?

taicchoumsft:
> Is the way to get Orchestrator running again is through package manager?

I will look into this.  In the meantime, Orchestrator team are attempting to fix this from the core.

boydc2014:
> > Is the way to get Orchestrator running again is through package manager?
> 
> I will look into this. In the meantime, Orchestrator team are attempting to fix this from the core.

Cool, before we have a solution to re-enabling orchestrator, i'm not sure should we just remove orchestrator packages. I'm under the impression that x64 is the new norms and S1 is also little and cheap. Not sure we have to remove it. 

@cwhitten what do you think?

cwhitten:
@boydc2014 there has been discussion in some other channels where we've felt this is the right point in time solution.

cwhitten (approved):

Issue #5427, opened 2020-12-31
Dialog name display wrong in Right side panel


rvinothrajendran:
## Describe the bug

Dialog name display wrong in Right side panel

**No idea is a bug or purposely display as “Adaptive dialog”**

## Version

1.3.0

## OS

- [ ] macOS
- [X] Windows
- [ ] Ubuntu

## To Reproduce

Create a new bot project and select the parent dialog 
Ex : parent dialog "WeatherBot"  right side panel display name as "Adaptive dialog"

## Expected behavior

Ex : parent dialog "WeatherBot"  right side panel display name as "WeatherBot"

## Screenshots

![image](https://user-images.githubusercontent.com/16264167/103420578-40404c80-4b98-11eb-8605-6db93351207b.png)

tonyanziano:
Thanks for pointing this out. We plan to fix this in the upcoming release.

mewa1024:
Is this related to bug #5383?

alanlong9278:
> Is this related to bug #5383?

Yes, this error is involved in #5383

yeze322:
duplicated, close this one and track the problem in #5383.

PR #5474, opened 2021-01-08
fix: unify display name function in client, visual editor

### LINKED ISSUE #5383: Trigger labels in bot tree, breadcrumbs, and property don't match
## Describe the bug

1.3.0

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [x] macOS
- [ ] Windows
- [ ] Ubuntu

## To Reproduce

I noticed this when I generated a form dialog:

image

image

But also in the LUIS To Do sample:

image

## Expected behavior

In general, the name should be the same in the tree, breadcrumbs, and property pane.
For a form dialog, the form dialog should not have "main dialog" in the breadcrumbs because it is not a main dialog.

alanlong9278:
## Description

bugfix: dialog name & trigger name mismatch in left menu,  breadcumbs & editors.

### solution

- [x]  Unify display name function. use `getFriedlyName` in client & editors, Shared the function in `bfc/shared` module.
- [x] Refactor breadcrumbs display name hook. It will auto change the display name when changing the name in form editor including dialog name, trigger name & action name.
 name
- [x] prefer using `designer.name` instead of dialog file name for dialog displayName in left menu  @zhixzhan 
  (if update name in property editor, for perf consideration, project tree won't real-time refresh )

## Task Item

closes #5383


Composer/packages/adaptive-flow/src/adaptive-flow-renderer/widgets/TriggerSummary/TriggerSummary.tsx
```diff
@@ -26,10 +25,15 @@ function getLabel(data: any): string {
 }
 
 function getName(data: any): string {
```
yeze322, line 27: It can be retired by the `getFriendlyName` in the @bfc/shared lib

yeze322:
Looks good to extract a new shared function `getFriendlyName`.

Composer/packages/client/src/pages/design/DesignPage.tsx
```diff
@@ -311,6 +310,30 @@ const DesignPage: React.FC<RouteComponentProps<{ dialogId: string; projectId: st
...
   }, [location]);
 
+  // get newest label for breadcrumbs
+  useEffect(() => {
+    if (currentDialog.content) {
+      setBreadcrumbs((prev) => {
+        return prev.map((b) => {
+          const separatorIndex = b.key.indexOf('-');
+          const type = b.key.substr(0, separatorIndex);
+          const name = b.key.substr(separatorIndex + 1);
```
beyackle, line 320: You can use a Regexp here - try something like
```
const regexp = /(?<type>[a-z]+)-(?<name>.*)/;
const { type, name } = match?.groups;
```

Since the `type` is something we have full control over, we know the match will succeed, but there might be some additional guards that need to be there so TS doesn't complain.

beyackle (dismissed):
Looks mostly good - just one change I'd like to see.

Composer/packages/client/src/pages/design/DesignPage.tsx
alanlong9278: I have checked the `type` variable. It is just set & use in the design page, we can add new type if have additional guards. I  create a const variable `types` to manager the type. 
```
const Types = {
  Dialog: 'dialog',
  Trigger: 'trigger',
  Action: 'action',
};
```

Composer/packages/client/src/pages/design/DesignPage.tsx
```diff
@@ -80,6 +80,13 @@ type BreadcrumbItem = {
   onClick?: () => void;
 };
 
+// field types
+const Types = {
+  Dialog: 'dialog',
+  Trigger: 'trigger',
+  Action: 'action',
+};
+
```
beyackle, line 89: See below comment - I'm wondering about changing this slightly so we don't have to worry about that "unsafe regex" linter rule.

Composer/packages/client/src/pages/design/DesignPage.tsx
```diff
@@ -313,17 +320,17 @@ const DesignPage: React.FC<RouteComponentProps<{ dialogId: string; projectId: st
     if (currentDialog.content) {
       setBreadcrumbs((prev) => {
         return prev.map((b) => {
-          const separatorIndex = b.key.indexOf('-');
-          const type = b.key.substr(0, separatorIndex);
-          const name = b.key.substr(separatorIndex + 1);
+          // eslint-disable-next-line security/detect-unsafe-regex
+          const regexp = new RegExp(/(?<type>[a-z]+)-(?<name>.*)/);
+          const { type, name } = b.key.match(regexp)?.groups || {};
```
beyackle, line 325: Seeing the linter need to be disabled here made me wonder if there were a simpler way to test this, and I think I have an idea how we could. This is an optional fix, but it will both get rid of the linter and make this a little faster.

Because we control the values of that Types enum and they aren't being exported anywhere, we could make the values a fixed length - say, single letters "D", "T", and "A" for dialog, trigger, and action (so a key would be `D-nameOfDialog`, for example). Then we'll always know the length of the type prefix and won't need any kind of searching or regex matching. The type would always be `b.key.charAt(0)` (or just `b.key[0]`), and the name would be `b.key.substr(2)`.

alanlong9278: Good idea, thanks

alanlong9278:
@beyackle Could u pls merge it if there is no problem? Thanks

yeze322 (approved):
Confirmed to be working

yeze322 (dismissed):
validated on my local PC

Composer/packages/client/src/pages/design/VisualEditorHeader.tsx
```diff
@@ -40,6 +35,13 @@ type VisualEditorHeaderProps = {
   pluginConfig?: PluginConfig;
 };
 
+// field types
+const Types = {
```
a-b-r-o-w-n, line 39: Let's give this a better name.

Composer/packages/client/src/pages/design/VisualEditorHeader.tsx
```diff
@@ -116,29 +117,35 @@ const useBreadcrumbs = (projectId: string, pluginConfig?: PluginConfig) => {
...
+            b.label = getFriendlyName(get(currentDialog.content, `triggers[${name}]`));
+            break;
+          case Types.Action:
+            b.label = getActionName(get(currentDialog.content, name));
+            break;
+        }
+        return b;
+      });
+    }
+  }, [currentDialog?.content]);
```
a-b-r-o-w-n, line 148: This looks like an anti-pattern. Should `breadcrumbArray` be in state?

Composer/packages/client/src/pages/design/VisualEditorHeader.tsx
yeze322: what about `BreadcrumbKeyPrefix`?

Composer/packages/client/src/pages/design/VisualEditorHeader.tsx
yeze322: @alanlong9278 `useMemo` should work

alanlong9278: Use `useMemo`

Composer/packages/client/src/pages/design/VisualEditorHeader.tsx
alanlong9278: use `BreadcrumbKeyPrefix`

yeze322:
#5427

Composer/packages/client/src/pages/design/VisualEditorHeader.tsx
```diff
@@ -85,10 +86,10 @@ const useBreadcrumbs = (projectId: string, pluginConfig?: PluginConfig) => {
   const focusPath = getFocusPath(selected, focused);
   const trigger = triggerIndex != null && dialogData.triggers[triggerIndex];
 
-  let breadcrumbArray: Array<BreadcrumbItem> = [];
+  const array: Array<BreadcrumbItem> = [];
```
yeze322, line 89: `array` is an ambiguous name! in some programming language it's even a reserved word

Composer/packages/client/src/pages/design/VisualEditorHeader.tsx
```diff
@@ -115,30 +116,37 @@ const useBreadcrumbs = (projectId: string, pluginConfig?: PluginConfig) => {
...
-
-    breadcrumbArray = [...breadcrumbArray.slice(0, 2), { key: 'action-' + actionName, label: actionName }];
-  }
-
+  // get newest label for breadcrumbs
+  const breadcrumbArray = useMemo(() => {
+    if (currentDialog.content) {
+      array.map((b) => {
+        const prefix = b.key.charAt(0);
+        const name = b.key.substr(2);
```
yeze322, line 132: I recommend that defining a pair of functions to parse / build the key, no matter how the key is parsed / built.

```
// Remember to make it typed
const buildKey = () => {};
const parseKey = () => {};
```

yeze322 (changes requested):
Generally good, left two comments.

Composer/packages/client/src/pages/design/VisualEditorHeader.tsx
alanlong9278: use parse/build key function

yeze322 (approved):

PR #5483, opened 2021-01-11
fix: validate dialog take a long time when updating lg in form

lei9444:
## Description
**Root cause**
The lg file update will trigger the validation function for all dialogs. A single dialog will take about 15ms, and if we have 10 dialogs in one bot, the validate time maybe 150ms.  

**Fix**
1. use atomFamily to replace the lgFilesState. So the dialog will only listen to the updating lg file and other dialogs will do nothing.
2. Separate the dialogs' diagnostics from the dialogState. 
3. add cache for the validateDialog function. Improve the time from 15ms to 8ms.

## Task Item
#minor

## Screenshots
Main branch:
image

This PR
image


boydc2014:
Looks good, this should be the last piece to include for a patch release.

Composer/packages/client/src/recoilModel/atoms/botState.ts
```diff
@@ -47,6 +47,22 @@ const emptyDialog: DialogInfo = {
   skills: [],
   isFormDialog: false,
 };
+type lgStateParams = { projectId: string; lgFileId: string };
```
hatpick, line 50: PascalCase type

Composer/packages/client/src/recoilModel/atoms/botState.ts
```diff
@@ -47,6 +47,22 @@ const emptyDialog: DialogInfo = {
...
+
+export const lgFileState = atomFamily<LgFile, lgStateParams>({
+  key: getFullyQualifiedKey('lg'),
+  default: () => {
+    return {} as LgFile;
```
hatpick, line 55: why does it return an empty object? If someone try to access a non-existing lg file this should return undefined not an empty object.

Composer/packages/client/src/recoilModel/dispatchers/lg.ts
```diff
@@ -29,37 +30,42 @@ const initialBody = '- ';
  *
  */
 
-const lgFilesAtomUpdater = (
+const lgFilesAtomUpdate = (
```
srinaath, line 33: it might be better rename this function to updateLgFiles?

Composer/packages/client/src/recoilModel/atoms/botState.ts
srinaath: How does this combination key work. Does recoil combine projectId and lgFileId into a single key on the back?

srinaath (changes requested):
@lei9444 try adding a remote skill to composer. It seems to break the app inside the validateDialogs

Composer/packages/client/src/recoilModel/atoms/botState.ts
hatpick: They serialize the keys

lei9444:
> @lei9444 try adding a remote skill to composer. It seems to break the app inside the validateDialogs

Thanks @srinaath I have fixed the issue

Composer/packages/client/src/pages/design/PropertyEditor.tsx
```diff
@@ -81,8 +82,7 @@ const PropertyEditor: React.FC = () => {
   }, [$schema, formUIOptions]);
 
   const errors = useMemo(() => {
-    const diagnostics = currentDialog?.diagnostics;
-    if (diagnostics) {
+    if (diagnostics && diagnostics.length) {
```
GeoffCoxMSFT, line 85:
```suggestion
    if (diagnostics?.length) {
```

Composer/packages/client/src/recoilModel/atoms/botState.ts
```diff
@@ -47,6 +47,32 @@ const emptyDialog: DialogInfo = {
...
+  content: '',
+  diagnostics: [],
+  templates: [],
+  allTemplates: [],
+  imports: [],
+};
+
+type LgStateParams = { projectId: string; lgFileId: string };
+
+export const lgFileState = atomFamily<LgFile, LgStateParams>({
```
GeoffCoxMSFT, line 62: This seems the same as lgFilesState.  Why make the name singular where lgFileIdsState is plural?

Composer/packages/client/src/recoilModel/atoms/botState.ts
```diff
@@ -47,6 +47,32 @@ const emptyDialog: DialogInfo = {
...
+  key: getFullyQualifiedKey('lg'),
```
GeoffCoxMSFT, line 63: This seems like a very short key - shouldn't it be lgFile (or lgFiles?)

Composer/packages/client/src/recoilModel/dispatchers/project.ts
```diff
@@ -251,6 +251,7 @@ export const projectDispatcher = () => {
           callback(projectId);
         }
       } catch (ex) {
+        console.log(ex);
```
GeoffCoxMSFT, line 254:
```suggestion
        console.error(ex);
```

Composer/packages/client/src/recoilModel/dispatchers/project.ts
```diff
@@ -251,6 +251,7 @@ export const projectDispatcher = () => {
...
         set(botProjectIdsState, []);
         removeRecentProject(callbackHelpers, path);
         handleProjectFailure(callbackHelpers, ex);
```
GeoffCoxMSFT, line 257: If we need to log errors to the console, I would hope that methods handleProjectFailure do it rather than each catch.

Composer/packages/client/src/pages/design/PropertyEditor.tsx
a-b-r-o-w-n: This seems like something the linter could catch. I'll investigate.

a-b-r-o-w-n: Found it. @GeoffCoxMSFT or @hatpick can you create a PR to update the lint rules with this?

https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/prefer-optional-chain.md

hatpick: I can do that.

Composer/packages/client/src/recoilModel/dispatchers/project.ts
srinaath: It already does that. The console.error can be safely removed

lei9444: This is used to test, I forget to remove it. I should be removed.

boydc2014 (approved):

Composer/packages/client/src/recoilModel/dispatchers/lg.ts
```diff
@@ -208,40 +211,27 @@ export const lgDispatcher = () => {
...
-              return (lgFile) => lgFile.id !== updatedFile.id;
-            } else {
-              return () => true;
-            }
-          })
-        );
+
+        updateLgFiles(callbackHelpers, projectId, { updates: updatedFiles }, (current, changed) => {
+          // compare to drop expired content already setted above.
+          return current?.content === changed?.content;
```
zhixzhan, line 233: We need also compare id, because "updateFiles" can be other lg file, and the content of course is different , then will always drop their changes. Fixed this in #5654

PR #5493, opened 2021-01-11
fix: telemetry fixes

tdurnford:
## Description

- Enable the capture of the help link URL
- Fix the unknown section for pageviews 
- Replace the URL captured for the section page views with the section name.  Right now we are capturing the actual URLs which could contain sensitive data as they contain dialog names.

## Task Item
#minor


Composer/packages/server/src/services/telemetry.ts
```diff
@@ -58,6 +58,10 @@ if (instrumentationKey) {
       if (typeof data.baseData.name === 'string') {
         data.baseData.name = data.baseData.name.replace(/\/\d+.\d+/, '/XXXXX.XXXXXXXXXX');
       }
+
+      if (typeof data.baseData.name === 'string') {
```
hatpick, line 62: L62-64 is the same as L58-60

GeoffCoxMSFT:
Reran the conventional PR job for you :)

Composer/packages/client/src/utils/getPageName.tsx
```diff
@@ -4,8 +4,8 @@
 import { PageNames } from '@bfc/shared';
 
 const PagesRegex = {
-  [PageNames.Design]: /\/dialogs/i,
-  [PageNames.Home]: /\/home/i,
+  [PageNames.Design]: /\/dialogs|\/bot\/d+.d+/i,
+  [PageNames.Home]: /\/|\/home|\/projects\/create|\/projects\/open/i,
```
a-b-r-o-w-n, line 8: So this now includes project creation? Would it be better to separate the project creation pages into it's own page category?

boydc2014:
@garypretty can you confirm do we want to include this in 1.3 patch?

Composer/packages/client/src/utils/getPageName.tsx
tdurnford: Per a conversation with @garypretty, since we are already capturing custom events for `/project/create` and `/projects/open` it would be redundant to capture them as a page view.

PR #5509, opened 2021-01-13
fix: showing correct error message in local publish

### LINKED ISSUE #5390: Bot fails to start after update to v1.3.0
## Describe the bug

After updating to v1.3.0 the bot does not start in the emulator. 
image

Publishing to azure is still going fine
image

## Version

1.3.0

## Browser

- [ ] Electron distribution
- Chrome
- [ ] Safari
- [ ] Firefox
- Edge

## OS

- [ ] macOS
- Windows
- [ ] Ubuntu

## To Reproduce

1. Open bot
2. Click on start bot or restart bot

## Expected behavior

Bot opens in emulator

VanyLaw:
## Description
some error doesn't serializate.

## Task Item

close #5390

### DIFF
--- a/extensions/localPublish/src/index.ts
+++ b/extensions/localPublish/src/index.ts
@@ -53,12 +53,10 @@ const isSkillHostUpdateRequired = (skillHostEndpoint?: string) => {
 };
 
 const stringifyError = (error: any): string => {
-  if (error instanceof Error) {
-    return error.toString();
-  } else if (typeof error === 'object') {
+  if (typeof error === 'object') {
     return JSON.stringify(error, Object.getOwnPropertyNames(error));
   } else {
-    return error;
+    return error.toString();
   }
 }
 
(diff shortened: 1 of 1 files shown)


extensions/localPublish/src/index.ts
```diff
@@ -52,6 +51,17 @@ const isLocalhostUrl = (matchUrl: string) => {
...
   return !skillHostEndpoint || isLocalhostUrl(skillHostEndpoint);
 };
+
+const stringifyError = (error: any): string => {
+  if (error instanceof Error) {
+    return `${error.name} - ${error.message}`;
+  } else if (typeof error === 'object') {
+    return JSON.stringify(error, Object.getOwnPropertyNames(error));
+  } else {
+    return error;
```
boydc2014, line 61: how about return error.toString()?

VanyLaw: if error is instanceof Error or typeof string, it can return toString(), but if it's an normal object, toString will throw error.

boydc2014: I mean the your default condition, line 61, should be error.toString() right? since your function are supposed to return a string

VanyLaw: done.

boydc2014 (approved):

PR #5513, opened 2021-01-13
fix: Revert back to previous bf-generate for 1.3

GeoffCoxMSFT:
## Description

The updated version of bf-generate has some bugs that we don't want to ship in this 1.3 patch update.

## Task Item

#minor

## Screenshots

N/A


GeoffCoxMSFT:
@boydc2014 @cwhitten I would be good to have this as part of the 1.3 patch today so we don't regress form dialogs.

hatpick (dismissed):

boydc2014:
I'm retargeting this into main, and merge everything we need to main, then take the latest main to 1.3patch.

GeoffCoxMSFT:
I'll do a rollback of the change that we don't want in 1.3 instead.

PR #5515, opened 2021-01-13
fix: l18n apostrophe errors

### LINKED ISSUE #5496: [MCS-DE] Electron App freezes when opening a bot
Whenever I try to open a Bot with the Composer Electron App v1.3.0 on Windows (BotFramework-Composer-1.3.0-windows-setup.exe) the application freezes.

Switching to any areas in the left navigation incl. "Home" is not possible.

This happens with an existing Bot as well as when creating a new empty bot.

Switching the Composer Application language to English (US), opening the bit and then switchint the Composer langague back to German solves this issue. But it needs to be done before opening a bot every time.

@cwhitten @benbrown

beyackle:
## Description

Some of our locale files had quote marks in places that broke ICU formatting. This should fix those.

## Task Item

fixes #5496

### DIFF
--- a/Composer/packages/server/src/locales/de.json
+++ b/Composer/packages/server/src/locales/de.json
@@ -1281,7 +1281,7 @@
     "message": "Get conversation members"
   },
   "get_started_50c13c6c": {
-    "message": "Los geht's!"
+    "message": "Los geht’s!"
   },
   "getting_help_ab6811b0": {
     "message": "Hilfe anfordern"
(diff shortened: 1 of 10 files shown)


tonyanziano (dismissed):
What generates these translations? If it's a script, can we modify the script to avoid inserting these double quotes so that we can prevent breaks like this from happening again?

beyackle:
@tonyanziano 
They're generated by human translators. While we can't directly script them, we can at least write a script to double-check their work and catch mistakes like this before they go into the product,

tonyanziano (approved):

PR #5524, opened 2021-01-14
fix: switching in project tree is lagging

lei9444:
## Description
Switching in project tree is causing too much re-rendering (including some full re-rendering), up to 8 times in a simple switch. This fix some simple refactoring that saves us at least 4 full rendering times, leaves us to 3-4 partial rendering. Will continue investigate the other unnecessary rendering.   

The goal eventually is to have only one rendering in one switch. 

### Root cause
1. We always set local state in useEffect when location change.
2. We are setting < breadcrumb /> when switching the item, but the state only used by breadcrumb component and put it in the < DesignPage />  will make the page full re-render.
3. The < App /> will listen to the url to initialize the logger, the location change will make the whole app re-render

### Fix
1. add value check, if no change, do nothing
2. move the < Breadscrum/> to a single component, and let it's own to manage the state
3. The same as 2

### next step
1. split the < DesignPage />
2. investigate the reach/router's state management, There is still a full re-render caused by Router

## Task Item
#minor

## Screenshots
before

image

after

image

### DIFF
--- a/Composer/packages/client/src/recoilModel/dispatchers/navigation.ts
+++ b/Composer/packages/client/src/recoilModel/dispatchers/navigation.ts
@@ -5,47 +5,55 @@
 //TODO: refactor the router to use one-way data flow
 import { useRecoilCallback, CallbackInterface } from 'recoil';
 import { PromptTab } from '@bfc/shared';
+import isEqual from 'lodash/isEqual';
 
 import { currentProjectIdState } from '../atoms';
 import { encodeArrayPathToDesignerPath } from '../../utils/convertUtils/designerPathEncoder';
 import { dialogsSelectorFamily, rootBotProjectIdSelector } from '../selectors';
+import { DesignPageLocation } from '../types';
 
 import { getSelected } from './../../utils/dialogUtil';
 import { designPageLocationState, focusPathState } from './../atoms/botState';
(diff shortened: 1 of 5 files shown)


Composer/packages/client/src/recoilModel/dispatchers/navigation.ts
```diff
@@ -5,47 +5,55 @@
...
+
+      await setCurrentProjectId(callbackInterface, projectId);
+
+      const preLocation = await snapshot.getPromise(designPageLocationState(projectId));
+
+      if (isEqual(preLocation, location)) return;
+
+      const { dialogId = '', selected = '', focused = '' } = location;
+      const focusPath = `${dialogId}#${focused ? `.${focused}` : selected ? `.${selected}` : ''}`;
       set(focusPathState(projectId), focusPath);
```
zhixzhan, line 39: Here we already have 'dialogId', 'selected', 'focused' from localtion.
Then the `focusPathState` seems can be delete from client & shellData.

zhixzhan (commented):
Looks good to me, split design page breadcrumb to a self render components really improve a lot.

Composer/packages/client/src/recoilModel/dispatchers/navigation.ts
lei9444: Thanks, @zhixzhan I will create a new PR to clean the data,

boydc2014 (approved):

PR #5528, opened 2021-01-14
fix: add check script for l10n files

beyackle:
## Description

This PR is to add a new "l10nCheck" script and associated yarn command which will inspect the locale files in `packages/server/src/locales` and find issues that tend to occur in them:

- Whitespace between the number and the = in a "plural" definition
- Missing "other" clause in a "plural" definition (this usually happens because a translator turns the word "other" into its translation in the target language)
- Single quotes (`'`) between letters; single quotes are semantic to ICU formatting and should be replaced by curly quotes in actual messages

This won't catch every possible issue, but it should be used as a starting point to find the most common errors.

## Task Item

refs #5496

### DIFF
--- a/Composer/scripts/l10nCheck.js
+++ b/Composer/scripts/l10nCheck.js
@@ -19,13 +19,16 @@ for (;;) {
   for (const [key, val] of Object.entries(json)) {
     const msg = val.message;
     const src = `${dirent.name} at ${key}:`;
-    if (/=\s[01]/.exec(msg)) {
-      console.log(src, `whitespace between 0/1 and =`);
-      errorCount += 1;
-    }
-    if (/=[01]/.exec(msg) && !/other/.exec(msg)) {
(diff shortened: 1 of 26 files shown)


Composer/scripts/l10nCheck.js
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+/* eslint-disable @typescript-eslint/no-var-requires */
+/* eslint-disable security/detect-non-literal-fs-filename */
+
+const fs = require('fs');
+const path = require('path');
+
+const localePath = 'packages/server/src/locales';
```
tonyanziano, line 9: does this need to be normalized for Windows?

beyackle: I'm running this on a Windows machine and it worked fine.

tonyanziano (approved):

PR #5529, opened 2021-01-14
chore: Updating to daily runtime for R12 development

GeoffCoxMSFT:
## Description

During R12 development on the main branch, we should pick up the daily builds of the runtime.  This will help us discover issues earlier.

This also updates to the latest bf-generate library as it uses the latest runtime.

## Task Item

#minor

## Screenshots

N/A


boydc2014:
@GeoffCoxMSFT do we have plans to move bf-generate to npm? which is probably our last dependency to myget today.

hatpick (approved):

benbrown (approved):
The changes to the runtime/ files will work for now, but the plan is to deprecate the local runtime in favor of the new runtime libraries as part of r12.

yeze322:
I found the `update-schema` script is broken after updating dependencies to 4.12.0-daily...

PR #5539, opened 2021-01-15
fix: Merge settings error

srinaath:
## Description
This PR fixes an issue where the root bot's keys are not propogated into the skills and we notice an error that the key is missing in the skill even though the key is available in the root bot.

Fixes #minor


a-b-r-o-w-n (approved):

liweitian:
The mergedSettings represents what a skill's bot has. If it does not have a key, then in the settings page, it will display nothing and says this bot is using root bot's key.
When we publish skill bots, for example, we are using the key stored in atom family settingState, settingState(skillbotId) will sync with root bots.

PR #5547, opened 2021-01-18
feat: support monaco editor offline usage and most fabric fonts offline display

liweitian:
## Description
Composer cannot use offline now due to monaco editor code and fabric fonts need to be downloaded from CDN. 

This PR will copy needed resouce to public folder during build time.  
FYI, https://fabricui.com/react/styles/icons https://github.com/suren-atoyan/monaco-react/issues/12

And from online discussion, one person said the fonts in public folder does not include all fonts, although this is not agreed by the fabric UI engineers and I did not find any font not showing in composer.
## Task Item
#minor
## Screenshots
The is tested without network
offline


hatpick:
Why do we need all these other files (objective-c, python, php, etc.) ?

a-b-r-o-w-n:
Is it possible to add some steps to our client build that fetches these files and then ignore them in the repo?

liweitian:
> Is it possible to add some steps to our client build that fetches these files and then ignore them in the repo?

Done. This pr is ready for another review.

Composer/packages/client/scripts/build.js
```diff
@@ -35,6 +35,16 @@ const WARN_AFTER_CHUNK_GZIP_SIZE = 1024 * 1024;
...
+//copy monaco editor resouce to public folder to support offline usage
+const monacoEditor = path.resolve(__dirname, '../../../node_modules/monaco-editor/min');
+const monacoEditorInPublicFolder = path.resolve(__dirname, '../public/min');
+fs.copySync(monacoEditor, monacoEditorInPublicFolder);
+
+//copy fabric fonts to public folder to support offline usage
+const fabricFonts = path.resolve(__dirname, '../../../node_modules/@uifabric/icons/fonts');
+const fabricFontsInPublicFolder = path.resolve(__dirname, '../public/fonts');
+fs.copySync(fabricFonts, fabricFontsInPublicFolder);
+
```
a-b-r-o-w-n, line 47: Doing this as part of the prod build only seems like it would break local development. Is that true? I was thinking it should happen as a postinstall script.

a-b-r-o-w-n:
@liweitian what is the status of this work? It seems to have stalled and there is some other work regarding monaco. Please close this PR if it is no longer needed. I will close in a few days if no activity.

liweitian:
Closed it for now since we are updating the version of monaco. I will open another pr if it is needed to solve issues on VMs cannot download monaco load.js from CDN

PR #5563, opened 2021-01-20
fix: Move persistence layer's delta computation into worker

### LINKED ISSUE #5561: Move persistence layer's delta computation into worker
Our profiling shows that, for a large bot, the file persistence layer's delta computation (calculating the delta to be saved to disk\storage) is taking a fair amount time and could cause the UI thread to be blocked. 

image

The 4th in the above picture is showing file persistence layer is taking about 100ms. 

The proposal is moving this part into worker, so that UI thread won't be blocked. 

Note: the reason we are calculating the state change between commits is we don't have a native support in recoil **yet**.  If we get a native support, we could save the whole delta computation process.

lei9444:
## Description
The persistence layer will check files' change after the state changed. When we use a big bot such as the [calendarSkill](https://github.com/microsoft/botframework-components/tree/calendarSkill_v2/skills/declarative/calendarSkill), the check function will block the main thread about 90ms every notify. 

**fix** 
Move the complex computation to worker

## Task Item
closes #5561 

## Screenshots
**before:**
image

**after:**
image

### DIFF
--- a/Composer/packages/client/src/recoilModel/parsers/calculator.ts
+++ b/Composer/packages/client/src/recoilModel/parsers/calculator.ts
@@ -1,0 +1,18 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { FileAsset } from '../persistence/types';
+
+import Worker from './workers/calculator.worker.ts';
+import { BaseWorker } from './baseWorker';
+import { FilesDifferencePayload, CalculatorType } from './types';
+
+// Wrapper class
+class Calculator extends BaseWorker<CalculatorType> {
+  difference(target: FileAsset[], origin: FileAsset[]) {
+    const payload = { target, origin };
+    return this.sendMsg<FilesDifferencePayload>('difference', payload);
+  }
+}
+
+export default new Calculator(new Worker());
--- a/Composer/packages/client/src/recoilModel/persistence/types.ts
+++ b/Composer/packages/client/src/recoilModel/persistence/types.ts
@@ -1,6 +1,17 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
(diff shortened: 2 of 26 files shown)


boydc2014:
I can see the differences, but do we know why after moving to worker, we still take about 10ms?

lei9444:
> I can see the perf gain here, just one more question, do we know why after moving to worker, we still take about 10ms?

The worker.postMessage will take some time here

srinaath:
This is awesome work @lei9444 ! Moving it a worker is so much better.

srinaath (approved):

Composer/packages/client/src/recoilModel/persistence/types.ts
```diff
@@ -30,3 +41,19 @@ export interface IFileChange {
...
+  | DialogInfo
+  | DialogSchemaFile
+  | LuFile
+  | QnAFile
+  | LgFile
+  | SkillManifestFile
+  | RecognizerFile
+  | FormDialogSchema;
+
+export type Difference = {
```
srinaath, line 55: Myabe name could be FileDifference?

Composer/packages/client/src/recoilModel/parsers/calculator.ts
```diff
@@ -0,0 +1,18 @@
...
+
+// Wrapper class
+class Calculator extends BaseWorker<CalculatorType> {
+  difference(target: FileAsset[], origin: FileAsset[]) {
+    const payload = { target, origin };
+    return this.sendMsg<FilesDifferencePayload>('difference', payload);
+  }
+}
+
+export default new Calculator(new Worker());
```
srinaath, line 18: FileDiffCalculator?

boydc2014 (approved):

Issue #5549, opened 2021-01-18
Unable to update Composer


arafattehsin:
## Describe the bug

I have attached the gif that'd give an overview of what's happening. This issue popped up in earlier versions as well [4813](https://github.com/microsoft/BotFramework-Composer/issues/4813)

For the context, I tried both options *Only me (user) or Everyone* while installing Composer. I am also the administrator of this new machine. 

## Version

1.3

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Go to 'Help'
2. Click on 'Check for Updates'
3. See error

## Expected behavior

It should have updated the Composer.

## Screenshots

![2021-01-19_9-39-17](https://user-images.githubusercontent.com/16351038/104968757-dac6d980-5a3a-11eb-869e-256367bdca55.gif)

After this, I got this error:

![image](https://user-images.githubusercontent.com/16351038/104968707-b10db280-5a3a-11eb-831b-d6000b223bb3.png)

cwhitten:
Hmm. Sorry to hear about your troubles @arafattehsin. I just picked up the latest nightly on Mac and it worked fine.

<img width="575" alt="Screen Shot 2021-01-18 at 4 01 44 PM" src="https://user-images.githubusercontent.com/1156704/104972383-7a865800-59a6-11eb-8100-80828d4ce4c5.png">

On Windows we ask users install the application for all users. 

The following is a heavier operation but should put you in a clean install state:

1. Uninstall Composer (you won't lose any Bot Projects)
2. Delete your entire `@bfc` directory under `C:\Users\AT\AppData\Local\`. Composer will recreate this for you.
3. Reinstall Composer 1.3.0 [here](https://aka.ms/bf-composer-download-win) and turn on Early adopters to pick up the latest nightly

hibrenda:
@cwhitten I can repro this on windows after delete entire @bfc and old composer as well.
![image](https://user-images.githubusercontent.com/43408889/104976372-098f8080-5a38-11eb-90ec-cfef3b71e0ab.png)

hibrenda:
![image](https://user-images.githubusercontent.com/43408889/104979845-32b40f00-5a40-11eb-9be8-3416392f142a.png)

boydc2014:
[...]

PR #5573, opened 2021-01-21
fix: electron update error

### LINKED ISSUE #5551: Data race in auto updating vs manual updating
## Describe the bug

We have a data race when user click "check for updates" with auto-update set to true, causing the updating fail (#5549). 

The reason is when auto-update is set to true
1. when composer starts up, we are checking the updates at backend and starts the downloading process, a temp file is created for the
2. when user click check for updates, we are trying to write a same temp file at the same location, then this operation fails because the permission is not allowed, since this file is already created and been actively used. 
3. when the operation in 2 fails, the clean process is trigger, and this file is totally removed, causing composer to be in a bad state where any other following auto & manual update won't work

## Version

lastest main
## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Install 1.3.0, turn on auto update
2. Run composer, click "help" => "check for updates"
3. See error

## Expected behavior

1. When auto update is on and is downloading, and when user click manual update, we should prompting user to show we are already doing a downloading now.
2. When auto update finished it's downloading, and user should get a notification to ask user to update or not
3. If user choose to ignore the notification. Then when user click manual update, we will show the same notification again asking for update, without have to download again (if there is not new version).

zhixzhan:
## Description
We have a data race when user click "check for updates" with auto-update set to true, causing the updating fail (#5549).

This PR let manual-update can continue auto-update progress.

1. if auto update is downloading, when click "check for updates" will show the progress.
2. if auto update is downloaded, when click "check for updates" will show quit-and-start modal.

## Task Item
close #5551 

## Screenshots

1. When auto update is downloading.

2. When auto update is downloaded.

### DIFF
--- a/Composer/packages/client/src/components/AppUpdater.tsx
+++ b/Composer/packages/client/src/components/AppUpdater.tsx
@@ -132,7 +132,7 @@ export const AppUpdater: React.FC<{}> = () => {
         }
 
         case 'update-in-progress': {
-          setAppUpdateStatus(AppUpdaterStatus.UPDATE_AVAILABLE, undefined);
+          setAppUpdateStatus(AppUpdaterStatus.UPDATE_AVAILABLE, payload.version);
           setAppUpdateShowing(true);
           break;
         }
--- a/Composer/packages/client/src/components/AppUpdater.tsx
+++ b/Composer/packages/client/src/components/AppUpdater.tsx
@@ -131,6 +131,12 @@ export const AppUpdater: React.FC<{}> = () => {
           break;
         }
 
+        case 'update-in-progress': {
+          setAppUpdateStatus(AppUpdaterStatus.UPDATE_AVAILABLE, undefined);
+          setAppUpdateShowing(true);
+          break;
+        }
+
         case 'update-not-available': {
           const explicit = payload;
           if (explicit) {
--- a/Composer/packages/electron-server/src/appUpdater.ts
+++ b/Composer/packages/electron-server/src/appUpdater.ts
@@ -48,13 +48,16 @@ export class AppUpdater extends EventEmitter {
    * and we will show UI if there are no available updates.
    */
   public checkForUpdates(explicit = false) {
-    if (!(this.checkingForUpdate || this.downloadingUpdate)) {
-      this.explicitCheck = explicit;
-      this.setFeedURL();
-      this.determineUpdatePath();
-      autoUpdater.autoDownload = this.settings.autoDownload;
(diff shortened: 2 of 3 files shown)


Composer/packages/electron-server/src/appUpdater.ts
```diff
@@ -108,6 +111,7 @@ export class AppUpdater extends EventEmitter {
 
   private onDownloadProgress(progress: any) {
     log('Got update progress: %O', progress);
+    this.downloadingUpdate = true;
```
zhixzhan, line 114: Auto update won't call `public downloadUpdate()`,  So seems here is the only place we can know current is downloading.

Composer/packages/electron-server/src/appUpdater.ts
```diff
@@ -116,10 +120,10 @@ export class AppUpdater extends EventEmitter {
   private onUpdateDownloaded(updateInfo: UpdateInfo) {
     log('Update downloaded: %O', updateInfo);
     this._downloadedUpdate = true;
-    this.resetToIdle();
     if (this.explicitCheck || !this.settings.autoDownload) {
       this.emit('update-downloaded', updateInfo);
     }
+    this.resetToIdle();
```
zhixzhan, line 126: reset will clean up `this.explictCheck`, so here place it in the end.

Composer/packages/client/src/components/AppUpdater.tsx
```diff
@@ -131,6 +131,12 @@ export const AppUpdater: React.FC<{}> = () => {
           break;
         }
 
+        case 'update-in-progress': {
+          setAppUpdateStatus(AppUpdaterStatus.UPDATE_AVAILABLE, undefined);
```
tonyanziano, line 135: the `undefined` argument here is supposed to be the version of the update -- if you look at the first video you posted, you can see that when this line is called and shows the "Update Available" dialog, the dialog shows "vundefined"

image

tonyanziano (changes requested):
Aside from the visual bug in my comment, I think this looks good! Nice job

Composer/packages/client/src/components/AppUpdater.tsx
zhixzhan: Nice catch!
I see the differences, now updated: 

image

tonyanziano (approved):

PR #5589, opened 2021-01-22
feat: extension events

a-b-r-o-w-n:
## Description

Adds simple mechanism to hook into Composer events in extensions.

Executes event listeners serially (in order the listener was registered) which allows things like modification to bot assets.

## Task Item

#minor

## Usage

There is example usage in the `sample-ui-plugin` that shows how to listen to the `project:opened` and `project:created` events.

```ts
module.exports = (composer) => {
  composer.context.on("project:opened", async (project) => {
    // do something with the project
  });
};
```


benbrown:
@a-b-r-o-w-n these event handlers should have access to the authed user if present.

a-b-r-o-w-n:
@benbrown do you have any suggestions on how to do that?

Composer/packages/server/src/models/extension/composerEventEmitter.ts
```diff
@@ -0,0 +1,73 @@
+// Copyright (c) Microsoft Corporation.
```
hatpick, line 1: You can do the following if you want it typed with vscode intellisense:
```typescript
type EventTable = {
  eventName: (arg1: string, arg2: string) => void;
  myEvent: (num: number, cb: () => void) => void;
};

type EventName = keyof EventTable;

type ArgsOf<T> = T extends (...args: infer U) => void ? U : never;

type Disposable = { dispose: () => void };

class Emitter<T extends EventTable> {
  private events: Record<string, Record<string, Function>>;

  private uid: number;

  constructor() {
    this.events = {};
    this.uid = 0;
  }

  public on<TName extends keyof EventTable>(
    eventName: TName,
    callback: (...args: ArgsOf<T[TName]>) => void
  ): Disposable {
    if (!this.events[eventName]) {
      this.events[eventName] = {};
    }
    const cbId = this.uid++;
    this.events[eventName][cbId] = callback;
    return {
      dispose: () => {
        delete this.events[eventName][cbId];
        if (!Object.keys(this.events[eventName]).length) {
          delete this.events[eventName];
        }
      },
    };
  }

  public emit<TName extends keyof EventTable>(
    eventName: TName,
    ...args: ArgsOf<T[TName]>
  ) {
    if (!this.events[eventName]) {
      throw `event: ${eventName} does not exist!`;
    }
    for (const cbId in this.events[eventName]) {
      this.events[eventName][cbId].apply(null, args);
    }
  }
}

const emitter = new Emitter<EventTable>();
const sub = emitter.on("eventName", (arg1, arg2) =>
  console.log(arg1, arg2)
);
emitter.emit("eventName", "test", "test");

sub.dispose();
```

a-b-r-o-w-n: This is cool. I'll make the change!

PR #5598, opened 2021-01-25
chore: splitting states & component

zhixzhan:
## Description

To reduce components re-render, it's better to splitting states to fine-gained level, so does components

## Task Item

#minor


boydc2014:
close this for now.

PR #5352, opened 2020-12-16
fix: active language callout on header should be closed once losing focus

liweitian:
## Description
active language callout on header should be closed once losing focus

## Task Item
Closes #5293

## Screenshots
callout

PR #5601, opened 2021-01-25
fix: active language callout bug

### LINKED ISSUE #5597: Header bar language select unable to click
## Describe the bug

When the bot has more than one language, the language selection call out is unable to click

## Version
latest main

## Browser

- [ ] Electron distribution
- [x ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [x] macOS
- [ ] Windows
- [ ] Ubuntu

liweitian:
## Description
Root cause: There is a focuszone inside the component TeachingBubble. By default we should press key esc to close the call out. Press tab cannot be used to jump out of the focuszone.
For #5352, when only one language exists, the dropdown is disabled, which causes it cannot enter the focuszone because there is nothing to focus. 

fix: use esc to close the callout. And enable the dropdown even if there is only one language exist.

## Task Item
Closes #5597
## Screenshots
tab


beyackle (approved):
LGTM

PR #5602, opened 2021-01-25
fix: keyboard focus error

### LINKED ISSUE #5292: Keyboard focus is invisible on the "Add" circle and the zoom in/out/center buttons
## Describe the bug

Tab navigation across the Design View buttons is difficult when the visible focus vanishes.

## Version

 1.3.0-rc3

## Browser

- [ ] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [x] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Open the Design view and make a new Trigger somewhere (just to keep the design view clean)
2. Press Tab until you go past the "Show code" button
3. Focus is on the "Add" thing, but it's invisible. If you stay there, the callout happens, but it's not immediate.
4. Hit Tab three more times - focus travels to Zoom In, Zoom Out, and Recenter, but it's still invisible
5. Hit Tab again and now it's visible in the property editor.

## Expected behavior

Buttons should have a distinctive appearance when focused.

## Screenshots

You can't see it, but focus is on the "zoom out" button here. If I hit Enter, it does the right thing.
image

alanlong9278:
## Description

Fix: focus wrong node when press tab. 
- Wrong behavior
 It will focus on action node & add button when pressing tab from `add` button

tab-err

- Expected behavior 
It should just focus on action node 
tab

## Task Item

closes #5292


Composer/packages/adaptive-flow/src/adaptive-flow-editor/renderers/NodeWrapper.tsx
```diff
@@ -38,6 +38,8 @@ export const ActionNodeWrapper: FC<NodeWrapperProps> = ({ id, tab, data, onEvent
   const nodeFocused = focusedId === id || focusedEvent === id;
   const nodeDoubleSelected = tab && nodeFocused && tab === focusedTab;
   const nodeSelected = selectedIds.includes(id);
+  const nodeId = `action-${selectableId}`;
+  const actionNode = document.getElementById(nodeId);
```
yeze322, line 42: Where is the `nodeId` built and can we use `actionRef` to focus the dom element?

alanlong9278: `nodeId` set as element id in `<div id={nodeId} ... />`
I see that `actionRef` is a callback and can not used at element. 
```
  const {
    shellApi: { addCoachMarkRef },
  } = useShellApi();
  const actionRef = useCallback((action) => {
    addCoachMarkRef({ action });
  }, []);
```

yeze322 (approved):

PR #5606, opened 2021-01-26
fix: Pop up the error message from persistence layer

### LINKED ISSUE #5578: failed writes do not cause an error in Composer
## Describe the bug

If a problem occurs while a file is being written, Composer does not throw an error or notify the user in any way.

We have a customer using a middleware plugin to add authentication, and when they throw errors, nothing happens!

## Version

1.3.0

## To Reproduce

1. Using the web app, open a bot in the dialog editor
2. Kill the webserver
3. continue to make edits in the client -- no errors will appear, though the writes will fail

## Expected behavior

* The UI should display an error
* Perhaps the UI should "lock" for further edits until a write is successful

lei9444:
## Description
Today our persistence layer do nothing when an error is thrown. In this PR add a pop up window to show the error message.

## Task Item
closes #5578

## Screenshots
error1


boydc2014:
Can we make the error message more clear? "Network error: fail to save the bots" or something else more specific

lei9444:
> Can we make the error message more clear? "Network error: fail to save the bots" or something else more specific

I have updated the dialog title 
image

benbrown:
If the server returns a specific error message, will this be displayed?

We have customers who need to throw specific errors, such as an access denied error with a custom message.

lei9444:
> If the server returns a specific error message, will this be displayed?
> 
> We have customers who need to throw specific errors, such as an access denied error with a custom message.

Yes, we support a custom message, and we need to put custom message like `    res.status(400).json({
      message: 'custom message',
    });`

srinaath:
@lei9444 @benbrown This message seems like it needs a little UX help. Eventhough, we are actually saving the contents of the bot "Failed to save the bot" seems a bit off. @sangwoohaan can you help here?

srinaath (changes requested):
@lei9444 @benbrown This message seems like it needs a little UX help. Even though, we are actually saving the contents of the bot in the background "Failed to save the bot" seems a bit off for the user to see.. @sangwoohaan can you help here?

@garypretty

boydc2014 (approved):

PR #5609, opened 2021-01-26
fix: yarn start:dev throw error

lei9444:
## Description
The version of react-dev-utils was upgraded to 11.0.1 in #5600 to fix immer  security error. But yarn start:dev will throw error at this version. 
image

I not sure why this will happen, but the version of immer(7.0.9) it used still has security issue. so I revert some of the code first to ensure the yarn start:dev is ok.

**Updates**
Add immer to resolutions to fix the version

## Task Item
refs #5600


VanyLaw:
Cool, It can work now.

srinaath (approved):

hatpick:
@srinaath can we have this merged asap? it keeps crashing on me.

PR #5620, opened 2021-01-27
feat: LG and LU folding range support in editors

### LINKED ISSUE #5332: Enhancement of LG and LU editing experience
# More comprehensive auto-completion
- [ ] Auto suggest entities from QnA and Luis recognizers results
Users can use entities from QnA and Luis recognizers, such as `@answer`(from QnA) or `@itemList`(from Luis), in LG, so the LG language server should be able to suggest these entities once user hit `@` in inline editor or all up view editor.

- [x] Dynamic property auto-completion
LSP should be able to complete the property user-defined via certain actions like Inputs, and SetPropertiy(s)

- [x] Support more static memory variables in LG language server auto suggestion. 
Currently, the static memory variables are missing some members, like class.xxx variables. Those variables can be found in the turnContext of a bot. We should add those missed ones to our static memory variables. 

# Correct return type info in hover
- [x] Hover on pre-built functions displays incorrect return type.

Decription: As shown in the image, the return type in hover actions is incorrect. It should display the correct type string instead of a integer. 
image
The mapping between an actual type and an integer is:
    Boolean = 1,
    Number = 2,
    Object = 4,
    String = 8,
    Array = 16

# Supporting folding templates in LG and folding utterance LU.
- [x] Add folding range actions for LG and LU language servers.

Description: As we did in the adaptive tool for VSCode extension. We should support folding range actions in Composer.
LG folding should based on the templates:
image

Lu should based on the utterances:
image

cosmicshuai:
fixes #5332 
implement  LG and LU folding range in language servers and enabled in the editor clients.

LG folding:
LG_folding

LU folding:
LU_folding


hatpick:
Nice, it would be great if you can add a short video of the code in action. ty

Composer/packages/tools/language-servers/language-generation/src/utils.ts
```diff
@@ -142,3 +142,18 @@ export const cardPropDict = {
   Attachment: ['contenttype', 'content'],
   Others: ['type', 'name', 'value'],
 };
+
+export function getCurrLine(document: TextDocument, lineCount: number, line: number) {
```
hatpick, line 146: you dont need to pass lineCount, document has it.

Better name for function is getLineByIndex

hatpick: Also, can you just ```document.getText().split('\n')``` to get the lines?

Composer/packages/tools/language-servers/language-understanding/src/utils.ts
```diff
@@ -70,3 +70,19 @@ export function convertDiagnostics(lgDiags: BFDiagnostic[] = [], document: TextD
   });
   return diagnostics;
 }
+
+export function getCurrLine(document: TextDocument, lineCount: number, line: number) {
```
hatpick, line 74: same comments as for LG util

Composer/packages/tools/language-servers/language-generation/src/utils.ts
cosmicshuai: Fixed the signature and naming. Also use `document.getText().split(/\r?\n/g)` to get line content.

Composer/packages/tools/language-servers/language-understanding/src/utils.ts
cosmicshuai: the same as LG util fix

hatpick (approved):

Composer/packages/lib/code-editor/src/BaseEditor.tsx
```diff
@@ -30,7 +30,7 @@ const defaultOptions = {
   lineDecorationsWidth: 10,
   lineNumbersMinChars: 3,
   glyphMargin: false,
-  folding: false,
+  folding: true,
```
a-b-r-o-w-n, line 33: Let's only override this setting in the LU and LG editors.

a-b-r-o-w-n (dismissed):

Composer/packages/lib/code-editor/src/BaseEditor.tsx
cosmicshuai: changed the lg folding option to LGEditor

boydc2014 (approved):

PR #5633, opened 2021-01-27
feat:  support channel enablement

### LINKED ISSUE #5632: Create: Allow Azure Bot Service channels to be enabled via settings
From the settings page, allow a user to toggle the teams, webchat and speech channels on and/off for each of their existing publishing profiles that contains a bot registration.

This will require the user to be logged in.

- [x] Require user to login
- [x] Display list of publishing profiles to select
- [x] Collect subscription id if not included
- [x] Query for current status of each channel
- [x] Allow channel to be toggled on or off

Still to do:
- [x] Collect speech key when enabling speech service
- [x] Customize the error messages, or remove all the case statements
- [x] Real aka.ms links for learn more
- [x] Testing

benbrown:
## Description

Allow a user to configure ABS channels, including webchat, teams and speech, directly from the settings page.

## Task Item

Fixes #5632

channels

## To do

- [ ] Collect speech key when enabling speech service
- [ ] Customize the error messages, or remove all the case statements
- [ ] Real aka.ms links for learn more
- [ ] Testing


Composer/packages/client/src/pages/botProject/ABSChannels.tsx
```diff
@@ -0,0 +1,674 @@
...
+            break;
+          case 'ResourceGroupNotFound':
+            // this resource group is not found - in other words, can't find a channel registration in the expected spot.
+            break;
+          case 'SubscriptionNotFound':
+            // the subscription is not found or invalid
+            break;
+          default:
+            // handle error.
+            break;
```
beyackle, line 237: Were you going to add some more error handling here?

Composer/packages/client/src/pages/botProject/ABSChannels.tsx
```diff
@@ -0,0 +1,674 @@
...
```
beyackle, line 335: Here too.

Composer/packages/client/src/pages/botProject/ABSChannels.tsx
```diff
@@ -0,0 +1,674 @@
...
+            // this resource group is not found - in other words, can't find a channel registration in the expected spot.
+            break;
+
+          case 'SubscriptionNotFound':
+            // the subscription is not found or invalid
+            break;
+
+          default:
+            // handle error.
+            break;
```
beyackle, line 376: With all this error handling, it might be easier to refactor this all into its own function.

Composer/packages/client/src/pages/botProject/ABSChannels.tsx
benbrown: I was going to consult with design on whether we want to provide custom error messages for these errors I observed.

So yes, probably!?

Composer/packages/client/src/pages/botProject/ABSChannels.tsx
benbrown: true!

PR #5635, opened 2021-01-28
fix: heap out of memory error when run yarn build

lei9444:
## Description
fix the immer version in resolutions will cause heap out of memory during yarn build.

remove the resolution first, but this may cause the security analysis fail 

## Task Item
#minor

### DIFF
--- a/Composer/package.json
+++ b/Composer/package.json
@@ -15,8 +15,7 @@
     "set-value": "^3.0.2",
     "terser-webpack-plugin": "^2.3.7",
     "axios": "^0.21.1",
-    "y18n": "^5.0.5",
-    "immer": "^8.0.1"
+    "y18n": "^5.0.5"
   },
   "engines": {
     "node": ">=12"


boydc2014:
@srinaath @hatpick can you help verify if this fixes this issue, i can't repro the original issue

boydc2014 (approved):

srinaath:
I dont think this fix is needed. I tried yarn build a couple of times on main. It seems to  work.

yeze322:
I got the same error on another branch, I bypassed it by increasing the max heap size.
https://github.com/microsoft/BotFramework-Composer/blob/3ac3ea526141bd702b889c94e6c64f88e6ef594c/Composer/packages/client/package.json#L11

srinaath:
@lei9444  @boydc2014 temporarily closing it. @hatpick and me tried again this morning on main and `yarn build` works. @yeze322 can you make sure to try again on main.

yeze322:
@srinaath I'm on the webchat branch, the error can appear. (encountered in two of my Windows PCs)
image

It's not a blocker to me, my workaround is increasing the heap size
image

PR #5636, opened 2021-01-28
fix: text input cursor err in runtime section

### LINKED ISSUE #5437: Text input jumps to last character after adding characters under runtime section
## Describe the bug
On the **Bot management and configurations** page, when entering text under the **Custom runtime** section the cursor jumps to the end of the input with any character. This does not match the behavior of similar text input above in sections like **External services**.
## Version
v1.3.0

## Browser
- [x] Electron distribution

## OS
- [x] Windows

## To Reproduce
1. Go to Bot management and configurations > Custom runtime
2. Add a character to the Runtime code location, in this sample adding 'A' as the first character.
image
3. Immediately after adding the character, the cursor moves to the end of the text input.
image

## Expected behavior
The cursor should stay where the user is typing, in this scenario I have a different custom runtime folder and wanted to reference it but had to move the cursor back for each character individually. This does not appear to be widespread but may be appearing in other instances.

Feel free to ping me for a demonstration if this doesn't make sense!

alanlong9278:
## Description

- Bugfix: Text input jumps to last character after adding characters as first character under runtime section.

- Reason: The value of `textField` is the props value, so the field component will be refresh after updating the props value.

- Solution：Set value as state value. 

## Task Item

closes #5437 

## Screenshots

1.  Wrong behavior
field

2. Expected behavior
correctfield


benbrown:
@tomlm this is what was happening to you the other day!! glad to see it fixed.

Composer/packages/client/src/pages/botProject/runtime-settings/RuntimeSettings.tsx
```diff
@@ -95,6 +102,12 @@ export const RuntimeSettings: React.FC<RouteComponentProps<{ projectId: string }
 
     setRuntimeField(projectId, field, newValue);
 
+    if (field === RuntimeType.Path) {
+      setRuntimePath(newValue);
+    } else {
+      setRuntimeCommand(newValue);
+    }
```
yeze322, line 109: I recommend that take the local state setters out side because when the onChange handler is triggered in the input box, it already knows where should the local state go.

Suggestion:
```jsx
// It doesn't return a function
const updateSetting = (field, val) => {
  // Error handling logic
  setRuntimeField(projectId, fieldId, val);
}

<TextField
  id="runtime-command"
  onChange={(e, val) => { // Can define this handler outside JSX
    setRuntimeCommand(val);
    updateSettings(val, 'command');
  }}}
/>
```

Composer/packages/client/src/pages/botProject/runtime-settings/RuntimeSettings.tsx
```diff
@@ -191,8 +204,8 @@ export const RuntimeSettings: React.FC<RouteComponentProps<{ projectId: string }
           errorMessage={formDataErrors.path}
           label={formatMessage('Runtime code location')}
           styles={name}
-          value={settings.runtime ? settings.runtime.path : ''}
-          onChange={updateSetting('path')}
+          value={runtimePath}
+          onChange={updateSetting(RuntimeType.Path)}
           onRenderLabel={onRenderLabel}
         />
```
liweitian, line 210: we can probably add a onBlur here, set setting only once.

liweitian (changes requested):

Composer/packages/client/src/pages/botProject/runtime-settings/RuntimeSettings.tsx
alanlong9278: > we can probably add a onBlur here, set setting only once.

Sync with other setting property.  Just update property when onBlur the text area.

Composer/packages/client/src/recoilModel/dispatchers/setting.ts
```diff
@@ -151,7 +151,7 @@ export const settingsDispatcher = () => {
   );
 
   const setRuntimeField = useRecoilCallback(
-    ({ set }: CallbackInterface) => async (projectId: string, field: string, newValue: boolean) => {
+    ({ set }: CallbackInterface) => async (projectId: string, field: string, newValue: boolean | string) => {
```
hatpick, line 154: is this necessary? how can a value be boolean or string?

alanlong9278: this is because runtime field could be string or boolean. We can set the string or boolean property in `setRuntimeField`
image

Composer/packages/client/src/pages/botProject/runtime-settings/RuntimeSettings.tsx
```diff
@@ -41,6 +41,11 @@ import {
   updateText,
 } from './style';
 
+const RuntimeType = {
```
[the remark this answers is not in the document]
alanlong9278: Sure

liweitian (approved):

PR #5638, opened 2021-01-28
feat: Alert a warning before remove a skill

### LINKED ISSUE #5132: When removing skills from project that's referenced, it should either cleanup the skill referenced in appsettings.json or show more meaningful error
## Description

When removing a connected skill from the bot project, any references in the root or other bot would error out with an error like this at the top level:

_"The skill 'dumbBot0' does not exist in in appsettings.jsonEchoBot-2"_

We can probably streamline this experience a bit by either doing:

1. Cleanup the project of any reference to the skill
2. Show more precisely inside the bot where things need to be fixed up, and maybe better phrasing because appsettings.json isn't something exposed readily in the Composer for people to edit nor is it actually what's required to fix the problem. Right now the error is top level, and in a large bot project this might be referenced in dozens of places. To fix, one simply has to delete the reference from the dialog, but it would help tremendously if the user knows where to go fix. 

## Version

                Bot Framework Composer

                Version:  1.2.1-nightly.191674.827857c
                Electron: 8.2.4
                Chrome: 80.0.3987.165
                NodeJS: 12.13.0
                V8: 8.0.426.27-electron.0
              
## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. From the main bot, connect to a new skill
2. Inside a dialog add reference to a skill (remote/local, doesn't matter)
3. Delete the skill

## Expected behavior

Either clear and concise way to tell the user how to resolve the problem and where to resolve the problem. Or
remove the skill from the dialog on behalf of the user.

zhixzhan:
## Description

When remove a skill from projectTree, check if it is used in rootBot and ask for confirm.

## Task Item
close #5132 

## Screenshots

Feb-01-2021 14-51-17


boydc2014:
Not sure it's related, but UT failed on this one. 

Also, @srinaath  can you confirm, is adding a warn is good enough.

hatpick:
What is the thinking behind not showing a dialog for a skill that is not used? Can we ask UX folks to comment on this?
(maybe different messaging compared to when the skill is used but confirming a destructive action such as delete is always good)

sangwoohaan:
Few changes : 
UX copy when removing a skill not being used
You are about to remove the skill from this project. Removing this skill won’t delete the files.

UX copy when removing a skill being used (I have modified the copy)
The skill you tried to remove from the project is currently used in the below bot(s). Removing this skill won’t delete the files, but it will cause your Bot to malfunction without additional action.

UI
Frame 53152

zhixzhan:
@sangwoohaan thanks for the updated design.
One more question is the modal buttons. the above UI are using "Cancel" / "Confirm",
but in composer other modals are using "Yes" / "Cancel", should we keep them aligns?

Composer/packages/client/src/recoilModel/selectors/skills.ts
```diff
@@ -79,3 +81,38 @@ export const skillNameIdentifierByProjectIdSelector = selector({
...
+      (result, skillId: string) => {
+        const usedInBots: { projectId: string; name: string }[] = [];
+        const skillNameIdentifier = get(botNameIdentifierState(skillId));
+
+        botIds.forEach((projectId: string) => {
+          const dialogIds = get(dialogIdsState(projectId));
+          let beenUsed = false;
+          dialogIds.forEach((dialogId) => {
+            const dialog = get(dialogState({ projectId, dialogId }));
+            console.log(dialog);
```
hatpick, line 101: remove this

Composer/packages/client/src/pages/design/SideBar.tsx
```diff
@@ -238,6 +238,27 @@ const SideBar: React.FC<SideBarProps> = ({ dialogId, projectId }) => {
...
+  async function handleRemoveSkill(skillId: string) {
+    // check if skill used in current project workspace
+    const usedInBots = skillUsedInBotsMap[skillId];
+    const confirmRemove = usedInBots.length
+      ? await OpenConfirmModal(formatMessage('Warning'), RemoveSkillDialog().subText, {
+          onRenderContent: () => {
+            return (
+              <div css={removeSkillDialogStyle}>
+                <div> {RemoveSkillDialog().subText} </div>
+                <div css={removeSkillDialogContentStyle}> {usedInBots.map(({ name }) => name).join('\r\n')} </div>
```
hatpick, line 250: \r ins not necessary

Composer/packages/client/src/constants.ts
```diff
@@ -305,6 +305,19 @@ export const repairSkillDialog = (name: string) => {
   };
 };
 
+export const RemoveSkillDialog = () => {
```
hatpick, line 308: pascalCase the name

hatpick (approved):

PR #5652, opened 2021-01-29
fix: fixes typos is lg and lu specs

hatpick:
## Description

fixes typos is lg and lu specs

## Task Item

#minor


hatpick:
@boydc2014 What do you think? If you think these changes are good, I can make this draft an actual PR for review.

beyackle (approved):
The changes look good - what effect do these typos have on the code editor itself?

hatpick:
@beyackle since the wrong spelling was used on both end of the "monarch" the syntax highlighting was doing what it was supposed to do, so no effect other than being spelled correct :)

PR #5654, opened 2021-01-29
fix: lg files auto add placeholder for missing templates

zhixzhan:
## Description

Our current lg CRUD operations are sync across multiple locales, to keep them structure aligns. but if the bot loaded initial status is not structure align (e.g migration from old version, edited in third party editor), authoring in Composer will meet a error, and so does running the bot.

For example a bot have two locale lg files:
 ```
<!-- a.en-us.lg -->                  |              <!-- a.zh-cn.lg --> 
# SendActivity_12345                 |
-hi                                  |
 ```
The searching scope of dialog used but missing templates can be <dialogId>.<*locale>.lg. for example 

And dialog 'a' is using a lg called 'SendActivity_12345' , will do auto add to lg files missing this template ( here is zh-cn, same to other locales fr-fr etc.).
 
After auto repair, two locale lg files will looks like:
 ```
<!-- a.en-us.lg -->                  |              <!-- a.zh-cn.lg --> 
# SendActivity_12345                 |              # SendActivity_12345
-hi                                  |              -
```

## Task Item
refs #5441 

## Screenshots

Real-time refreshing "readOnly" editor:

preview

### CODE REVIEW

File: Composer/packages/client/src/utils/lgUtil.ts [hunk 1]
```diff
@@ -1,10 +1,45 @@
...
+
 /**
- * lgUtil.ts is a single place use lg-parser handle lg file operation.
- * it's designed have no state, input text file, output text file.
- *
+ * Auto create placeholder template used in dialog but did not exist in <dialog>.<*locale>.lg
  */
 
-export * from '@bfc/indexers/lib/utils/lgUtil';
+export const createMissingLgTemplatesForDialogs = async (
```
srinaath, line 15: This fn is a good candidate for unit test.

zhixzhan: updated

boydc2014 (approved):

PR #5667, opened 2021-02-01
refactor: update message when resource already exists

irwinwilliams:
## Description

This is a nitpick 
Changing the error statement for when a resource group already exists.

#minor


a-b-r-o-w-n:
@irwinwilliams thanks for the PR! This code was recently refactored and the error message no longer applies. I am closing this PR for now, but feel free to open another if the current error message is not sufficient.

PR #5669, opened 2021-02-01
feat: abs handoff in composer

### LINKED ISSUE #5404: ABS handoff to Composer
Doc: https://microsoft-my.sharepoint-df.com/:w:/p/cwhitten/ERQSd0P1ciVGqQbTvBOZDE4Bl6upSFyDElAwNNgqwbi6iQ?e=xymxMM

Here is detailed work items to track this feature:
- [ ] Implement bfcomposer:// protocol for abs source, and take abs payload.
- [ ] Implement the logic of create new bot, or pull from abs based on the payload.
- [ ] Update the provision and publish part to bind the webapp and channel registration together.

VanyLaw:
## Description

- [x] create a bot with profile when user click "open in composer" in azure, implement through create protocol
- [x] open the existed bot when the alias is the same.

## Task Item

close #5404


extensions/azurePublish/src/node/deploy.ts
```diff
@@ -195,4 +208,158 @@ export class BotProjectDeploy {
...
+
+  /**
+   * Bind key vault settings to webapp
+   * @param absSettings
+   */
+  private async BindKeyVault(absSettings: any, hostname: string) {
+    const subscriptionId = absSettings.subscriptionId;
+    const resourceGroupName = absSettings.resourceGroup;
+    const webAppName = hostname;
+    const hint = absSettings.appPasswordHint;
```
luhan2017, line 282: @zidaneymar , we should check if the appPasswordHint has value and if the value is valid to generate a keyvault URL here.

extensions/azurePublish/src/node/provision.ts
```diff
@@ -270,7 +271,9 @@ export class BotProjectProvision {
               displayName: config.hostname, // todo: this may be wrong!
               endpoint: `https://${provisionResults.webApp.hostname}/api/messages`,
               appId: provisionResults.appId,
+              webAppHostname: provisionResults.webApp.hostname
             });
+            provisionResults.bot = config.hostname;
```
luhan2017, line 276: Why we added a bot field here? What is this used for, what's the difference with hostname?

extensions/azurePublish/src/node/deploy.ts
```diff
@@ -195,4 +208,158 @@ export class BotProjectDeploy {
...
+      throw createCustomizeError(AzurePublishErrors.ABS_ERROR, botUpdateResult._response?.bodyAsText);
+    }
+  }
+
+  /**
+   * Bind key vault settings to webapp
+   * @param absSettings
+   */
+  private async BindKeyVault(absSettings: any, hostname: string) {
+    const subscriptionId = absSettings.subscriptionId;
```
luhan2017, line 279: Is it possible that the keyvault hint are in difference subscriptionId or resource group? I think to keep this part clean, we should only use the keyvault hint to generate the keyvault url. in the keyvault hint, we did have subscription and resource group information. Let's use that.

extensions/azurePublish/src/node/deploy.ts
luhan2017: @zidaneymar , also the hintsList.length should be a fixed value, if not, just return.  and pop up error, failed to parse the passwordHint.

Composer/packages/client/src/components/CreationFlow/CreateOptions.tsx
```diff
@@ -246,6 +248,34 @@ export function CreateOptions(props: CreateOptionsProps) {
...
+          })
+          .catch((e) => {
+            setIsOpen(true);
+          });
+      } else {
+        setIsOpen(true);
+      }
+    } else {
+      setIsOpen(true);
+    }
```
tonyanziano, line 276: can this be simplified to:

```
if (props.location.search) {

  ...

  if (payload && typeof payload === 'string' && source === 'abs') {
  
    ...
  
    axios.get
      .then(
        ...
        navigate
        return
      )
      .catch(setIsOpen(true))
   }
}
setIsOpen(true)

```

It seems like the two `else` blocks are redundant since we only want to navigate if we find a matching alias, otherwise we want to `setIsOpen(true)`

Composer/packages/client/src/components/CreationFlow/DefineConversation.tsx
```diff
@@ -72,6 +72,8 @@ interface DefineConversationFormData {
   schemaUrl: string;
   location?: string;
 
+  profile?: any; // abs payload to create bot
```
tonyanziano, line 75: can this be typed?

Composer/packages/client/src/recoilModel/dispatchers/project.ts
```diff
@@ -308,6 +309,25 @@ export const projectDispatcher = () => {
...
+            hostname: profile.tag?.webapp || '',
+            runtimeIdentifier: 'win-x64',
+            settings: {
+              MicrosoftAppId: appId,
+            },
+            ...profile,
+          }),
+        };
+        dispatcher.setPublishTargets([newProfile], projectId);
+      }
```
tonyanziano, line 330: Can this be more generalized? What if several months from now, we do another creation flow integration and need to add a provisioning profile after creation that is not an azure publish profile?

If you look at the import flow I wrote for PVA, I tried to make the import flow itself generic, but abstracted away the PVA-specific logic into the `powerVirtualAgentsProvider.ts` 

Could a similar thing be done here where the publish profile type to be added is specified in the deep link payload, and then based on the publish profile type we fill out the rest of the publishing profile in another function / class?

I just worry about scale.

Composer/packages/server/src/externalContentProvider/azureBotServiceProvider.ts
```diff
@@ -0,0 +1,103 @@
+import { IContentProviderMetadata, ExternalContentProvider } from './externalContentProvider';
+// import { authService } from '../services/auth/auth';
+// import logger from '../logger';
```
tonyanziano, line 3: remove

Composer/packages/server/src/externalContentProvider/azureBotServiceProvider.ts
```diff
@@ -0,0 +1,103 @@
...
+import { join } from 'path';
+import { remove, ensureDirSync } from 'fs-extra';
+// import { urlencoded } from 'body-parser';
```
tonyanziano, line 6: remove

Composer/packages/server/src/externalContentProvider/azureBotServiceProvider.ts
```diff
@@ -0,0 +1,103 @@
...
+
+function prettyPrintError(err: string | Error): string {
+  if (typeof err === 'string') {
+    return err;
+  }
+  if (err && err.message) {
+    return err.message;
+  }
+  return '';
+}
```
tonyanziano, line 16: This exact same function exists in `powerVirtualAgentsProvider.ts`. Extract it out into a separate file so it can be reused and we don't have duplicate function definitions.

Composer/packages/server/src/externalContentProvider/azureBotServiceProvider.ts
```diff
@@ -0,0 +1,103 @@
...
+    }
+    ensureDirSync(this.tempBotAssetsDir);
+    const zipPath = join(this.tempBotAssetsDir, 'EchoBot-1.zip');
+
+    // const writeStream = createWriteStream(zipPath);
+    // await new Promise((resolve, reject) => {
+    //   writeStream.once('finish', resolve);
+    //   writeStream.once('error', reject);
+    //   result.body.pipe(writeStream);
+    // });
```
tonyanziano, line 52: remove

Composer/packages/server/src/externalContentProvider/azureBotServiceProvider.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  public async downloadBotContent() {
+    const { resourceId } = this.metadata;
+    /**
+     * TODO: get download url from payload, then download zip
+     * then write the zip to disk
+     */
+    if (resourceId) {
+      // get resource by resourceId
+    }
```
tonyanziano, line 43: needed?

Composer/packages/server/src/externalContentProvider/azureBotServiceProvider.ts
```diff
@@ -0,0 +1,103 @@
...
+  }
+  private async getAccessToken(): Promise<string> {
+    try {
+      // const accessToken = await authService.getAccessToken({
+      //   targetResource: 'https://management.core.windows.net/',
+      // });
+      // if (accessToken === '') {
+      //   throw 'User cancelled login flow.';
+      // }
+      // return accessToken;
```
tonyanziano, line 77: remove

extensions/azurePublish/src/node/deploy.ts
```diff
@@ -43,9 +47,19 @@ export class BotProjectDeploy {
     name: string,
     environment: string,
     hostname?: string,
-    luisResource?: string
+    luisResource?: string,
+    absSettings?: any
   ) {
     try {
+      console.log(absSettings);
```
tonyanziano, line 54: remove or use `this.logger`

tonyanziano (changes requested):

extensions/azurePublish/src/node/provision.ts
luhan2017: Let's keep this bot field for now, I will follow up with you offline. @zidaneymar @VanyLaw

extensions/azurePublish/src/node/deploy.ts
```diff
@@ -43,9 +47,19 @@ export class BotProjectDeploy {
...
+      if (absSettings) {
+        await this.linkBotWithWebapp(absSettings, hostname);
```
luhan2017, line 56: @zidaneymar don't make it abs specific, we should do this as long as the subscriptionid, resourceGroup and botName is available, let's follow up on this offline.

extensions/azurePublish/src/node/deploy.ts
```diff
@@ -197,4 +208,185 @@ export class BotProjectDeploy {
...
+   * link the bot channel registration with azure web app service
+   * @param absSettings the abs settings
+   * @param hostname the hostname of webapp which bot service would be linked to
+   */
+  private async linkBotWithWebapp(settings: any, absSettings: any, hostname: string) {
+    let subscriptionId = settings.subscriptionId;
+    let resourceGroupName = settings.resourceGroup;
+    let botName = settings.botName;
+
+    if (!absSettings.resourceId || !hostname) {
```
luhan2017, line 222: Remove this logic, we want to do this even if the resourceId is not available.

extensions/azurePublish/src/node/deploy.ts
```diff
@@ -197,4 +208,185 @@ export class BotProjectDeploy {
...
+    const creds = new TokenCredentials(this.accessToken);
+
+    const keyVaultApiConfig = {
+      creds: creds,
+      logger: this.logger,
+      subscriptionId: subscriptionId
+    } as KeyVaultApiConfig;
+    const keyVaultApi = new KeyVaultApi(keyVaultApiConfig);
+
+    await keyVaultApi.WebAppAssignIdentity(resourceGroupName, webAppName);
```
luhan2017, line 327: Do we still need this logic after we change the set password logic?

Composer/packages/client/src/recoilModel/dispatchers/project.ts
```diff
@@ -308,6 +311,13 @@ export const projectDispatcher = () => {
       );
       set(botProjectIdsState, [projectId]);
 
+      if (profile) {
+        // ABS Create Flow, update publishProfile after create project
+        const dispatcher = await snapshot.getPromise(dispatcherState);
+        const newProfile = getPublishProfileFromPayload(profile, source);
+
+        newProfile && dispatcher.setPublishTargets([newProfile], projectId);
+      }
```
tonyanziano, line 320: Nice! 👍

Composer/packages/electron-server/src/utility/url.ts
```diff
@@ -10,14 +10,27 @@ export const parseDeepLinkUrl = (deeplinkUrl: string) => {
   try {
     const convertedUrl = new URL(deeplinkUrl);
     const action = lowerCase(convertedUrl.hostname);
+    console.log(action);
```
tonyanziano, line 13: remove

tonyanziano (approved):

luhan2017 (dismissed):

luhan2017 (approved):

PR #5689, opened 2021-02-03
fix: the recognizer dialog created is for LUIS even if Orchestrator is selected in dropdown

lei9444:
## Description
**In this PR**
 check the dialog recognizer type, if it is Orchestrator recognizer type, generate the recognizer dialog with orchestrator recognizer template

orch

## Task Item
#minor


Composer/packages/client/src/recoilModel/Recognizers.tsx
```diff
@@ -148,40 +165,43 @@ export const Recognizer = React.memo((props: { projectId: string }) => {
   const qnaFiles = useRecoilValue(qnaFilesState(projectId));
   const settings = useRecoilValue(settingsState(projectId));
   const curRecognizers = useRecoilValue(recognizersSelectorFamily(projectId));
+  const curRecognizersRef = useRef(curRecognizers);
   const filePersistence = useRecoilValue(filePersistenceState(projectId));
+  curRecognizersRef.current = curRecognizers;
```
taicchoumsft, line 170: Not sure if we should move this to the `useEffect` just to be safe?

taicchoumsft (approved):
Tested and it works as expected.  Signing off.

Composer/packages/client/src/recoilModel/Recognizers.tsx
```diff
@@ -148,40 +165,43 @@ export const Recognizer = React.memo((props: { projectId: string }) => {
...
-        setRecognizers(recognizers);
-      }
+    if (isEmpty(filePersistence)) return;
+    let recognizers: RecognizerFile[] = [];
+
+    dialogs
+      .filter((dialog) => isCrossTrainedRecognizerSet(dialog) || isLuisRecognizer(dialog))
+      .forEach((dialog) => {
+        const luProvide = getLuProvider(dialog.id, curRecognizersRef.current);
+        const filtedLus = luFiles.filter((item) => getBaseName(item.id) === dialog.id);
```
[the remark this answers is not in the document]
lei9444: Thanks @taicchoumsft , I think I need to improve my english

Composer/packages/client/src/recoilModel/Recognizers.tsx
lei9444: I'm afraid that the setRecognizer will trigger the  calculation again if I put it to the useEffect dependency

boydc2014 (approved):

Issue #4370, opened 2020-10-14
Add electron build & test into the nightly CI and/or PR CI pipelines


cwhitten:
Too often we get unhealthy electron builds in the nightly feed because we don't test this distribution in our CI environments. Let's address that by adding some automation.

P0: Add steps to install & test the nightly package before publishing. Upon failure, exit the build and report the error to a Teams channel
P1: Add steps to install & test the PR bits as part of the CI suite. Upon failure, exit the build and report the error to a Teams channel

We can take the opportunity to add a new Teams channel for these types of system notifications

PR #5692, opened 2021-02-04
chore: add recognizer files for sample bots

lei9444:
## Description
our samples don't have recognizers today, the samples can't start directly

**In this PR**
Add recognizer dialog if the bot needs them

## Task Item
refs #4370


boydc2014:
Note: this is a step which we didn't close when we are adding recognizers as a top level folder. 

And this is now needed when we do the daily CI pipeline, which are requiring the capability to test a bot directly.

boydc2014 (approved):

cwhitten:
Closing. Samples are being deprecated

PR #5693, opened 2021-02-04
fix: add 100px marign right for branching nodes (SwitchConditon, IfCondition)

### LINKED ISSUE #5637: Wide switch statements don't show last condition
## Describe the bug
If I have a switch statement with many conditions the UI doesn't display the last condition.

## Version
1.3

## Browser

- [ ] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [x] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Create a switch statement with several conditions (6 or more)

image

2. The last condition won't show on the design surface and you can't scroll to see it.

It will show once you add an action to it but you won't know the condition unless you read the switch statement or until you add that action. 

image

## Expected behavior

I would expect to be able to scroll and see the condition without having to add an action.

yeze322:
## Description
closes #5637 

With an extra 100px margin, the last case label of a SwitchConditon can be visible by scrolling right.
image


boydc2014 (dismissed):

boydc2014:
@yeze322 UT for boundary test seems to be failed

FAIL adaptive-form packages/adaptive-flow/__tests__/adaptive-flow-renderer/layouters/calculateNodeBoundary.test.ts
  ● calculateSwitchCaseBoundary › should return a box whose property be calcalated by conditionBoundary, choiceBoundary, branchBoundaries

boydc2014 (approved):

PR #5700, opened 2021-02-04
fix: Pass access token for publishing through request body instead of header

### LINKED ISSUE #5650: Provision Failure Error When Attempting to Create a Publishing Profile
## Version

1.3

## Browser

N/A

## Operating System

- [X] macOS
- [X] Windows
- [ ] Ubuntu

## To Reproduce

1. Go to "Bot Management and Configuration".
2. Scroll to the "Publish Targets" section at the bottom of the screen.
3. Click "Add new publish profile" link.
4. Enter profile name and select "Publish bot to Azure Web App" for the "Publish Target".
5. Accept the default "Create new Azure Resources" option.
6. Select Azure subscription.
7. Enter name of resource group.
8. Enter location.
9. Accept default values on the "Review + Create" screen.
10. See error

## Expected behavior

I expected to have each of the selected resources (resource group and accompanying required services) provisioned in Azure.

## Screenshots

Screen Shot 2021-01-28 at 3 53 53 PM

Screen Shot 2021-01-28 at 3 55 35 PM

Screen Shot 2021-01-28 at 3 58 09 PM

3C9B2A94-E2FE-4942-9434-25CF45437DB2_4_5005_c

tonyanziano:
## Description

Due to a limitation in the version of NodeJS that Composer runs on (< 13), requests sent to the server are limited to a header size of 8 KB. This limitation was blocking Azure publishing and provisioning requests from the client to the server, which was ultimately preventing publishing from happening in certain cases where the Bearer token received from Azure was larger than 8 KB.

This threshold can be increased manually by starting Composer with the environment variable of `NODE_OPTIONS` set to `"--max-http-header-size=16000"`.

This fix resolves that issue by embedding the access token in the body of the request from the client to the server, instead of in the Authorization header. This should bypass the limitation and allow Composer's server to make Azure publishing requests on behalf of the client using the provided token.

## Task Item

fixes #5650


Composer/packages/server/src/controllers/publisher.ts
```diff
@@ -68,9 +68,7 @@ export const PublishController = {
     const profile = profiles.length ? profiles[0] : undefined;
     const extensionName = profile ? profile.type : ''; // get the publish plugin key
 
-    // get token from header
-    const accessToken = req.headers.authorization?.substring('Bearer '.length);
-    log('access token get from header: %s', accessToken);
+    log('access token retrieved from body: %s', accessToken || 'no token provided');
```
a-b-r-o-w-n, line 71: Is this wise to log the access token?

a-b-r-o-w-n (approved):
Approved, but I have 1 question.

Composer/packages/server/src/controllers/publisher.ts
tonyanziano: We do it in the auth service as well. Since you have to enable logs via the `DEBUG=composer*` environment variable it seems ok to me. But we can remove it if you think otherwise.

PR #5710, opened 2021-02-07
chore: fix several uischema hook issues

### LINKED ISSUE #6014: Unexpected list size for Teams triggers UI
## Describe the bug

When adding a trigger for teams, the displayed popup appears larger than expected (extra white space).

## Version

latest from main branch, local build, commit d9ee12c

## Browser

- [ ] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [x] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. select "Add a Trigger"
2. Choose "Microsoft Teams" for the type
3. Click the "Which Teams Event"

## Expected behavior

A popup no larger than the list (like the others).

## Screenshots

composer-teams-events

yeze322:
## Description
#minor
Partially fixes #6014 

Check in necessary changes to support sdk customization in Composer.

**Changes**:
- Resolve title from sdk.schema automatically
- Hide `actions` field from Form Editor
- Update UT


srinaath:
@yeze322 @garypretty  Teams menu under actions needs to be renamed "Microsoft Teams" to give users more context

cwhitten (approved):

Composer/packages/ui-plugins/teams/src/triggers.ts
```diff
@@ -0,0 +1,157 @@
...
+
+import { UISchema } from '@bfc/extension-client';
+import { SDKKinds } from '@bfc/shared';
+import formatMessage from 'format-message';
+
+export default {
+  [SDKKinds.OnAppBasedLinkQuery]: {
+    trigger: {
+      submenu: {
+        label: formatMessage('Microsoft Teams'),
```
beyackle, line 12: It might be better to define constants before this object (e.g. `const TEAMS_LABEL = formatMessage('Microsoft teams')`) and then use those for messages that repeat a lot.

Composer/packages/ui-plugins/teams/src/actions.ts
```diff
@@ -0,0 +1,161 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { UISchema } from '@bfc/extension-client';
+import { SDKKinds } from '@bfc/shared';
+import formatMessage from 'format-message';
+
```
beyackle, line 7: See the comment below (on triggers.ts) about defining constants for common values.

Composer/packages/lib/shared/src/lgUtils/stringBuilders/buildLgType.ts
```diff
@@ -13,7 +13,7 @@ import LgType from '../models/LgType';
  */
 export default function buildLgType(lgType: LgType): string {
   const { hostKind, hostField } = lgType;
-  const kindPrefix = hostKind.replace('Microsoft.', '');
+  const kindPrefix = hostKind.replace('Microsoft.', '').replace('.', '');
```
beyackle, line 16: This seems like an odd change to me. Can you give me an example of where we'd need to get rid of more periods?

Composer/packages/lib/shared/src/luNameBuilder/stringBuilders/buildLuType.ts
```diff
@@ -12,7 +12,7 @@ import { LuIntentName } from '../models/stringTypes';
  */
 export default function buildLuType(luType: LuType): LuIntentName {
   const { hostKind, type } = luType;
-  const kindPrefix = hostKind.replace('Microsoft.', '');
+  const kindPrefix = hostKind.replace('Microsoft.', '').replace('.', '');
```
beyackle, line 15: Same question here.

beyackle (changes requested):

Composer/packages/lib/shared/src/lgUtils/stringBuilders/buildLgType.ts
EricDahlvang: Should this be `.replace('Teams.','');` ?

yeze322: @benbrown I escaped the `.` here because $kind is used to generate lg template name, however, lg utils can only recongize \w+ pattern, so there is a trailing replace statement to remove all dots.

@EricDahlvang I didn't write `replace('Teams.')` here is because of Teams namespace logic shouldn't be exposed to Composer domain. We can remove 'Microsoft.' suffix because we know all builtin $kinds are in the `Microsoft.` namespace, but for the namespace `Teams.`, I consider it as an action extension.

yeze322:
@beyackle Thanks for the careful review Ben! I planned to resolve your comments but according to some recent discussions around Teams scenario, I'm going to check in those uischema to dotnet repo directly.

yeze322:
Teams is out of the scope of R12. This PR aims at fixing several issues found during Teams development

yeze322:
validated the Teams scenario
image

cwhitten (approved):

PR #5711, opened 2021-02-07
fix: Add subscription id to publish profile

### LINKED ISSUE #5577: Collect the Azure subscription ID as part of the publishing profile
The subscription id is needed to take follow-up actions on the bot registration and other resources.  For example, managing channels on ABS will require this value, and if not present, we will have to prompt the user.

This field should be added to the publishing profile at the root level of the object.

- [ ] In Composer
- [ ] In azureProvision.js cli tool

zidaneymar:
## Description
add subscription id to publish profile

## Task Item
fixes #5577


luhan2017 (approved):

benbrown:
@luhan2017 we also need to add this file to the schema so that the json editor knows about it...

PR #5717, opened 2021-02-08
fix: client build fail in docker [TEST]

lei9444:
## Description
test

## Task Item
#minor


boydc2014:
Looks it still fails, may due to memory?

lei9444:
This PR can't fix the issue

PR #5718, opened 2021-02-08
fix: Composer nav + button discoverability

### LINKED ISSUE #5680: Composer nav + button discoverability
The + button in the composer nav is not discoverable and is creating friction while trying to add skills.
the create dialog and trigger icons in the three dot menu though contextual is hard to discover 

image

Suggested fix:

image

lei9444:
## Description
new Design: https://www.figma.com/file/deFhMPBnRWxTYiYDYqYxrc/Composer---Vivek?node-id=10809%3A0
image

filter1

This PR also fix one bug. The bug is the filter is used in some function, and the function can only get the stale value, so when we input the text into the search box, The filter doesn't work
image
In this PR use Ref to make sure the value is latest.

## Task Item
closes #5680


boydc2014 (approved):

hatpick:
Does this PR also solve the following issue:
(header of project tree is scrolling with it)
https://user-images.githubusercontent.com/1483492/107267236-74aaf200-69fb-11eb-98fe-25a59ecddde4.mov

liweitian:
> Does this PR also solve the following issue:
> (header of project tree is scrolling with it)
> https://user-images.githubusercontent.com/1483492/107267236-74aaf200-69fb-11eb-98fe-25a59ecddde4.mov

From the design, the header is removed. @vivekkshankar could you help verify on this? The search bar is also changed on lu/lg/qna page. Should we do that?

Composer/packages/client/src/components/ProjectTree/ProjectTreeHeader.tsx
```diff
@@ -51,54 +118,47 @@ export const ProjectTreeHeader: React.FC<ProjectTreeHeaderProps> = ({ menu }) =>
...
+          ariaLabel={formatMessage('Type trigger name')}
+          componentRef={searchBoxRef}
+          iconProps={{ iconName: 'Filter' }}
+          placeholder={formatMessage('Filter by trigger name')}
+          styles={searchBox}
+          onBlur={handleSearchBoxBlur}
+          onChange={onFilter}
+        />
+      ) : (
         <OverflowSet
```
hatpick, line 140: why do you need this overflow set? Simply used a commandbar button
(https://developer.microsoft.com/en-us/fluentui#/controls/web/button)

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
```diff
@@ -197,6 +187,8 @@ export const ProjectTree: React.FC<Props> = ({
   const setPageElement = (name: string, value) => setPageElementState('dialogs', { ...pageElements, [name]: value });
 
   const [filter, setFilter] = useState('');
+  const filterRef = useRef(filter);
```
hatpick, line 190: I dont understand what's happening with this ref, what does it represent? and why is it needed?

Composer/packages/client/src/components/ProjectTree/ProjectTreeHeader.tsx
```diff
@@ -34,9 +51,59 @@ export interface ProjectTreeHeaderMenuItem {
 
 export interface ProjectTreeHeaderProps {
   menu: ProjectTreeHeaderMenuItem[];
+  onFilter?: (_e, newValue?: string) => void;
```
hatpick, line 54: if you dont want to use the event why are you including it in the callback?

Composer/packages/client/src/components/ProjectTree/ProjectTreeHeader.tsx
```diff
@@ -3,27 +3,44 @@
...
+    width: '100%',
+  },
+};
+
+const buttonStyle = css`
+  height: 100%;
+`;
 
 const headerText = css`
+  border-bottom: 1px solid #edebe9;
```
hatpick, line 28: use a color from Fluent instead of hard coding it

Composer/packages/client/src/components/ProjectTree/ProjectTreeHeader.tsx
```diff
@@ -3,27 +3,44 @@
...
-import { TooltipHost, DirectionalHint } from 'office-ui-fabric-react/lib/Tooltip';
+import { CommandButton } from 'office-ui-fabric-react/lib/Button';
+import { IOverflowSetItemProps, OverflowSet } from 'office-ui-fabric-react/lib/OverflowSet';
 import { IContextualMenuItem } from 'office-ui-fabric-react/lib/ContextualMenu';
+import { ISearchBox, ISearchBoxStyles, SearchBox } from 'office-ui-fabric-react/lib/SearchBox';
 
-import { moreButton, overflowSet } from './treeItem';
+const searchBox: ISearchBoxStyles = {
+  root: {
+    borderBottom: '1px solid #edebe9',
```
hatpick, line 16: same here for the color
https://developer.microsoft.com/en-us/fluentui#/styles/web/colors/neutrals
gray30

Composer/packages/client/src/components/ProjectTree/ProjectTreeHeader.tsx
```diff
@@ -34,9 +51,59 @@ export interface ProjectTreeHeaderMenuItem {
...
+      text={item.name}
+      onClick={item.onClick}
+    />
+  );
+};
+
+const onRenderOverflowButton = (isActive: boolean) => {
+  const moreLabel = formatMessage('Actions');
+  return (overflowItems: IContextualMenuItem[] | undefined) => {
+    if (overflowItems == null) return null;
```
hatpick, line 76: if(!overflowItems) {
   return null;
}

nit: avoid using Abstract Equality Comparison (==) vs. ===

hatpick (changes requested):

hatpick:
> > Does this PR also solve the following issue:
> > (header of project tree is scrolling with it)
> > https://user-images.githubusercontent.com/1483492/107267236-74aaf200-69fb-11eb-98fe-25a59ecddde4.mov
> 
> @hatpick From the design, the header is removed. @vivekkshankar could you help verify on this? The search bar is also changed on lu/lg/qna page. Should we do that?

When the project tree list is long and user scrolls, the "header" that you have add button and filtering in it will scroll oit of view with it, @vivekkshankar is that what you wanted for ux?

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
lei9444: Hi, @hatpick , There is bug in main branch, I fixed it in this PR. The bug is the filter doesn't work when we input some text. The reason is I found the filter text is used in a function, but react will use the stale value anytime
image

lei9444:
> > > Does this PR also solve the following issue:
> > > (header of project tree is scrolling with it)
> > > https://user-images.githubusercontent.com/1483492/107267236-74aaf200-69fb-11eb-98fe-25a59ecddde4.mov
> > 
> > 
> > @hatpick From the design, the header is removed. @vivekkshankar could you help verify on this? The search bar is also changed on lu/lg/qna page. Should we do that?
> 
> When the project tree list is long and user scrolls, the "header" that you have add button and filtering in it will scroll out of the viewport, @lei9444 @vivekkshankar is that what you wanted for ux?

Hi, @hatpick I think we need to fix the header, I have fixed in this PR 
image

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
hatpick: I have created a hook in my local branch for another feature:
```
export const useDebounce = <T>(value: T, delay: number) => {
  const [debouncedValue, setDebouncedValue] = useState<T>(value);

  useEffect(() => {
    const handler = setTimeout(() => {
      setDebouncedValue(value);
    }, delay);

    return () => {
      clearTimeout(handler);
    };
  }, [value]);

  return debouncedValue;
};
```
you can consume it for this exact case like:
```
const [query, setQuery] = React.useState<string | undefined>();
const debouncedQuery = useDebounce<string | undefined>(query, 300);
```

and then you deal with ```debouncedQuery```

wanna give it a try? if it works as expected you can add that hook to share library so others can use it too.

srinaath:
@hatpick Are we good with this PR? It removes a P0 of our plate.

Composer/packages/client/src/components/ProjectTree/ProjectTree.tsx
lei9444: Thanks @hatpick , I think I didn't explain the issue very clearly. 
I think there is a wrong usage in < ProjectTree />: Some children components will use the filter message to re-render the UI, but all these components will call a function from parent to filter their items and they don't depends on the filter message change, so the function will only use the stale value. 

So I think I just need to listen to the filter state and re-render these components when filter change. I fixed it in the latest commit and removed the Ref.

I think it's a good idea to share the useDebounce to the library, the filter state in ProjectTree also use a  similar implementation with throttle. Can we do this in a new PR?

cwhitten (approved):

PR #5721, opened 2021-02-08
feat: Add SetSpeakMiddleware to WebApp and Functions runtimes

### LINKED ISSUE #5720: Add SetSpeakMiddleware
To support the directline speech and telephony channels, we need a mechanism to ensure the appropriate SSML tags are set on activity Speak properties.  This will be added as middleware and will work as below.

- There should be a flag in settings to allow a user to automatically set the activity Speak property, if empty, to the value of the Text property.
- If it is not already present, wrap the speak property value with a speak and a voice tag, setting the voice font (specified within settings). If the speak and voice tags are already present, then the value is left alone, allowing a developer to override.

The same middleware will be added to the new runtime and this is being tracked at https://github.com/microsoft/botframework-components/issues/488

garypretty:
Fixes #5720. Adds SetSpeakMiddleware to the current runtime for webapp and functions.


luhan2017:
@garypretty overall looks good to me, the only thing is that, do we need to have a separated language configuration for speech? can we just leverage the language of LG? Composer support multiple language today, I think the speech language should also be dynamic.

luhan2017 (approved):

luhan2017 (approved):

cwhitten (approved):

PR #5733, opened 2021-02-09
fix: update bf-lu version to resolve app settings missing issue in luis publishing

### LINKED ISSUE #5732: Update bf-lu package version to resolve the app settings missing issue
## Describe the bug

Current bf-lu has a bug that app settings defined in lu files may be lost when doing luis publishing. There was a fix in bf-cli side https://github.com/microsoft/botframework-cli/pull/1108 to resolve the issue. Composer need to upgrade the bf-lu version to overcome this issue.

## Version

Latest main

## Browser

- [ ] Electron distribution
- [x] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Add a lu file and add app settings in the header and then import another lu file
2. Publish this lu file
3. You will see app settings didn't take effect in luis portal

## Expected behavior

App settings should work for luis publishing

feich-ms:
## Description

Upgrade bf-lu package to resolve app settings missing issue in luis publishing

## Task Item
Closes #5732


srinaath:
@chrimc62 In this PR we are upgrading bf-lu which in turn upgrades the bf-cli to incorporate the microsoft/botframework-cli#1108.

feich-ms:
Hi @srinaath, I have another bf-lu fix that was just merged a few hours ago and the nightly build supposed to be released soon. I will update this PR to that latest bf-lu version once it comes out so that the composer can consume all the latest R12 bf-cli fixes. I will remove [Do Not Merge] in title once it's done. Thanks.

feich-ms:
This PR is ready for merging.

srinaath:
@feich-ms can we pick the latest bf-lu and get this PR merged.

feich-ms:
@srinaath I have updated the bf-lu package version to the latest 4.12.0-rc0. But the Docker Build Validation and Security Analysis validation failed in CI, do you have any insights how to make them pass.

srinaath:
@feich-ms Lets get it merged. The docker build failure is not because of this PR.

srinaath:
@feich-ms it looks like static-eval 2.0.2 is the comp governance issue. Can u use yarn resolutions to update it? Right here[BotFramework-Composer/package.json at main · microsoft/BotFramework-Composer (github.com)](https://github.com/microsoft/BotFramework-Composer/blob/main/Composer/package.json#L5)

feich-ms:
@srinaath the static-eval 2.0.2 package is in runtime/node folder that is using npm to manage the packages. I tried add the resolution in runtime/node/package.json, but it seems npm doesn't support resolutions. I also tried added it to the resolution of Composer/packagejson, but it yarn install didn't update the package-lock.json in runtime/node folder. Do you have any suggestions?

feich-ms:
@srinaath we did much investigation today. It looks like not a good choice to switch from npm to yarn for node runtime package management since that will require users to install yarn additionally if they eject js runtime. To fix the `static-eval 2.0.2 package` issue completely, we suggest to fix that from the root package that used statuc-eval. Here is the fix PR from js sdk https://github.com/microsoft/botbuilder-js/pull/3331. We already asked @chon219 to cherry-pick the fix to next js sdk rc release. After the fix is out, we can update the packages in runtime/node from composer side to consume the latest js sdk which will not have dependency on static-eval any more.

boydc2014 (approved):

PR #5736, opened 2021-02-09
fix: auto-complete behavior for memory variables and add support of suggest Luis entities

### LINKED ISSUE #5332: Enhancement of LG and LU editing experience

### LINKED ISSUE #5523: Support auto-complete in LG editor when user is typing characters
As a user, when I editing in LG editor with memory variables, the auto-complete worked with "." as below 
image

The auto-complete should work if user is typing some characters.
image

cosmicshuai:
closes #5523 
closes #5332 

1. Fix auto-complete behavior for memory variables
2. Support suggest Luis entities defined in the .lu files.

The annimation of these two fixes
fix suggest incomplete


Composer/packages/lib/indexers/src/dialogUtils/extractLUEntities.ts
```diff
@@ -0,0 +1,2 @@
+// Copyright (c) Microsoft Corporation.
```
hatpick, line 1: what is this file?

Composer/packages/server/src/server.ts
```diff
@@ -164,13 +164,13 @@ export async function start(electronContext?: ElectronContext): Promise<number |
...
 
-  const { getLgResources, luImportResolver, staticMemoryResolver } = BotProjectService;
+  const { getLgResources, luImportResolver, staticMemoryResolver, staticEntityResolver } = BotProjectService;
 
   function launchLanguageServer(socket: rpc.IWebSocket) {
     const reader = new rpc.WebSocketMessageReader(socket);
     const writer = new rpc.WebSocketMessageWriter(socket);
     const connection: IConnection = createConnection(reader, writer);
-    const server = new LGServer(connection, getLgResources, staticMemoryResolver);
+    const server = new LGServer(connection, getLgResources, staticMemoryResolver, staticEntityResolver);
```
hatpick, line 173: does LG server need an entity resolver?

Composer/packages/tools/language-servers/language-generation/src/LGServer.ts
```diff
@@ -496,6 +502,33 @@ export class LGServer {
       return Promise.resolve(null);
     }
 
+    const wordRange = getEntityRangeAtPosition(document, params.position);
+    const word = document.getText(wordRange);
+    const projectId = this.getLGDocument(document)?.projectId;
+    let suggestEntities: string[] = [];
+    if (projectId && this.entitiesResolver) {
+      const luConetnts = this.entitiesResolver(projectId);
```
hatpick, line 510: typo: luContents

Composer/packages/tools/language-servers/language-generation/src/LGServer.ts
```diff
@@ -496,6 +502,33 @@ export class LGServer {
...
+      if (luConetnts) {
+        for (const conetnt of luConetnts) {
```
hatpick, line 512: typo: content

Composer/packages/server/src/services/project.ts
```diff
@@ -120,6 +120,11 @@ export class BotProjectService {
     return [...defaultProperties, ...userDefined];
   }
 
+  public static staticEntityResolver(projectId: string): string[] | undefined {
+    const contents = BotProjectService.getIndexedProjectById(projectId)?.luFiles.map((file) => file.content);
+    return flatten(contents);
```
hatpick, line 125: this would be the entire content of the lu files not the entities

Composer/packages/tools/language-servers/language-generation/src/LGServer.ts
```diff
@@ -496,6 +502,33 @@ export class LGServer {
...
+
+      this.luisEntities = suggestEntities;
+    }
+
+    const startWithAt = word.startsWith('@');
+
+    const completionEntityList = this.luisEntities.map((entity: string) => {
+      return {
+        label: entity,
+        kind: CompletionItemKind.Keyword,
```
hatpick, line 526: I dont think this should be keyword, it will cause a certain ui treatment on the monaco side.
keyword usually is used for language grammar keywords such as if else, for loop, etc.

Composer/packages/tools/language-servers/language-generation/src/LGServer.ts
```diff
@@ -496,6 +502,33 @@ export class LGServer {
...
+        insertText: entity,
+        documentation: `Registerd Luis Entity of ${entity}`,
```
hatpick, line 528: localize this

Composer/packages/tools/language-servers/language-generation/src/utils.ts
```diff
@@ -60,6 +63,25 @@ export function getRangeAtPosition(document: TextDocument, position: Position):
   return undefined;
 }
 
+export function getEntityRangeAtPosition(document: TextDocument, position: Position): Range | undefined {
+  const text = document.getText();
+  const line = position.line;
+  const pos = position.character;
+  const lineText = text.split('\n')[line];
```
hatpick, line 70: .split(/\r?\n/g)

hatpick (changes requested):

Composer/packages/server/src/server.ts
cosmicshuai: Yes, since LG server needs to suggest Luis entity. We have to provide a way to extract all enitites defined in user's .lu files.

Composer/packages/server/src/services/project.ts
cosmicshuai: Yes. Since the extract luis entities is a async process. So I pass all contents of .lu file to LG server and do the parseFile operation. This way can simplify the code.

Composer/packages/lib/indexers/src/dialogUtils/extractLUEntities.ts
cosmicshuai: deleted

Composer/packages/tools/language-servers/language-generation/src/LGServer.ts
cosmicshuai: modified

Composer/packages/tools/language-servers/language-generation/src/LGServer.ts
cosmicshuai: fixed

Composer/packages/tools/language-servers/language-generation/src/LGServer.ts
cosmicshuai: Switch to CompletionItemKind.Property. How about this? Since it does not have a explicit mapping of entity to CompletionItemKind

Composer/packages/tools/language-servers/language-generation/src/LGServer.ts
cosmicshuai: changed to formatMessage()

Composer/packages/tools/language-servers/language-generation/src/utils.ts
cosmicshuai: changed

Composer/packages/server/src/services/project.ts
hatpick: I dont understand why dont you find all the entities here? why language server needs to do that? It seems like your code in language server calculates the entities every time there is a request for autocomplete, how is that more efficient?

cosmicshuai: I think they are the same. Defining the function of calculating entites here. This resolver method is passed to LGServer. The actual computing stage is still occurred in the LGServer side. The bf-lu package for extract entities are optimized before. In LUServer, we also use the extracting entities for every request the server recieved.

hatpick: You need to cache the results, imagine you have lots of lu files, sending the contents of all those files and sending over to the language server, then extracting entities every time user type in a @ sign is not feasible imo.
@boydc2014 what do you think?

boydc2014: yes, agree, we can't treat the extracted entities as "hot" data, instead, you should cache it somewhere and update it in a lower frequency than typing. Either with a explicit cache, or use something like debounce, throttle which might make your code simpler

cwhitten: @cosmicshuai do you have what you need here to adjust this accordingly?

cosmicshuai: @cwhitten Yes. I talked with Dong. I had a minor change in calculating entities. But there was some issue to push the commits yesterday.

Composer/packages/tools/language-servers/language-generation/src/LGServer.ts
```diff
@@ -553,6 +561,38 @@ export class LGServer {
...
+            const luisJson = await extractLUISContent(content);
+            suggestEntities = suggestEntities.concat(getSuggestionEntities(luisJson, suggestionAllEntityTypes));
+          }
+        }
+      }
+
+      this.luisEntities = suggestEntities;
+    };
+
+    setTimeout(getEntities);
```
lei9444, line 583: I think if this will take a long time, we can move it to the lgWorker.ts

cosmicshuai: moved to lgworker to do the lu file parse

cosmicshuai:
@hatpick I added a cache for lu entities in LG server. Also, now the parse process of luis files is moved to lgworker. This will unblock the completion if the luis files are quite large. The cache won't be updated unless a five-second interval between completion requestes.
Could you help review this?

Composer/packages/tools/language-servers/language-generation/src/LGServer.ts
```diff
@@ -553,6 +559,32 @@ export class LGServer {
       return Promise.resolve(null);
     }
 
+    const wordRange = getEntityRangeAtPosition(document, params.position);
+    const word = document.getText(wordRange);
+    const projectId = this.getLGDocument(document)?.projectId;
+    let luContents: string[] = [];
+    if (projectId && this.entitiesResolver) {
+      luContents = this.entitiesResolver(projectId) || [];
```
hatpick, line 567: Can you fetch this data from server only if your time check on L#572 allows it?

Composer/packages/tools/language-servers/language-generation/src/lgParser.ts
```diff
@@ -29,6 +31,19 @@ class LgParserWithoutWorker {
...
+  public async extractLuisEntity(luContents: string[]): Promise<{ suggestEntities: string[] }> {
+    let suggestEntities: string[] = [];
+    if (luContents) {
+      for (const content of luContents) {
+        const luisJson = await extractLUISContent(content);
+        suggestEntities = suggestEntities.concat(getSuggestionEntities(luisJson, suggestionAllEntityTypes));
+      }
+    }
+
+    console.log(suggestEntities);
```
hatpick: remove log

hatpick:
> @hatpick I added a cache for lu entities in LG server. Also, now the parse process of luis files is moved to lgworker. This will unblock the completion if the luis files are quite large. The cache won't be updated unless a five-second interval between completion requestes.
> Could you help review this?

One thing to note is that when a user is editing a LG file and communicating with LG server during, LU entities are not going to change during that time, so why not calculate and cache the whole thing per editing session? 
For example, You can expire the cache and re-calculate it when document id changes.
You can take advantage of this and make 

p.s: I couldn't find the cache you are referring to, would you please send me. link to line number?

Composer/packages/tools/language-servers/language-generation/src/lgParser.ts
cosmicshuai: removed

Composer/packages/tools/language-servers/language-generation/src/LGServer.ts
cosmicshuai: changed the location of fetch lu files contents

cosmicshuai:
@hatpick 
I've had some discussions with Leilei about the Composer file storage. 
The cache for luis entity exists in LGServer. https://github.com/microsoft/BotFramework-Composer/blob/e6d5702f8f7e96e4983ec8db107a810fe5cdf91c/Composer/packages/tools/language-servers/language-generation/src/LGServer.ts#L57
The cache strategy is defined at: https://github.com/microsoft/BotFramework-Composer/blob/e6d5702f8f7e96e4983ec8db107a810fe5cdf91c/Composer/packages/tools/language-servers/language-generation/src/LGServer.ts#L570

I understood your suggestions, however there are some limitations of language server protocol. 
1.  The language servers start with the composer server and the language server only takes language feature requests, like completion, hover, etc. So, the LG server won't know users had switched to editing luis file and then switched back to a new LG editing session. The server must check the luis files by itself. 
2. The FileInfo type does not provide an event of file change.  We have to compare the name, content, lastModified to check whether it is modified. 

Therefore, I added a new cache to store the lu contents. When a new completion request is received, and it meets the requirement of five-seconds interval, it will compare the current lu files contents with the cached lu file contents. The parse will happen when the file contents are not equal. This will help reduce the computation.

cosmicshuai:
@hatpick 
You are right. Every time a user starts a new LG editing session. The LG server will receive an 'initializeDocuments' request. I put the extract luis entities here: https://github.com/microsoft/BotFramework-Composer/blob/41bc1f92bef7b753252a36544958c58f852b647c/Composer/packages/tools/language-servers/language-generation/src/LGServer.ts#L112

And store the entities in _luisEntities propety.  The completion responses will only read from this. 
Also before computing luis entites, it will compare the cached luis content with the current luis content. Only any change is detected, the luis parse will happen.

hatpick (dismissed):
It's much better now, thanks

cwhitten (approved):

PR #5737, opened 2021-02-09
fix: add subscription id to json editor schema

zidaneymar:
## Description
add subscription id to json editor schema

## Task Item
#minor


extensions/azurePublish/src/node/schema.ts
```diff
@@ -30,6 +30,10 @@ const schema: JSONSchema7 = {
       title:
         'Runtime identifier for hosting bot, default to win-x64, please refer to https://docs.microsoft.com/en-us/dotnet/core/rid-catalog',
     },
+    subscriptionId: {
+      type: 'string',
+      title: 'id of your subscription'
+    },
     settings: {
```
luhan2017, line 37: Upper case I

luhan2017: I found that QnA is also missing from the schema, let's add it.

zidaneymar: > I found that QnA is also missing from the schema, let's add it.

added

extensions/azurePublish/src/node/schema.ts
```diff
@@ -30,6 +30,31 @@ const schema: JSONSchema7 = {
...
+    },
+    resourceGroup: {
+      type: 'string',
+      title: 'Name of your resource group'
+    },
+    botName: {
+      type: 'string',
+      title: 'Name of your bot channel registration service'
+    },
+    abs: {
```
luhan2017, line 45: QnA is missing from schema and there is no abs section now, all of the settings from abs are under root.

benbrown:
It would be great if this PR could include all necessary schema updates for the settings!

extensions/azurePublish/src/node/schema.ts
zidaneymar: qna and other resources added.

luhan2017 (approved):

PR #5740, opened 2021-02-09
fix: Include structured response parse data in lg template

lei9444:
## Description
Add structure LG parsed result to lg template.

For example:

`# template
[Activity
 Text=good
 Speak=hi
]`
The parsed Template would contains a property named properties, in which would contains:

`properties: {Text: "good", Speak:"hi", $type:'Activity'}`

## Task Item
#minor
refs https://github.com/microsoft/botbuilder-js/issues/3273


hatpick:
This is a PR to feature branch, I'm merging to unblock the feature.

PR #5741, opened 2021-02-09
feat: appended provision with a profile

### LINKED ISSUE #5403: Provision V2: Edit publish profile(add/update/delete services) through UX
For provision V2, we want to support editing publish profile through UX. Which includes the follow work items: 

- [ ] support edit publish profile with UX

Here is the figma design: 
https://www.figma.com/file/J3oIdXDmeBQbAe4a4PYbQc/Jaskeerat-Bedi?node-id=2200%3A70869

VanyLaw:
## Description
Add the ability to do the provision in an existed profile.

- [x] update the edit profile UI.
- [x] update the add profile UI.
- [x] do the appended provision e2e successfully.
- [x] can publish successfully with new profile.

In this PR we also did the following updates:
- Add subscriptionId, resourceGroup and botName in the publish profile. When edit the publish profile, subscriptionId and resourceGroup is not editable at this moment, you can only append new resources under the same subscription and resourceGroup. botName is the name of bot channel registration. 
- Decouple the resourceGroup and hostname, previously in publish profile, we use hostname(name-environment) for both resourceGroup and hostname, now this value is decoupled, you can create new resources with different name under an existing resourceGroup. This is more convenient for users who don't have the permission to create a new resourceGroup. 
 

## Task Item

close #5403


Composer/packages/client/src/pages/botProject/create-publish-profile/PublishProfileDialog.tsx
```diff
@@ -72,7 +70,11 @@ export const PublishProfileDialog: React.FC<PublishProfileDialogProps> = (props)
   useEffect(() => {
     PluginAPI.publish.closeDialog = closeDialog;
     PluginAPI.publish.onBack = () => {
-      setPage(Page.AddProfile);
+      setPage(Page.ProfileForm);
+      setTitle({
+        title: current ? formatMessage('Edit a publishing profile') : formatMessage('Add a publishing profile'),
```
carlosscastro, line 75: Shouldn't / can't these messages be localized?

Composer/packages/client/src/pages/botProject/create-publish-profile/PublishProfileDialog.tsx
```diff
@@ -28,19 +27,18 @@ type PublishProfileDialogProps = {
...
 
 export const PublishProfileDialog: React.FC<PublishProfileDialogProps> = (props) => {
   const { current, types, projectId, closeDialog, targets, setPublishTargets } = props;
-  const [page, setPage] = useState(current ? Page.EditProfile : Page.AddProfile);
+  const [page, setPage] = useState(Page.ProfileForm);
   const [publishSurfaceStyles, setStyles] = useState(defaultPublishSurface);
 
   const [dialogTitle, setTitle] = useState({
-    title: formatMessage('Add a publishing profile'),
+    title: current ? formatMessage('Edit a publishing profile') : formatMessage('Add a publishing profile'),
```
carlosscastro, line 41: Can this be localized?

luhan2017:
@zidaneymar we have three known issues now, need to fix after we get the permission back.
1. remove the dependency between webapp and bot channel registration while provisioning. 
2. two issue of region, one is UX display error, even I select westus for luis, it will still show the region of resourcegroup.  The other is whatever region I select, it will always provision on the westus. 
3. need to implement QnA in getExistResource function.

Composer/packages/client/src/pages/botProject/create-publish-profile/PublishProfileDialog.tsx
carlosscastro: Nevr mind, formatMessage does that :)

Composer/packages/client/src/pages/botProject/create-publish-profile/PublishProfileDialog.tsx
carlosscastro: n/a

luhan2017:
> @zidaneymar we have three known issues now, need to fix after we get the permission back.
> 
> 1. remove the dependency between webapp and bot channel registration while provisioning.
> 2. two issue of region, one is UX display error, even I select westus for luis, it will still show the region of resourcegroup.  The other is whatever region I select, it will always provision on the westus.
> 3. need to implement QnA in getExistResource function.

All the three issues are fixed now.

luhan2017 (approved):

boydc2014 (approved):

PR #5746, opened 2021-02-09
fix: get projectId from link in tree navigation

### LINKED ISSUE #5727: Clicking on root dialog across/bot skill loads the root dialog for active project, not the clicked on one
## Describe the bug

When I have a bot and a skill loaded.
```
Bot
    bot.dialog
Skill
    skill.dialog
```
When click on skill.dialog, and click on bot.dialog, it flashes, and stays on skill.dialog.

## Version
image

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. load project with bot and skill.
2. Click on any dialog for the skill
3. click on any dialog for the bot
The root dialog for the skill will be selected, not the dialog you click on in #3.

## Expected behavior

To show the dialog I click on.

beyackle:
## Description

There was a bug in the sidebar navigation; the links did include the correct project ID (for both the root bot and skills), but navigation was only using the project ID of the currently active bot/skill, not the one the link was actually meant to point at.

## Task Item

closes #5727


beyackle:
The other change was to remove a now-redundant React.Fragment wrapper; there's only one thing inside there, so it was a no-op.

Composer/packages/client/src/pages/design/SideBar.tsx
```diff
@@ -107,7 +107,7 @@ const SideBar: React.FC<SideBarProps> = React.memo(({ projectId }) => {
     if (link.botError) {
       setBrokenSkillInfo(link);
     }
-    const { skillId, dialogId, trigger } = link;
+    const { skillId, dialogId, trigger, projectId } = link;
```
a-b-r-o-w-n, line 110: The diff makes it hard to see how this value is being used.

a-b-r-o-w-n: Can you point me to where its being used?

beyackle: It's used immediately below in the calls like `navTo(skillId ?? projectId, dialogId);`. Before this change, the value of projectId was obtained from the outside scope - a prop that was handed to the SideBar component - and that was defined out in DesignPage.tsx also as `skillId ?? projectId`, so every dialog link the side bar ended up navigating to whichever projectId was currently "active" instead of their own internal projectId.

a-b-r-o-w-n: Can we consider renaming some variables for clarity? It's very difficult to know what `projectId` is when used in the various places.

cwhitten (approved):

PR #5758, opened 2021-02-10
fix: Orchestrator API changes from SDK and CLI

taicchoumsft:
## Description
OrchestratorRecognizer (Dotnet SDK) has changed names of `modelPath` and `snapshotPath` to `modelFolder` and `snapshotFile` respectively to be less ambiguous about what should be passed in, folder or file.  

BF-Orchestrator (CLI side) also updated the API for pulling in the list of models and default model field to support the incoming multilang models.  

Updating these name changes here in this PR.

## Task Item
#minor

### DIFF
--- a/Composer/packages/server/package.json
+++ b/Composer/packages/server/package.json
@@ -70,7 +70,7 @@
     "@microsoft/bf-dispatcher": "^4.11.0-beta.20201016.393c6b2",
     "@microsoft/bf-generate-library": "^4.10.0-daily.20210129.207826",
     "@microsoft/bf-lu": "^4.12.0-dev.20210129.82760de",
-    "@microsoft/bf-orchestrator": "^4.11.0-beta.20210210.116dcad",
+    "@microsoft/bf-orchestrator": "4.11.0-beta.20210210.116dcad",
     "applicationinsights": "^1.8.7",
     "archiver": "^5.0.2",
     "axios": "^0.21.1",
--- a/Composer/packages/server/package.json
+++ b/Composer/packages/server/package.json
@@ -50,11 +50,11 @@
     "@types/rimraf": "^2.0.2",
     "@types/yeoman-environment": "^2.10.2",
     "fs-extra": "7.0.1",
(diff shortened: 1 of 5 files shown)


tsuwandy (dismissed):

pavolum (approved):

Composer/packages/server/package.json
```diff
@@ -70,7 +70,7 @@
     "@microsoft/bf-dispatcher": "^4.11.0-beta.20201016.393c6b2",
     "@microsoft/bf-generate-library": "^4.10.0-daily.20210129.207826",
     "@microsoft/bf-lu": "^4.12.0-dev.20210129.82760de",
-    "@microsoft/bf-orchestrator": "4.11.0-beta.20201208.165a6d6",
+    "@microsoft/bf-orchestrator": "^4.11.0-beta.20210210.116dcad",
```
cwhitten, line 73: why the carrot?

Composer/packages/server/src/models/bot/builder.ts
```diff
@@ -160,7 +160,7 @@ export class Builder {
     if (!luFiles.filter((file) => !emptyFiles[file.name]).length) return;
 
     const nlrList = await this.runOrchestratorNlrList();
-    const defaultNLR = nlrList.default;
+    const defaultNLR = nlrList.defaults.en_intent;
```
cwhitten, line 163: I wonder if we should add a TODO here to hoist this hardcoded default into something configurable.. What is the user behavior? When would they _not_ want to default to English here?

taicchoumsft: I broke this change apart from upcoming Orchestrator multilang changes to keep this PR small and focused on the SDK/CLI name changes, and just maintain parity with what we currently have.

The multilang PR will hopefully address this concern of defaulting to English only when there are English locale LU files.

Composer/packages/server/package.json
taicchoumsft: Thanks @cwhitten, will remove carrot to fix this version in place.

PR #5772, opened 2021-02-11
feat: support multibot templates during in new creation flow

benbrown:
## Description

This change allows any number of bots to be created by a template, enabling multi-skill scenarios. 

Previously our code assumed there would be one bot project resulting from the file -> new action. 

This change detects whether the result of the create is a single project or multiple projects. It then processes any projects found, resulting in N-number of fully instantiated projects available to open in Composer.

It then returns the root bot to the client so that it can open it and all child bots.

#minor

## Screenshots

https://www.loom.com/share/e6c545242a9e4f409eb54d28cb7cc2f2

### DIFF
--- a/Composer/packages/client/src/recoilModel/dispatchers/project.ts
+++ b/Composer/packages/client/src/recoilModel/dispatchers/project.ts
@@ -2,7 +2,6 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-import { BotProjectFile } from '@bfc/shared';
 import formatMessage from 'format-message';
 import findIndex from 'lodash/findIndex';
 import { RootBotManagedProperties } from '@bfc/shared';
@@ -22,7 +21,6 @@ import {
   botNameIdentifierState,
   botOpeningMessage,
   botOpeningState,
-  botProjectFileState,
   botProjectIdsState,
   botProjectSpaceLoadedState,
   botStatusState,
@@ -53,6 +51,7 @@ import {
   navigateToSkillBot,
   openLocalSkill,
   openRemoteSkill,
+  openRootBotAndSkills,
   openRootBotAndSkillsByPath,
   openRootBotAndSkillsByProjectId,
   removeRecentProject,
@@ -484,11 +483,8 @@ export const projectDispatcher = () => {
             if (settingStorage.get(projectId)) {
               settingStorage.remove(projectId);
             }
-            const currentBotProjectFileIndexed: BotProjectFile = botFiles.botProjectSpaceFiles[0];
-            callbackHelpers.set(botProjectFileState(projectId), currentBotProjectFileIndexed);
 
-            const mainDialog = await initBotState(callbackHelpers, projectData, botFiles);
-            callbackHelpers.set(botProjectIdsState, [projectId]);
+            const { mainDialog } = await openRootBotAndSkills(callbackHelpers, { botFiles, projectData });
 
             // Post project creation
             callbackHelpers.set(projectMetaDataState(projectId), {
--- a/Composer/packages/server/src/models/asset/assetManager.ts
+++ b/Composer/packages/server/src/models/asset/assetManager.ts
@@ -113,6 +113,7 @@ export class AssetManager {
         throw new Error('already have this folder, please give another name');
       }
 
+      log('About to create folder', dstDir);
       mkdirSync(dstDir, { recursive: true });
 
       // find selected template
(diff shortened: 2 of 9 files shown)


Composer/packages/server/src/models/asset/assetManager.ts
```diff
@@ -159,9 +160,10 @@ export class AssetManager {
     dstDir: string,
     projectName: string
   ): Promise<boolean> {
+    log('About to instantiate a template!', dstDir, generatorName, projectName);
     this.yeomanEnv.cwd = dstDir;
```
pavolum, line 164: Ah I see what happened, this.yeomenEnv.cwd was not changing the process cwd as expected hence your addition of proecces.chdir. 

Does that mean we no longer need this.yeomenEnv.cwd  as its not really doing anything?

Composer/packages/server/src/utility/project.ts
```diff
@@ -77,7 +77,7 @@ export async function ejectAndMerge(currentProject: BotProject, jobId: string) {
     // run the merge command to merge all package dependencies from the template to the bot project
     BackgroundProcessManager.updateProcess(jobId, 202, formatMessage('Merging Packages'));
     const realMerge = new SchemaMerger(
-      [manifestFile],
+      [manifestFile, '!**/imported/**', '!**/generated/**'],
```
pavolum, line 80: Nice! Was this the fix for the multiple file creation issue?

pavolum (approved):
Only thing I see is the removal of yeomen.env.cmd as I believe that wasnt doing what is expected and your process.chDir is. 
Thats my code though and I am happy to remove it as a seperate small clean up PR.

Composer/packages/server/src/services/project.ts
```diff
@@ -452,36 +451,81 @@ export class BotProjectService {
...
+                if (schemaUrl && !createFromPva) {
+                  await currentProject.saveSchemaToProject(schemaUrl, botRef.path);
+                }
+
+                log('Init project', id);
+                await currentProject.init();
+              }
+              resolve(id);
+            } catch (err) {
+              return reject(err);
```
srinaath, line 508: How does this error get handled inside Promise.all?

Composer/packages/server/src/services/project.ts
```diff
@@ -452,36 +451,81 @@ export class BotProjectService {
...
-        if (schemaUrl && !createFromPva) {
-          await currentProject.saveSchemaToProject(schemaUrl, locationRef.path);
-        }
+      const botsToProcess: { storageId: string; path: string; name: string }[] = [];
 
-        await currentProject.init();
+      // The outcome of our creation might be > 1 bot! We need to determine how many bots we find in this folder.
+      // is this a single bot?
+      if (await StorageService.checkIsBotFolder(newProjRef.storageId, newProjRef.path, user)) {
+        botsToProcess.push({ ...newProjRef, name: name });
```
srinaath, line 459: nit can use shorthand name instead of name:name

srinaath (dismissed):

Composer/packages/server/src/services/project.ts
benbrown: the reject should call the promise.all to fail, then the try/catch wrapper will bubble that error up to the client.  I believe this is the desired behavior, but in the future we may want to do something more graceful since the user will need to go to the command line and debug the build failure.

Composer/packages/server/src/services/project.ts
benbrown: done

srinaath (approved):

Composer/packages/client/src/recoilModel/dispatchers/project.ts
```diff
@@ -484,11 +483,8 @@ export const projectDispatcher = () => {
             if (settingStorage.get(projectId)) {
               settingStorage.remove(projectId);
             }
-            const currentBotProjectFileIndexed: BotProjectFile = botFiles.botProjectSpaceFiles[0];
-            callbackHelpers.set(botProjectFileState(projectId), currentBotProjectFileIndexed);
 
-            const mainDialog = await initBotState(callbackHelpers, projectData, botFiles);
-            callbackHelpers.set(botProjectIdsState, [projectId]);
```
beyackle, line 491: Question (not blocking, just curious) - why are these lines being deleted?

benbrown: this happens inside the call to openRootBotandSkill

PR #5778, opened 2021-02-12
fix: Add AddNewKnowledgeBaseCompleted telemetry event to create KB from scratch

tdurnford:
## Description
When the user creates a new Knowledge Base from scratch, we should log a AddNewKnowledgeBaseCompleted event.

## Task Item
#minor


beyackle (approved):
LGTM.

PR #5790, opened 2021-02-15
feat: Webchat Integration

### LINKED ISSUE #5565: Webchat UI Integration
This ticket encapsulates all the UI work to incorporate Webchat implementation in Composer.

Figma: https://www.figma.com/file/HwYEQoCmWbNrIDdCOmGvzN/bot-testing-BRAINS-storm?node-id=931%3A9852

*   [ ] Restart conversation with the same user ID/different user ID. 
*   [ ] Saving a conversation as a transcript. 
*   [ ] Allow uploads and attachments to a conversation

srinaath:
## Description

This PR integrates the webchat control to Composer. The controller turns on when the bot is started.  
Adding `botframework-webchat` forces Composer to upgrade Node > 14.0. The pipeline currently uses Node 13.8 and will be upgraded to LTS Node version 14.15.4/x64.

**Functionalities exposed**

1.  Restart conversation with the same user ID/different user ID.
2.  Saving a conversation as a transcript.
3.  Allow uploads and attachments to a conversation

Functionalities not supported yet

1.  Bots that have OAuth actions (No support for creating tunnel yet in Composer)
2.  Remote Bots (No support for creating tunnel yet in Composer)
3.  Showing REST errors happening between Bot and Mock Directline Server (Coming in next PR as a tab on chat panel)

## Task Item

Fixes #5565

## Screenshots

recording

recording 2


Composer/packages/webchat-client/package.json
```diff
@@ -0,0 +1,55 @@
...
+  "license": "MIT",
+  "dependencies": {
+    "@bfc/extension-client": "*",
+    "@bfc/shared": "*",
+    "@bfc/ui-shared": "*",
+    "@emotion/core": "^10.0.27",
+    "@emotion/styled": "^10.0.27",
+    "axios": "^0.21.1",
+    "botframework-webchat": "^4.11.0",
+    "lodash": "^4.17.19",
```
hatpick, line 35: this can be moved to peer

Composer/packages/webchat-client/package.json
```diff
@@ -0,0 +1,55 @@
...
+    "prop-types": "^15.7.2"
```
hatpick, line 36: is this really used?

Composer/packages/webchat-client/src/WebChatPanel.tsx
```diff
@@ -0,0 +1,178 @@
...
+export interface WebChatPanelProps {
+  botUrl: string;
+  secrets: BotSecrets;
+  /** Directline host url. By default, set to Composer host url. */
+  directlineHostUrl?: string;
+  botName: string;
+  projectId: string;
+  isWebChatPanelVisible: boolean;
+  activeLocale: string;
+  appLifecycleHandler: any;
```
hatpick, line 25: avoid any, give it the appropriate type

Composer/packages/webchat-client/src/WebChatPanel.tsx
```diff
@@ -0,0 +1,178 @@
...
+  const [chats, setChatData] = useState<Record<string, ChatData>>({});
+  const [currentConversation, setCurrentConversation] = useState<string>('');
+  const conversationServiceRef = useRef<ConversationService>(new ConversationService(directlineHostUrl));
+  const conversationService = conversationServiceRef.current;
+  const downloadLinkRef = useRef<HTMLAnchorElement>(null);
+  const [isConversationStartQueued, queueConversationStart] = useState<boolean>(false);
+  const ipcRenderer = useRef<any>(appLifecycleHandler);
+
+  useEffect(() => {
+    ipcRenderer.current?.on('conversation-cleanup', () => {
```
hatpick, line 49: how can this ref be undefined?

Composer/packages/webchat-client/src/WebChatPanel.tsx
```diff
@@ -0,0 +1,178 @@
...
+      conversationService.cleanupAll();
+    });
+  }, []);
+
+  useEffect(() => {
+    if (botUrl) {
```
hatpick, line 55: nit: queueConversationStart(!!botUrl);

Composer/packages/webchat-client/src/WebChatPanel.tsx
```diff
@@ -0,0 +1,178 @@
...
+    if (isWebChatPanelVisible && isConversationStartQueued) {
+      const startConversation = async () => {
+        const chatData: ChatData = await conversationService.startNewConversation(
+          botUrl,
+          secrets,
+          projectId,
+          activeLocale
+        );
+
+        setCurrentConversation(chatData.conversationId);
```
hatpick, line 72: these to setters below can cause issue with React under the following scenario:
Imagine your async function is awaiting on the line 65 when component gets unmounted and now it tries to update the state on an unmounted component on line 72. You'll get warning in the dev tools.

Composer/packages/webchat-client/src/WebChatPanel.tsx
```diff
@@ -0,0 +1,178 @@
...
+        setChatData({
+          [chatData.conversationId]: chatData,
+        });
+      };
+      startConversation();
+      queueConversationStart(false);
+    }
+  }, [isWebChatPanelVisible]);
+
+  const webchatContent = useMemo(() => {
```
hatpick, line 82: Create a component for this instead of useMemo

Composer/packages/webchat-client/src/utils/ConversationService.ts
```diff
@@ -0,0 +1,316 @@
+// Copyright (c) Microsoft Corporation.
```
hatpick, line 1: nit: file name should be camleCase, PascalCase ins only for Components

Composer/packages/webchat-client/src/utils/ConversationService.ts
```diff
@@ -0,0 +1,316 @@
...
+// Licensed under the MIT License.
+
+import { v4 as uuidv4 } from 'uuid';
+import axios, { AxiosInstance } from 'axios';
+import { createDirectLine } from 'botframework-webchat';
+
+export const headers = {
```
hatpick, line 8: why is this exported?

Composer/packages/webchat-client/src/utils/ConversationService.ts
```diff
@@ -0,0 +1,316 @@
...
+  'Content-Accept': 'application/json',
```
hatpick, line 9: did you mean Content-Type?

Composer/packages/webchat-client/src/utils/ConversationService.ts
hatpick: please check the exported stuff down below to make sure they really need to be exported.

Composer/packages/webchat-client/src/utils/ConversationService.ts
```diff
@@ -0,0 +1,316 @@
...
+  channelServiceType: ChannelService;
+  members: ConversationMember[];
+  mode: WebChatMode;
+  locale: string;
+  msaAppId?: string;
+  msaPassword?: string;
+};
+
+// All activity types. Just parsing for Trace currently.
+export enum ActivityTypes {
```
hatpick, line 40: nit: ActivityType (singular)
for reference:
https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-classes-structs-and-interfaces#naming-enumerations

Composer/packages/webchat-client/src/utils/ConversationService.ts
```diff
@@ -0,0 +1,316 @@
...
+  MessageDelete = 'messageDelete',
+  InstallationUpdate = 'installationUpdate',
+  MessageReaction = 'messageReaction',
+  Suggestion = 'suggestion',
+  Trace = 'trace',
+  Handoff = 'handoff',
+}
+
+export type ChatData = {
+  chatMode: 'livechat' | 'transcript';
```
hatpick, line 60: use WebChatMode type

Composer/packages/webchat-client/src/utils/ConversationService.ts
```diff
@@ -0,0 +1,316 @@
...
+  directline: any;
```
hatpick, line 61: avoid any

Composer/packages/webchat-client/src/utils/ConversationService.ts
```diff
@@ -0,0 +1,316 @@
...
+  projectId: string;
+  user: User;
+  conversationId: string;
+};
+
+export default class ConversationService {
```
hatpick, line 67: It's better if the methods of the class are prototypal (method(args){...}) vs. arrow (method: (args) => {...})
I highly recommend converting your code to use prototype.

extensions/webchat/src/node/directline/middleware/attachmentHandler.ts
```diff
@@ -0,0 +1,120 @@
...
+
+import { DLServerState } from '../store/dlServerState';
+import { BotErrorCodes, createAPIException, sendErrorResponse } from '../utils/apiErrorException';
+
+type AttachmentParams = {
+  attachmentId: string;
+  viewId: string;
+};
+
+export function createGetAttachmentInfoHandler(state: DLServerState) {
```
hatpick, line 15: nit: either createAttachmentInfoHandler or getAttachmentInfoHandler

extensions/webchat/src/node/directline/middleware/conversationHandler.ts
```diff
@@ -92,3 +95,22 @@ export const createUpdateConversationHandler = (state: DLServerState) => {
     });
   };
 };
+
+export const cleanupConversation = () => {
```
hatpick, line 99: localize strings below

Composer/packages/webchat-client/src/WebChatPanel.tsx
srinaath: This is the ipcRenderer sent from electron. If we are running in web context this will be undefined.

Composer/packages/webchat-client/package.json
```diff
@@ -0,0 +1,55 @@
+{
```
a-b-r-o-w-n, line 1: I don't think this deserves it's own package. Each package added adds time to our build pipeline. We are getting to the point now that that is becoming a burden.

Composer/packages/webchat-client/src/WebChatPanel.tsx
a-b-r-o-w-n: Is this driven by a menu item or something? Is there similar functionality for the web-based composer?

Composer/packages/webchat-client/src/WebChatPanel.tsx
```diff
@@ -0,0 +1,178 @@
...
+  directlineHostUrl = BASEPATH,
+  botName,
+  isWebChatPanelVisible,
+  openBotInEmulator,
+  activeLocale,
+  appLifecycleHandler,
+}) => {
+  const [chats, setChatData] = useState<Record<string, ChatData>>({});
+  const [currentConversation, setCurrentConversation] = useState<string>('');
+  const conversationServiceRef = useRef<ConversationService>(new ConversationService(directlineHostUrl));
```
a-b-r-o-w-n, line 42: `useMemo` is a better hook for this pattern.

```suggestion
  const conversationService = useMemo(() => new ConversationService(directlineHostUrl), [directlineHostUrl]);
```

Composer/packages/webchat-client/src/WebChatPanel.tsx
```diff
@@ -0,0 +1,178 @@
...
+          openBotInEmulator(projectId);
+        }}
+      />
+      <div data-testid="WebChat-Content" style={{ height: 'calc(100% - 36px)' }}>
+        {webchatContent}
+      </div>
+      {/* A shadow download link to trigger browser native API on saving transcript JSON.  */}
+      <a ref={downloadLinkRef} aria-hidden="true" href="#save" style={{ display: 'none' }}>
+        {formatMessage('Save')}
+      </a>
```
a-b-r-o-w-n, line 175: If this is hidden in the UI, we don't even need to attach it to the dom until a transcript is downloaded.

```js
const btn = document.createElement('a');
btn.style.display = 'none';
someContainerRef.appendChild(btn);
// ... do download stuff
btn.click();
someContainerRef.removeChild(btn);
```

a-b-r-o-w-n (changes requested):

a-b-r-o-w-n:
@srinaath I'm a little confused why the webchat client code is "first-party" to the client app, but the server code lives in an extension. I assume that if the extension were disabled or removed, the client app would be broken. Is that true?

If this is the case, I would suggest moving the webchat server code into the server package -- no need to separate it out. OR come the client code into the extension.

srinaath:
> @srinaath I'm a little confused why the webchat client code is "first-party" to the client app, but the server code lives in an extension. I assume that if the extension were disabled or removed, the client app would be broken. Is that true?
> 
> If this is the case, I would suggest moving the webchat server code into the server package -- no need to separate it out. OR come the client code into the extension.

@a-b-r-o-w-n  Thats correct. The webchat client  presently would break without the DirectLine extensions server but of course in future we could let Webchat communicate over DL ASE protocol rather than the Direct Line protocol. Initially, the thought process was to have the webchat-client also in an extension. But, as I started working on the feature, it felt more in line with a package for ex (webchat client feature is like form-dialogs and integral to Composer client) and also the ease of working with a package vs an extension under our timelines prompted that decision. The DirectLine server routes were external to Composer's server and didn't want to pollute the Composer Server.

a-b-r-o-w-n:
@srinaath Without a concrete timeline of porting the webchat client into an extension, I would suggest moving the client code into the client package. There is almost no benefit from having it in a separate package at this point in time and only downsides (build times, dev experience [how to watch files], etc).

I believe the same can be said about the server component as well. It just doesn't fit the extension model because all it does it add api routes and is expected to be enabled always.

Composer/packages/webchat-client/package.json
srinaath: prop-types is removed now.

Composer/packages/webchat-client/src/WebChatPanel.tsx
srinaath: yep. Using a boolean flag mounted now and updating state only if component remains mounted.

Composer/packages/webchat-client/src/utils/ConversationService.ts
srinaath: Using react.memo now

Composer/packages/webchat-client/package.json
srinaath: Moved it to Composer client now

srinaath:
@a-b-r-o-w-n Mock DL server and Webchat client are now part of Composer core until we have a solid plan to move all 3 parts (DL server, webchat client, Debug Panel Inspector tab) to an extension.

Composer/packages/webchat-client/src/utils/ConversationService.ts
srinaath: Removed this header

Composer/packages/client/src/components/BotRuntimeController/BotController.tsx
```diff
@@ -56,23 +55,28 @@ const startPanelsContainer = css`
 
 const transparentBackground = 'rgba(255, 255, 255, 0.5)';
 
-const BotController: React.FC = () => {
+type BotControllerProps = {
+  hideController: (isHidden: boolean) => void;
```
hatpick, line 59: nit: onHideController

Composer/packages/client/src/components/Header.tsx
```diff
@@ -140,20 +150,34 @@ export const Header = () => {
...
 
   const onUpdateAvailableClick = useCallback(() => {
     setAppUpdateShowing(true);
   }, []);
 
+  useEffect(() => {
+    if (!hideBotController && isWebChatPanelVisible) {
+      toggleWebChatPanel(false);
+    }
+  }, [hideBotController]);
```
hatpick, line 179: missing isWebChatPanelVisible

cwhitten (approved):

PR #5798, opened 2021-02-16
feat: Adding QNA Maker template to template list in component creation flow

### LINKED ISSUE #5797: QnA Maker template in new creation flow
**Friendly Name** 
QnA Maker

**Description**
A template for creating a bot that connects to a QnA Maker knowledge base.

**Purpose**

A basic bot for folks that want to connect to an existing KB, or create a new one.

**Functionality**

- [x] QNA template appears as an option in template selector
  - [x]  ReadMe is manually added since that QNA template is not pulled from npm
- [x] Creates an empty bot template in the users desired location
- [x] Post creation routes the user to the QnA modal
- [x] Following completion of the QnA modal
  - [x] QNA source is parsed and instantiated
  - [x]  QNA dialog is scaffolded using QnA intent recognizer

pavolum:
## Description

For R12 we want to ensure that users can create a bot using the component model that fulfills a similar function to the existing creation flows 'create from qna' experience. 

The end experience is similar, however, under the hood, the component model creation goes about this in a slightly different way:
- Manually appends qna template to template state as opposed to having that template populated via npm feed as the other templates are populated
  - Since qna template is not itself an npm package under the hood it does not have its respective package readMe to show in the creation flow. Instead its readMe is hosted in the botframework-components repo and Composer makes an API call to github to grab that readMe for that template. 
  - Misc change: If readMe fetch ever throws an error (regardless of getting readMe from npm or github) it now returns a back up markdown response and logs the error. That way readMe fetch failure does not interrupt creation flow
- Server has override to check if that template was selected and creates an empty bot template 
- Post creation, client workspace routes to the qna modal experience where the user can indicate KBs to pull qna content from
- Post QNA modal completion, QNA Maker parses qna content, adds the KB and the routing to that KB via QNA intent recognition.

## Task Item

fixes #5797


Composer/packages/client/src/recoilModel/dispatchers/storage.ts
```diff
@@ -165,6 +166,16 @@ export const storageDispatcher = () => {
...
 
       if (data && Array.isArray(data) && data.length > 0) {
+        (data as BotTemplate[]).push({
+          id: QnABotTemplateId,
+          name: 'generator-qna-bot',
+          description: 'Empty bot template that routes to qna configuration',
+          package: {
+            packageName: 'generator-empty-bot',
+            packageSource: 'npm',
+            packageVersion: '0.0.1',
```
srinaath, line 176: should these values be driven from the server response rather than the client?

srinaath (approved):

Composer/packages/client/src/recoilModel/dispatchers/storage.ts
```diff
@@ -165,6 +166,16 @@ export const storageDispatcher = () => {
       const data = response?.data;
 
       if (data && Array.isArray(data) && data.length > 0) {
+        (data as BotTemplate[]).push({
+          id: QnABotTemplateId,
+          name: 'generator-qna-bot',
+          description: 'Empty bot template that routes to qna configuration',
```
hatpick, line 172: localize

Composer/packages/server/src/models/asset/assetManager.ts
```diff
@@ -334,4 +335,11 @@ export class AssetManager {
 
     return templates;
   }
+
+  public async getRawGithubFileContent(owner: string, repo: string, branch: string, path: string) {
+    const githubUrl = new URL(`https://raw.githubusercontent.com/${owner}/${repo}/${branch}/${path}`);
```
hatpick, line 340: nit: no need to use URL, use the string template literal is good enough

Composer/packages/client/src/recoilModel/dispatchers/storage.ts
pavolum: Good point, I have moved it to the server side

srinaath (approved):

PR #5818, opened 2021-02-18
chore: update package versions from alerts

beyackle:
## Description

Got a few Component Governance alerts about old versions - let's see if this fixes them.

## Task Item

#minor


boydc2014:
Seems like the security alerts have been fixed.

PR #5825, opened 2021-02-18
feat: add debug panel to lu/lg/qna page

alanlong9278:
## Description

- Add debug panel to lu/lg/qna/publish page
- Delete diagnostic page
image

### Question
What dose `shouldShowEditorError` mean? I see it just calculate the height of content component.
 
## Task Item
#minor


srinaath:
@alanlong9278 If we are removing the diagnostics panel we still need to add a filter to filter diagnostics by bots.

srinaath (changes requested):
@yeze322 @alanlong9278  There seems to be a bug when i try this branch. Essentially, the debug panel wont show the errors of the currently seletected project inside the LG/LU/Qna page. It shows the errors and warnings for the current project only inside design page

alanlong9278:
> @yeze322 @alanlong9278 There seems to be a bug when i try this branch. Essentially, the debug panel wont show the errors of the currently seletected project inside the LG/LU/Qna page. It shows the errors and warnings for the current project only inside design page

Fix the bug about show selected project errors/warninigs in the LG/LU/Qna page, but can not do that in publish page because we can select multible project. And revert diagnostics page to the left nav. @srinaath  @yeze322

yeze322 (approved):

srinaath (approved):

Issue #5708, opened 2021-02-07
Improve the bot loading performance


boydc2014:
## Describe the bug

Composer is taking too much time to load a big bot. For example, in Calendar bot, it's taking up to 30-60s to see the first screen.

## Version

Lastest main

## Browser

- [X] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [X] Windows
- [ ] Ubuntu

## To Reproduce

1. open a big bot
2. wait the design page to show up

## Expected behavior

Expect the first screen will show up very soon, within a few second, even the bot is considered as big. 

## Additional context

We had a proposal for hot/cold splitting which we can leverage the idea to break down the processing when loading.

hibrenda:
addressed in #5853

PR #5841, opened 2021-02-19
fix: load bot will do full index twice (step 1: improve the load bot perf)

lei9444:
## Description
When we open a new bot(calendar skill), composer will do twice full index today. Every full index will take about 9s, and the loading time is about 22s today.
image

**In this PR**
reduce one full index, the loading time is about 14s
image

**Next step**
Put the full index to the background. 
If we need to show UI during the indexing stage, only parse the resource(lg, lu, qna) for current dialog  

## Task Item
refs #5708

### DIFF
--- a/Composer/packages/client/src/recoilModel/dispatchers/project.ts
+++ b/Composer/packages/client/src/recoilModel/dispatchers/project.ts
@@ -478,7 +478,7 @@ export const projectDispatcher = () => {
             // Bot creation successful
             clearInterval(timer);
             callbackHelpers.set(botOpeningMessage, response.data.latestMessage);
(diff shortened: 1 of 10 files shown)


boydc2014:
I thought this should be a one-line change, just not waiting the worker's indexing.  Did you include something else in this PR? or you found that's not enough?

lei9444:
> I thought this should be a one-line change, just not waiting the worker's indexing. Did you include something else in this PR? or you found that's not enough?

I just think we don't need to do twice index, so I move all calculation to worker, and this is also preparation for next step. I think I can close this PR first, I will create a draft PR tomorrow.

PR #5848, opened 2021-02-19
fix: Orchestrator Download Notification

### LINKED ISSUE #5615: Add configuration settings to select / download a non-default Orchestrator model.
The Orchestrator feature will use the default basemodel. This is to track the request to allow selection of alternate model via settings, likely in package configuration.

taicchoumsft:
## Description
Add a notification when Orchestrator is downloading models.   

This is also the start of the work to move Orchestrator model downloading to outside of the publishing pipeline.   It will be possible to trigger the download of the models anytime with the use of the `downloadModel` dispatcher.

## Task
fixes #5615


Composer/packages/client/src/components/Orchestrator/DownloadNotification.tsx
```diff
@@ -0,0 +1,43 @@
...
+`;
+
+const cardDetail = css`
+  margin-left: 8px;
+  flex-grow: 1;
+`;
+
+const infoType = css`
+  margin-top: 4px;
+  color: #0078d4;
```
lei9444, line 26: We can use a color from Fluent, I think this color is SharedColors.CyanBlue10

lei9444:
Tested and LGTM

Composer/packages/client/src/components/Orchestrator/DownloadNotification.tsx
taicchoumsft: Thanks @lei9444 , resolved.

boydc2014 (approved):

PR #5851, opened 2021-02-20
fix: error message specificity, tempalteID state management, runtime selection data handling

### LINKED ISSUE #5556: Pass 'runtime type' from Composer UI, through creation server call, to yeomen template
The creation UI has fields for runtime language to be used when bootstrapping the template. Currently this field selection is ignored. 

- [x] When the runtime generator is set up to accept a runtime flag then implement the data passing to the runtime instantiation from the ui
- [x] in R12 we do not support multi local component model templates so force the user to only be able to select english

### LINKED ISSUE #5783: Send more detailed error If any step in creation fails
If any step in creation fails (build, dialog merge, etc) then the current flow sends back a general 'server sent 500 response' error. 

Change the logic so a more detailed error response is sent back to the user.

### LINKED ISSUE #5850: Change Composer routing logic to handle templates with '@microsoft/' prefix
Component model template IDs correspond to their respective package manager names. With recent packages now in the Microsoft npm namespace there are package ids with the '@microsoft/' prefix.

Template IDs were previously passed through URL route paths which is no longer viable with these new npm package names.

- [x] Handle selected templateID through local state in Creation flow as opposed to routing path
- [x] Normalize display name to not include microsoft prefix

pavolum:
## Description

Various small items for the new creation experience are handled in this PR. More details outline in each task item below 

## Task Item

- fixes #5850
  - New logic for templateID state management and template display implemented to account for new '@microsoft/' npm pacakges that represent the template yeomen generators
- fixes #5556 
  - Users selection for desired runtime now passed to template generators 
  - Template generators still need to do something with this info, tracked here (issue 601 in components repo)  
- fixes #5783
  - Change has been made so any errors in the creation flow that arise are surfaced to the user as opposed to the previous generic '500 server call' that was previously surfaced


pavolum:
TODO: handle strong typing of runtime selection (azureWebApp and azureFunction), currently it is just passing the string from the front end, to the server, to the yeomen generators

Composer/packages/server/src/models/asset/assetManager.ts
```diff
@@ -281,8 +282,11 @@ export class AssetManager {
     if (packageName) {
       return packageName
         .replace('generator-', '')
+        .replace('@microsoft/', '')
+        .replace('microsoft', '')
```
benbrown, line 286: this seems really brittle! we can make this more specific i think, at least like ^generator and ^@microsoft/.

Why the second removal of microsoft?

Composer/packages/server/src/models/asset/assetManager.ts
```diff
@@ -281,8 +282,11 @@ export class AssetManager {
...
         .split('-')
-        .reduce((a, b) => a.charAt(0).toUpperCase() + a.slice(1) + ' ' + b.charAt(0).toUpperCase() + b.slice(1));
+        .reduce((a, b) => a.charAt(0).toUpperCase() + a.slice(1) + ' ' + b.charAt(0).toUpperCase() + b.slice(1))
```
benbrown, line 288: this all seems complicated. what is this for?

benbrown: oh i see it is to generate the "pretty" name.

ok maybe we should try to refactor this in the future to use some other mechanism.

Composer/packages/client/src/components/CreationFlow/v2/CreationFlow.tsx
```diff
@@ -146,9 +148,9 @@ const CreationFlowV2: React.FC<CreationFlowProps> = () => {
     }
   };
 
-  const handleCreateNext = async (data: string) => {
+  const handleCreateNext = async () => {
     setCreationFlowStatus(CreationFlowStatus.NEW_FROM_TEMPLATE);
-    navigate(`./create/${data}`);
+    navigate(`./create/template`);
```
benbrown, line 153: I wonder if some of this could have been avoided by doing an encodeURIcomponent on the name value...

benbrown: I confirmed this was the case.

benbrown:
DO NOT MERGE.

This PR is replaced by 3 smaller ones:
#5865
#5880
#5881

PR #5875, opened 2021-02-22

GeoffCoxMSFT:

### DIFF
--- a/Composer/packages/server/schemas/sdk.fr.uischema
+++ b/Composer/packages/server/schemas/sdk.fr.uischema
@@ -1,11 +1,11 @@
 {
   "Microsoft.AdaptiveDialog": {
     "form": {
-      "description": "Cela configure un dialogue piloté par les données via une collection d'événements et d'actions.",
+      "description": "Cela configure un dialogue piloté par les données via une collection d\"événements et d\"actions.",
       "label": "Dialogue adaptatif",
       "properties": {
         "recognizer": {
-          "description": "Pour comprendre ce que dit l'utilisateur, votre dialogue a besoin d'un « module de reconnaissance » qui comprend des exemples de mots et de phrases utilisables par les utilisateurs.",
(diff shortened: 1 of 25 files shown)


Composer/packages/server/schemas/sdk.fr.uischema
```diff
@@ -1,11 +1,11 @@
 {
   "Microsoft.AdaptiveDialog": {
     "form": {
-      "description": "Cela configure un dialogue piloté par les données via une collection d'événements et d'actions.",
+      "description": "Cela configure un dialogue piloté par les données via une collection d\"événements et d\"actions.",
       "label": "Dialogue adaptatif",
       "properties": {
         "recognizer": {
-          "description": "Pour comprendre ce que dit l'utilisateur, votre dialogue a besoin d'un « module de reconnaissance » qui comprend des exemples de mots et de phrases utilisables par les utilisateurs.",
+          "description": "Pour comprendre ce que dit l\"utilisateur, votre dialogue a besoin d\"un « module de reconnaissance » qui comprend des exemples de mots et de phrases utilisables par les utilisateurs.",
```
a-b-r-o-w-n, line 8: This looks a little off. I expect that `'` is the right punctuation here.

GeoffCoxMSFT: @beyackle How do we handle escaping that is compatible with OneLocBuild again?

beyackle: Apostrophes should be replaced by the curly "smart" version (which is `’`, U+2019).

beyackle (changes requested):
See line comment; apostrophes should become U+2019, not an escaped double quote.

Composer/packages/server/schemas/sdk.fr.uischema
GeoffCoxMSFT: I'll PR updates shortly.  I used this regex to find cases of the problem. formatMessage\(".*'.*". 
@a-b-r-o-w-n  or @beyackle  - Is there a way to check for matches pre-PR or during CI to prevent future problems?

GeoffCoxMSFT: PR is out, so I'm closing this PR.

PR #5880, opened 2021-02-22
fix: return more verbose error during new creation for display in the client

### LINKED ISSUE #5783: Send more detailed error If any step in creation fails

benbrown:
## Description

This PR exposes the error message from inside background build processes to the client for display.

## Task Item
fixes #5783

### DIFF
--- a/Composer/packages/server/src/controllers/status.ts
+++ b/Composer/packages/server/src/controllers/status.ts
@@ -9,7 +9,11 @@ const getStatus = async (req, res) => {
     if (jobId) {
       const result = BackgroundProcessManager.getProcessStatus(jobId);
       if (result) {
-        res.status(result.httpStatusCode).json(result);
+        res.status(result.httpStatusCode).json({
+          ...result,
+          statusCode: result.httpStatusCode,
+          message: result.latestMessage,
+        });
       } else {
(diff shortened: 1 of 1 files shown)


Composer/packages/server/src/controllers/status.ts
```diff
@@ -9,7 +9,11 @@ const getStatus = async (req, res) => {
     if (jobId) {
       const result = BackgroundProcessManager.getProcessStatus(jobId);
       if (result) {
-        res.status(result.httpStatusCode).json(result);
+        res.status(result.httpStatusCode).json({
```
srinaath, line 12: are we ok with sending the additional properties httpStatusCode and latestMessage in addition to the statusCode and message. Feels like the response will have duplicate fields for status code and message.

benbrown: I think that's ok -- the client is looking for statusCode and message only in order to display them in the UI. No harm in leaving the other fields in there as there other fields in that object as well that are used at times in the UI such as the history and log output of the background process.

PR #5881, opened 2021-02-23
feat: enable new creation flow as default

### LINKED ISSUE #5556: Pass 'runtime type' from Composer UI, through creation server call, to yeomen template

### LINKED ISSUE #6416: Add support for JS runtime during creation flow
* Expose the JS feed
* Add support for package manager
* Add support for local start
* Add support for publishing
* plumb through the choice to the yeoman generator stuff

benbrown:
## Description

This includes the functionality necessary for creating and using the new Adaptive Runtime in Javascript, and Azure Functions functionality in both Javascript and Dotnet.

In addition, this makes the "new create flow" the default and ONLY way to create new bots.  This removes the "new creation flow" feature flag and enables all features previously hidden.

## Task Item

fixes #5556 
fixes  #6416


Composer/packages/types/src/creation.ts
```diff
@@ -22,3 +22,8 @@ export const nodeFeedKey = 'firstPartyNode';
 export const defaultFeeds = [nodeFeedKey, csharpFeedKey] as const;
 export type FeedName = typeof defaultFeeds[number];
 export type FeedType = 'npm' | 'nuget';
+
+export const webAppRuntimeKey = 'azureWebApp';
+export const functionsRuntimeKey = 'azureFunctions';
+export const availableRunTimes = [webAppRuntimeKey, functionsRuntimeKey] as const;
```
a-b-r-o-w-n, line 28: This looks like a case for an enum.

a-b-r-o-w-n (dismissed):

Composer/packages/types/src/creation.ts
pavolum: I have used enums for this in the past and have been told by the team to use this method instead to enforce the 'types  > enums' standard. 

@cwhitten & @hatpick I believe we discussed this in the past and this method was advised. Please correct me if I am wrong.

a-b-r-o-w-n: I suppose at the end of the day the enum is used as a value instead of a type, however, what I'm concerned with is requiring the types package to be a prod dependency instead of a dev dependency. Maybe sense @hatpick has opinions here he can chime in with some more detail, but ideally we would be able to convert the types package into a .d.ts file -- I'm not sure if using `const` would negate that ability.

hatpick: That's a valid point @a-b-r-o-w-n, but isn't js for enum basically an IIFE that assigns enum values to the exported value as well, like this:

enum MyEnum {
  A = "A",
  B = "B"
}

becomes:
var MyEnum;
(function (MyEnum) {
    MyEnum["A"] = "A";
    MyEnum["B"] = "B";
})(MyEnum = exports.MyEnum || (exports.MyEnum = {}));

a-b-r-o-w-n: This is true @hatpick. Do you know how @types packages handle enums?

hatpick: I think typescript inlines all enums, for example ```MyEnum.A``` is replace by ```'A'```, so there is no reference to ```MyEnum``` at runtime.

Composer/packages/client/src/components/CreationFlow/v2/CreateBot.tsx
```diff
@@ -200,27 +192,39 @@ export function CreateBotV2(props: CreateBotProps) {
   };
 
   const getTemplate = (): BotTemplate | undefined => {
-    const currentTemplate = templates.find((t) => {
-      return t.id === currentTemplateId;
+    const currentTemplate = displayedTemplates.find((t) => {
+      if (t?.id) {
+        return t.id === currentTemplateId;
```
hatpick, line 197: without ```if``` it could be ```t.?id === currentTemplateId```

Composer/packages/client/src/components/CreationFlow/v2/CreateBot.tsx
```diff
@@ -200,27 +192,39 @@ export function CreateBotV2(props: CreateBotProps) {
...
+        setDisplayedTemplates(newTemplates);
       }
     }
-  }, [templates]);
+  }, [templates, selectedProgLang]);
 
   useEffect(() => {
-    if (selectedFeed?.props?.itemKey) {
-      props.fetchTemplates([feedDictionary[selectedFeed.props.itemKey]]);
+    if (displayedTemplates.length > 1) {
```
hatpick, line 222: ```if(displayedTemplates?.[0]?.id) ...```

Composer/packages/client/src/components/CreationFlow/v2/DefineConversation.tsx
```diff
@@ -70,8 +72,8 @@ type DefineConversationFormData = {
   name: string;
   description: string;
   schemaUrl: string;
-  primaryLanguage: string;
-  runtimeChoice: string;
+  runtimeLanguage: string;
+  runtimeChoice: RuntimeType;
```
hatpick, line 76: nit: runtimeType seems like an obvious choice for the prop name

Composer/packages/client/src/components/CreationFlow/v2/DefineConversation.tsx
```diff
@@ -289,6 +293,28 @@ const DefineConversationV2: React.FC<DefineConversationProps> = (props) => {
...
+    const currentTemplate = templateProjects.find((t) => {
+      if (t?.id) {
+        return t.id === templateId;
+      }
+    });
+
+    if (currentTemplate) {
+      if (runtimeLanguage === csharpFeedKey) {
+        currentTemplate.dotnetSupport?.functionsSupported &&
+          result.push({ key: functionsRuntimeKey, text: 'Azure Functions' });
```
hatpick, line 307: formatMessage (if needed, there are 3 more below)

extensions/azurePublish/src/node/deploy.ts
```diff
@@ -174,7 +175,7 @@ export class BotProjectDeploy {
   private async deployZip(token: string, zipPath: string, name: string, env: string, hostname?: string) {
     this.logger({
       status: BotProjectDeployLoggerType.DEPLOY_INFO,
-      message: 'Uploading zip file...',
+      message: `Uploading zip file... to ${hostname ? hostname : name + (env ? '-' + env : '')}`,
```
hatpick, line 178: does this need localization?

hatpick (dismissed):

Composer/packages/client/src/components/CreationFlow/v2/DefineConversation.tsx
benbrown: good suggestion i changed this everywhere.

Composer/packages/client/src/constants.ts
```diff
@@ -399,25 +399,22 @@ export const triggerNotSupportedWarning = () =>
     'This trigger type is not supported by the RegEx recognizer. To ensure this trigger is fired, change the recognizer type.'
   );
 
-export const feedDictionary: { [key in FeedName]: string } = {
-  firstPartyCsharp:
-    'https://registry.npmjs.org/-/v1/search?text=conversationalcore+scope:microsoft&size=100&from=0&quality=0.65&popularity=0.98&maintenance=0.5',
-  firstPartyNode: '',
-};
+export const firstPartyTemplateFeed =
+  'https://registry.npmjs.org/-/v1/search?text=generator+keywords:bf-template+scope:microsoft'; // +maintainer:botframework
```
beyackle, line 403: Is there a reason the extra query parameters (like quality and popularity) are going away here?

Composer/packages/server/src/models/asset/assetManager.ts
```diff
@@ -276,6 +281,21 @@ export class AssetManager {
...
+            if (keywords.indexOf('bf-dotnet-functions') != -1 || keywords.indexOf('bf-dotnet-webapp') != -1) {
+              templateToReturn.dotnetSupport = {
+                functionsSupported: keywords.indexOf('bf-dotnet-functions') != -1,
+                webAppSupported: keywords.indexOf('bf-dotnet-webapp') != -1,
+              };
+            }
+            if (keywords.indexOf('bf-js-functions') != -1 || keywords.indexOf('bf-js-webapp') != -1) {
+              templateToReturn.nodeSupport = {
+                functionsSupported: keywords.indexOf('bf-js-functions') != -1,
+                webAppSupported: keywords.indexOf('bf-js-webapp') != -1,
```
beyackle, line 294: Should be able to use `keywords.includes` here instead of indexOf.

Composer/packages/client/src/components/CreationFlow/v2/CreateBot.tsx
```diff
@@ -148,21 +148,13 @@ export function CreateBotV2(props: CreateBotProps) {
...
-    }
-
-    if (option === optionKeys.createFromQnA) {
-      routeToTemplate = QnABotTemplateId;
-    }
-
-    TelemetryClient.track('CreateNewBotProjectNextButton', { template: routeToTemplate });
+    TelemetryClient.track('CreateNewBotProjectNextButton', { template: currentTemplateId });
 
+    const runtimeLanguage = selectedProgLang?.props?.itemKey ? selectedProgLang.props.itemKey : csharpFeedKey;
```
beyackle, line 153: If this is a check if itemKey is null, then I think you can use:
```suggestion
    const runtimeLanguage = selectedProgLang?.props?.itemKey ?? csharpFeedKey;
```

Composer/packages/client/src/components/CreationFlow/v2/CreateBot.tsx
```diff
@@ -200,27 +192,35 @@ export function CreateBotV2(props: CreateBotProps) {
...
+        const newTemplates = templates.filter((template) => {
+          return template.dotnetSupport;
+        });
+        setDisplayedTemplates(newTemplates);
+      } else if (itemKey === nodeFeedKey) {
+        const newTemplates = templates.filter((template) => {
+          return template.nodeSupport;
+        });
+        setDisplayedTemplates(newTemplates);
       }
```
beyackle, line 215: If `csharpFeedKey` and `nodeFeedKey` will never themselves be null, I think you can get rid of the outer `if`.

beyackle (changes requested):
Mostly looks good, just needs a few small changes.

Composer/packages/client/src/constants.ts
pavolum: They were unneeded, we want to ensure that every generator in npm within the Microsoft scope, with the bf-template keyword is returned by this feedUrl. Its on our backlog to add the Maintainer:botframework to this search query to enforce a stricter contract. These at the only qualifiers we need for the template feed so things like quality and popularity are not valid qualifiers for our use case.

Composer/packages/server/src/models/asset/assetManager.ts
pavolum: Good catch, I have changed it to use .includes()

Composer/packages/client/src/components/CreationFlow/v2/CreateBot.tsx
pavolum: Sounds good, I have removed the outer if, the logic should ensure that it is never null

Composer/packages/client/src/components/CreationFlow/v2/CreateBot.tsx
pavolum: Good catch, I have made the change

Composer/packages/server/src/models/asset/assetManager.ts
```diff
@@ -282,13 +282,13 @@ export class AssetManager {
...
           if (isArray(keywords)) {
-            if (keywords.indexOf('bf-dotnet-functions') != -1 || keywords.indexOf('bf-dotnet-webapp') != -1) {
+            if (keywords.includes('bf-dotnet-functions') || keywords.indexOf('bf-dotnet-webapp') != -1) {
               templateToReturn.dotnetSupport = {
                 functionsSupported: keywords.indexOf('bf-dotnet-functions') != -1,
                 webAppSupported: keywords.indexOf('bf-dotnet-webapp') != -1,
               };
             }
-            if (keywords.indexOf('bf-js-functions') != -1 || keywords.indexOf('bf-js-webapp') != -1) {
+            if (keywords.includes('bf-js-functions') || keywords.indexOf('bf-js-webapp') != -1) {
```
beyackle: You missed a couple here.

beyackle (commented):
One minor fix and I'll give it the ✅.

Composer/packages/server/src/models/asset/assetManager.ts
pavolum: Resolved

Composer/cypress/integration/CreateNewBot.spec.ts
```diff
@@ -1,35 +0,0 @@
-// Copyright (c) Microsoft Corporation.
-// Licensed under the MIT License.
-
-context('Creating a new bot', () => {
```
cwhitten, line 4: can this be replaced with an integration test on the new creation experience?

pavolum: Yeah I am working on one now

pavolum: So I added a simple integration test for testing the creation of empty bot e2e. It does have to wait a bit since the creation takes a bit prior to asserting that the bot was created successfully. I suppose we could mock the server response and bot creation but that would probably defeat the purpose of the integration test a bit.

pavolum: new creation Integration test runs fine locally but breaking the build on the CI machine, I am unable to see the logs to see what the CI machine has issues with when running the test, the logs link to a 404 page. 

My guess is the CI machines do not have node? Or perhaps the amount of time the UT waits prior to asserting on the post creation view is not long enough since the CI machine is installing the template each time from npm which takes a bit the first time it is run. I dont think it is that though since I have played with large wait times and the CI machine still fails. 

I need to figure out how best to run this test as it is the only blocker from getting this merged

pavolum:
Status update: We have fixed the integration test issue by ensuring dotnet is installed on the CI machine prior to running the unit tests. That was our blocker to get this through to main. Validating the fix now. 

I will also update the integration test shortly so it does not wait an arbitrary amount of time before making assertions on post creation app state and instead makes those assertions right when the operation completes. Besides that we are looking good to merge.

beyackle (approved):

boydc2014 (approved):

PR #5883, opened 2021-02-23
fix: add static QnA and Luis entities and remove duplicate entities

### LINKED ISSUE #5882: LG Language Server suggested entities does not work correctly.
1. Missing QnA and Luis static members: `@answer`, `@itemList` 
2. Suggest duplicate entities from multiple files

MicrosoftTeams-image

cosmicshuai:
closes #5882

1. Add `@answer` and `@itemList` to suggested entities

2. remove duplicate entities from multiple lu files.

image

### DIFF
--- a/Composer/packages/tools/language-servers/language-generation/src/lgParser.ts
+++ b/Composer/packages/tools/language-servers/language-generation/src/lgParser.ts
@@ -7,8 +7,9 @@ import path from 'path';
 import { lgImportResolverGenerator, LgFile, ResolverResource } from '@bfc/shared';
 import uniqueId from 'lodash/uniqueId';
 import { lgUtil } from '@bfc/indexers';
+import uniq from 'lodash/uniq';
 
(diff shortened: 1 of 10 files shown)


Composer/packages/tools/language-servers/language-generation/src/lgParser.ts
```diff
@@ -33,15 +33,15 @@ class LgParserWithoutWorker {
...
+    let suggestEntities: string[] = ['answer', 'itemList'];
     if (luContents) {
       for (const content of luContents) {
         const luisJson = await extractLUISContent(content);
         suggestEntities = suggestEntities.concat(getSuggestionEntities(luisJson, suggestionAllEntityTypes));
       }
     }
 
-    return { suggestEntities: suggestEntities };
+    return { suggestEntities: uniq(suggestEntities) };
```
lei9444, line 44: we can use uniq from lodash direttly

cosmicshuai: changed

boydc2014 (approved):

PR #5936, opened 2021-02-24
fix: Lg Editor bus and ux improvement from bug bash

### LINKED ISSUE #5868: LG Editor: Rename 'Add new variation' on Suggested Actions tab to 'Add new action'
The text link allowing you to add a Suggested Action needs to be renamed to 'Add Suggested Action' instead of 'Add new variation'.

### LINKED ISSUE #5869: LG Editor: Tabbing from a variation input box selects an invisible delete icon
Tabbing out of a variation input box navigates to the invisible delete icon for the next item. Continuing to tab moves to the next invisible delete icon. 

Assuming this might also affect the use of the same control in the QnA KB section?

### LINKED ISSUE #5870: LG Editor: Removing a response tab leaves behind redundant LG templates
When removing a tab from the LG Editor (e.g. Attachments) any additional LG templates that were generated to support the tab's content are left behind and are not removed, which can leave a bot in a broken state.

e.g.  

1. Add a Send A Response action
2. Set a text reponse
3. Add an attachments tab and select a Hero Card template (don't complete the placeholders) which leaves an LG syntax error.
4. Remove the tab - error still exists as template was not removed from LG file.

Removing a tab should remove the generated templates (providing they are not being referenced anywhere else).

### LINKED ISSUE #5871: LG Editor: Suggested Actions tab disappears if only one Suggested Action added
1. Add Suggested Actions tab and then add 3 suggested actions
2. Switch to code editor and remove all but 1
3. Switch back to Response editor, 
4. The Suggested Actions tab has disappeared. 

Suggested Action is still there in the Code Editor, so looks like the tab just isn't being displayed for a single Suggested Action. 

Also confirmed that using the editor to add 1 action, navigate to another action and then back again, the tab has disappeared.

### LINKED ISSUE #5872: LG Editor: Adaptive Card template comment causes LG errors.
Use the LG Editor to add an Attachments tab to a response and then create an attachment, selecting the Adaptive Card template.  At this point, if a user does not replace the template content with an adaptive card JSON they have multiple LG errors.  

Is there a way we can have the Adaptive Card template either be valid at the outset or maybe some form of placeholder instead?

### LINKED ISSUE #5873: LG Editor: Pre-Built functions should be shown alphabetically
The pre-built functions list shown when you choose the functions helper on the LG editor shows a long list of functions.  Currently they don't appear to be in any sort of order, making them potentially hard to locate.  We should re-order the main and sub menus to be displayed alphabetically.

Note, if we take this then we should also update the functions list shown as a dropdown when editing conditions.

### LINKED ISSUE #5874: LG Editor: Attachment templates copy / order
A few amendments to the Attachments reponse tab.

1. Given that you only create one attachment at a time, I think we need to rename the 'Create from templates' menu button to 'Create from template'.

2. Move the 'Create from template' button to above the 'Custom' button.

3. Can we please re-order the attachment templates as follows (moving the lesser used / known templates down - I did consider ordering alphabetically here, but given the short list and the high degree of confidence that the top 3 below are used more than audio / video, I think this order is sensible)

	- Adaptive card
	- Hero card
	- Sign-in card
	- Audio card
	- Video card
        - Animation card

hatpick:
## Description

This PR fixes bugs from bug bash.
It also has UX improvements.

## Task Item
closes #5873 
closes #5869 
closes #5874 
closes #5872
closes #5871 
closes #5870 
closes #5868 

@tdurnford


Composer/packages/lib/code-editor/src/lg/modalityEditors/ModalityEditorContainer.tsx
```diff
@@ -105,10 +105,30 @@ export const ModalityEditorContainer: React.FC<Props> = ({
...
+                          {
+                            key: 'remove',
+                            text: formatMessage(
+                              'Yes, I’d like to remove the the content of this tab completely from the LG file.'
+                            ),
+                          },
+                          {
+                            key: 'keep',
+                            text: formatMessage(
+                              'No, I want to keep the content, just de-reference from this response.'
```
GeoffCoxMSFT, line 122: de-reference could be confusing to some users.  How about this?

No, I want to keep the content in the LG file and only remove the reference from this response.

GeoffCoxMSFT (approved):

Composer/packages/lib/code-editor/src/lg/modalityEditors/ModalityEditorContainer.tsx
tdurnford: @sangwoohaan Can you weigh in?

PR #5945, opened 2021-02-24
fix: Fixed issue with adding structured choices to choice prompts

### LINKED ISSUE #5751: Cannot Add Structured Choice in Prompt For Confirmation
Its impossible to add structured choice inputs with a defined action when configuring a Prompt for Confirmation.
the whole choice disappears when clicking out of the Action Json field.  

## Version

V1.3.0
image

## OS

Windows 10

## To Reproduce

1. Add a prompt for confirmation, 
2. go to user input
3. add a confirm choice of type structured choice
4. Then add data to an "Action" field
5. click anywhere else.

Results in the whole structured choice disappearing

## Expected behavior

Structured Choice should be Added

## Screenshots
Please see Gif of the Issue which clearly shows the problem.

ConfirmChoice-StructuredInput Bug
)

tdurnford:
## Description
The `<BaseEditor />`'s `onBlur` handler was not updating correctly which caused the `<ArrayItemField />` to incorrectly remove the structured choice. Adding `onBlur` and `onFocus` to the `useEffect` deps array resolved the issue.

## Task Item
Closes #5751


hatpick (approved):

Composer/packages/lib/code-editor/src/BaseEditor.tsx
```diff
@@ -188,7 +188,7 @@ const BaseEditor: React.FC<BaseEditorProps> = (props) => {
         onBlurListener.dispose();
       };
     }
-  }, [editorRef.current]);
+  }, [editorRef.current, onBlur, onFocus]);
```
a-b-r-o-w-n, line 191: One thing that could catch this is the eslint rule for exhaustive deps. We should look into enabling that.

a-b-r-o-w-n (approved):

PR #5838, opened 2021-02-19
fix: remove translate relative path to absolute path for runtime path

alanlong9278:
## Description
fix: remove translate relative path to absolute path for runtime path.

Have commented the translate function. Call me if there are any question I ignore. @pavolum 

```
 // if (settings?.runtime?.customRuntime && settings.runtime.path && !Path.isAbsolute(settings.runtime.path)) {
    //   const absolutePath = Path.resolve(this.dir, 'settings', settings.runtime.path);

    //   if (fs.existsSync(absolutePath)) {
    //     settings.runtime.path = absolutePath;
    //     await this.updateEnvSettings(settings);
    //   }
    // }
```

## Task Item

closes #5786

PR #5950, opened 2021-02-24
fix: support relative path to runtime

### LINKED ISSUE #5786: Composer translates relative paths in runtime settings to absolute paths
## Describe the bug

Composer incorrectly changes relative paths to absolute paths. This defeats the purpose of allowing relative paths since if a project folder moves, the path breaks.  

## Version

Current main

## To Reproduce

1. navigate to settings
2. change the runtime path to "../"
3. reload the page. the path will be replaced with an absolute path

## Expected behavior

The relative path should remain unmodified.

## Screenshots
sRmGK578RD

benbrown:
## Description

* Remove translation of the runtime relative path to an absolute path
* Update package manager, local publish and azurePublish to properly handle the relative path when used.

## Task Item

Fixes #5786

Replaces #5838


Composer/packages/server/src/models/bot/botProject.ts
```diff
@@ -202,16 +202,6 @@ export class BotProject implements IBotProject {
...
 
-    // Resolve relative path for custom runtime if the path is relative
-    if (settings?.runtime?.customRuntime && settings.runtime.path && !Path.isAbsolute(settings.runtime.path)) {
-      const absolutePath = Path.resolve(this.dir, 'settings', settings.runtime.path);
-
-      if (fs.existsSync(absolutePath)) {
-        settings.runtime.path = absolutePath;
-        await this.updateEnvSettings(settings);
-      }
-    }
```
a-b-r-o-w-n, line 213: Why was this code added in the first place? Will we have a regression if we remove it?

benbrown:
@alanlong9278  can you check this out and double check it for me? I *think* I've addressed all the spots where the runtime path is used.

extensions/packageManager/src/node/index.ts
```diff
@@ -352,7 +351,7 @@ export default async (composer: IExtensionRegistration): Promise<void> => {
 
       let runtimePath = currentProject.settings?.runtime?.path;
       if (runtimePath && !path.isAbsolute(runtimePath)) {
-        runtimePath = path.resolve(currentProject.dir, runtimePath);
+        runtimePath = path.resolve(currentProject.dir, 'settings', runtimePath);
       }
```
a-b-r-o-w-n, line 355: I'm a little worried about the amount of duplicated code. Can we add a method to the project that centralizes this?

Composer/packages/server/src/models/bot/botProject.ts
benbrown: This was incorrectly added during an earlier part of the creation work. It was introduced after the r11 release.

extensions/packageManager/src/node/index.ts
benbrown: good idea.

benbrown: @a-b-r-o-w-n I addressed this.  Want to give it another look?

PR #5983, opened 2021-02-25
fix: generate friendlier project names from new long package names.

### LINKED ISSUE #5948: In new create flow, default name for new projects if long and unwieldy
The default project name for the bot is based on the template name.

However, with the new templates that have names like "@microsoft/generator-microsoft-bot-calendar" we end up with very long and awkard default names!

We should update the new creation flow to use the "display name" as the default name, so something like "empty_bot" instead of "_microsoft_generator_microsoft_bot_empty"

### LINKED ISSUE #5987: Copy updates for creation modal
## Describe the bug

1) Should say "Select a template" rather than "Choose a template for your bot".
2) Description copy should have a period at the end.

## Version

v1.3.1 special creation release

## Browser

- [ ] Electron distribution

## OS

- [ ] macOS

## Screenshots

image

benbrown:
## Description

this applies a similar heuristic used by the server to generate pretty names for the project name based on the package name.

@microsoft/generator-microsoft-bot-conversational-core becomes "conversational_core"

this should preserve the current naming behavior, but apply it to the new flows.

## Task Item

fixes #5948 
fixes #5987 

## Screenshots

30nzOLRWFK


Composer/packages/client/src/components/CreationFlow/v2/DefineConversation.tsx
```diff
@@ -121,23 +120,30 @@ const DefineConversationV2: React.FC<DefineConversationProps> = (props) => {
   // template ID is populated by npm package name which needs to be formatted
   const normalizeTemplateId = (templateId?: string) => {
     if (templateId) {
-      return templateId.replace(invalidNameCharRegex, '_');
+      console.log('NORMALIZE ', templateId);
```
cwhitten, line 123: remove?

benbrown: removed!

cwhitten (approved):

PR #5999, opened 2021-02-26
chore: remove $kind restriction on Flow Editor layouter

### LINKED ISSUE #5395: Declarative UI widgets in Custom Actions
Sometimes one needs to ask the customer for a list of options, eg which products would you like to add. The user would be able to add products / text until they are ready and hit the I'm done button. So this is basically a loop that keeps on running, asking a value, adding value to array and repeat. Repeating the dialog / re-submitting events is in this case to coarse and would make the dialog structure quite bad.

Currently I fix this using a foreach loop with a dummy array as loop value, like the image below.

This is ofocurse a bit of a poormans solution. Having a Do While loop with a condition and as bonus an internal counter for first iteration behavior etc, would be a real clean approach for this. The condition could be falsified by the activity.value of the button (or any other of the million usecases).

image

yeze322:
## Description
closes #5395 

**Summary**
Be able to reuse builtin branching widgets in custom Actions. (`IfConditionWidget` | `SwitchConditionWidget` | `ForeachWidget`)

**Changes**
1. Remove $kind checking in transfomers of `IfConditionWidget` | `SwitchConditionWidget` | `ForeachWidget` to reuse them.
2. Implicitly add `nowrap` flag to all above widgets in case uischema author forgets to add it. (without this flag, the layout will be messed up)

**Explaination**:

Now that we already supported uischema driven Action customization, the Flow Editor's layouter shouldn't restrict the input $kind anymore. By removing the $kind checking logic, those branching widgets are opened to external $kinds.

External Action $kind can then declare the `widget` field as one of `IfConditionWidget` | `SwitchConditionWidget` | `ForeachWidget` in uischema flow part.

**Follow-up**:
@zxyanliu let's document it!

## Screenshots
1. Reuse `IfConditonWidget`
image

image

2. Reuse ForeachWidget
image
image


yeze322:
@cwhitten @EricDahlvang please review

boydc2014:
a lot of adaptive-flow UT seems to be failing

yeze322:
@boydc2014 Oops, it's the circular reference problem

boydc2014 (approved):

PR #6016, opened 2021-02-26
fix: update default nuget search feed url

benbrown:
## Description

This updates the default feed URL used by Composer to search nuget to include the `microsoft.bot.components` prefix and the appropriate semVerLevel parameter.

#minor


carlosscastro (approved):

carlosscastro:
Note that in the future, I'd love these not having query string urls here and just be 'https://api.nuget.org/v3/index.json' and have the query part be hidden from users (but potentially overridable through UI). Dotnet community uses the package feed url only, so we cannot ask them to manage query strings to azure search instances down the road.

PR #6020, opened 2021-02-26
fix: exclude new runtime from ejection list

### LINKED ISSUE #5979: Bugbash (private): Runtime language selection have duplicate C# entries
From private daily build?

image

benbrown:
This PR removes the new runtime from the eject list, removing the duplicated "C#" option.

behind the scenes, this is done by filtering on the existence of the rutnime's eject method, which has been (appropriately) removed from the new v2 runtime.

## Task Item

Fixes #5979


extensions/runtimes/src/index.ts
```diff
@@ -452,33 +452,6 @@ export default async (composer: any): Promise<void> => {
       // return the location of the build artifiacts
       return publishFolder;
     },
-    eject: async (project, localDisk: IFileStorage, isReplace: boolean) => {
```
cwhitten, line 455: is this not the `eject` implementation for the current runtime?

benbrown: no this is the one in the new runtime definition

cwhitten (approved):

PR #6032, opened 2021-02-28
fix: Orchestrator downloads multi-language model when it is not needed

### LINKED ISSUE #5976: Bug Bash v1.3.1 - starting with existing english only bot, but multilingual model is being downloaded
## Describe the bug
Using nightly build v1.3.1
1. Open Eyal's School Navigator bot which is English only bot
2. Start bot
3. Once model is downloaded (start bot status is now 'Starting bot...', go to %APPDATA%\BotFrameworkComposer\models

Expected:
Only english model folder

Actual:
English and multilingual model folders are created

## Version
1.3.1

## Browser

- [ ] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [X] Windows
- [ ] Ubuntu

taicchoumsft:
## Description
The previous multilanguage check was a simple array length check that failed when multiple english recognizer dialogs/settings were present (e.g. when Orchestrator is used in multiple dialogs)  

Now we strengthen the check and check against valid UI locales before downloading the multilanguage model. 

## Task Item
closes #5976


Composer/packages/client/src/recoilModel/dispatchers/orchestrator.ts
```diff
@@ -33,46 +35,55 @@ export const downloadModel = (addr: string, model: 'en' | 'multilang') => {
   });
 };
 
-export const orchestratorDispatcher = () => {
-  const downloadLanguageModels = useRecoilCallback(
-    ({ set, snapshot }: CallbackInterface) => async (projectId: string) => {
-      const recognizers = await snapshot.getPromise(recognizersSelectorFamily(projectId));
+export const availableLanguageModels = (recognizerFiles: RecognizerFile[]) => {
+  const dialogsUsingOrchestrator = recognizerFiles?.filter(
```
hatpick, line 39: nit: if the recognizerFiles  as arg is not optional, ?. is not necessary

Composer/packages/client/src/recoilModel/dispatchers/orchestrator.ts
```diff
@@ -33,46 +35,55 @@ export const downloadModel = (addr: string, model: 'en' | 'multilang') => {
...
 
-      if (isUsingOrchestrator) {
-        // pull out languages that Orchestrator has to support
-        const botLanguages = recognizers
-          .filter(({ id, content }) => content.$kind === SDKKinds.OrchestratorRecognizer)
-          .map(({ id, content }) => id.split('.')?.[1]);
+    if (
+      multiLangLuFiles
+        .map((r) => r.id.split('.')?.[1])
+        ?.some((lang) => Locales.map((l) => l.locale).includes(lang?.toLowerCase()))
```
hatpick, line 56: nit: mapper maps to the array with the same size, not need to use ?.
however, I recommend filtering out undefined values from the resulting array.

Composer/packages/client/src/recoilModel/dispatchers/orchestrator.ts
```diff
@@ -33,46 +35,55 @@ export const downloadModel = (addr: string, model: 'en' | 'multilang') => {
...
-        }
+export const orchestratorDispatcher = () => {
+  const downloadLanguageModels = useRecoilCallback(({ snapshot }: CallbackInterface) => async (projectId: string) => {
+    const recognizers = await snapshot.getPromise(recognizersSelectorFamily(projectId));
 
-        // TODO: We need to get a table of Orchestrator supported languages
-        if (botLanguages.length > 1) {
-          languageModels.push('multilang');
-        }
+    const isUsingOrchestrator = recognizers?.some(
```
hatpick, line 68: nit: can recognizers not have value?

taicchoumsft: Thanks @hatpick , addressed all your comments.  Yes, you're right the recognizer array will always have some value. Removed `?` from this line as well.

boydc2014 (approved):

PR #6034, opened 2021-02-28
[DO NOT MERGE][WIP] Package manager: tactical fix for minimum bar myget support

### LINKED ISSUE #6033: Package manager: Myget usage blocked when using service index (which is the standard)
The standard way of providing a nuget feed is passing the service index, adn this is what dotnet developers do on a daily basis. Currently, package manager relies on passing the search query for things to work, which is not the standard. Furthermore, if a standard service index is passed, the results are not filtered so it is impossible to get reasonable results. Also, it is not possible to get prerelease packages from a mygetr feed, given that all bot packages will be prerelease until //BUILD, this would block usage of myget all up.

carlosscastro:
## Description

Currently adding NuGet feeds 'the usual way', such as passing a service index `https://api.nuget.org/v3/index.json` is not working properly. A more advanced approach to fixing this is exposed in this[ alternate PR](https://github.com/microsoft/BotFramework-Composer/pull/6035), however this fix represents the minimum tactical fix to overcome this issue in r12.

## Task Item

Closes #6033

### DIFF
--- a/extensions/packageManager/src/node/index.ts
+++ b/extensions/packageManager/src/node/index.ts
@@ -48,7 +48,7 @@ export default async (composer: IExtensionRegistration): Promise<void> => {
       // this is actually a myget feed that points to the feed we want...
       const queryEndpoint = feed.resources.find((resource) => resource['@type'] === 'SearchQueryService');
       if (queryEndpoint) {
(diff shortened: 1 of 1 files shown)


carlosscastro:
We opted for the more complete version.

PR #6035, opened 2021-02-28
refactor: Package manager: enhance nuget protocol support, refactor functionality and prepare for future ui improvements

### LINKED ISSUE #6033: Package manager: Myget usage blocked when using service index (which is the standard)

carlosscastro:
Closes #6033 

## Description

- Adds improved support for NuGet protocol
- Refactors package feed functionality into future testeable units (tests incoming in R13)
- Adds concept of query that can be passed to feed
- Server side set up ready for adding search functionality in the UI: search text bot, pre-release check box and paging. Work needed is just build the UI and pass query object
- Next steps for full cleanup:
  - Only allow pointers to service index in package manager
  - Implement searchbox, prerelease check box and paging in package maanger UI
  - When users add a feed, need to specify npm or nuget to avoid using heuristics to decide the type of feed
A more minimal / tactical fix for the myget issue can be found in this [WIP PR](https://github.com/microsoft/BotFramework-Composer/pull/6034)

## Screenshots

Testing: 
Npm feed
image

NuGet 
image

MyGet: automatically filtered with tags: bf-component, prerelease, semver version
image

Local

image

Quick design for PR readability

image


extensions/packageManager/src/node/index.ts
```diff
@@ -247,30 +132,30 @@ export default async (composer: IExtensionRegistration): Promise<void> => {
...
+          const packageQuery: IPackageQuery = {
+            prerelease: true,
+            semVerLevel: '2.0.0',
+            query: 'tags:bf-component',
+          };
+
+          const packages = await feed.getPackages(packageQuery);
+
+          if (Array.isArray(packages)) {
+            combined = combined.concat(packages);
```
beyackle, line 150: I think `combined.push(...packages)` is the more idiomatic way to do this.

extensions/packageManager/src/node/feeds/nuget/nugetFeed.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    // If we don't have a service index provided at construction, get it from the url
+    if (!serviceIndex || !serviceIndex.resources) {
+      const httpResponse = await axios.get(this.packageSource.url);
+      serviceIndex = httpResponse?.data;
+    }
+
+    // Now that we have a service index either from seed data or from querying the NuGet feed,
+    // we get the query endpoint, build query parameters, call feed api and return query results to caller
+    const queryEndpoint = serviceIndex?.resources?.find((resource) => resource['@type'] === 'SearchQueryService');
```
beyackle, line 58: Is it possible for the array of resources to contain multiple search query services? If so, we should use `filter` instead of `find` and then iterate over all of them.

extensions/packageManager/src/node/feeds/nuget/nugetFeed.ts
```diff
@@ -0,0 +1,127 @@
...
+    if (queryEndpoint) {
+      const nuGetSearchUrl = this.buildNuGetSearchUrl(queryEndpoint['@id'], query);
+
+      this.composer.log(`Retrieving NuGet packages with computed search url ${nuGetSearchUrl}`);
+      const httpResponse = await axios.get(nuGetSearchUrl);
+
+      if (!httpResponse || !httpResponse.data) {
```
beyackle, line 65: `if (httpResponse?.data != null)` would work here (or `if (httpResponse?.data)` if there's no way this can be tricked by a falsy non-null value).

extensions/packageManager/src/node/feeds/nuget/nugetFeed.ts
```diff
@@ -0,0 +1,127 @@
...
+   * Build NuGet search url based on the NuGet search spec.
+   * Spec: https://docs.microsoft.com/en-us/nuget/api/search-query-service-resource#search-for-packages
+   * @param baseUrl The NuGet search service url.
+   * @param query The desired query parameters.
+   */
+  private buildNuGetSearchUrl(baseUrl: string, query?: IPackageQuery): string {
+    // If a query is provided, it takes precedence. Otherwise, default to the package source one.
+    const effectiveQuery = query ?? this.packageSource?.defaultQuery;
+    let url: string = `${baseUrl}?prerelease=${effectiveQuery?.prerelease ?? true}&semVerLevel=${
+      effectiveQuery?.semVerLevel ?? '2.0.0'
```
beyackle, line 110: Why is 2.0.0 the default here?

beyackle (changes requested):
Mostly looks good - just a few suggestions.

extensions/packageManager/src/node/index.ts
carlosscastro: This is the original code that just moved and I'm trying to reduce risk. So, I'll change it as part of this PR if I get unit tests in. Otherwise I'll make this change along with unit tests to be sure I exercise this path, since in the current UI this is always a single element array  in practical terms.

extensions/packageManager/src/node/feeds/nuget/nugetFeed.ts
carlosscastro: No, just one as per the spec. Thankfully they clarify that.

extensions/packageManager/src/node/feeds/nuget/nugetFeed.ts
carlosscastro: This is the default value we chose based on the docs. I can add a link to the spec here.

extensions/packageManager/src/node/feeds/nuget/nugetFeed.ts
carlosscastro: So here I want the opposite logical result (enter if something is null) but I get your point of doing a single conditional statement.

extensions/packageManager/src/node/feeds/nuget/localNuGetFeed.ts
```diff
@@ -0,0 +1,126 @@
...
+   */
+  private async crawlLocalFeed(url: string): Promise<IPackageDefinition[]> {
+    // The local feed is expected to be in folders using the structure defined here:
+    // https://docs.microsoft.com/en-us/nuget/hosting-packages/local-feeds
+    // the line below will:
+    // * get a list of all the files at the specified url
+    // * extract only folders from that list
+    // * pass each one through the getPackageInfo function, which extracts metadata from the package
+    // * return a feed in the form that is used by nuget search API
+    const packages = readdirSync(url, { withFileTypes: true }).filter((f) => f.isDirectory());
```
a-b-r-o-w-n, line 49: Why not use the async version here?

extensions/packageManager/src/node/feeds/nuget/localNuGetFeed.ts
```diff
@@ -0,0 +1,126 @@
...
+   * @param rootDir Root directory.
+   * @param packageName Package name to be retrieved.
+   */
+  private async getPackageInfo(rootDir: string, packageName: string): Promise<IPackageDefinition> {
+    // Available versions should be folders underneath this folder.
+    const packageDir = path.join(rootDir, packageName);
+
+    let versions;
+    try {
+      versions = readdirSync(packageDir, { withFileTypes: true })
```
a-b-r-o-w-n, line 69: Use the async version here as well.

extensions/packageManager/src/node/feeds/nuget/localNuGetFeed.ts
```diff
@@ -0,0 +1,126 @@
...
+    // the line below will:
+    // * get a list of all the files at the specified url
+    // * extract only folders from that list
+    // * pass each one through the getPackageInfo function, which extracts metadata from the package
+    // * return a feed in the form that is used by nuget search API
+    const packages = readdirSync(url, { withFileTypes: true }).filter((f) => f.isDirectory());
+    const feed = [];
+    for (const p of packages) {
+      feed.push(await this.getPackageInfo(url, p.name));
+    }
```
a-b-r-o-w-n, line 53: This looks like it should be pushed in to an array of promises and then return `Promise.all(...)`.

a-b-r-o-w-n (changes requested):
Overall this is a great refactor. Can you use the async file apis in the local nuget feed?

extensions/packageManager/src/node/feeds/nuget/localNuGetFeed.ts
carlosscastro: Thanks Andy! Great suggestions. I left this code unchanged to take it in stages, and just changed the structure for now. I'll try this out and depending on how much it changes things andthe risk equation, I'll either include in this iteration or in a separate pr

carlosscastro: Actually looks much better and seems low risk, so I'll test it and push these suggestions shortly.

carlosscastro:
Thanks for the review @a-b-r-o-w-n and @beyackle. Pushed code that addresses your feedback. Happy to keep iterating. Also note that this is not the end game, this is the first step towards the end game, mostly enabling some more scenarios for R12, setting up the foundation for unit testing and upcoming functionality in r13.

extensions/packageManager/src/node/index.ts
```diff
@@ -249,30 +132,30 @@ export default async (composer: IExtensionRegistration): Promise<void> => {
...
-            // cast this to the form of the http response from nuget
-            raw = { data: { data: rawlocal } };
-          } else {
-            raw = await axios.get(url);
-          }
-          const feed = await normalizeFeed(raw.data);
-          if (Array.isArray(feed)) {
-            combined = combined.concat(feed);
+          const feed: IFeed = await new FeedFactory(composer).build(packageSource);
+          const packageQuery: IPackageQuery = {
```
benbrown, line 141: this blocks the use of the default query, so it undoes the narrower default query.
this should only be used if a default query is not specified.

carlosscastro: Yep, changed to this logic. Added a todo since in reality we want the provided query to further filter with the package query as basis. However we don't need that right now.

carlosscastro: Pushing a fix shortly.

extensions/packageManager/src/node/index.ts
carlosscastro: Added this one after all, thanks for the suggestion!

extensions/packageManager/src/node/index.ts
benbrown: thanks!

benbrown (approved):

extensions/packageManager/src/node/feeds/nuget/localNuGetFeed.ts
```diff
@@ -0,0 +1,125 @@
...
+    // the line below will:
+    // * get a list of all the files at the specified url
+    // * extract only folders from that list
+    // * pass each one through the getPackageInfo function, which extracts metadata from the package
+    // * return a feed in the form that is used by nuget search API
+    const packages = await readdir(url, { withFileTypes: true });
+    const feedPromises: Promise<IPackageDefinition>[] = [];
+    for (const p of packages.filter((f) => f.isDirectory())) {
+      feedPromises.push(this.getPackageInfo(url, p.name));
+    }
```
beyackle, line 53: `const feedPromises = packages.filter((f) => f.isDirectory()).map((f) => this.packageInfo(url, f.name));` should work too.

beyackle (changes requested):
One minor thing - otherwise looks good.

beyackle (approved):

a-b-r-o-w-n (approved):

cwhitten (approved):

PR #6044, opened 2021-03-01
fix: provision process shown in publish list

alanlong9278:
### Bug
 provision process shown in publish list
MicrosoftTeams-image (4)

### Reason
The process poor is shared by provision & publish. So the provision process will be shown in publish page when get publish status without jobId

### Solution
Store jobId in client and add job id in get publish status api

## Task Item

#minors

### DIFF
--- a/Composer/packages/client/src/recoilModel/dispatchers/publisher.ts
+++ b/Composer/packages/client/src/recoilModel/dispatchers/publisher.ts
@@ -21,12 +21,12 @@ import { openInEmulator } from '../../utils/navigation';
 import { botEndpointsState } from '../atoms';
 import { rootBotProjectIdSelector, dialogsSelectorFamily } from '../selectors';
 import * as luUtil from '../../utils/luUtil';
+import { ClientStorage } from '../../utils/storage';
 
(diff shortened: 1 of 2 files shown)


Composer/packages/client/src/recoilModel/dispatchers/publisher.ts
```diff
@@ -99,6 +102,13 @@ export const publisherDispatcher = () => {
     if (data == null) return;
     const { set, snapshot } = callbackHelpers;
     const { endpointURL, status } = data;
+
+    // remove job id in publish storage if published
+    if (status === PUBLISH_SUCCESS || status === PUBLISH_FAILED) {
+      const publishJobIds = publishStorage.get('jobIds') || {};
+      publishJobIds[`${projectId}-${target.name}`] = '';
```
VanyLaw, line 109: maybe delete publishJobIds[`${projectId}-${target.name}`] is better?

alanlong9278: Yep, refactor it

PR #6061, opened 2021-03-01
fix: read default values from schema when building modal

### LINKED ISSUE #5906: Adapters: boolean defaults from schema not working in modal

beyackle:
## Description

The adapter settings modal wasn't reading the values for defaults from the schema; it was just initializing the values as empty if there was no existing value object to load.

## Task Item

closes #5906

## Screenshots

image

### DIFF
--- a/Composer/packages/client/src/pages/botProject/adapters/ExternalAdapterModal.tsx
+++ b/Composer/packages/client/src/pages/botProject/adapters/ExternalAdapterModal.tsx
@@ -14,6 +14,7 @@ import { DialogSetting } from '@bfc/shared';
 import { JSONSchema7 } from '@botframework-composer/types';
 import { EditorExtension, PluginConfig } from '@bfc/extension-client';
 import mapValues from 'lodash/mapValues';
+import { JSONSchema4Type } from 'json-schema';
 
 import { settingsState, dispatcherState } from '../../../recoilModel';
 import { useShell } from '../../../shell';
@@ -25,8 +26,6 @@ export type AdapterRecord = {
   enabled: boolean;
 };
 
-type ConfigValue = string | number | boolean;
-
 type Props = {
   adapterKey: string;
   packageName: string;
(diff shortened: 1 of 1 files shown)


beyackle:
Hang on, need to fix a type error.

Composer/packages/client/src/pages/botProject/adapters/ExternalAdapterModal.tsx
```diff
@@ -35,18 +35,24 @@ type Props = {
   projectId: string;
   schema: JSONSchema7;
   uiSchema: JSONSchema7 & { helpLink?: string };
-  value?: { [key: string]: ConfigValue };
+  value?: { [key: string]: JSONSchema4Type };
```
a-b-r-o-w-n, line 38: Why use `JSONSchema4Type` instead of the already imported `JSONSchema7` type?

Also, this could be replaced by `SchemaDefinitions` which is also exported by the types package.

Composer/packages/client/src/pages/botProject/adapters/ExternalAdapterModal.tsx
```diff
@@ -35,18 +35,24 @@ type Props = {
...
+  const { properties } = schema;
+
+  return mapValues(properties, 'default');
+}
+
 const AdapterModal = (props: Props) => {
   const { isOpen, onClose, schema, uiSchema, projectId, adapterKey, packageName, isFirstTime } = props;
 
-  const [value, setValue] = useState(props.value);
+  const [value, setValue] = useState(props.value == null ? makeDefault(schema) : props.value);
```
a-b-r-o-w-n, line 55: A couple of questions:
1. What if the value is `null` and that is a valid value?
2. Depending on the answer to the above -- could we use the nullish operator instead?

Composer/packages/client/src/pages/botProject/adapters/ExternalAdapterModal.tsx
beyackle: Good idea - this was to fix a typecheck error, so I pulled the first thing I found that looked like what it was complaining about.

a-b-r-o-w-n: Ok. Not a big deal if that doesn't correct the issue.

Composer/packages/client/src/pages/botProject/adapters/ExternalAdapterModal.tsx
beyackle: The only possible things `value` can be are `null` or an object with settings in it; if it's `null` then we're in the case where we tried to find settings in appSettings.json and they weren't there. That being said, it makes sense to use the coalesce operator here, so I'll do that.

carlosscastro (approved):

PR #6076, opened 2021-03-02
fix: Don't prompt for LUIS key if not needed

### LINKED ISSUE #6074: Orchestrator: Should not show warning for missing LUIS keys (Regression)
## Describe the bug
When we're only using Orchestrator as the LU Provider, Composer still asks for LUIS keys.  

## Version
1.3.1

## Expected behavior
When not using LUIS anywhere, Composer should not prompt for a LUIS key

taicchoumsft:
## Description
`botIndexer::shouldUseLuis` function is used to check if LUIS is needed. Inside, it checks against the `dialog.luProvider` property. If it is not present or is `Microsoft.LuisRecognizer`, `LUIS` is assumed to be used.  

But the `luProvider` property is not added to the dialog unless using the `dialogsWithLuProviderSelectorFamily` selector. This PR switches all needed selectors that make this LUIS check to use `dialogsWithLuProviderSelectorFamily` so this property is filled.  This allows Orchestrator to be selected, and if Luis is not used, no keys will be prompted for.

If  LUIS is used anywhere else, the key gets prompted again as expected.

## Task Item
closes #6074


srinaath (approved):

srinaath:
@taicchoumsft  do u any of these selectors need the update?

[BotFramework-Composer/localRuntimeBuilder.ts at main · microsoft/BotFramework-Composer (github.com)](https://github.com/microsoft/BotFramework-Composer/blob/main/Composer/packages/client/src/recoilModel/selectors/localRuntimeBuilder.ts)

taicchoumsft:
> @taicchoumsft  do u any of these selectors need the update? Basially, these are operations used by the start bot controller. They all now operate under the context that LUIS or no LUIS. They dont have idea about Orchestrator when building the bot.
> 
> [BotFramework-Composer/localRuntimeBuilder.ts at main · microsoft/BotFramework-Composer (github.com)](https://github.com/microsoft/BotFramework-Composer/blob/main/Composer/packages/client/src/recoilModel/selectors/localRuntimeBuilder.ts)

I looked through all the selectors, and none of the callers require the `LuProvider` property.  Also tested building locally just to be sure, the start bot controller works correctly with Orchestrator now, and no regressions so far with LUIS.

boydc2014:
@lei9444  can you help confirm this fix?

lei9444 (approved):

PR #6105, opened 2021-03-03
fix: hide BotController in home page

### LINKED ISSUE #6079: Bot start toolbar show up in home page
## Describe the bug

 In Home page,  bot start toolbar will show up when open a bot but should not be there.

## Browser

- [ ] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

## To Reproduce

1. Open a bot
2. Go to home page
3. Click `open` 
4. See `Bot start toolbar` show up

## Expected behavior
`Bot start toolbar`  should not show up

## Screenshots
preview

zhixzhan:
## Task Item
close #6079 

## Screenshots
Mar-03-2021 11-21-19

### DIFF
--- a/Composer/packages/client/src/components/Header.tsx
+++ b/Composer/packages/client/src/components/Header.tsx
@@ -29,8 +29,8 @@ import {
 } from '../recoilModel';
 import composerIcon from '../images/composerIcon.svg';
 import { AppUpdaterStatus } from '../constants';
-import { useLocation } from '../utils/hooks';
 import TelemetryClient from '../telemetry/TelemetryClient';
+import { useBotControllerBar } from '../hooks/useControllerBar';
 
 import { WebChatPanel } from './WebChat/WebChatPanel';
 import { languageListTemplates } from './MultiLanguage';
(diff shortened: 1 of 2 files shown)


Composer/packages/client/src/components/Header.tsx
```diff
@@ -167,9 +167,16 @@ export const Header = () => {
...
-    const hideCondition = !pathname.endsWith('/home') || pathname.includes('/bot/');
-    setStartBotsWidgetVisible(hideCondition);
+    /**
+     * hide it on the /home /projects/* page, show it in /bot/*
+     * home: /home
+     * create: /projects/create
+     * open: /projects/open
+     * save as: /projects/{projectId}/{botName}/save
+     */
+    const hideCondition =
```
hatpick, line 177: @srinaath @zhixzhan This solution of checking the current path seems to be flaky, can we decide to show/hide the control based on our app state instead? for example, if root project id has a value or something of that sort?

srinaath: I agree. When i first wrote it i had one pathname to check. But, currently there is no appstate that signifies this use case. RootBotId is set even in homepage

zhixzhan: I made some updates, created a new `activePageMode` state in global.

boydc2014: either directly judging on url patterns, or creating a hook to encapsulate that logic

hatpick (approved):

srinaath:
Nice @zhixzhan. This fix is perfect.

PR #6219, opened 2021-03-04
fix: allow multiple feeds with same url to co-exist

### LINKED ISSUE #6071: Package manager - editing behavior in manage feeds modal
## Describe the bug

When editing a feed in the manage feeds modal, the interaction can be confusing for users who don't realize they need to save changes by clicking the checkmark or hitting enter. In fact, the version of this component used in the QnAMaker section of Composer doesn't require a save action. We should create consistency and eliminate this confusion by using that component here.

## Expected behavior

When a user edits the feed name or URL, their changes should save automatically without having to click a checkmark or hit enter. The component we use here should be consistent with QnA section and throughout the app.

## Screenshots

Current solution requires the user to hit the checkmark to save changes:
image

In the QnAMaker section, the user doesn't have to take a save action:
image

## Additional context

Please let me see and interact with the new implementation before committing to ensure desired behavior.

### LINKED ISSUE #6072: Package manager - add a package modal design updates + errors
## Describe the bug

Design for Fluent consistency and the addition of error states in the "Install" or "Add a package" modal in Package manager.

add a package modal and errors

### LINKED ISSUE #6205: bug: package manager feed loading can't handle 2 feeds with same url
## Describe the bug

When package manager loads a feed, it passes in the feed url from the client.
However, this means if multiple feeds are set up with the same feed url, however different query preferences, they will conflict and only the last config will be used.

## Version

nightly 

## To Reproduce

1. create a new feed with the url `https://api.nuget.org/v3/index.json`
2. notice that the `nuget` feed now includes non-microsoft packages
 
## Expected behavior

Feed-specific settings should work. I should be able to have 2 feeds with the same url.

I suspect the fix here is to include the feed name as part of the feed request API.

### LINKED ISSUE #6226: nonstandard button for saving new feed is super confusing
## Describe the bug

I tried to add a new feed.  The Green check box doesn't communicate that you need to push it to commit the change. 

I couldn't figure out why it wasn't saving my change.  It said "you haven't saved it" and I STILL didn't realize the green check mark was to save the change. 

We should use an industry standard icon for saving.

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Go to Packagemanager add feeds
2. add a feed "c:\nuget"  This green check box totally failed to communicate what it was for.  I thought it was telling me that I had typed in a valid path.  It didn't look like a button and it didn't communicate that it was required to save. 
image

## Expected behavior
I think a button with "SAVE" or standard SAVE icon should be used.
image

## Screenshots
Once I save we use industry standard icons for edit and delete, and it's completely clear.
image

Also, look at the usage of green checkmark on this page, where it's not a button for saving at all.  
image

benbrown:
## Description

Instead of passing the literal URL, pass the key to the feed definition.  This allows multiple feeds to share a url, but have different query parameters.

* Also: remove JS feed for now
* Also: add community feed to display all bf-component packages

## Task Item

Fixes #6205 
Fixes #6072 
Fixes #6071 
closes #6226 

## Screenshots

BIMvYeGiJq
Screen Shot 2021-03-04 at 4 37 37 PM


benbrown:
@clearab -- do you want to change the labels of these default feeds?

benbrown:
new edit feed behavior

extensions/packageManager/src/node/index.ts
```diff
@@ -63,13 +63,14 @@ export default async (composer: IExtensionRegistration): Promise<void> => {
...
-            readonly: true,
-          },
+          // TODO: Re-enable the NPM feed when we have a JS runtime
+          // {
+          //   key: 'npm',
+          //   text: 'npm',
+          //   url: 'https://registry.npmjs.org/-/v1/search?text=keywords:bf-component&size=100&from=0',
+          //   searchUrl: 'https://registry.npmjs.org/-/v1/search?text={{keyword}}+keywords:bf-component&size=100&from=0',
+          //   readonly: true,
+          // },
```
a-b-r-o-w-n, line 73: On my machine, the npm feed is still present. We may need to actually remove it from the store for now.

a-b-r-o-w-n (changes requested):

extensions/packageManager/src/node/index.ts
benbrown: This will only happen to people who used the pre-release version.

cwhitten:
@benbrown there's a conflict here

a-b-r-o-w-n (approved):

benbrown:
> @benbrown there's a conflict here

resolved!

cwhitten (approved):

PR #6235, opened 2021-03-05
fix: reset errors when switching between bots

### LINKED ISSUE #5785: The runtime settings section of the settings page is incorrectly tied to the root bot
## Describe the bug

When editing or viewing the runtime settings, only the root bot's settings are ever displayed, regardless of what bot is selected on the sidebar.

## Version

Current main

## To Reproduce

1. open a multibot project
2. navigate to settings
3. switch to the child bot
4. change runtime settings
5. change back to root bot, note that the settings were actually changed there and not in child bot.

## Expected behavior

i should be able to edit the settings for each bot.

## Screenshots

kXJFkfYNLt

## Additional context

this was introduced recently.

liweitian:
## Description
Need to recheck errors while switching between bots. 
## Task Item
Closes #5785 

## Screenshots
dddd


Composer/packages/client/src/pages/botProject/runtime-settings/RuntimeSettings.tsx
```diff
@@ -203,6 +212,16 @@ export const RuntimeSettings: React.FC<RouteComponentProps<{ projectId: string }
     );
   };
 
+  const errorElement = (errorText: string) => {
+    if (!errorText) return '';
+    return (
+      <div css={errorContainer}>
```
hatpick, line 218: I think using div here might violate DOM rules, since Fabric uses p tag for error messages. Please check dev tools for potential errors and if there are any, use span here. (or use Fabric Text component)

liweitian: it should be good now.

boydc2014 (approved):

PR #6236, opened 2021-03-05
feat: update dotnet sdk version from 4.12.0-rc1 -> 4.12.0

yeze322:
## Description
#minor
No functional changes.


Composer/packages/server/src/locales/en-US.json
```diff
@@ -1553,8 +1553,8 @@
   "get_started_with_bot_framework_composer_57a6d38b": {
     "message": "Get started with Bot Framework Composer"
   },
-  "get_started_with_your_new_bot_da9aaf6a": {
-    "message": "Get started with your new bot"
+  "get_your_bot_up_and_running_418dd8ea": {
```
hatpick, line 1556: do you want this change to be a part of this PR?

yeze322: Yes, it's an intended change, this string was modified in some other PR but locale json was not updated

hatpick (approved):

cwhitten (approved):

PR #6238, opened 2021-03-05
fix: import lg/lu not find all up view

### LINKED ISSUE #6188: Bot response page's imported LG files are not clickable
## Describe the bug

Imported LG file nodes are not clickable in bot response page

## Version
latest main

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Create an ToDoWithLuisSample bot
2. Go to bot response page
3. Expand on the DeleteItem view
4. Click the import link below

## Expected behavior

The "import" link should show the imported file 

## Screenshots

image

### LINKED ISSUE #6189: LU import resolver can't handler import statement without specific locale
## Describe the bug

if have a common.en-us.lu been reported by a.en-us.lu with the statement below

`[import] (common.en-us.lu)` works but
`[import](common)` doesn't work
`[import](common.lu)` doesn't work

## Version

latest main

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Create an ToDoWithLuisSample
2. Go to LanguageUnderstanding/en-us folder
3. Create an common.en-us.lu file with few intents and add `[import](common.lu)` into maindialog's lu file
4. Go to LU all up view, the import link is not clickable. 

## Expected behavior

It works no matter what kind of importing format.

## Screenshots

image
image

zhixzhan:
## Description

- locate import file with correct id
- add intent whitespace on imported project tree item

## Task Item
close #6188 
close #6189 

## Screenshots
Mar-05-2021 16-12-06


Composer/packages/client/src/recoilModel/selectors/dialogImports.ts
```diff
@@ -36,10 +36,11 @@ export const getLanguageFileImports = <T extends LgFile | LuFile | QnAFile>(
       continue;
     }
     const currentImports = file.imports.map((item) => {
+      const importedFile = getFile(getBaseName(item.id));
       return {
         displayName: item.description,
         importPath: item.path,
-        id: getBaseName(item.id),
+        id: importedFile ? importedFile.id : '',
```
hatpick, line 43: ```importedFile?.id ?? ''```

hatpick:
@zhixzhan can you change the icon so all of them use the same icon?

Composer/packages/client/src/recoilModel/selectors/dialogImports.ts
zhixzhan: the `id` needs to be string type, otherwise ESlint would complained of it

zhixzhan:
> @zhixzhan a couple of notes:
> 
> 1. can you change the icon so all of them use the same icon?

Agree, will update it.

> 2. I think we just want to show a flat list of LG files here, for example if file1.lg is imported in multiple of other l files, we dont want to see it many times, but one time, can we get UX feedback on this?

So if several dialog.lg has referred a 'hello.lg'

in dialogA.lg
```
[import1](hello.lg)
```

in dialogB.lg
```
[myHello](hello.lg)
```

currently will display
```
- dialogA
     -import1
- dialogB
     -myHello
```

if we flatten it, it would be
```
- dialogA
- dialogB
- {name}
```
here need to choose a {name} to display. 

Yes we do need some feedback from designer.

zhixzhan:
Updated the icon and no indent on 'import'

image

boydc2014:
@hatpick Agreed about the the ux design here, let's start a thread with designer, for now, i think it's good for us to take this as a fix, what do you think?

boydc2014 (approved):

hatpick:
@boydc2014 I dont think we should show the alias for the import, but we should show the actual imported file name, similar to other languages.

For example, imagine there is a constant.ts file in your typescript project that a bunch of other files import, such as:

file1.ts content:
```import * as constants1 from './constant.ts'```
file2.ts content
```import * as constant2 from './constant.ts'```

As you can see in the example above same content from constant.ts has been used with two different aliases, but is that the good UX to show two different items as constant1 and constant2 in the tree view?

**Comment by srinaath** [comment]
Looping @mewa1024  here.

**Comment by mewa1024** [comment]
Yes, I think agree with @hatpick above, but I want to check that I understand the suggestion. Back to the original example, 
image

this would appear as

```
bot name
- dialogA
- dialogB
- hello
```
yes?

hatpick:
Correct, that's my suggestion.

mewa1024:
@sangwoohaan fyi, this is the discussion about imported LG in the all up view. I believe the same behavior is modeled for LU.

PR #6240, opened 2021-03-05
fix: move orchestrator build to worker

lei9444:
## Description
**issue**
The orchestrator build will take a long time if the lu file is big. If we use the electron, the UI thread will be blocked by orchestrator build. 

**fix**
Move it to the child thread.

## Task Item
#minor


Composer/packages/server/src/models/bot/process/orchestratorWorker.ts
```diff
@@ -0,0 +1,86 @@
...
+    isDialog,
+    '',
+    null,
+    fullEmbedding
+  );
+}
+
+export async function writeSnapshot(output: IOrchestratorBuildOutput, generatedFolderPath: string) {
+  // write snapshot data into /generated folder
+  const snapshots: { [key: string]: string } = {};
```
hatpick, line 53: nit: Record<string, string>

hatpick (approved):

taicchoumsft (dismissed):
Thanks Leilei, awesome stuff.

taicchoumsft (approved):

boydc2014 (approved):

PR #6250, opened 2021-03-05
fix: Fix issue in path that caused multibot templates to be misinterpreted as single bots

benbrown:
## Description

when using a new template, there may be > 1 bot created.

the code will crawl the folder looking for multiple bots.

this change ensures that the template generator returns the parent folder which includes all created bot projects.

also included, expose errors in creation more directly.

#minor


srinaath (dismissed):

cwhitten:
@benbrown there's a conflict here

benbrown:
@cwhitten will resolve.

cwhitten (approved):

PR #6285, opened 2021-03-06
fix: ordered template list

### LINKED ISSUE #6026: Create modal - Template naming and list order
## Describe the bug

The list of templates in the create modal should be arranged intentionally, potentially from simple to more complex for R12 release, and the names themselves should be revised. @Dewain27 @emivers8 @benbrown and @mareekuh to agree on naming. Initial recommendations below, but final TBD.

Initial recommendations:
1. Blank bot (empty bot)
2. Basic conversational bot (conversational core)
3. QnAMaker bot (generator-qna-bot)

(Is multibot or calendar part of release for R12?)

## Screenshots

image

### LINKED ISSUE #6126: Cleanup feature flag
The feature flag needs to be reworked to 

Name:
The Component Model

Description:
Enable the Component Model including the new creation experience, the Adaptive Runtime, and Package Manager. [Learn More](https://github.com/microsoft/botframework-components).

### LINKED ISSUE #6174: New Creation Experience preview feature is missing a "Learn More" link
## Describe the bug
Every preview feature in composer must follow a format that is a "title" a "description" and a "Learn More" link.  the Learn More link navigates to an external link that describes what the preview feature does and why you might want to enable it and try it out.

## Version
1.3.x the R12 bug bash build

## Browser
- [X] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS
- [X] macOS
- [ ] Windows
- [ ] Ubuntu

## To Reproduce

1. Launch Composer
2. Click on the Composer Settings icon in the lefthand nav to launch the Application Settings page
3. Scroll down to the Preview features section
4. Note that the "New Creation Experience" feature has a title and a description, but is missing the Learn More link

## Expected behavior
working or at least a stubbed Learn More link

pavolum:
## Description

This PR implements a few small bug fixes. 
- Adds sorting logic on the server side after templates are returned from the npm registry feed
  - Sorting logic changes the order of the templates returned to the client for display in the UI using a hardcoded sorting list for reference
  - Populates the display name for templates defined in the sorting list
- Removes redundant and unneeded BotTemplateV2 type
- Changes text for NEW_CREATION_FLOW feature flag
- Adds documentation link for NEW_CREATION_FLOW feature flag

Found an additional but with the fix in the last commit in this PR. When an existing feature flags content was changed (display name, description or documentation link), it was not picked up by the users instance of feature flag. Made the change to always take the default content for these fields. 

## Task Item

fixes #6026 
fixes #6126 
fixes #6174 

## Screenshots

image

image


Composer/packages/lib/shared/src/featureFlagUtils/index.ts
```diff
@@ -6,10 +6,11 @@ import { FeatureFlagMap } from '@botframework-composer/types';
 
 export const getDefaultFeatureFlags = (): FeatureFlagMap => ({
   NEW_CREATION_FLOW: {
-    displayName: formatMessage('New Creation Experience'),
-    description: formatMessage('Preview the new adaptive runtime and component system.'),
+    displayName: formatMessage('The Component Model'),
+    description: formatMessage('Enable the Component Model including the new creation experience, the Adaptive Runtime, and Package Manager.'),
     isHidden: false,
     enabled: false,
+    documentationLink: 'https://aka.ms/AAbf42g',
```
a-b-r-o-w-n, line 13: Why the obfuscated aka link?

Composer/packages/server/src/constants.ts
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+import { QnABotTemplateId } from "@bfc/shared";
```
a-b-r-o-w-n, line 4: nit: this should be single quoted, and should have been formatted pre-commit by prettier. Do you know why it's not?

Composer/packages/server/src/controllers/asset.ts
```diff
@@ -2,12 +2,14 @@
...
 import { Request, Response } from 'express';
-import { BotTemplateV2, QnABotTemplateId } from '@bfc/shared';
+import { BotTemplate, QnABotTemplateId } from '@bfc/shared';
 import formatMessage from 'format-message';
 
 import AssetService from '../services/asset';
 import { getNpmTemplates } from '../utility/npm';
 import log from '../logger';
+import { templateSortOrder } from '../constants';
+import _ from 'lodash';
```
a-b-r-o-w-n, line 12: Import just the funcs you need from lodash. Not really a problem on the server, but its best to remain consistent with client & server.

```ts
import union from 'lodash/union';
```

Composer/packages/server/src/services/featureFlags.ts
```diff
@@ -28,6 +28,9 @@ export class FeatureFlagService {
       FeatureFlagService.currentFeatureFlagMap[key] = {
         ...FeatureFlagService.defaultFeatureFlags[key],
         ...FeatureFlagService.currentFeatureFlagMap[key],
+        displayName: FeatureFlagService.defaultFeatureFlags[key].displayName,
+        description: FeatureFlagService.defaultFeatureFlags[key].description,
+        documentationLink: FeatureFlagService.defaultFeatureFlags[key].documentationLink,
```
a-b-r-o-w-n, line 33: Is it true that the only value we want to persist is the enabled state? I think we can refactor this code to reflect that.

```ts
FeatureFlagService.currentFeatureFlagMap[key] = {
  ...FeatureFlagService.defaultFeatureFlags[key],
  enabled: FeatureFlagService.currentFeatureFlagMap[key].enabled ?? FeatureFlagService.defaultFeatureFlags[key].enabled
}
```

Composer/packages/server/src/constants.ts
```diff
@@ -20,3 +22,5 @@ export const APPINSIGHTS_INSTRUMENTATIONKEY = process.env.APPINSIGHTS_INSTRUMENT
 export const piiProperties = ['projectId', 'rootProjectId'];
 
 export const COMPOSER_VERSION = '1.3.1';
+
+export const templateSortOrder = [{ generatorName: '@microsoft/generator-microsoft-bot-empty', displayName: 'Blank bot' }, { generatorName: '@microsoft/generator-microsoft-bot-conversational-core', displayName: 'Basic conversational bot' }, { generatorName: QnABotTemplateId, displayName: 'QnAMaker bot' }]
```
a-b-r-o-w-n, line 26: I think this (and the sorting func) could be extracted into a util and tested. It's not really a shared constant.

Composer/packages/server/src/controllers/asset.ts
```diff
@@ -48,8 +50,24 @@ export async function getProjTemplatesV2(req: any, res: any) {
...
 
+    let sortedTemplateList: BotTemplate[] = [];
+
+    // Sort incoming template array and reassign display name based on sort list
+    templateSortOrder.forEach((tempSortEntry, index) => {
+      const templateIndex = templates.findIndex((template) => {
+        return template.id === tempSortEntry.generatorName;
+      })
+      if (templateIndex != -1){
+      templates[templateIndex].name = tempSortEntry.displayName;
```
a-b-r-o-w-n, line 61: Can this be read from the BotTemplate somehow? Could it be added to the package manifest?

a-b-r-o-w-n (changes requested):

Composer/packages/lib/shared/src/featureFlagUtils/index.ts
pavolum: I feel like that is best practice, In case we want to change the location of the documentation post release we can do it through the aka portal. Not sure how likely that is but better have the ability to do then not.

a-b-r-o-w-n: Most (if not all) of our other aka links use a friendly name (ex. https://aka.ms/bfcomposer).

Composer/packages/server/src/controllers/asset.ts
pavolum: Handled

Composer/packages/server/src/constants.ts
pavolum: Formatter fixed, thanks for the help!

Composer/packages/server/src/services/featureFlags.ts
pavolum: You could do the above if you always saved to storage each time updateFeatureFlags is called which I believe is on app start. Even if the feature flags have not changed. The current code is a bit more robust but checks if something has in fact changed and only saves if it has. So we would lose that feature with the suggested refactor but would result in slightly less and more simplified code.

Composer/packages/server/src/controllers/asset.ts
pavolum: Are you suggesting having an index in botTemplate and the sort function just populates that index instead of actually sorting what is returned to the client? That would be possible, I initially implemented it that way but it required a bit of additional logic to pass that index order from the BotTemplate obj to the index prop in the detailsList where this is later populated. I figured sorting led to less front end changes.

pavolum: As for the package manifest. We could add sort data, down the line, to the meta data for each package that is provided from the npm registry. Not sure if that is the best approach, it does open up changing sorting outside of Composer release schedule but does add a dependency from Composers visual layer to some meta data in individual npm packages that are external to Composer

a-b-r-o-w-n: The sort data is fine to live in Composer, what I'm suggesting is 
1. Move the `templateSortOrder` const and this sorting function into another module and add a unit test.
2. I was just asking if we can put the display name in the package manifest.

Composer/packages/server/src/constants.ts
pavolum: Handled

Composer/packages/server/src/controllers/asset.ts
pavolum: Ah I see, yeah I will create a task for that. I have handled (1)

Composer/packages/lib/shared/src/featureFlagUtils/index.ts
pavolum: Ah I see, so use aka but a friendly name, gotcha, I have made the change and pushing latest iteration now

Issue #6220, opened 2021-03-04
Orchestrator: Wrong model settings created when publishing to Azure


taicchoumsft:
## Describe the bug
Orchestrator settings, and model copy logic need to be fixed for Azure. When we added multilanguage support to Orchestrator, the logic for publishing was not updated.  

## Version
Nightlies 217487+

## Browser
All
## OS
All

## To Reproduce
1. Create a bot with Orchestrator as a recognizer (in any dialog)
2. Try to publish the bot to Azure.
3. Observe `orchestrator.settings.json` file is set wrongly
4. There is another publishing bug ([Components #675](https://github.com/microsoft/botframework-components/issues/675)) that causes the model to not be uploaded.

## Expected behavior
Model should be uploaded, and settings should be correct, for both english, and multilanguage based bots.

PR #6307, opened 2021-03-08
fix: Orchestrator publishing settings were wrong (legacy runtime)

### LINKED ISSUE #6305: Publishing (Legacy Runtime): Publishing to Azure likely broken for legacy runtime if using LUIS/Orchestrator/QNA
## Describe the bug
Change introduced [here](https://github.com/microsoft/BotFramework-Composer/commit/b64b438edd22cdcc837722517994692f948858c7) definitely breaks Orchestrator publishing on the legacy runtime, and likely also breaks Luis/QNA.

The change is in `LuisAndQna.ts` - `ComposerDialogs`  was removed from path.

When publishing using the legacy runtime, the generated folder is a child of `ComposerDialogs` (i.e. `generated` folder is in `<BotFolder>/ComposerDialogs/generated`.  The C# runtime then picks up the settings from the generated folder, which would be `<BotFolder/ComposerDialogs/generated/orchestrator.settings.json`, `<BotFolder/ComposerDialogs/generated/luis.westus.settings.json`  etc.

With this change, the `generated` folder is now in `<BotFolder>/generated`, but the bot root is still `<BotFolder>/ComposerDialogs`.   So any artifact in the generated folder will not be found by the legacy runtime. 

## Version
Post 1.3.1 nightlies

## Browser
All

## OS
All 

## To Reproduce
1.  Create a bot using the legacy runtime
2.  Enable LUIS and/or Orchestrator
3. Eject the runtime
4.  Publish the bot to Azure using the Composer Publishing flow
 
## Expected behavior
Luis/Orchestrator/Qna should work normally.  With Orchestrator, you should not see the `"ModelFolder" cannot be null` message

taicchoumsft:
## Description
Orchestrator settings were published wrongly due to multiple reasons: 
1) Publishing location for the `generated` folder was changed when the new runtime was added (#6305)
2) Orchestrator multilanguage support for publishing to Azure was missing (#6220)

This PR moves the `generated` folder from `<BotFolder>/generated` back to `<BotFolder>/ComposerDialogs/generated` for the old runtime.  Awaiting PR comments to see if this will adversely affect the new runtime.

Also copies the Orchestrator en/multilang models as necessary and updates `orchestrator.settings.json` accordingly with the new relative filepaths when publishing. 

## Task Item
closes #6305,  #6220


extensions/azurePublish/src/node/luisAndQnA.ts
```diff
@@ -10,7 +10,7 @@ import { ILuisConfig, FileInfo, IBotProject } from '@botframework-composer/types
 import { AzurePublishErrors } from './utils/errorHandler';
 import { BotProjectDeployLoggerType } from './types';
 
-const botPath = (projPath: string) => path.join(projPath); // there should not be a need to specify the sub directory
+const botPath = (projPath: string) => path.join(projPath, 'ComposerDialogs'); // there should not be a need to specify the sub directory
```
benbrown, line 13: yes, this will break the new runtime. 

we will need to figure out a way to do this conditionally.

benbrown (changes requested):

extensions/azurePublish/src/node/luisAndQnA.ts
cwhitten: why not add a guard on the new creation feature flag here?

benbrown: that would work.

Composer/packages/server/src/models/bot/builder.ts
```diff
@@ -259,22 +260,28 @@ export class Builder {
...
-    await this.updateOrchestratorSetting();
+    await this.updateOrchestratorSetting(featureFlags);
   }
 
   /**
    * Orchestrator: Update Orchestrator Settings for publishing
    *
-   * Orchestrator file paths are updated relative to the project root.
+   * Models are located in <project root>/model
+   * In the Adaptive Runtime, Orchestrator snapshot files are located in <project root>/generated.
```
boydc2014, line 281: Is this the correct assumption @benbrown

extensions/azurePublish/src/node/luisAndQnA.ts
taicchoumsft: After discussing with @benbrown this morning, we will use the runtime key to decide between using the old vs new `generated` path.  

This is because there's an edge case using feature flags - a user could potentially be publishing a bot with old runtime even with new creation flag on (or vice versa).

Composer/packages/server/src/models/bot/builder.ts
benbrown: Yes

benbrown (dismissed):

PR #6309, opened 2021-03-08
fix: save extension state

### LINKED ISSUE #6086: Back button on Configure Resources does not keep entered information
## Describe the bug

When creating a publishing profile, when on the Configure Resource page of the publishing profile and you hit the back button it does not retain the information you entered on the prior screen.

## Version

image

## Browser

- [X] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [X] macOS
- [X] Windows
- [ ] Ubuntu

## To Reproduce

1. Go to create a publishing profile
2. Add Name and Publishing target
3. Click Next: Configure resources
4. Click Back
5. All data is gone that was entered before

## Expected behavior

Information provided should have persisted to allow me to change without re-entering all data.

## Screenshots

image

VanyLaw:
## Description

save extension state

## Task Item

close #6086


a-b-r-o-w-n:
Why can't the extension save its own state? I'm concerned that by exposing extension state to the client, the extensions become un-encapsulated. Why not at that point just move the publish extension into the client itself? What is the point of the extension anymore?

extensions/azurePublish/src/components/azureProvisionDialog.tsx
```diff
@@ -814,7 +818,24 @@ export const AzureProvisionDialog: React.FC = () => {
             />
           ) : null}
           <div>
-            <DefaultButton style={{ margin: '0 4px' }} text={'Back'} onClick={onBack} />
+            <DefaultButton
+              style={{ margin: '0 4px' }}
+              text={'Back'}
```
hatpick, line 823: formatMessage

extensions/azurePublish/src/components/azureProvisionDialog.tsx
```diff
@@ -624,7 +628,7 @@ export const AzureProvisionDialog: React.FC = () => {
 
   const PageFormConfig = (
     <Fragment>
-      <ChoiceGroup defaultSelectedKey="create" options={choiceOptions} style={{}} onChange={updateChoice} />
+      <ChoiceGroup defaultSelectedKey={choice.key} options={choiceOptions} style={{}} onChange={updateChoice} />
```
hatpick, line 631: remove style if empty

Composer/packages/client/src/pages/botProject/create-publish-profile/PublishProfileDialog.tsx
```diff
@@ -31,13 +31,30 @@ type PublishProfileDialogProps = {
...
+type ExtensionState = AzureExtensionState; // can expand to pva like | PVAExtensionState
+
+type AzureExtensionState = {
+  subscriptionId: string;
+  resourceGroup: string;
+  hostName: string;
+  location: string;
+  luisLocation: string;
+  enableResources: any;
+  requireResources: any;
```
hatpick, line 43: nit: rename to the following:
requiredResources
enabledResources

use types instead of any.

Composer/packages/client/src/pages/botProject/create-publish-profile/ProfileFormDialog.tsx
```diff
@@ -17,21 +16,18 @@ import { TooltipHost } from 'office-ui-fabric-react/lib/Tooltip';
...
   types: PublishType[];
   onNext: () => void;
-  updateSettings: (name: string, type: string, configuration: string) => Promise<void>;
-  projectId: string;
   setType: (value) => void;
+  name: string;
+  targetType: string;
+  setName: (value: string) => void;
+  setTargetType: (value: string) => void;
   current?: { index: number; item: PublishTarget } | null;
```
hatpick, line 31: I know this is not a part of this PR but would you please rename ```current``` to ```currentTarget```

Composer/packages/client/src/pages/botProject/create-publish-profile/PublishProfileDialog.tsx
```diff
@@ -31,13 +31,30 @@ type PublishProfileDialogProps = {
...
+  choice: any;
```
hatpick, line 44: avoid any

Composer/packages/client/src/plugins/api.ts
```diff
@@ -35,6 +35,8 @@ interface PublishAPI {
   savePublishConfig?: (config: PublishConfig) => void;
   getTokenFromCache?: () => { accessToken: string; graphToken: string };
   isGetTokenFromUser?: () => boolean;
+  getExtensionState?: () => any;
```
hatpick, line 38:
```
getExtensionState?: <T>() => T;
setExtensionState?: <T>(state: T) => void;
```

extensions/azurePublish/src/components/azureProvisionDialog.tsx
```diff
@@ -306,10 +305,13 @@ export const AzureProvisionDialog: React.FC = () => {
     getType,
     getTokenFromCache,
     isGetTokenFromUser,
+    setExtensionState,
+    getExtensionState,
   } = usePublishApi();
   // set type of publish - azurePublish or azureFunctionsPublish
   const publishType = getType();
   const currentConfig = removePlaceholder(publishConfig);
+  const extensionState = getExtensionState();
```
hatpick, line 314: you can do:
```const extensionState = getExtensionState() || defaultExtensionState;```
where your default extension state represents all the values you have used down below when the state doesn't exists, then you can avoid all the ```|| something``` expressions down below

**Comment by VanyLaw** [comment]
> Why can't the extension save its own state? I'm concerned that by exposing extension state to the client, the extensions become un-encapsulated. Why not at that point just move the publish extension into the client itself? What is the point of the extension anymore?

@a-b-r-o-w-n Your are right, but now the profile adding page[like below] is in client and share with other type of profile. I'm not sure how to keep the form state in client and extension during we moving forward and backward is better? Could you give me some advises? Thanks.

page in client:
image

page in extension:
image

File: extensions/azurePublish/src/components/azureProvisionDialog.tsx (hunk 7 above)
Comment (VanyLaw) [reply extensions/azurePublish/src/components/azureProvisionDialog.tsx:314]: good idea, Thanks.

**Comment by VanyLaw** [comment]
Hi @a-b-r-o-w-n, I use localStorage to save the extension state now. could your help to review the PR? Thanks.

File: Composer/packages/client/src/pages/botProject/create-publish-profile/PublishProfileDialog.tsx [hunk 8]
```diff
@@ -73,7 +76,9 @@ export const PublishProfileDialog: React.FC<PublishProfileDialogProps> = (props)
 
   // setup plugin APIs
   useEffect(() => {
-    PluginAPI.publish.closeDialog = closeDialog;
+    PluginAPI.publish.closeDialog = () => {
+      closeDialog();
+    };
```
a-b-r-o-w-n, line 81: Any reason for this change?

Composer/packages/client/src/pages/botProject/create-publish-profile/PublishProfileDialog.tsx
```diff
@@ -176,7 +179,11 @@ export const PublishProfileDialog: React.FC<PublishProfileDialogProps> = (props)
         modalProps={{
           isBlocking: true,
         }}
-        onDismiss={closeDialog}
+        onDismiss={() => {
+          closeDialog();
+          // remove extension state when parent component destroy
+          window.localStorage.removeItem(`${targetType}:state`);
```
a-b-r-o-w-n, line 185: This should be up to the extension when to clear its local storage.

a-b-r-o-w-n (changes requested):
This isn't what we discussed. What we discussed is adding an api in the extension client package called `useLocalStorage` that returns scoped apis for extensions (and more specifically, bundles) to access local storage. 

The extension client will manage scoping those apis correctly. In order to do so we discussed adding a `__bundleId` to window in order to create a local storage item with the following shape: 

```js
{ "composer:extensions": { [__extensionId]: { [__bundleId]: {} } }
```

extensions/azurePublish/src/components/util.ts
```diff
@@ -60,3 +60,52 @@ export const getExistResources = (config) => {
...
+  }
+};
+
+export const setExtensionState = (value) => {
+  window.localStorage.setItem(`${window.Composer?.__extensionId}:state`, JSON.stringify(value));
+};
+
+export const clearExtensionState = () => {
+  window.localStorage.removeItem(`${window.Composer?.__extensionId}:state`);
+};
```
tonyanziano, line 111: This should be at the Composer level and the extension internals (`__extensionId`) should be hidden from the extension. This way all extensions can call these APIs and we don't have to implement them for each individual extension.

tonyanziano (changes requested):

cwhitten:
Removing from 1.4

VanyLaw:
Hi @a-b-r-o-w-n, I resolved all the comments, could you take a look please?

Composer/packages/extension-client/src/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,82 @@
...
+    () => (key: string) => {
+      const allStates = window.localStorage.getItem(KEY);
+      if (allStates) {
+        const extensionState = JSON.parse(allStates)?.[__extensionId];
+        return extensionState?.[__bundleId]?.[key];
+      }
+      return undefined;
+    },
+    [__extensionId, __bundleId]
+  );
```
a-b-r-o-w-n, line 21: The proper hook to use here is `useCallback`

Composer/packages/extension-client/src/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,82 @@
...
+        return JSON.parse(allStatesString)?.[__extensionId]?.[__bundleId];
+      }
+    },
+    [__extensionId, __bundleId]
+  );
+
+  const setItem = useMemo(
+    () => (key: string, value: any) => {
+      const allStatesString = window.localStorage.getItem(KEY);
+      const allStates = allStatesString ? JSON.parse(allStatesString) : {};
```
a-b-r-o-w-n, line 36: Just call `getAll()`

Composer/packages/extension-client/src/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,82 @@
...
+
+      const extensionState = allStates?.[__extensionId] || {};
+      const bundleState = extensionState?.[__bundleId] || {};
```
a-b-r-o-w-n, line 39: These should also be extracted into a helper method or const since it is used frequently.

Composer/packages/extension-client/src/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,82 @@
...
+    () => (value: any) => {
+      const allStatesString = window.localStorage.getItem(KEY);
+      const allStates = allStatesString ? JSON.parse(allStatesString) : {};
+
+      const extensionState = allStates?.[__extensionId] || {};
+      // eslint-disable-next-line no-underscore-dangle
+      extensionState[__bundleId] = value;
+      // eslint-disable-next-line no-underscore-dangle
+      allStates[__extensionId] = extensionState;
+      window.localStorage.setItem(KEY, JSON.stringify(allStates));
```
a-b-r-o-w-n, line 61: This can be extracted into a helper method since it is duplicated above.

Composer/packages/extension-client/src/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,82 @@
...
+    [__extensionId, __bundleId]
+  );
+
+  const clearAll = useMemo(
+    () => () => {
+      const allStatesString = window.localStorage.getItem(KEY);
+      if (allStatesString) {
+        const allStates = allStatesString ? JSON.parse(allStatesString) : {};
+        const extensionState = allStates?.[__extensionId] || {};
+        // eslint-disable-next-line no-underscore-dangle
```
a-b-r-o-w-n, line 72: Just go ahead and disable this for the whole file.

extensions/azurePublish/src/components/azureProvisionDialog.tsx
```diff
@@ -546,14 +553,16 @@ export const AzureProvisionDialog: React.FC = () => {
...
       startProvision(options);
+      clearAll();
       closeDialog();
     },
     []
   );
 
   const onSave = useMemo(
-    () => () => {
+    () => async () => {
```
a-b-r-o-w-n, line 563: This method doesn't look async to me.

extensions/azurePublish/src/components/azureProvisionDialog.tsx
```diff
@@ -31,19 +31,19 @@ import {
 } from 'office-ui-fabric-react';
 import { JsonEditor } from '@bfc/code-editor';
 import { SharedColors } from '@uifabric/fluent-theme';
+import { get } from 'request';
```
a-b-r-o-w-n, line 34: This looks like a mistake. I don't see it being used anywhere.

a-b-r-o-w-n (changes requested):

extensions/azurePublish/src/components/util.ts
```diff
@@ -60,3 +60,35 @@ export const getExistResources = (config) => {
...
+  description: string;
+  text: string;
+  tier: string;
+  group: string;
+  key: string;
+  required: boolean;
+  [key: string]: any;
+};
+
+type AzureExtensionState = {
```
a-b-r-o-w-n, line 85: These types aren't exported, are they being used anywhere?

PR #6323, opened 2021-03-08
fix: moved long running creation steps to worker threads to prevent main process overload 

### LINKED ISSUE #6098: App goes into unresponsive mode during new creation 'npm install' and 'dialog merge' portions of the creation flow
## Describe the bug

The app becomes unresponsive for a short period when you create a new bot in the new creation flow during the template installation phase as well as the dialog merge phase.

This is likely caused by those portions being run within the app process as opposed to externally in a separately spawned process as, for example, is done with the `dotnet build` commands. 

## Version, browser and OS

all versions, browsers and OS

## To Reproduce

Go through the new creation flow and create any template for the first time

### LINKED ISSUE #6099: Update loading spinner text in new creation to be more granular
## Overview

The creation time for new bots in Composer can range anywhere from 30 seconds to 1.5 minutes. The loading spinner incrementally updates the user on the process of the creation flow. 

Initially it says 'loading' when it should be more granular. It should instead say 'installing template' and 'instantiating template' during the respective phases of the creation flow.

pavolum:
## Description

The main app process was getting overloaded during the new creation experience due to two long running processes that occur in the new creation flow:

- Yeomen template installation
- Dialog merge

This caused the application to go unresponsive for a short period of time which, among other things, inhibited the status endpoint to update the front end from indicating where in the creation process the process was in (displayed as a status message under the loading spinner)

This PR creates worker threads that run these processes away from the main app process and communicates with the parent process when the worker thread completes successfully, has a polling update or errors out. 

This ensures the main process does not get overloaded and the creation status is updated in a more granular manner.

## Task Item

fixes #6098 
fixes #6099 

## Screenshots

image

Note - Have fixed yeoman typo in screenshot to say 'Yeoman' as opposed to 'Yeomen'


Composer/packages/server/package.json
```diff
@@ -23,7 +23,7 @@
   },
   "author": "",
   "nodemonConfig": {
-    "exec": "cross-env TS_NODE_FILES=true node --max-http-header-size=16000 --inspect=9228 -r ts-node/register src/init.ts",
+    "exec": "yarn build && yarn start",
```
benbrown, line 26: why this change?

Composer/packages/server/src/models/asset/assetManager.ts
```diff
@@ -110,30 +108,9 @@ export class AssetManager {
...
-
-      if (remoteTemplateAvailable) {
-        await this.instantiateRemoteTemplate(yeomanEnv, generatorName, dstDir, projectName);
-      } else {
-        throw new Error(`error hit when installing remote template`);
-      }
+
+      await startYeomanTemplateWork(npmPackageName, templateVersion, dstDir, projectName, jobId);
 
       ref.path = `${ref.path}/${projectName}`;
```
benbrown, line 115: when this gets merged, let's make sure we don't revert the change needed for multibots, seen here:

https://github.com/microsoft/BotFramework-Composer/pull/6250/files#diff-bab4ce8cda0d657cc0bb0d75fc68eaea1d94673aed6b0ec3a807b50284d45101L138

Composer/packages/server/src/workers/merge_worker.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export function startMerge(manifestFile: string, currentProject: BotProject) {
+  return new Promise<void>((resolve, reject) => {
+    const w = new Worker(__filename, {
+      workerData: { manifestFile, dataDir: currentProject.dataDir },
+    });
+    w.on('exit', (returnCode) => {
+      if (returnCode === 0) {
+        resolve();
+      }
```
benbrown, line 42: shouldn't we have an else here?

benbrown (approved):
See comments

Composer/packages/server/src/workers/templateInstallation_worker.ts
```diff
@@ -0,0 +1,114 @@
...
+  dstDir: string,
+  projectName: string
+): Promise<boolean> => {
+  log('About to instantiate a template!', dstDir, generatorName, projectName);
+  yeomanEnv.cwd = dstDir;
+
+  await yeomanEnv.run([generatorName, projectName], {}, () => {
+    log('Template successfully instantiated', dstDir, generatorName, projectName);
+  });
+  return true;
```
srinaath, line 47: Is this return required since the catch handler of the caller will hit if there is an error.

Composer/packages/server/src/workers/merge_worker.ts
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) Microsoft Corporation.
```
srinaath, line 1: Is this the file naming convention for worker files. We use the \*.worker.ts file naming pattern in other places.

Composer/packages/server/package.json
a-b-r-o-w-n: I can chime in here- There are a couple of problems with the dev mode (using `ts-node` instead of `node`). 

1. Workers can only be instantiated with `*.js` files. Node throws an error if the filename arg to the worker is not js. This was the case when running with `ts-node`; the filename would be the .ts file instead of a .js one.
2. Ok, so just make the worker file a js one, problem solved. A new error was thrown about node not being able to resolve `ts-node` when executing the worker. 

TBH, we didn't investigate this really hard and made getting the workers working a priority. I will try to take a look at this to see if we can revert to the original dev script, but I would suggest taking this PR in the meantime.

Composer/packages/server/src/workers/merge_worker.ts
a-b-r-o-w-n: That naming convention exists for webpack loaders. I'm not sure it's worth snapping to that convention for node workers as it seems redundant.

`import { someThing } from '../workers/some-thing.worker'

Composer/packages/server/src/workers/merge_worker.ts
a-b-r-o-w-n: No, the rejection is handled in the message handler.

Composer/packages/server/src/workers/merge_worker.ts
```diff
@@ -0,0 +1,50 @@
...
+    .then(() => {
+      process.exit(0);
+    })
+    .catch((err) => {
+      parentPort?.postMessage({ error: err });
+      process.exit(1);
+    });
+}
+
+export function startMerge(manifestFile: string, currentProject: BotProject) {
```
hatpick, line 34: Please name this exported function in a way so the reader on the caller end knows this is a worker thread doing something.
+ (It's not just starting the work, it's doing it)

Composer/packages/server/src/workers/templateInstallation_worker.ts
```diff
@@ -0,0 +1,114 @@
...
+    parentPort?.postMessage({ status: formatMessage('Instantiating Yeoman template') });
+
+    await instantiateRemoteTemplate(yeomanEnv, generatorName, dstDir, projectName);
+  } else {
+    // handle error
+    throw new Error(`error hit when installing remote template`);
+  }
+};
+
+export function startYeomanTemplateWork(
```
hatpick, line 80: same comment for naming here

Composer/packages/server/src/workers/merge_worker.ts
a-b-r-o-w-n: To add to this, the caller doesn't even need to know that this is a 'worker'. It can be named something like `mergeSchemas` and the caller can just await it.

Composer/packages/server/src/models/asset/assetManager.ts
pavolum: Sounds good, will ensure that that when I handle merge conflicts

Composer/packages/server/src/workers/merge_worker.ts
srinaath: @a-b-r-o-w-n 

Then camelCase naming it \`import { someThing } from '../workers/mergeSchemas' seems a good option correct?

Composer/packages/server/src/workers/templateInstallation_worker.ts
pavolum: good catch, it has been removed

Composer/packages/server/src/workers/merge_worker.ts
pavolum: Handled

Composer/packages/server/src/workers/merge_worker.ts
pavolum: Handled

Composer/packages/server/src/workers/templateInstallation_worker.ts
pavolum: handled

Composer/packages/server/src/models/asset/assetManager.ts
pavolum: no merge conflicts, the bit of code you outlined is safe and sound in the merge output

a-b-r-o-w-n (dismissed):

a-b-r-o-w-n (approved):

PR #5853, opened 2021-02-20
fix: improve the loading bot performance

lei9444:
## Description
**issue**
1. The lg files are parsed twice
2. composer will wait for all resources process to finish

for calendar skill, parsing time is 19s, some other tasks' (render, state set) time is 3s, the total time to finish is 22s
image

**fix**
1. reduce one lg files' index. move all parsing operations to the worker.
2. put the lg files index to background. If we open a new dialog, check the lg file state and do a single parse if the file is not yet parsed during the loading stage.

The total time now to show the design page is about 5s
image

**next step**
If this solution is ok, we can let lu and qna use the same solution during loading.

## Task Item
refs #5708

## Screenshots
loadtime

PR #6334, opened 2021-03-09
feat: suggest user defined memory variables in LG files

### LINKED ISSUE #6137: Wrong predictive text options shown for turn.activity
## Describe the bug

Autocomplete options are sometimes not helpful.

## Version

1.3.1-nightly.219396.61c944a

## To Reproduce

1. Open a Send a Response action
2. Add a new response to the LG file
3. Start typing a variable name (e.g. `turn.activity.text`)
4. Watch the autocomplete options; `activity` is correctly a field in `turn`, but then `text` is not found as a field on `activity`.

## Expected behavior

I would expect the options to be aware of the correct type of "activity" and offer suggestions based on it.

## Screenshots

image

cosmicshuai:
closes #6137

Now in LG all up view editor, user wrote some variables like `user.predefined.props`, the language server will suggest this variable when user try to enter it again.

Here is a short demo. (`user.group.name` is defined is dialog files)
userDefinedVa

Also in this PR, the calculation of memory variables has been refactored, this will reduce the calculation.


Composer/packages/tools/language-servers/language-generation/src/LGServer.ts
```diff
@@ -817,6 +824,50 @@ export class LGServer {
...
+    if (projectId && fileId) {
+      const lgTextFiles = projectId ? this.getLgResources(projectId) : [];
+      const lgContents: string[] = [];
+      lgTextFiles.forEach((item) => {
+        if (item.id !== fileId) {
+          lgContents.push(item.content);
+        }
+      });
+
+      setTimeout(async () => {
```
hatpick, line 838: why the setTimeout?

Composer/packages/tools/language-servers/language-generation/src/lgParser.ts
```diff
@@ -44,6 +44,19 @@ class LgParserWithoutWorker {
 
     return { suggestEntities: uniq(suggestEntities) };
   }
+
+  public async extractLGVariables(curCbangedFile: string | undefined, lgFiles: string[]) {
```
hatpick, line 48: does this need to be async?

Composer/packages/tools/language-servers/language-generation/src/LGServer.ts
cosmicshuai: removed, it's not necessary.

Composer/packages/tools/language-servers/language-generation/src/lgParser.ts
cosmicshuai: Removed async.

hatpick (approved):

cosmicshuai:
Since this PR is related to add something in LG worker, it can wait until https://github.com/microsoft/BotFramework-Composer/pull/5853 is merged.

PR #6346, opened 2021-03-09
fix: Change identifier used for the new runtime

### LINKED ISSUE #6338: Rename the new runtime profile
The new runtime has a quirky identifier `csharp-azurewebapp-v2`

Should we rename this to something more future friendly and more descriptive such as `adaptive-runtime-dotnet-webapp`

If so, we need to:

- [x] Update the instances of this string in the Composer code [See here](https://github.com/microsoft/BotFramework-Composer/search?q=csharp-azurewebapp-v2) [Done](https://github.com/microsoft/BotFramework-Composer/pull/6346)
- [x] Update the default settings created by the generators [See here](https://github.com/microsoft/botframework-components/search?q=csharp-azurewebapp-v2) [Done](https://github.com/microsoft/botframework-components/pull/721)
- [x]  To avoid problems for users, patch this function to alias the old name ot the new name [see here](https://github.com/microsoft/BotFramework-Composer/blob/197dda8f22372f235b44b5858397436804abf3dc/Composer/packages/server/src/models/extension/extensionContext.ts#L87)
- [ ] Republish all the generators to npm

benbrown:
## Description

Change the identifier of the new runtime from `csharp-azurewebapp-v2` to the more descriptive and accurate `adaptive-runtime-dotnet-webapp`

Also includes backwards compatibility for the old name so that people who have created bots in the pre-release period will still work.

## Task Item

fixes #6338

Component side is here:
https://github.com/microsoft/botframework-components/pull/721


beyackle (approved):

cwhitten:
@benbrown is the name set? I recall talking about a rename for R12

benbrown:
> @benbrown is the name set? I recall talking about a rename for R12

This is the change we discussed.

cwhitten (approved):

PR #6347, opened 2021-03-09
chore: restart dev server when extensions change

a-b-r-o-w-n:
## Description

Restarts the server's dev server when extension is re-built.

## Task Item

#minor


beyackle (approved):
Looks good.

PR #6349, opened 2021-03-10
fix: enable deep link from ABS to work as expected with new creation flow

### LINKED ISSUE #6190: We need to support abs handoff for new creation flow.
## Describe the bug

Turn on the Create New Flow flag and Open in composer from abs. 

It is v2 templates with v1 UX 
image

## Browser

- [ ] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

pavolum:
## Description

Since the new creation flow code was cloned, updates have been made to the legacy creation code to handle deep linking and config passing from ABS. This led to the issue #6190 since the new creation flow had not implemented that logic. 

This PR handles the routing to the new creation flow so users who have the component model FF enabled will always be redirected to v2/project/create even if they click a deep link to Composers legacy creation flow project/create.

This PR also handles the passed data from the deep link and applies it to the end users project in the same way the existing creation flow handles that project data. 

## Task Item

fixes #6190


pavolum:
Do not merge this PR, the logic that applies the ABS data to the users project should work but has not been tested. I am waiting to hear from @luhan2017 on the best way to test the deep link with a local composer build to validate that this PR handles that project data correctly.

luhan2017 (approved):

cwhitten (approved):

PR #6354, opened 2021-03-10
chore: project tree flatten imports

zhixzhan:
## Description
update project tree ux in all up view

1.  Use file id as display name instead of previously '[import]'
2. Flatten imports (except form dialog)

```
- Echo-Bot
   - Echo-Bot.lg
- DialogA
   - DialogA.lg
      - [import](greeting.lg) 
```

will present as

```
- Echo-Bot
   - Common
   - greeting 
   - Echo-Bot
   - DialogA
```

## Task Item

refs #6238 

## Screenshots
Screen Shot 2021-03-10 at 3 38 48 PM


boydc2014 (approved):

Composer/packages/client/src/recoilModel/selectors/project.ts
```diff
@@ -271,12 +274,9 @@ export const projectDialogsMapSelector = selector<{ [key: string]: DialogInfo[]
   },
 });
 
-export const projectTreeSelectorFamily = selectorFamily<
-  TreeDataPerProject[],
-  { showLgImports: boolean; showLuImports: boolean }
->({
+export const projectTreeSelectorFamily = selector<TreeDataPerProject[]>({
```
beyackle, line 277: You should rename this now that it's no longer a family (just `projectTreeSelector` is fine).

Composer/packages/client/src/recoilModel/selectors/dialogImports.ts
```diff
@@ -37,8 +37,9 @@ export const getLanguageFileImports = <T extends LgFile | LuFile | QnAFile>(
     }
     const currentImports = file.imports.map((item) => {
       const importedFile = getFile(getBaseName(item.id));
+      const displayName = item.id.substring(0, item.id.indexOf('.'));
       return {
-        displayName: item.description,
+        displayName,
```
beyackle, line 42: You can keep this mostly as it is and just write `displayName: item.id.substring(0, item.id.indexOf('.'))` as a field on the object (replacing the `item.description` part).

beyackle (changes requested):
The code mostly looks good - I have a few style comments, plus I noticed the spacing in the Design View tree is a little off. The horizontal distance between the cubical "bot" icon and the dialog icon should match the distance from the dialog to the trigger icons.
image

PR #6360, opened 2021-03-10
feat: Add project ids, subscription id, and Microsoft app ids to telemetry

tdurnford:
## Description
- Enable telemetry app settings during dev mode to allow developers to verify new telemetry
- Add project ids to telemetry properties
- Add feature flags to global telemetry properties
- add Microsoft App Id and Subscription Id to NewPublishingProfileSaved and PublishingProfileStarted

## Task Item
#minor


Composer/packages/server/src/controllers/publisher.ts
```diff
@@ -68,6 +65,23 @@ export const PublishController = {
...
+        subscriptionId: configuration.subscriptionId,
+      });
+    } catch (error) {
+      TelemetryService.trackEvent('PublishingProfileStarted', {
+        target,
+        projectId,
+      });
+    }
+
+    TelemetryService.startEvent('PublishingProfileCompleted', target + projectId, { target, projectId });
```
hatpick, line 83: what is target + projectId? If string concat, use ````${}````

hatpick (approved):

PR #6367, opened 2021-03-10
fix: Hide deprecated "custom runtime" related features for adaptive rutime

### LINKED ISSUE #6321: Custom Runtime section still in settings menu
A bot created with the new creation flow does not need this section, and it is very confusing for it to be in the bot settings menu still. This should either be removed or disabled when the component model feature flag is enabled.

benbrown:
## Description

Since new Adaptive Runtime comes pre-ejected, we no longer support the ability to eject, nor do we support the ability to toggle this feature off.    All that remains in the runtime control section is the customizable start command that is used when the bot is started.  It is very useful for this to remain available as it allows the developer to pass additional options along to the runtime.

* Removes "Use custom runtime" option from settings for new-style Adaptive runtime bots
* Removes ability to eject for new-style Adaptive runtime bots
* Removes ability to change the path to the runtime

## Task Item

Fixes #6321

## Screenshots

For "old" bots, no change:

Screen Shot 2021-03-10 at 5 04 56 PM

For "new" bots, removed features:

Screen Shot 2021-03-10 at 5 11 12 PM


Composer/packages/client/src/pages/botProject/runtime-settings/RuntimeSettings.tsx
```diff
@@ -43,6 +45,14 @@ import {
 
 type RuntimeType = 'path' | 'command';
 
+/** Determine if a bot is configured to use the adaptive runtime vs the legacy runtime */
+export const isAdaptiveRuntime = (settings: DialogSetting): boolean => {
```
GeoffCoxMSFT, line 49: @bfc/shared now has the isAdaptiveRuntime check you can use.

Composer/packages/client/src/pages/botProject/runtime-settings/RuntimeSettings.tsx
```diff
@@ -43,6 +45,14 @@ import {
...
+  return (
+    settings?.runtime?.key === 'csharp-azurewebapp-v2' ||
+    (settings?.runtime?.key.match(/^adaptive-runtime/) ? true : false)
```
GeoffCoxMSFT, line 52: I didn't do a fancy regex, I just used startsWith.  Let me know if you think the regex is necessary and I can update the function I wrote.

Composer/packages/client/src/pages/botProject/runtime-settings/RuntimeSettings.tsx
```diff
@@ -67,11 +77,13 @@ export const RuntimeSettings: React.FC<RouteComponentProps<{ projectId: string }
   const [templateKey, setTemplateKey] = useState('');
   const [runtimePath, setRuntimePath] = useState(settings.runtime?.path ?? '');
   const [runtimeCommand, setRuntimeCommand] = useState(settings.runtime?.command ?? '');
+  const [isAdaptive, setIsAdaptive] = useState(false);
   const [usingCustomRuntime, setUsingCustomRuntime] = useState(settings.runtime?.customRuntime ?? false);
 
   useEffect(() => {
     // check the status of the boilerplate material and see if it requires an update
     if (projectId) getBoilerplateVersion(projectId);
+    setIsAdaptive(isAdaptiveRuntime(settings));
```
GeoffCoxMSFT, line 86: I think you want to useMemo here:
const isAdaptive = useMemo(isAdaptiveRuntime(settings),[settings]).

Otherwise you'll have to add settings to the dependency list of your effect.

GeoffCoxMSFT (changes requested):
Need to update the isAdaptive dependencies (or useMemo) and consider calling the shared method of isUsingAdaptiveRuntime.

Composer/packages/client/src/pages/botProject/runtime-settings/RuntimeSettings.tsx
benbrown: updated!

Composer/packages/client/src/pages/botProject/runtime-settings/RuntimeSettings.tsx
benbrown: nope that's fine, this regex is doing the same thing as startswith

GeoffCoxMSFT (approved):

PR #6368, opened 2021-03-11
feat: Add telemetry to track if Composer was opened from PVA or ABS

### LINKED ISSUE #5937: Telemetry: Track Handoff from ABS / PVA
We need a way of tracking a telemetry event when Composer is opened from ABS or PVA (or any other source in the future).  Ideally when opened we would either

- add an additional property onto the existing session started / Composer opened event indicating the referrer if any
or
- we add a new event (named something like HandoffToComposerCompleted) where we add properties to indicate the referrer

For each referrer we may want to track additional properties from the data payload sent from the referrer.

tdurnford:
## Description
Add telemetry to track if Composer was opened from PVA or ABS
## Task Item
Closes #5937


hatpick (dismissed):

Composer/packages/client/src/telemetry/useInitializeLogger.ts
```diff
@@ -46,6 +47,15 @@ export const useInitializeLogger = () => {
     TelemetryClient.pageView(page, page);
   }, [page]);
 
+  useEffect(() => {
+    // Track if Composer was opened from PVA or ABS
+    if (pathname === '/projects/import' || pathname === '/projects/create') {
+      const url = new URL(href);
+      const source = url.searchParams.get('source');
```
hatpick, line 54: source can be null which doesn't match your telemetry type, maybe do ```?? ''```

cwhitten (approved):

PR #6382, opened 2021-03-11
fix: refactor workers into their own package

### LINKED ISSUE #6372: Can't create bot with component model flag on
## Describe the bug

Can't create a bot with component flag on

## Version

Latest nightly
image

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Turn component model on
2. Create a bot
3. See error

image

benbrown:
## Description

refactors the server workers into their own package.

cuts direct dependencies on server internals.

this allows the workers to be unpacked in the electron repo and used.

## Task Item

fixes #6372


boydc2014:
@lei9444 can you help take a look at this one, and verify if there is impact to the way other workers are used or packed?

Composer/packages/server-workers/src/templateInstallation.worker.ts
```diff
@@ -8,23 +8,22 @@ import yeoman from 'yeoman-environment';
 import TerminalAdapter from 'yeoman-environment/lib/adapter';
```
hatpick, line 8: remove commented code
a bunch of awaits down below are not necessary, linter shows them.

Composer/packages/server-workers/src/dialogMerge.worker.ts
```diff
@@ -1,13 +1,11 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+import * as Path from 'path';
```
hatpick, line 4: nit: camelCase for import names, import * as path from 'path'

Composer/packages/server/src/models/asset/assetManager.ts
```diff
@@ -18,7 +19,19 @@ import { copyDir } from '../../utility/storage';
...
+let workerPath = '@bfc/server-workers/lib/templateInstallation.worker';
+if (isElectron) {
+  workerPath = path.join(
+    process.resourcesPath,
+    'app.asar.unpacked/node_modules/@bfc/server-workers/lib/templateInstallation.worker'
+  );
+}
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires, security/detect-non-literal-require
+const { runYeomanTemplatePipeline } = require(workerPath);
```
hatpick, line 34: nit: this can be await import(path) right before it's used to avoid the linter complaints.

Composer/packages/server/src/utility/project.ts
```diff
@@ -7,17 +7,28 @@ import { remove } from 'fs-extra';
...
+let workerPath = '@bfc/server-workers/lib/dialogMerge.worker';
+if (isElectron) {
+  workerPath = Path.join(
+    process.resourcesPath,
+    'app.asar.unpacked/node_modules/@bfc/server-workers/lib/dialogMerge.worker'
+  );
+}
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires, security/detect-non-literal-require
+const { runDialogMerge } = require(workerPath);
```
hatpick, line 30: nit: same comment

Composer/packages/server-workers/package.json
```diff
@@ -0,0 +1,28 @@
...
+    "build": "yarn clean && tsc",
+    "clean": "rimraf lib",
+    "lint": "eslint --quiet src",
+    "prepublishOnly": "npm run lint && npm run build"
+  },
+  "author": "benbro@microsoft.com",
+  "license": "MIT",
+  "dependencies": {
+    "yeoman-environment": "^2.10.3",
+    "format-message": "^6.2.3",
```
hatpick, line 19: let's remove this to reduce the size as much as we can

Composer/packages/server-workers/package.json
```diff
@@ -0,0 +1,28 @@
...
+    "@microsoft/bf-dialog": "4.11.0-dev.20201025.69cf2b9"
```
cwhitten, line 20: should this be 4.12.x?

Composer/packages/server-workers/src/templateInstallation.worker.ts
```diff
@@ -81,12 +87,15 @@ export function runYeomanTemplatePipeline(
   templateVersion: string,
   dstDir: string,
   projectName: string,
-  jobId: string
+  jobId: string,
+  templateGeneratorPath: string,
+  BackgroundProcessManager: any
```
cwhitten, line 92: can this have a better type?

Composer/packages/server/src/models/asset/assetManager.ts
```diff
@@ -18,7 +19,19 @@ import { copyDir } from '../../utility/storage';
 import StorageService from '../../services/storage';
 import { IFileStorage } from '../storage/interface';
 import { BotProject } from '../bot/botProject';
-import { runYeomanTemplatePipeline } from '../../workers/templateInstallation.worker';
+import { templateGeneratorPath } from '../../settings/env';
+import { BackgroundProcessManager } from '../../services/backgroundProcessManager';
+
+let workerPath = '@bfc/server-workers/lib/templateInstallation.worker';
```
cwhitten, line 25: preference to extract this into some utility and abstract the electron logic on L26. this sets up an easy extension in a hosted scenario where we may want some additional flexibility into where workers are held.

This would reduce to the following on L34
```typescript
const { runYeomanTemplatePipeline } = require(getWorkerPath('templateInstallation'));
```

Composer/packages/server/src/utility/project.ts
```diff
@@ -7,17 +7,28 @@ import { remove } from 'fs-extra';
...
 import settings from '../settings';
 import log from '../logger';
 import AssetService from '../services/asset';
 import { BotProject } from '../models/bot/botProject';
 import { BackgroundProcessManager } from '../services/backgroundProcessManager';
-import { runDialogMerge } from '../workers/dialogMerge.worker';
 
 import { Path } from './path';
 
+let workerPath = '@bfc/server-workers/lib/dialogMerge.worker';
```
cwhitten, line 21: preference to extract this into some utility and abstract the electron logic on L26. this sets up an easy extension in a hosted scenario where we may want some additional flexibility into where workers are held.

This would reduce to the following on L30
```typescript
const { runYeomanTemplatePipeline } = require(getWorkerPath('dialogMerge'));
```

cwhitten (changes requested):

Composer/packages/server/src/models/asset/assetManager.ts
a-b-r-o-w-n: I would suggest going a step further and adding an abstraction in the `server-workers` package that hides this. 

```ts
import { ServerWorker } from '@bfc/server-workers';

// somewhere in code
await ServerWorker.enqueue('templateInstallation', { ...args });
```

a-b-r-o-w-n:
@benbrown and I discussed offline, I am going to finish up this work today.

Composer/packages/server/src/models/asset/assetManager.ts
hatpick: enqueue doesn't reflect the job being done, await on enqueue is confusing. A verb like executeJob or runJob is more appropriate imho. With enqueue, it feels like a job is ready to be dispatched synchronously (like in Android)

Composer/packages/server-workers/src/templateInstallation.worker.ts
a-b-r-o-w-n: @hatpick my recent change upgrade the yeoman-environment package, but the @types package has not been updated upstream- so the linter may warn about unnecessary `await`s but the methods being invoked do return promises.

cwhitten (approved):

PR #6394, opened 2021-03-12
fix: Allow multiline variations for LG text and speech modalities

hatpick:
## Description

Adds multiline support for LG response editor (variations under Text and Speech)

## Task Item

#minor

## Screenshots

https://user-images.githubusercontent.com/1483492/110911185-35581580-82c7-11eb-96c4-04f13c476fc0.mp4


Composer/packages/lib/code-editor/src/lg/__tests__/TextModalityEditors.test.tsx
```diff
@@ -13,6 +13,7 @@ describe('TextModalityEditor', () => {
   it('should render the value if it is not a template reference', async () => {
     const { findByText } = render(
       <TextModalityEditor
+        focusOnMount={false}
```
tdurnford, line 16: Why are we setting focusOnMount to false? Should we just remove it?

tdurnford (dismissed):

Composer/packages/lib/code-editor/src/lg/hooks/useStringArray.ts
```diff
@@ -16,10 +18,30 @@ const getInitialItems = <T extends ArrayBasedStructuredResponseItem>(
...
-    ? template?.body?.replace(/- /g, '').split(/\r?\n/g) || []
+    ? template?.body
+        // Split by non-escaped -
+        // eslint-disable-next-line security/detect-unsafe-regex
+        ?.split(/(?<!\\)- /g)
+        // Ignore empty or newline strings
+        ?.filter((s) => s !== '' && s !== '\n')
+        ?.map((s) => s.replace(/\r?\n$/g, ''))
+        // Remove LG template multiline block symbol
+        ?.map((s) => s.replace(/```/g, '')) || []
```
beyackle, line 29: I think a normal `.` will work here instead of `?.` on lines 24 to 29 - since we only reach here in the case that both `response.value` and `template.body` are non-null, then we know that the `split` call will work, and so on down the line. You might need to change the checks on line 20 to expliticly `response.value != null && template.body != null` just to work around the falsey values.

Composer/packages/lib/code-editor/src/lg/__tests__/TextModalityEditors.test.tsx
hatpick: @tdurnford 
linter said it was required, so I set it to false. It's a test.

beyackle (changes requested):
Just one minor thing, but this looks overall good.

Composer/packages/lib/code-editor/src/lg/hooks/useStringArray.ts
hatpick: good call on ```.```
As for the check, I think it's better to avoid non-strict equality checks (== vs. ===).

beyackle: Most of the time, yes, but `== null` and `!= null` are special cases that also check for `undefined`.

tdurnford (approved):

beyackle (approved):

PR #6403, opened 2021-03-13
fix: restore dropdown triangles in ProjectTree

### LINKED ISSUE #6181: Project tree is missing the expand/collapse at the root bot
## Describe the bug

In the Design page, the arrow that collapses and expands the root bot disappears sometimes.

## Version

Version:  1.3.1-nightly.219396.61c944a
Electron: 8.2.4
Chrome: 80.0.3987.165
NodeJS: 12.13.0
V8: 8.0.426.27-electron.0
## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [x] macOS
- [ ] Windows
- [ ] Ubuntu

## To Reproduce

1. Open a bot.
2. Go to the design page. Look at the project tree. The arrow by the root bot is missing:
image

## Expected behavior

Should look like this (seems to work in either older bots or projects with only one bot):
image

beyackle:
## Description

This restores the missing hide/reveal triangles to the tree, also bringing their shape more in line with the original designs (which were small right triangles instead of the default equilateral ones with clumsy CSS scaling applied to them). There are still remaining alignment issues, but this does solve the larger problem and will serve as a starting point for more refinements.

## Task Item

closes #6181

## Screenshots

image


beyackle:
Discarding this; I found a cleaner way to do it.

PR #6412, opened 2021-03-15
fix: lg create error with multi-language

### LINKED ISSUE #6409: error on creating dialog
## Describe the bug
The main operation is not affected. Only a weird error shows up.
1. create a new bot.
2. add more languages.
3. create a new dialog.

It seems that it is trying to create a .lg (without a name) file and then to delete it. 

微信图片_20210315161724

lei9444:
## Description
**Root cause**
when we create a new lg file, today's flow is:
1. add id to lgFileIds state 
2. FilePersistence find the ids changes and create new file for lg
3. add single lg file state by id
in the step 2, FilePersistence will use the ids to find the new file and get the default value (empty) from the state. In fact, we need the real value from step 3.

**fix**
update the state change order. set file state first, then set the ids state. FilePersistence will do nothing when the id doesn't been add to the ids state.

 

## Task Item
closes #6409

### DIFF
--- a/Composer/packages/client/src/recoilModel/dispatchers/lg.ts
+++ b/Composer/packages/client/src/recoilModel/dispatchers/lg.ts
@@ -56,12 +56,12 @@ const updateLgFiles = (
 
   // adds
   if (adds?.length) {
-    set(lgFileIdsState(projectId), (ids) => ids.concat(adds.map((file) => file.id)));
     adds.forEach((lgFile) => {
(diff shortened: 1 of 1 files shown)


boydc2014 (approved):

boydc2014:
Not sure related to this PR, but CI failed because some path.ts error

beyackle:
The `path` error is weird - that code has been there for two years with no problems. Did the CI flow change something recently?

PR #6467, opened 2021-03-17
fix: Added retry logic to OneAuth login flows when an interactive sign in is required

### LINKED ISSUE #5436: Publishing 1.3.0 to Web App hangs after logging in
## Describe the bug

I'm trying to publish as an Azure web app.  I created just a basic echo bot template.  I then started the process to add a new publish profile and select the web app option.  It asked me to login to Azure which I did successfully.  It then hung on the "Configure resources" page.  It says "Loading" with the spinning wheel but nothing ever actually loads.

## Version

1.3.0

## Browser
Not really meaningful for this point since I'm just trying to publish.

- [ ] Electron distribution
- [x] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

Create a new publish profile 
Give it a name and select the publish to Azure Web App option
Login
It hangs on the next screen trying to load resources

I belong to a lot of different directories so is it possible it's related to that?

## Expected behavior

I expect it to show the create resources UI as documented in the publish documentation:
https://docs.microsoft.com/en-us/composer/how-to-publish-bot

## Screenshots
It's stuck on the below screen:

image

## Additional context

I have previously been using 1.2.1 and was able to use the provision script there to publish bots. I was also able to publish bots to the provisioned resources from Composer for older versions.  I tried opening one of those bots with 1.3.0 and publishing and I receive an error message saying, "Error - Required field `accessToken` is missing from publishing profile.".  The accessToken field is definitely in there, though, in the same place it was in 1.2.1.  I ran an "az account get-access-token" and pasted the new token in to make sure it had a valid one.

That may be a separate issue with 1.2.1 publish profiles not working with 1.3.0, though.  Which isn't as big of a deal right now since I could ideally create a new profile if needed.  The interesting thing, though, is that 1.2.1 bots can be published but it runs into an issue with 1.3.0.  So, my Azure account is setup with everything it needs to publish bots and it just seems to be an issue with the UI retrieving Azure information for some reason.

tonyanziano:
## Description

There was retry logic for the OneAuth `getAccessToken()` flow, but it was trying to catch the `InteractionRequired` error in the wrong place. A login error _actually throws_, which ends up in the `catch` block.

Moved retry logic to the `catch` block, and also added it to the ARM login flow.

## Task Item

This potentially fixes the issue we were seeing with Non-federated accounts:

Fixes #5436


Composer/packages/electron-server/src/auth/oneAuthService.ts
```diff
@@ -34,6 +34,27 @@ type GetTenantsResult = {
   value: AzureTenant[];
 };
 
+enum Status {
```
a-b-r-o-w-n, line 37: Where do these values come from? Maybe add a comment explaining where to find this list?

a-b-r-o-w-n (approved):

Composer/packages/electron-server/src/auth/oneAuthService.ts
tonyanziano: So we have them in our copied version of OneAuth's  `d.ts` file ([Source here](https://github.com/microsoft/BotFramework-Composer/blob/main/Composer/packages/electron-server/src/auth/oneauth.d.ts)). However, the problem is that, at run time, the `OneAuth` object that we initialize does not actually have these enums defined as properties on the object even though it is in the package's definition file.

So I extracted those enums out into this file, but I can add a comment pointing to the original definition file for reference.

cwhitten (approved):

ehsan2018:
Hello, I am having a similar issue upon publishing a bot from composer V1.4.
Would you please provide a plain language instruction as I am not a developer and have a hard time following technical terms. Thank you.

ehsan2018:
Thank you for your quick reply. I do not see the 'early adopters' under the setting page. However, it seems my application is up to date as I clicked on help/ check for update. 

image

image

a-b-r-o-w-n:
@ehsan2018 check on the application settings page (the gear icon on the bottom left).

PR #6471, opened 2021-03-18
fix: get microsoftAppPassword from profile appPasswordHint

VanyLaw:
## Description
get microsoftAppPassword from profile appPasswordHint, update the generated profile when create from abs.

## Task Item

#minor


Composer/packages/client/src/pages/publish/Publish.tsx
```diff
@@ -254,6 +255,9 @@ const Publish: React.FC<RouteComponentProps<{ projectId: string; targetName?: st
         token = await AuthClient.getARMTokenForTenant(tenant);
       }
     }
+    // get keyvault token
```
luhan2017, line 258: remove this part.

Composer/packages/server/src/externalContentProvider/azureBotServiceProvider.ts
```diff
@@ -0,0 +1,158 @@
...
+    const vaultNames = hint.match(/vaults\/([^/]*)/);
+    const secretNames = hint.match(/secrets\/([^/]*)/);
+    const resourceGroupNames = hint.match(/resourceGroups\/([^/]*)/);
+    const vaultName = (vaultNames && vaultNames.length > 0 && vaultNames[1]) || '';
+    const secretName = (secretNames && secretNames.length > 0 && secretNames[1]) || '';
+    const resourceGroupName = (resourceGroupNames && resourceGroupNames.length > 0 && resourceGroupNames[1]) || '';
+    const tenantId = this.metadata.tenantId;
+
+    // get token
+    const armToken = await this.getArmAccessToken(tenantId);
```
luhan2017, line 65: There might be some exceptions here:
1. get token failure
2. SetPolicy failure
3. get secrets failure

we should add logic to handle those exceptions.

Composer/packages/client/src/utils/electronUtil.ts
```diff
@@ -23,44 +27,25 @@ export type Profile = ABSProfile; // can include PVAProfile or other type of pro
...
-          ...temp,
-        }),
-      };
-    }
+export async function getPublishProfileFromPayload(
+  profile: Profile,
+  source: string
+): Promise<PublishTarget | undefined> {
+  try {
+    const resutl = await httpClient.post(`/import/${source}/generateProfile`, profile);
```
tonyanziano, line 35: `result`

Composer/packages/client/src/utils/electronUtil.ts
```diff
@@ -23,44 +27,25 @@ export type Profile = ABSProfile; // can include PVAProfile or other type of pro
...
+    console.log(resutl);
```
tonyanziano, line 36: remove

Composer/packages/client/src/pages/publish/Publish.tsx
```diff
@@ -254,6 +255,9 @@ const Publish: React.FC<RouteComponentProps<{ projectId: string; targetName?: st
...
+    const result = await AuthClient.getAccessToken(vaultScopes);
```
tonyanziano, line 259: This should not be done for every type of publish. Publishing to PVA should not require a call to get a KeyVault access token

Composer/packages/client/src/utils/electronUtil.ts
```diff
@@ -23,44 +27,25 @@ export type Profile = ABSProfile; // can include PVAProfile or other type of pro
...
-  switch (source) {
-    case 'abs':
-      if (payload) {
-        const profile: ABSProfile = JSON.parse(payload);
-        const alias = `${source}-${profile.botName}-${profile.appId}`;
-        return alias;
-      }
+export async function getAliasFromPayload(source: string, payload: string): Promise<string | undefined> {
+  try {
+    const resutl = await httpClient.post(`/import/${source}/getAlias`, JSON.parse(payload));
```
tonyanziano, line 45: `result`

Composer/packages/client/src/utils/electronUtil.ts
```diff
@@ -23,44 +27,25 @@ export type Profile = ABSProfile; // can include PVAProfile or other type of pro
...
+    console.log(resutl);
```
tonyanziano, line 46: remove

Composer/packages/server/src/externalContentProvider/powerVirtualAgentsProvider.ts
```diff
@@ -141,6 +142,10 @@ export class PowerVirtualAgentsProvider extends ExternalContentProvider<PowerVir
     return this.getAccessToken();
   }
 
+  public async generateProfile(): Promise<PublishTarget> {
+    throw new Error('PVA Not support');
+  }
```
tonyanziano, line 147: you have defined this as an optional function inside of `externalContentProvider.ts` so you shouldn't need to implement it for `PowerVirtualAgentsProvider`

Composer/packages/server/src/externalContentProvider/externalContentProvider.ts
```diff
@@ -53,4 +54,9 @@ export abstract class ExternalContentProvider<T extends IContentProviderMetadata
    * (Optional) Performs any necessary authentication for the service and returns an access token.
    */
   abstract authenticate?(): Promise<string>;
+
+  /**
+   * get different type of profiles base on provider
```
tonyanziano, line 59: "(Optional) Generates a publish profile based on the provider."

Composer/packages/server/src/services/auth/electronAuthProvider.ts
```diff
@@ -60,6 +60,24 @@ export class ElectronAuthProvider extends AuthProvider {
...
+  async getArmAccessToken(tenantId: string): Promise<string> {
+    const { getARMTokenForTenant } = this.electronContext;
+    if (isLinux()) {
+      log('Auth login flow is currently unsupported in Linux.');
+      return '';
+    }
+    log('Getting access token.');
+    try {
+      // otherwise get a fresh token
+      log('Did not find cached token. Getting fresh token.');
```
tonyanziano, line 72: remove

Composer/packages/server/src/services/auth/electronAuthProvider.ts
```diff
@@ -60,6 +60,24 @@ export class ElectronAuthProvider extends AuthProvider {
     }
   }
 
+  async getArmAccessToken(tenantId: string): Promise<string> {
+    const { getARMTokenForTenant } = this.electronContext;
+    if (isLinux()) {
+      log('Auth login flow is currently unsupported in Linux.');
+      return '';
+    }
+    log('Getting access token.');
```
tonyanziano, line 69: `log('Getting ARM access token.');`

tonyanziano (changes requested):

Composer/packages/server/src/externalContentProvider/externalContentProvider.ts
VanyLaw: Seem like abstract methods must be implemented in child class even it's optional, So I remove the abstract and make it just optional.

Composer/packages/client/src/pages/publish/Publish.tsx
VanyLaw: This code is only for testing, sorry we forget to remove these code. Currently, we implement the get keyvault token in azure BotServiceProvider in server.

VanyLaw: I just remove it

tonyanziano (approved):
Nice job!

cwhitten:
@VanyLaw @luhan2017 I'm trying to test this and I'm seeing the following behavior:

1. I don't see @alanlong9278 's dialog that was merged
2. No publishing profile is created
3. I can't publish back to Azure

luhan2017:
> * No publishing profile is created

Did you set COMPOSER_ENABLE_ONEAUTH=true in your build pipeline? I can't reproduce this issue in my side, the issue could because the oauth is not set correctly so, it failed to get the keyvault and generate the profile.

cwhitten:
> > * No publishing profile is created
> 
> Did you set COMPOSER_ENABLE_ONEAUTH=true in your build pipeline? I can't reproduce this issue in my side, the issue could because the oauth is not set correctly so, it failed to get the keyvault and generate the profile.

Yes, it was set as I used the main mac build/signing pipeline.

boydc2014:
1. fix the ci
2. testing the scenario with an existing name?

luhan2017:
@cwhitten , The ut and notification issue is fixed now. 
But we can't reproduce the alias not working issue as you mentioned before. Could you please help to figure out the reproduceable steps and check if there is anything wrong in data.json. I did see this issue once, and I found that it because there are multiple bots associated with the same bot (which is probably because in the middle of the implementation, the alias logic changes, so there are some bad data in the data.json), I don't see that again after I fix the data.json.

cwhitten (approved):

PR #6475, opened 2021-03-18
fix: Edit actions for Lg with structured responses (cut/copy/paste)

### LINKED ISSUE #6447: Using Cut / Paste with an action created with the new LG editor results in loss of underlying data
## Describe the bug

Adding a Send an Activity response (or any other action that contains an LG response, such as a prompt) that has related generated LG templates, such as the user adding multiple text / speech variations and then attempting to cut / paste that action results in data loss.  The related auto generated template is deleted on cut.

## Version

Latest Composer nightly.

## To Reproduce

1. Add Send a Response action
2. Add multiple text variations
3. Cut the action and then immediately paste it back into the dialog
4. LG error occurs because the referenced template no longer exists
5.

hatpick:
## Description
Adds cut/copy/paste for nested Lg template with structured responses.

## Task Item
closes #6447

https://user-images.githubusercontent.com/1483492/111669627-39f36100-87d4-11eb-8236-d5e8bba6ba4e.mp4


hatpick:
Wrong branch

PR #6506, opened 2021-03-22
fix: Remove nuget.config and cleanup

### LINKED ISSUE #6509: Remove nuget.config from dotnet runtime
## Describe the bug

- Nuget.config makes sense for the nightly, but not the production runtime, since everything is present on nuget.

- Remove copywrite headers, per Steve Ickman.

- Remove superfluous package references from projects.

EricDahlvang:
## Description

~~Remove nuget.config (all packages are on nuget, no need to reference myget)~~
-No need to reference packages which are inside Core, from the projects consuming Core.
-Remove copywrite headers, per Steve Ickman

## Task Item

closes #6509


boydc2014:
@EricDahlvang the reason we are on myget, is because composer often consume the daily packages from SDK, which is in myget.  And the current company policy don't allow us to use both myget + nuget at the same time, so eventually the call made is to use myget only with a feature "upstream source" configured to nuget. 

So unless we won't use daily packages anymore, we should still keep using myget. 

@mrivera-ms @cwhitten knew the previous context here.

boydc2014:
All looking good, just not sure about the nuget.config change.

EricDahlvang:
@boydc2014 I've added nuget.config back to this PR

boydc2014 (dismissed):

boydc2014 (approved):

PR #6444, opened 2021-03-17
fix: hidden showCode button when url is all & show code for form dialog

alanlong9278:
## Description

fix: hidden showCode button when url is all & show code for form dialog
luEdit

## task
closes #6440

PR #6516, opened 2021-03-23
fix: navigation of imports in project tree

zhixzhan:
## Description

Fix some imports in project tree did match correct link.

## Task Item

refs #6444
refs #6354 

## Screenshots
form dialog

imports


boydc2014:
did this also fix the hosit view of lu intent?

zhixzhan:
> did this also fix the hosit view of lu intent?

Do you mean reference lu intent not found in inline editor? not include that, I will make another PR fix it.

PR #6526, opened 2021-03-23
feat: detect "old" bots and migrate them to new runtime

### LINKED ISSUE #6680: Migrate old projects to the new adaptive runtime
When we launch the new adaptive runtime, we must provide a way to update old projects to the new format.

### LINKED ISSUE #7373: Errors converting bot from 1.4 to adaptive runtime
## Describe the bug

I was testing the part of the Composer 1.4 to Composer 2.0 migration in which the user opens a bot created in 1.4 in 2.0 and chooses to convert to the new format. (Using a build created for testing this flow--not one of the regular nightlies)

3 issues:
- (minor, but unexpected) In the "Create a bot project" modal, the Name field should be autopopulated with the name of the old bot. Intstead it is a lowercase version of the old bot.
- (already reported in #7091) In the "Create a bot project" modal, the "Runtime" field is empty, and the dropdown menu does not open
- converting was unsuccessful and I got an error (shown below)
- the folder intended for the converted bot was left on disk - related to closed item #6726 ?

## Version

Using Frankenbuild from Chris, Friday 4/23:

Version:  1.4.0
Electron: 8.2.4
Chrome: 80.0.3987.165
NodeJS: 12.13.0
V8: 8.0.426.27-electron.0

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [x] macOS
- [ ] Windows
- [ ] Ubuntu

## To Reproduce

1. Open the new build of Composer. 
2. Select a 1.4 bot .
3. In the "Convert your project to the latest format" modal, select "Convert."
4. In the "Create a bot project", the name is autopopulated with the original name, but the uppercase has been changed to lowercase, as shown here (my 1.4 bot was "ActionsSample-0":
image

It's a common pattern when duplicating a file for the UI to autopopulate with the new name, so that part is expected. But the changing the case is not expected.

5. The "Runtime type" field is empty and opening the dropdown menu (click on the down arrow), does nothing:
image

6. Click OK.

7. The progress indicators appear are working fine:
image

8. until this error:
image

Expanded:
image

I think I do have node installed: 
image

9. Click OK and return to the home menu. The converted bot project does not appear in the MRU, but it does appear in the Finder:
image

Had the same errors with the Actions Sample bot and the ToDoWithLUIS bot samples from 1.4.

## Expected behavior

- for the project name, the only change is not to lowercase the project name
- if conversion fails, unused folder should be removed

benbrown:
## Description

* Detect an old bot by looking at the settings.runtime.key field when the bot is open
* If NOT an adaptive-runtime bot, redirect user to migrate flow
* Create a new project using the adaptive-runtime generator, then copy all declarative assets over
* Reformat the settings 

Fixes #6680
Fixes #7373


Composer/packages/server/src/services/project.ts
```diff
@@ -204,7 +204,11 @@ export class BotProjectService {
     }
   };
 
-  public static openProject = async (locationRef: LocationRef, user?: UserIdentity): Promise<string> => {
+  public static openProject = async (
+    locationRef: LocationRef,
+    user?: UserIdentity,
+    allowPartialBots?: boolean
```
cwhitten, line 210: I'm mixed on this new interface. This is a critical path module and I'm wondering if we can preserve the interface.

options:

- change interface to accept `options` instead of `allowPartialBots` declared initially with `{ allowPartialBots: boolean }` so that it is future-proofed a bit.
- try/catch the error at the call-site and check the message. if we're migrating pass, re-throw if not
- something else?

Composer/packages/server/src/services/project.ts
```diff
@@ -432,6 +436,126 @@ export class BotProjectService {
...
+        // add all original files to new project
+        for (let f = 0; f < originalFiles.length; f++) {
+          await currentProject.migrateFile(
+            originalFiles[f].name,
+            originalFiles[f].content,
+            originalProject.rootDialogId
+          );
+        }
+
+        const newSettings = {
```
cwhitten, line 486: @carlosscastro can you look at this settings transform? 

I think should we whip up a json-schema definition and pull in a runtime validator and announce failures for the cases that users have modified settings into something we don't expect.

Composer/packages/server/src/services/project.ts
```diff
@@ -432,6 +436,126 @@ export class BotProjectService {
...
+        BackgroundProcessManager.updateProcess(jobId, 200, 'Migrated successfully', {
+          id,
+          ...project,
+        });
+      } else {
+        BackgroundProcessManager.updateProcess(jobId, 500, 'Could not find source project to migrate.');
+      }
+    } catch (err) {
+      BackgroundProcessManager.updateProcess(jobId, 500, err instanceof Error ? err.message : err, err);
+      // TelemetryService.trackEvent('CreateNewBotProjectCompleted', { template: templateId, status: 500 });
```
cwhitten, line 555: Instrumenting this is a good idea

cwhitten (changes requested):

Composer/packages/client/src/recoilModel/dispatchers/utils/project.ts
```diff
@@ -848,6 +870,22 @@ export const saveProject = async (callbackHelpers, oldProjectData) => {
   return result;
 };
 
+export const migrateProject = async (callbackHelpers, oldProjectData) => {
+  const { oldProjectId, name, description, location } = oldProjectData;
+  const response = await httpClient.post(`/projects/${oldProjectId}/project/saveAs`, {
```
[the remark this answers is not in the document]
benbrown: this function was actually unused, and has been removed

Composer/packages/client/src/recoilModel/dispatchers/utils/project.ts
```diff
@@ -848,6 +870,22 @@ export const saveProject = async (callbackHelpers, oldProjectData) => {
...
+    storageId: 'default',
+    name,
+    description,
+    location,
+  });
+  const data = loadProjectData(response.data);
```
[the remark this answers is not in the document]
benbrown: removed completely

Composer/packages/server/src/services/project.ts
```diff
@@ -432,6 +436,126 @@ export class BotProjectService {
...
+      const locationRef = getLocationRef(location, storageId, name);
+
+      await BotProjectService.cleanProject(locationRef);
+
+      // Update status for polling
+      BackgroundProcessManager.updateProcess(jobId, 202, formatMessage('Getting template'));
+
+      const newProjRef = await AssetService.manager.copyRemoteProjectTemplateToV2(
+        '@microsoft/generator-microsoft-bot-adaptive',
+        '1.0.0-preview-20210310.8ee9434', // use any available version
```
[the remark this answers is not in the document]
benbrown: yes, I reverted this to '*' which will get the latest version of the generator.

Composer/packages/server/src/services/project.ts
benbrown: what this flag allows is for it to *not* throw if we are migrating -- otherwise, it will reject the project as a non-bot because it doesn't YET include the necessary files.

I will go with option 1 here -- adding an options param instead of the boolean.

Composer/packages/server/src/models/bot/botStructure.ts
```diff
@@ -170,7 +175,7 @@ export const defaultFilePath = (
   }
 
   const DIALOGNAME = dialogId;
-  const isRootFile = BOTNAME === DIALOGNAME.toLowerCase();
+  const isRootFile = BOTNAME === DIALOGNAME.toLowerCase() || rootDialogId === DIALOGNAME;
```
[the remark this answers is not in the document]
benbrown: i updated this to lowercase things just in case.

Composer/packages/client/src/recoilModel/dispatchers/project.ts
```diff
@@ -281,8 +292,12 @@ export const projectDispatcher = () => {
     try {
       await flushExistingTasks(callbackHelpers);
       set(botOpeningState, true);
-      await openRootBotAndSkillsByProjectId(callbackHelpers, projectId);
-
+      const { requiresMigrate } = await openRootBotAndSkillsByProjectId(callbackHelpers, projectId);
```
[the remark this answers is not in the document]
benbrown: The reason I had this handled externally is that navigating inside these methods causes issues with the calling function.

That is, where we call openRootBot*, we are doing lots of stuff AFTER it resolves.  So if we navigate half-way, the remainder of that stuff is left in a weird place.

Hence, returning this status allows us to STOP the process and navigate rather than continue...

benbrown:
Includes a change related to #6678 -- renames any `azureFunctionsPublish` publish types to `azurePublish` as those will be combined into a single publishing mechanism when #6678 merges.

GeoffCoxMSFT (dismissed):

Composer/packages/server/src/services/project.ts
```diff
@@ -435,6 +442,129 @@ export class BotProjectService {
...
+              showTyping: originalProject.settings?.feature?.UseShowTypingMiddleware || false,
+              useInspection: originalProject.settings?.feature?.UseInspectionMiddleware || false,
+              removeRecipientMentions: originalProject.settings?.feature?.RemoveRecipientMention || false,
+              setSpeak: originalProject.settings?.feature?.useSetSpeakMiddleware || false,
+              blobTranscript: originalProject.settings?.blobStorage?.connectionString
+                ? originalProject.settings?.blobStorage
+                : {},
+            },
+            telemetry: {
+              instrumentationKey: originalProject.settings?.applicationInsights?.instrumentationKey,
```
benbrown, line 502: the i in instrumentationKey should be capitalized

Composer/packages/client/src/recoilModel/dispatchers/project.ts
```diff
@@ -281,8 +304,23 @@ export const projectDispatcher = () => {
...
+      if (requiresMigrate) {
```
carlosscastro, line 308: Can we make this code and the code in line 236 be in a single location and reuse?

Composer/packages/client/src/recoilModel/dispatchers/project.ts
```diff
@@ -448,6 +486,32 @@ export const projectDispatcher = () => {
...
+      const { set, snapshot } = callbackHelpers;
+      try {
+        const dispatcher = await snapshot.getPromise(dispatcherState);
+        set(botOpeningState, true);
+
+        // starts the creation process and stores the jobID in state for tracking
+        const response = await migrateToV2(callbackHelpers, oldProjectId, name, description, location);
+
+        if (response.data.jobId) {
+          dispatcher.updateCreationMessage(response.data.jobId, '', '', '', '');
```
carlosscastro, line 505: CAn we make the strings optional parameters so that this line looks like `dispatcher.updateCreationMessage(response.data.jobId)`?

Composer/packages/server/src/services/project.ts
carlosscastro: That is a great idea!

Composer/packages/server/src/services/project.ts
```diff
@@ -435,6 +442,131 @@ export class BotProjectService {
...
+        }
+
+        const newSettings = {
+          ...currentProject.settings,
+          runtimeSettings: {
+            features: {
+              showTyping: originalProject.settings?.feature?.UseShowTypingMiddleware || false,
+              useInspection: originalProject.settings?.feature?.UseInspectionMiddleware || false,
+              removeRecipientMentions: originalProject.settings?.feature?.RemoveRecipientMention || false,
+              setSpeak: originalProject.settings?.feature?.useSetSpeakMiddleware || false,
```
carlosscastro, line 498: This one was added by @tracyboehrer and it is now an object instead of a boolean. @tracyboehrer could you help us get this one right?

I think it will look something like (pseudo logic, Tracy should help us with the precise one): `setSpeak: { }` based on the properties here: https://github.com/microsoft/botbuilder-dotnet/blob/main/libraries/Microsoft.Bot.Builder.Dialogs.Adaptive.Runtime/Settings/SpeakSettings.cs

So the presence of the object enables setSpeak. I don't know what are the sensible default values to enter when we get, from the migrated file, a 'true'.

Composer/packages/server/src/services/project.ts
```diff
@@ -435,6 +442,131 @@ export class BotProjectService {
...
+                ? originalProject.settings?.blobStorage
+                : {},
+            },
+            telemetry: {
+              instrumentationKey: originalProject.settings?.applicationInsights?.InstrumentationKey,
+            },
+            skills: {
+              allowedCallers: originalProject.settings?.skillConfiguration?.allowedCallers,
+            },
+            storage: originalProject.settings?.cosmosDb?.authKey ? 'CosmosDbPartitionedStorage' : undefined,
```
carlosscastro, line 509: I would say that if CosmosDbPartitionedStorage is not there, the result we want is that the storage key is not there at all.

carlosscastro: I think this code does this already, right?

benbrown: Correct.

Composer/packages/client/src/recoilModel/dispatchers/project.ts
benbrown: added a forceMigrate function that encapsulates this for reuse!

Composer/packages/client/src/recoilModel/dispatchers/project.ts
benbrown: yup

hibrenda:
As discussed in today's stand up, runtime migration also needs to take care of .botproj as we do have multi-bots scenarios .
Following  is the issue we run into : 
old .botproj (root bot with local skill)
image
new .botproj (after migration)
image

Composer/packages/client/src/components/CreationFlow/v2/CreationFlow.tsx
```diff
@@ -150,13 +154,21 @@ const CreationFlowV2: React.FC<CreationFlowProps> = () => {
     saveProjectAs(projectId, formData.name, formData.description, formData.location);
   };
 
+  const handleMigrate = (formData) => {
+    handleDismiss();
```
GeoffCoxMSFT, line 158: handleDismiss sets the creation flow to CLOSE and then navigates home.  Is that all synchronous? This then sets the status to MIGRATE.  Do you want to only set the status to one state or does this need to go through the CLOSE state?

Composer/packages/client/src/components/CreationFlow/v2/CreationFlow.tsx
```diff
@@ -150,13 +154,21 @@ const CreationFlowV2: React.FC<CreationFlowProps> = () => {
...
+    setCreationFlowStatus(CreationFlowStatus.MIGRATE);
+    migrateProjectTo(projectId, formData.name, formData.description, formData.location);
```
GeoffCoxMSFT, line 160: If formData has type any, do you want to check name, description, and/or location have values before calling migrateProjectTo?

Composer/packages/client/src/recoilModel/dispatchers/project.ts
```diff
@@ -236,6 +238,22 @@ export const projectDispatcher = () => {
     }
   );
 
+  const forceMigrate = async (projectId: string) => {
+    if (
+      await OpenConfirmModal(
+        formatMessage('Convert your project to the latest format'),
+        formatMessage(
+          'This project was created in an older version of Composer. To open this project in Composer 2.0, we must copy your project and convert it to the latest format. Your original project will not be changed.'
```
GeoffCoxMSFT, line 246: How about 'To open this project in this version of Composer ...'

This way we don't have to think about the version number stuff in the message.

Composer/packages/client/src/components/CreationFlow/v2/CreationFlow.tsx
```diff
@@ -217,6 +229,16 @@ const CreationFlowV2: React.FC<CreationFlowProps> = () => {
           onDismiss={handleDismiss}
           onOpen={openBot}
         />
+        <DefineConversationV2
+          createFolder={createFolder}
+          focusedStorageFolder={focusedStorageFolder}
+          path="migrate/:projectId"
```
GeoffCoxMSFT, line 235: I'm not sure what the path is for but do you need to add a route to router.tsx for migration?

GeoffCoxMSFT (dismissed):

Composer/packages/client/src/components/CreationFlow/v2/CreationFlow.tsx
srinaath: Since this is is inside the creation flow I think this would take the creation route as prefix followed by migrate/:projectId as a sub routes

cwhitten (approved):

Issue #6512, opened 2021-03-23
Support referenced LU authoring when adding a trigger 


hibrenda:
Based on the context of https://github.com/microsoft/BotFramework-Composer/issues/6489 , the calendar bot is using the referenced LU content for trigger phrases, so we need to support how to add trigger for referenced LU content from UX .

Option A : Auto attached trigger phrases if trigger name match with the intent name defined in lu (current or imported) file. 
Option B : Drop down list for user to choose to bind them together. 

Similarly what should be showed in the LU editor & inline LU view should be aligned with the action.
eg. for choosing from drop down list, we may display the referenced file path ?
      for auto attached trigger phrases we may extract the trigger phrase from corresponding lu file ?

@mewa1024 @sangwoohaan  @boydc2014 @zhixzhan

sangwoohaan:
This is new to me – didn't know you can import lu files and reference them from activities. Instead of coming up short-term solution, I'd like to take some more time to investigate. Could this be an item for R14?

mewa1024:
yes, I agree with @sangwoohaan. For R13, the new creation experience (with templates that import LU) is a preview experience.

mewa1024:
@hibrenda, @sangwoohaan, I'm so sorry--I got R13 mixed up with R12 in my previous comment. I think we should consider this for R13 since many of the new templates will encounter this. Related issue filed by Lauren: #5358

sangwoohaan:
As a temporary solution for R13, the solution would be as simple as this 
![Frame 53156](https://user-images.githubusercontent.com/50963385/112387784-6c5afd80-8caf-11eb-9948-425edfcc5454.png)

zhixzhan:
Hi @hibrenda @sangwoohaan , before this new design, I figure out a solution to support display/edit referenced intent as well as it is in current file inside, no matter where actually the intent is, it behavior same. Do you think it's reasonable or we should take this new design? 
This is the PR 
https://github.com/microsoft/BotFramework-Composer/pull/6536

mewa1024:
Hi @zhixzhan yes, I think we should go with your solution.

mewa1024:
For the new trigger dialog, I think we should go with Option A for now. While Option B would make it easier to reuse trigger intents, I think it might confuse non-advanced users.

mewa1024:
@srinaath okay to move this to R15 milestone?

PR #6536, opened 2021-03-24
fix: use import resolver in LU parsing

### LINKED ISSUE #6489: inline lu editor did not show as expected for reference intent
- Enable new creation experience
- create calendar bot
- Reference intent in LU file
image
- Inline lu content did not show correctly
image
- LU all up view show as expected
image

zhixzhan:
## Description

Support lu parsing file with imports, get all intents from imports.

 a.lu
```
[import](b.lu)
# a
- 1
```

 b.lu
```
[import](c.lu)
# b
- 2
```

 c.lu
```
# c
- 1
```

parsing result for `a.lu` is 
```
{
intent : [{ name: 'a', body :'1' }]
allIntent: [ { name: 'a', body :'1' }, { name: 'b', body :'2' }, { name: 'c', body :'3' } ]
...
}
```

## Task Item
close #6489 
refs #6512 

## Screenshots

inline editor show referenced intent 
image
image


boydc2014 (changes requested):
The read scenario works for me, the write scenario of inline lu editor doesn't behave as expect, it is not writing at the right location. 
The inline lu editor always write to the dialog lu, instead of the imported lu when should.

zhixzhan:
> The read scenario works for me, the write scenario of inline lu editor doesn't behave as expect, it is not writing at the right location.
> The inline lu editor always write to the dialog lu, instead of the imported lu when should.

Fixed. now will write to correct file.

boydc2014 (approved):

PR #6542, opened 2021-03-24
feat: allow users to select their tenant when creating publish profiles

a-b-r-o-w-n:
## Description

If a user is a member of more than 1 tenant, a new tenant selection UI is shown before they can configure a new publish profile. 

*Breaking change*: New publish profiles will encode the tenant that was selected to be used for fetching arm tokens. Old profiles will continue to work ONLY if a user is a member of a single tenant. Users that are in multiple tenants must either add the correct tenant id to the existing profile or create a new profile.


extensions/azurePublish/src/components/azureProvisionDialog.tsx
```diff
@@ -758,7 +783,29 @@ export const AzureProvisionDialog: React.FC = () => {
...
+              style={{ margin: '0 4px' }}
+              text={formatMessage('Back')}
+              onClick={() => {
+                onBack();
+              }}
+            />
+            <PrimaryButton
+              disabled={!selectedTenant}
+              style={{ margin: '0 4px' }}
+              text="Next: Configure Resources"
```
tonyanziano, line 800: format message?

extensions/azurePublish/src/components/azureProvisionDialog.tsx
```diff
@@ -885,7 +932,7 @@ export const AzureProvisionDialog: React.FC = () => {
               text={'Done'}
               onClick={async () => {
```
tonyanziano, line 933: does this need to be `async` anymore?

tonyanziano (changes requested):
A few minor comments. Otherwise looks good 👍

cwhitten:
@a-b-r-o-w-n 

regarding:
> Breaking change: New publish profiles will encode the tenant that was selected to be used for fetching arm tokens. Old profiles will continue to work ONLY if a user is a member of a single tenant. Users that are in multiple tenants must either add the correct tenant id to the existing profile or create a new profile.

How is this presented to the user?

tonyanziano (dismissed):

a-b-r-o-w-n:
@cwhitten there is an in-app notification: https://github.com/microsoft/BotFramework-Composer/pull/6542/files#diff-f10b35fc6b5c1783f689105d901c1e1619f594476b689ffdcd7be2eb3111f670R252

cwhitten:
@a-b-r-o-w-n is this the behavior today without this change, or are you adding it as a constraint?

> Old profiles will continue to work ONLY if a user is a member of a single tenant

For example, I have access to more than one tenant, but I can also successfully publish on existing profiles. Will I be broken?

a-b-r-o-w-n:
@cwhitten I don't fully understand all the scenarios and why some were working while others were failing, but I suspect that the tenant being used implicitly was the correct one. This change forces you to be explicit about which tenant to publish to. Not sure if that clears anything up. 

I could revert back to the old method (just choosing the first tenant) and not show the notification which would be in parity with what we do today. What do you think?

cwhitten:
Makes sense. I'm of the mind to preserve existing behavior so we don't break people, and when we see an error about tenant mismatch we render the notification as you have and suggest they modify their profile or create a new one. 

@mewa1024 @emivers8 we could use a UX assist here.

We need to bring this functionality in so users with access to multiple tenants can publish deterministically. Without it there are potential mismatches with other information in the provisioning profile resulting in publishing failures. We also don't want to break people where this is working as expected. It's likely due to the "default" tenant that they have set up happens to be the right one with all the services their provisioning and publishing to. 

Attached is an example mis

match

mewa1024:
@cwhitten sorry for the delay in reply. Just want to make sure I understand the solution:

- For existing profiles, if the user had only one tenant when they created it and continues to have only one tenant, **there is no error.**
- For existing profiles, if the user had only one tenant when they created it and now has more than one tenant, **an error appears telling them to edit their profile and specify the tenant.**
- For new profiles, for users with one tenant or with more tenants, **they have to choose the tenant**. 

So in this screen (from Composer 1.4 nightly), we would have an additional dropdown for "Azure directory" above "Subscription" for both new profiles and existing profiles that a user edits:

image

Is that summary right?

So when a user is signed in to Azure, we don't sign them into a specific tenant? In other words, unlike the Azure portal or LUIS.ai, where when you sign in you choose a tenant that you use for the entire session. There is a control for switching tenants separate from the UI where you create/choose resources:

image

If we need an Azure sign in control, it could go in the in the main menu above the gear, similar to this from VS Code:
image

cwhitten:
@mewa1024 see below:

> For existing profiles, if the user had only one tenant when they created it and continues to have only one tenant, there is no error.

Correct

> For existing profiles, if the user had only one tenant when they created it and now has more than one tenant, an error appears telling them to edit their profile and specify the tenant.

Not always. If the information in the profile is associated with the "default" tenant, it works successfully. If the default tenant was changed, they would see an error.

> For new profiles, for users with one tenant or with more tenants, they have to choose the tenant.

Correct

The tenant/directory selection currently happens when you sign-in to azure, not when you are selecting resources.  @a-b-r-o-w-n are you able to share a gif or screenshot of this for @mewa1024 ?

a-b-r-o-w-n:
@cwhitten 

> The tenant/directory selection currently happens when you sign-in to azure, not when you are selecting resources.

This is sort of true. There is a new page in between selecting resources and the initial sign-in. The flow works like this: 
1. User clicks on "Add new profile" and selects an Azure profile.
2. The user will be prompted to sign in.
3. If the user is part of multiple tenants, a new screen appears with a tenant selection UI.
4. After the user chooses the tenant, they have to log in again to that tenant. (I don't know of a way around this, neither did @tonyanziano).
5. The provisioning / publishing flow works as normal.


https://user-images.githubusercontent.com/3619060/112508511-da093700-8d4c-11eb-814e-19fe5ed4ae48.mov

cwhitten:
Ahh, I got it. Following up w/ you on Teams

mewa1024:
fyi, Andy, Jas, and I talked offline. Andy figured out how to make the sign in work without the second approval step, so we can move the Azure tenant selector to the same screen where you pick the subscription etc should work. Jas updated screen here:
https://www.figma.com/file/J3oIdXDmeBQbAe4a4PYbQc/Jaskeerat-s-files?node-id=5319%3A87048

(Zooming out in the Figma file shows the latest provisioning flow.)

cwhitten:
> fyi, Andy, Jas, and I talked offline. Andy figured out how to make the sign in work without the second approval step, so we can move the Azure tenant selector to the same screen where you pick the subscription etc should work. Jas updated screen here:
> https://www.figma.com/file/J3oIdXDmeBQbAe4a4PYbQc/Jaskeerat-s-files?node-id=5319%3A87048
> 
> (Zooming out in the Figma file shows the latest provisioning flow.)

Nice!

GeoffCoxMSFT:
@a-b-r-o-w-n @cwhitten Can we get this checked in soon? We have a whole set of work in these files for publishing and one involves deferring sign in until later in the process so we can't really work on that until this is checked in.  :)

cwhitten:
> @a-b-r-o-w-n @cwhitten Can we get this checked in soon? We have a whole set of work in these files for publishing and one involves deferring sign in until later in the process so we can't really work on that until this is checked in. :)

Yep, planning to get it in this morning

cwhitten (approved):

PR #6545, opened 2021-03-24
a11y: fix ARIA labels on a lot of components

### LINKED ISSUE #6464: A11y: ARIA labels and roles should be defined
## Describe the bug

We have a large number of elements with incorrect ARIA roles, missing ARIA labels, and other related issues. This is a general problem throughout the app, and the Accessibility Insights extension is good at pointing these places out.

## Version

1.3.1

beyackle:
## Description

This is a wide-ranging set of changes that adds missing ARIA labels everywhere I could get it to work properly. It also fixes some small problems with the "select a trigger" text in the empty adaptive-flow pane (including a contrast issue that I hadn't noticed before). There are still some remaining issues that I haven't been able to resolve:
- The ProjectTree roles are tied to a larger suite of remodeling that it needs; I'd like to take that on as a future task
- Boolean fields in the adaptive forms come up as missing an ARIA label, even though the Dropdown is specifying an `ariaLabel` prop, and I don't know what's going on here

## Task Item

closes #6464


cwhitten:
@beyackle there is a conflict in this change

tonyanziano (approved):
LGTM!

cwhitten (approved):

PR #6549, opened 2021-03-25
fix: add options to support doing luis and qna cross-train separately

lei9444:
## Description
add crosstrain settings to composer setting:
  inner           Only performs the inner dialog cross train, default is true
  intra            Only performs the intra dialog cross train, default is true

## Task Item
#minor

### DIFF
--- a/Composer/packages/server/src/models/settings/defaultSettingManager.ts
+++ b/Composer/packages/server/src/models/settings/defaultSettingManager.ts
@@ -21,7 +21,6 @@ const newSettingsValuePath = [
   'luis.authoringEndpoint',
   'skillConfiguration',
   'customFunctions',
-  'crossTrain',
 ];
 
 const discardedSettingsValuePath = ['downsampling.maxUtteranceAllowed'];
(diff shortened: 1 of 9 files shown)


taicchoumsft (approved):

Composer/packages/server/src/models/settings/defaultSettingManager.ts
```diff
@@ -103,6 +104,10 @@ export class DefaultSettingManager extends FileSettingManager {
         isSkill: true,
         allowedCallers: [],
       },
+      crossTrain: {
```
scheyal, line 107: Terminology: 
Inter --> across
Intra --> Inner/within/internal
So we should have:
Inter / Intra  or 
cross/inner

mewa1024:
(@vivekkshankar fyi)

cosmicshuai:
I tested this PR, the behavior is expected as the result from the bf-cli. Here is a demo
for
```
inner: true
intra: false
```
image
image

for
```
inner: false
intra: true
```
image
image

cosmicshuai:
I tested the latest branch, it works correctly.

boydc2014 (approved):

PR #6550, opened 2021-03-25
fix: Warm up Orchestrator build cache from previous Composer session

### LINKED ISSUE #6481: Orchestrator: Initialize LabelResolverCache from blu files when in a new Composer Session
Orchestrator training (i.e. creating embeddings for utterances) takes a long time. To alleviate as much pain as possible, try to cache results between builds.

We already cache build results between builds. We store an in-memory map of `LabelResolvers` so that incremental training is very quick (only different utterances will need to be recalculated). 

But when we start a new Composer session, the cache is cold and recalculation happens again, even though the embeddings have essentially been cached in the `.blu` files. 

This request is to warm up the cache with data from the `.blu` files, so that training Orchestrator in a new Composer session will be quick, so long as we've trained before.

taicchoumsft:
## Description
If there are embeddings (in `.blu` files) from a previous Composer session, we can use them to warm up the Orchestrator `LabelResolversCache`. This dramatically speeds up subsequent build times in a cold Composer session. 

If Orchestrator isn't used, the warmup does nothing. 

## Task Item
closes #6481


Composer/packages/server/src/models/bot/process/orchestratorWorker.ts
```diff
@@ -25,9 +26,89 @@ export class LabelResolversCache {
...
+          async (f) =>
+            [f.replace('.blu', '.lu'), new Uint8Array(await readFile(Path.join(generatedFolderPath, f)))] as [
+              string,
+              Uint8Array
+            ]
+        )
+      );
+
+      return modelData.model && snapshotData.length
+        ? await Orchestrator.getLabelResolversAsync(modelData.model, '', new Map(snapshotData), false)
```
tsuwandy, line 102: I assume this matches the lu object id "f.replace('.blu', '.lu')" pass in the buildAsync method, correct?

tsuwandy (dismissed):

Composer/packages/server/src/models/bot/process/orchestratorWorker.ts
taicchoumsft: Yes, these are also the ids we get back from the results object from `buildAsync`,

boydc2014:
By code session, we mean, maybe it was closed before, right? If for a hot session, we will just memorize that in memory, is that correct?

taicchoumsft:
> By code session, we mean, maybe it was closed before, right? If for a **hot** session, we will just memorize that in memory, is that correct?

Yes, that is right. This cache warmup is for a user who has trained Orchestrator in a previous session, then closed Composer, then re-opened it.  In the same hot session, we use the existing in-memory cache logic.

boydc2014 (approved):

lei9444:
Tested it and LGTM

boydc2014 (approved):

PR #6551, opened 2021-03-25
fix: docker build failed

lei9444:
## Description
**Root cause**
The docker build use multi-stage builds today. There are three stages and will take more than 21G device size before the final image created. Our virtual machine's disk available space is 21G today. 

image

so when we do copy will throw the error: 'no space left on device'

**Fix**

Only use the first stage  to do the docker build ci

## Task Item
#minor

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -30,7 +30,8 @@ jobs:
           push: false
           context: .
           file: Dockerfile
-          tags: botframework-composer
+          tags: build
+          target: build
 
       - name: Health check
         run: |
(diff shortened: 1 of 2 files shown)


boydc2014:
If we only first stage, what about the size of the image?

boydc2014:
I see, you make the CI stop at stage 1, but normal run should work. That's a good idea.

boydc2014:
Looks good to me. Please add some comments in case it can be confusing.

boydc2014 (dismissed):

cwhitten (approved):

PR #6557, opened 2021-03-25
feature: Add delete button for publish targets

### LINKED ISSUE #5642: Can't Delete Publish Target
## Describe the bug

There is no way to delete a Publish Target short of editing the json file.

## Version

v1.3.0

## OS
- [ ] Windows

## To Reproduce

1. Create a Publish Profile
2. Attempt to Delete it

## Screenshots
image

GeoffCoxMSFT:
## Description

Add the delete button to the publish targets listed on the settings page.  
Includes a confirmation prompt.

## Task Item

closes #5642 

## Screenshots

image

image


benbrown (dismissed):

Composer/packages/client/src/pages/botProject/PublishTargets.tsx
```diff
@@ -83,6 +84,26 @@ export const PublishTargets: React.FC<PublishTargetsProps> = (props) => {
     }
   }, [scrollToSectionId]);
 
+  const onDeletePublishTarget = useCallback(
+    async (publishTarget: PublishTarget) => {
+      if (publishTargets) {
```
hatpick, line 89: how can it not have value, user just click delete from a list of them

Composer/packages/client/src/pages/botProject/PublishTargets.tsx
```diff
@@ -116,6 +138,15 @@ export const PublishTargets: React.FC<PublishTargetsProps> = (props) => {
                     {formatMessage('Edit')}
                   </ActionButton>
                 </div>
+                <div css={tableRowItem(columnSizes[2])}>
+                  <ActionButton
+                    data-testid={'deletePublishProfile'}
+                    styles={editPublishProfile}
+                    onClick={() => onDeletePublishTarget(p)}
```
hatpick, line 145: what's the point of use callback if it's used in an anonymous function (re-defines with every render)

hatpick (approved):

Composer/packages/client/src/pages/botProject/PublishTargets.tsx
GeoffCoxMSFT: Since it is async, I guess if the component was unmounted just as the callback happened it could be undefined.  It should not really be possible, but I get a compile error if I don't check.

Composer/packages/client/src/pages/botProject/PublishTargets.tsx
GeoffCoxMSFT: Good point.

hatpick (approved):

PR #6576, opened 2021-03-26
fix: Display error w/o token on auth failure

### LINKED ISSUE #6359: if access token is invalid, provision process fails silently when loading subscriptions
## Describe the bug

An invalid access token, such as one pasted in manually, causes the call to load azure subscriptions to fail silently.

Errors in this API should be caught and messaged to the user or handled appropriately rather than silently failing.

When this happens, an incorrect error message about an empty subscription list is displayed.

See this line:
https://github.com/microsoft/BotFramework-Composer/blob/9215e4f79581193ca8050c2d7e0590738c758780/extensions/azurePublish/src/components/api.tsx#L44

GeoffCoxMSFT:
## Description

getSubscriptions was failing silently.  Error on page indicated empty list rather than error.
Now error shows up on page.  Bearer token is excluded from error message and logger file.

## Task Item

fixes #6359


hatpick (dismissed):

extensions/azurePublish/src/components/azureProvisionDialog.tsx
```diff
@@ -376,11 +383,24 @@ export const AzureProvisionDialog: React.FC = () => {
...
+              setSubscriptionsErrorMessage(
+                formatMessage(
+                  'Your subscription list is empty, please add your subscription, or login with another account.'
+                )
+              );
+            }
+          }
+        })
+        .catch((err) => {
+          setSubscriptionsErrorMessage(err.message);
```
hatpick, line 401: to be complete, you need the mount check here too

extensions/azurePublish/src/components/azureProvisionDialog.tsx
```diff
@@ -885,7 +903,7 @@ export const AzureProvisionDialog: React.FC = () => {
               onClick={() => {
                 setPage(PageTypes.ReviewResource);
                 setTitle(DialogTitle.REVIEW);
-                let selectedResources = enabledResources.slice();
+                let selectedResources = requireResources.concat(enabledResources);
```
hatpick, line 906: nit: [...arr1, ...arr2]

hatpick (dismissed):

benbrown (approved):

PR #6578, opened 2021-03-26
fix: check for node on app start on users machine and notify if not installed

### LINKED ISSUE #6097: NPM and Node.js not installed needs to prompt you to install Node.js and NPM
## Problem

Currently, Composer doesn't tell the user during install to install NPM.  We need a way to communicate and/or install Node.js and NPM as part of Composer.

When Composer installs or runs the first time, that it looks for NPM and Node.js and if not installed it installs it kicks off the install for the user.

If you start Composer and NPM and Node.js are not installed at the proper versions or at all... Composer communicates with a link to the installer for Node.js and NPM.  A warning should continue to be on your Composer interface until the issue is resolved.

pavolum:
## Description

Adding server call that runs 'node -v' on the users machine. If the user does not have node the code catches the error and responds to the caller that node is not installed. If the user does have node it returns to the caller that node is installed along with he node version. 

Added caller on app start to new server endpoint. If it returns that node is not installed it prompts the user with a modal on app start with a link to the node installation page. 

## Task Item

fixes #6097 

## Screenshots

image


pavolum:
OPEN QUESTIONS: 
- The modal allows the user to dismiss the modal and continue using the app even if node is not installed, should this be allowed or if node is not installed should we prevent further interaction with the app even though main app components like creation will not work
  - If we do let them navigate the app then maybe e should add some sort of banner to remind them that the app will have issues since node is not installed. At current state they will get an informative error if they run creation without Node, maybe that will suffice.

emivers8:
> OPEN QUESTIONS:
> 
> * The modal allows the user to dismiss the modal and continue using the app even if node is not installed, should this be allowed or if node is not installed should we prevent further interaction with the app even though main app components like creation will not work
>   
>   * If we do let them navigate the app then maybe e should add some sort of banner to remind them that the app will have issues since node is not installed. At current state they will get an informative error if they run creation without Node, maybe that will suffice.

Yeah, I wasn't sure what would be standard practice here. If we let it stay open, your suggestion is probably best. We don't want to have to throw up an error dialog every time someone attempts to use a feature that isn't going to work or be surprised when it crashes or something. Maybe @cwhitten can weigh in--would you expect the app to just close if you didn't install node, or would you expect to still use it, just with a banner as @pavolum suggests?

Composer/packages/client/src/App.tsx
```diff
@@ -21,19 +22,29 @@ const Logger = () => {
...
+  const userHasNode = useRecoilValue(userHasNodeInstalledState);
 
+  const { fetchExtensions, fetchFeatureFlags, checkNodeVersion } = useRecoilValue(dispatcherState);
+  const [showNodeModal, setShowNodeModal] = useState(false);
   useEffect(() => {
     loadLocale(appLocale);
   }, [appLocale]);
 
   useEffect(() => {
+    checkNodeVersion();
```
hatpick, line 34: Should we care about this check only if we are running on electron?

pavolum: We need to check if node is installed in all environments due to the dependency taken on node to create a new bot under the component model which is present in both electron and non electron environments

hatpick: @pavolum on a web hosted composer, your check basically asks server if it has node installed, how's that useful?

boydc2014:
I think we should not allow user to continue, if node is not found. 

Also, i'm not sure the approach to detect node. If i understand correctly, this is for electron, right? because if it's web version, without node, the server won't even run. Even for electron, the underlying there is a node there to start server. So it's bit weird to **detect node using node**. 

I would prefer a simpler approach, like adding some scripts in electron startup and abort when no node is detected.

cwhitten:
1. Node isn't required for the PVA scenarios. We definitely should not block use of Composer. The node requirement is a constraint to create new runtimes in Composer, not use Composer entirely. The two are different and we should treat it as such. I'd recommend we throw up the Node error when you try and step through the creation experience and we aren't in the PVA scenarios.
2. As far as node recognizing node, it is weird I agree but generally if you replace node in this sense with python (if/when the SDK supported declarative/runtime) we would be using node to detect python, which makes a lot more sense. 

My two cents would to let users click around Composer, but not let them create new bots.

Composer/packages/client/src/App.tsx
hatpick: @pavolum ping

hatpick (dismissed):

cwhitten (approved):

PR #6579, opened 2021-03-26
chore: Package manager: Update tags for default search terms to msbot-component

carlosscastro:
#minor


beyackle (dismissed):

benbrown (approved):

extensions/packageManager/src/node/index.ts
```diff
@@ -57,47 +57,48 @@ export default async (composer: IExtensionRegistration): Promise<void> => {
...
-            url: 'https://api.nuget.org/v3/index.json',
-            readonly: true,
-            defaultQuery: {
-              prerelease: true,
-              semVerLevel: '2.0.0',
-              query: 'microsoft.bot.components+tags:bf-component',
-            },
-            type: PackageSourceType.NuGet,
+      // Read the list of sources from the config file.
+      let storedSources: IPackageSource[] = composer.store.read('feeds') as IPackageSource[];
```
cwhitten, line 61: prefer const over let

cwhitten: would userStoredSources or userProvidedSources be a better name?

extensions/packageManager/src/node/index.ts
```diff
@@ -57,47 +57,48 @@ export default async (composer: IExtensionRegistration): Promise<void> => {
...
-              prerelease: true,
-              semVerLevel: '2.0.0',
-              query: 'tags:bf-component',
-            },
-            type: PackageSourceType.NuGet,
+          type: PackageSourceType.NuGet,
+        },
+        {
+          key: 'nuget-community',
+          text: 'community packages',
```
cwhitten, line 79: is this string rendered in the ui?

extensions/packageManager/src/node/index.ts
```diff
@@ -57,47 +57,48 @@ export default async (composer: IExtensionRegistration): Promise<void> => {
...
+      let packageSources: IPackageSource[] = [
+        {
+          key: 'nuget',
+          text: 'nuget',
+          url: 'https://api.nuget.org/v3/index.json',
+          readonly: true,
+          defaultQuery: {
+            prerelease: true,
+            semVerLevel: '2.0.0',
+            query: 'microsoft.bot.components+tags:msbot-component',
```
cwhitten, line 73: nit: pulling out `tags:msbot-component` into a variable would give a bit of reuse and give the opportunity to describe it a bit more

extensions/packageManager/src/node/index.ts
carlosscastro: This was non-const and then changed the structure a bit, thanks

carlosscastro: made it const and named `userStoredSources` so it's more clear!

extensions/packageManager/src/node/index.ts
carlosscastro: Agreed. When we support querying in the UI we would do one more round of refactoring here to use these queries in a smoother way and simplify the code.

extensions/packageManager/src/node/index.ts
carlosscastro: @cwhitten I think it is the text that shows up in the drop down where you select feeds and also in the list when you manage your feeds.

cwhitten: We need to wrap this in our translation api `formatMessage`. Here is an example in the client: https://github.com/microsoft/BotFramework-Composer/blob/main/Composer/packages/adaptive-flow/src/adaptive-flow-editor/renderers/NodeMenu.tsx#L33

We typically try and encapsulate view data from the server code, of which this is defined. I don't see an issue pulling in `format-message` into scope here, though.

extensions/packageManager/src/node/index.ts
```diff
@@ -57,47 +57,50 @@ export default async (composer: IExtensionRegistration): Promise<void> => {
...
-            defaultQuery: {
-              prerelease: true,
-              semVerLevel: '2.0.0',
-              query: 'microsoft.bot.components+tags:bf-component',
-            },
-            type: PackageSourceType.NuGet,
+      // Read the list of sources from the config file.
+      const userStoredSources: IPackageSource[] = composer.store.read('feeds') as IPackageSource[];
+
+      const botComponentTag: string = 'msbot-component';
```
cwhitten, line 63: We have a lint rule that is grumpy about the type annotation here:

`Type string trivially inferred from a string literal, remove type annotation`

carlosscastro: Lint angry at me for not flowing with the untyped world :)

carlosscastro:
Linter seems to be happy now. Had some issues running it locally, fixed now.

cwhitten:
new lint error:

>/home/runner/work/BotFramework-Composer/BotFramework-Composer/extensions/packageManager/src/node/index.ts
Error:   7:1  error  There should be no empty line within import group  import/order

incidentally, you can run this static analysis locally with `yarn lint`

carlosscastro:
@cwhitten yes I've been running that. I just found the problem. I was running the linter but it was not honoring the rules in the project because of environment stuff in my computer. Sorry for the extra churn!

carlosscastro:
Checks were failing due to package source outages, re-running

cwhitten (approved):

Issue #6340, opened 2021-03-09
Instrumentation for Orchestrator


scheyal:
Request for instrumentation of Language & Orchestrator in Composer for Opt-In users (assuming privacy, legal approval).  (@garypretty)

Metric | Type | P | Comment
-- | -- | -- | --
Orchestrator   package is installed | Count | 0 |  
Orchestrator is   used as recognizer | Count | 0 |  
Plus Context: Root or not Root dialog | Count | 0 |  
No. of intents per   LU | Count | 0 |  
No. of entities per   LU | Count | 0 | Even if ignored in   Orch. it is a good to identify import use patterns & future use
No. of examples per   LU | Count | 0 |  
Skills used | Count | 0 | Dispatch children   profile
LUIS dialogs | Count | 0 | Dispatch children   profile
QnAMaker dialogs | Count | 0 | Dispatch children   profile
Basemodel used | Value | 0 |  
Basemodel download   time | Timespan | 0 |  
Build time: first | Timespan | 0 |  
Buildtime:   not-first | Timespan | 0 |  
Prediction time | Timespan | 0 |  
CPU | Value | 2 | Computer Hardware.   Should be interesting beyond Orchestrator.
Memory | Value | 2 | Computer Hardware.   Should be interesting beyond Orchestrator.
Architecture | Value | 2 | Computer Hardware.   Should be interesting beyond Orchestrator.
OS | Value | 1 | Linux, Mac,   Windows
Runtime Env. | Value | 1 | Dotnet, nodejs +   version info
Hosting Env. | Value | 3 | Azure, AWS…Should   be interesting beyond Orchestrator.

garypretty:
We met and discussed the above requirements.  Some are already in place, some we will push out and some we will tackle as part of R13.  Here is my summary.

**To implement in R13**

- Add two events ***OrchestratorDownloadStarted*** and ***OrchestratorDownloadCompleted***.

- Add two events ***OrchestratorBuildStarted*** and ***OrchestratorBuildCompleted***. These events should have two additional custom properties added to them, 'basemodel' (name of the model used) and 'firstbuild' (true / false).

Regarding additional properties around CPU / Memory / Architecture - @tdurnford is confirming if it is even possible for us to capture these.  If it is then we would add these to the global properties attached to all events or to the session started event.

**Already in place**

- Orchestrator package installed should be captured from the telemetry within the package manager.

**Pushing to a future milestone**

[...]

PR #6590, opened 2021-03-29
feat: Add LU editor telemetry and 

### LINKED ISSUE #6541: LU editor telemetry
We need to add telemetry for the new LU editor features being introduced in R13.

The following events will be required.

LUEditorEntityDefinitionAdded - fires when the user inserts a new entity using the new UI.
Properties should be added to this event indicating EntityType (e.g. ML or DateTimev2)

LUEditorEntityTagAdded - fires when the user inserts a new entity using the new UI.
Properties should be added to this event indicating EntityType (e.g. ML or DateTimev2)

tdurnford:
## Description
- Add telemetry to the LU Editor Toolbar
- Add Architecture / Memory / and CPU to the SessionStarted event

## Task Item
Closes #6541
Refs #6340

### DIFF
--- a/Composer/packages/lib/code-editor/src/LuEditor.tsx
+++ b/Composer/packages/lib/code-editor/src/LuEditor.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-import { LuFile } from '@botframework-composer/types';
+import { LuFile, TelemetryClient } from '@botframework-composer/types';
 import styled from '@emotion/styled';
 import { EditorDidMount, Monaco } from '@monaco-editor/react';
 import { FluentTheme, NeutralColors } from '@uifabric/fluent-theme';
@@ -76,6 +76,7 @@ export interface LULSPEditorProps extends BaseEditorProps {
       }
     | string;
   toolbarHidden?: boolean;
+  telemetryClient: TelemetryClient;
   onNavigateToLuPage?: (luFileId: string, luSectionId?: string) => void;
 }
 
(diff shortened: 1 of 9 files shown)


Composer/packages/lib/code-editor/src/LuEditor.tsx
```diff
@@ -239,7 +241,7 @@ const LuEditor: React.FC<LULSPEditorProps> = (props) => {
           if (luEdits?.scrollLine) {
             editor.revealLineInCenter(luEdits?.scrollLine);
           }
-
+          telemetryClient.track('LUEditorEntityDefinitionAdded', { entityType });
```
hatpick, line 244: I think ```LUEditorToolbar``` should be the prefix for all the toolbar events, the keyword is toolbar

Composer/packages/lib/code-editor/src/LuEditor.tsx
```diff
@@ -248,11 +250,12 @@ const LuEditor: React.FC<LULSPEditorProps> = (props) => {
   );
 
   const insertEntity = useCallback(
-    (entityName: string) => {
+    (entityName: string, entityType: string) => {
       if (editor) {
         const edits = computeInsertLuEntityEdits(entityName, editor);
         if (edits) {
           editor.executeEdits('toolbarMenu', edits);
+          telemetryClient.track('LUEditorEntityTagAdded', { entityType });
```
hatpick, line 258: We should differentiate between insert vs. tag
We should differentiate between tag via toolbar vs floating labeling menu

hatpick: entityType here is either: 'ml' or 'prebuilt', if you want the real prebuilt entity such as ```age```, you need to check if it's prebuilt, use entity name

hatpick (approved):
I'd change 'method' to 'source' but that's a nit

hatpick (approved):

Issue #6559, opened 2021-03-25
a11y: incorrect/misleading/uninformative screen-reader labels


beyackle:
## Describe the bug

This ticket describes the current state of the screen-reader text; there are a number of issues as noted below:

- "No item in view" on triggers in tree
- Error display just reads numbers (e.g. "6 3")
- focus moves backward when collapse/expand chevron is selected
- "one bot have been found"
- Just "button" for expand/collapse on bot project settings
- "westus" read instead of LUIS region
- Copy content for translation dialog is missing explanatory labels
- read out "Installing plugin" text from dialog
- adapter section: "Manage profiles" doesn't read ABS channels
- app settings page: toggles don't say what they do

## Version

1.4.0 

## Browser

- [ ] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [x] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

beyackle:
I think this can be closed once those two PRs merge; I've double-checked the list and the one remaining issue that looks unsolved (the "focus moves backward" one) I wasn't able to duplicate.

PR #6593, opened 2021-03-29
a11y: announce dialog text in package manager

beyackle:
## Description

This adds screen-reader text to the loading-spinner box in the package manager, giving SR-only users feedback that they've successfully gotten a package to be installed or removed.

## Task Item

refs #6559


extensions/packageManager/src/components/WorkingModal.tsx
```diff
@@ -5,14 +5,25 @@
...
 }
 export const WorkingModal: React.FC<WorkingModalProps> = (props) => {
+  const { announce } = useApplicationApi();
+
+  useEffect(() => {
+    async function doAnnouncement() {
+      announce(props.title);
+    }
+    doAnnouncement();
+  }, [props.title]);
```
beyackle, line 25: note: this does look strange, but it's the correct way to use an async function inside useEffect.

extensions/packageManager/src/components/WorkingModal.tsx
```diff
@@ -5,14 +5,25 @@
 import { jsx } from '@emotion/core';
 import { Dialog, DialogType } from 'office-ui-fabric-react';
 import { LoadingSpinner } from '@bfc/ui-shared';
+import { useApplicationApi } from '@bfc/extension-client';
 
 import { modalControlGroup } from './styles';
+import { useEffect } from 'react';
```
a-b-r-o-w-n, line 11: These annotations are cool!

extensions/packageManager/src/components/WorkingModal.tsx
a-b-r-o-w-n: This definitely seems incorrect though. What's wrong with this?

```ts
useEffect(() => {
  announce(props.title);
}, [props.title]);
```

If you wanted to await something, that would be the way to do it, but just invoking an async function shouldn't require doing this.

a-b-r-o-w-n (changes requested):

extensions/packageManager/src/components/WorkingModal.tsx
beyackle: I agree it shouldn't, but useEffect has a restriction where it doesn't like something that returns a promise. This is a workaround documented in places like https://www.robinwieruch.de/react-hooks-fetch-data and https://github.com/facebook/react/issues/14326 (where Dan Abramov himself weighs in).

hatpick: You can't await a promise within a useEffect block, you can just call it.

(use callbacks to handle results if required OR use IIFE to await it like:
```typescript
(async () => {
   await yourAsyncFunction(args);
})();
```

tdurnford: According to the type definition, `announce` isn't async.

https://github.com/microsoft/BotFramework-Composer/blob/90a332028c93280c32cbe21be78fb6cebc0078eb/Composer/packages/types/src/shell.ts#L70

beyackle: I know it isn't, but trying to do this the obvious way makes the app crash with the same error that blog post mentions: `Warning: useEffect function must return a cleanup function or nothing. Promises and useEffect(async () => ...) are not supported, but you can call an async function inside an effect.` Pretending it's async and using the IIFE workaround makes this work.

a-b-r-o-w-n: I think there is a misunderstanding here @beyackle 
1. is the type definition for `announce` wrong and it actually does return a promise?
2. Even if it does return a promise, it is still ok to invoke an async function inside an effect without using await. [(First example I found, but there are others)](https://github.com/microsoft/BotFramework-Composer/blob/main/Composer/packages/client/src/components/CreationFlow/v2/CreationFlow.tsx#L83)

@hatpick and I are just saying that in this case, the IIFE is unnecessary because we do not need to await the function.

hatpick (approved):

a-b-r-o-w-n (approved):

PR #6628, opened 2021-03-31
fix: do not require multiple sign ins for tenant selection

### LINKED ISSUE #6595: Azure provision in webapp is broken
## Describe the bug

main branch start composer webapp

1. add a new publish profile.
2. input arm token and graph token.
3. the provision page is blank.

image

## Browser

- [ ] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

### LINKED ISSUE #6614: [3/29 bugbash] [Cumulative issue] Repeating auth prompts - 2 main issues
1. people reported being prompted too many times for auth during the flow
2. specific to the ABS flow, if users open composer from Azure (clicking on "open in composer" from ABS), then they should net be prompted for auth at all. Can't we pass a auth token or something like that?

## Version
v1.4.0-nightly.228616.da1ac55

ADO issues
https://fuselabs.visualstudio.com/BotFramework/_queries/edit/58583/?triage=true
https://fuselabs.visualstudio.com/BotFramework/_queries/edit/58633/?triage=true

a-b-r-o-w-n:
## Description

- Caches the token required to fetch tenants so that multiple sign ins are not required.
- Fixes an issue provisioning and publishing with the web app

## Task Item

fixes #6595
fixes #6614


extensions/azurePublish/src/components/azureProvisionDialog.tsx
```diff
@@ -591,143 +604,118 @@ export const AzureProvisionDialog: React.FC = () => {
...
-            {subscriptionOption?.length > 0 && choice.key === 'create' && (
-              <form style={{ width: '100%' }}>
+          {subscriptionOption?.length > 0 && formData.choice === 'create' && (
+            <form style={{ width: '100%' }}>
+              {!userShouldProvideTokens() && (
                 <Dropdown
                   required
+                  ariaLabel={formatMessage(
+                    'Switching your Azure directory will change which subscriptions you can access'
+                  )}
```
beyackle, line 621: 👍

beyackle (changes requested):
One tiny thing, but otherwise this looks great.

beyackle (approved):

extensions/azurePublish/src/components/ResourceGroupPicker.tsx
```diff
@@ -169,7 +171,7 @@ export const ResourceGroupPicker = ({
         ariaLabel={formatMessage(
           'A resource group is a collection of resources that share the same lifecycle, permissions, and policies'
         )}
-        disabled={loading}
+        disabled={loading || disabled}
```
GeoffCoxMSFT, line 174: Nit: We should probably get rid of the loading part and let the caller control the disabled state entirely.  I assume this is because the caller needs to disable this during auth even if the list of names is populated?

a-b-r-o-w-n: This is to support editing an existing profile. At the moment, we are not allowing users to change the resource group.

extensions/azurePublish/src/components/azureProvisionDialog.tsx
```diff
@@ -50,6 +48,18 @@ import { ChooseResourcesList } from './ChooseResourcesList';
...
 
+type ProvisionFormData = {
+  choice: string;
+  tenantId: string;
+  subscriptionId: string;
+  resourceGroup: string;
+  hostname: string;
+  region: string;
+  luisLocation: string;
+  enabledResources: ResourcesItem[];
```
GeoffCoxMSFT, line 59: This looks a lot cleaner.  Nit: If we are going to clean this up, then I think we eventually need to have the list of possible resources (required and optional) outside this type and only have the list of selected resource IDs as form data.

extensions/azurePublish/src/components/azureProvisionDialog.tsx
```diff
@@ -211,6 +221,28 @@ const reviewCols: IColumn[] = [
...
 
+const getHostname = (config) => {
+  if (config?.hostname) {
+    return config.hostname;
+  } else if (config?.name) {
+    return config?.environment ? `${config.name}-${config.environment}` : config.name;
+  }
+};
+
+const getDefaultFormData = (currentProfile, defaults) => {
```
GeoffCoxMSFT, line 232: This really clarifies how the profile get overlaid on the extension state.  It was really confusing before.

extensions/azurePublish/src/components/azureProvisionDialog.tsx
```diff
@@ -211,6 +221,28 @@ const reviewCols: IColumn[] = [
...
+  return {
+    choice: defaults.choice ?? 'create',
+    tenantId: currentProfile?.tenantId ?? defaults.tenantId,
```
GeoffCoxMSFT, line 235: This is probably overthinking this, but I noticed that you clear the other fields when the tenantId changes.  Here, this chooses between the profile tenant and the default.  If we choose one or the other, would we want to use all the tenant-bound values from that choice?  (i.e. I choose the profile tenantId and so I want to choose the subscriptionId from the profile and not the default?)

extensions/azurePublish/src/components/azureProvisionDialog.tsx
```diff
@@ -506,48 +525,38 @@ export const AzureProvisionDialog: React.FC = () => {
...
+      updateFormData('hostname', newName);
       // debounce name check
       checkNameAvailability(newName);
     },
     [checkNameAvailability]
   );
 
-  const updateCurrentLocation = useMemo(
-    () => (_e, option?: IDropdownOption) => {
+  const updateCurrentLocation = useCallback(
```
GeoffCoxMSFT, line 540: Good catch.

extensions/azurePublish/src/components/azureProvisionDialog.tsx
```diff
@@ -579,9 +588,10 @@ export const AzureProvisionDialog: React.FC = () => {
 
       // set review list
       const requireList = result.filter((item) => item.required);
-      setRequireResources(requireList);
       const optionalList = result.filter((item) => !item.required);
-      setEnabledResources(optionalList);
+      updateFormData('requiredResources', requireList);
```
GeoffCoxMSFT, line 592: The list of required and optional resources are more like the list of subscriptions - they are what is possible to have selected.  Maybe they shouldn't be part of the form data like I said up top, but it can be done later.

extensions/azurePublish/src/components/azureProvisionDialog.tsx
```diff
@@ -659,140 +666,108 @@ export const AzureProvisionDialog: React.FC = () => {
...
+                placeholder={formatMessage('Select one')}
+                selectedKey={formData.subscriptionId}
+                styles={{ root: { paddingBottom: '8px' } }}
+                onChange={(_e, o) => {
+                  updateFormData('subscriptionId', o.key as string);
+                }}
+                onRenderLabel={onRenderLabel}
+              />
+              <ResourceGroupPicker
+                disabled={currentConfig?.resourceGroup}
```
GeoffCoxMSFT, line 709: I don't understand this value for disable - if there isn't already a resource group selected then the group picker is disabled?

extensions/azurePublish/src/components/azureProvisionDialog.tsx
```diff
@@ -211,6 +221,28 @@ const reviewCols: IColumn[] = [
...
+  if (config?.hostname) {
+    return config.hostname;
+  } else if (config?.name) {
+    return config?.environment ? `${config.name}-${config.environment}` : config.name;
+  }
+};
+
+const getDefaultFormData = (currentProfile, defaults) => {
+  return {
+    choice: defaults.choice ?? 'create',
```
GeoffCoxMSFT, line 234: Nit: A more descriptive name than choice would help here.  I can find a better name in a later PR.

Composer/packages/client/src/plugins/api.ts
```diff
@@ -35,7 +35,7 @@ interface PublishAPI {
   getName?: () => string;
   savePublishConfig?: (config: PublishConfig) => void;
   getTokenFromCache?: () => { accessToken: string; graphToken: string };
-  isGetTokenFromUser?: () => boolean;
+  userShouldProvideTokens?: () => boolean;
```
GeoffCoxMSFT, line 38: @benbrown mentioned that other extensions outside Composer exist (I'm not sure). I don't know if this name change would mean a breaking contract change for them.

GeoffCoxMSFT (approved):

Composer/packages/client/src/plugins/api.ts
a-b-r-o-w-n: This is a good point. What I'll do is add the old method back, but tag it as deprecated.

extensions/azurePublish/src/components/azureProvisionDialog.tsx
a-b-r-o-w-n: I'd like to defer this until later.

extensions/azurePublish/src/components/azureProvisionDialog.tsx
a-b-r-o-w-n: I think this could actually just be `tenantId: currentProfile?.tenantId` because we won't ever set the tenant id with the defaults.

extensions/azurePublish/src/components/azureProvisionDialog.tsx
a-b-r-o-w-n: I was super confused as well.

Composer/packages/electron-server/__tests__/setupTests.js
```diff
@@ -1,4 +1,5 @@
-process.env.DEBUG = 'composer*';
```
a-b-r-o-w-n, line 1: @tonyanziano I removed this. Was it there for a good reason? Just trying to cut down on the noise in the test output!

GeoffCoxMSFT (approved):

cwhitten (approved):

PR #6637, opened 2021-03-31
fix: Fixed minor typo in the Composer Project ReadMe file:

filipelteixeira:
createCosmosDB to createCosmosDb

## Description

Found a small typo in the project generated readme file, that prevented the Cosmos DB resource to not be created, if used as described. The value is case sensitive, and is written as createCosmosDB instead of createCosmosDb.

I was provisioning the Azure infrastructure for an article I wrote about MBF and found out the Cosmos DB resource was still being created although I used the correct described value set as false.
After looking into the script code, I realized that the value had a typo, and recreated with the correct parameter name. It worked. :)

## Task Item

Didn't found an issue related to it. #minor


a-b-r-o-w-n (approved):
Thanks for the contribution!

cwhitten:
@filipelteixeira there is now a conflict on this change. Can you address it?

filipelteixeira:
@cwhitten reviewed and addressed:
- Removed the file on the merge overlap;
- Added the minor fix to the file in the new structure.

cwhitten:
@filipelteixeira I don't think all the component file renames should be in this diff. 

Screen Shot 2021-04-07 at 4 32 31 PM

filipelteixeira:
> @filipelteixeira I don't think all the component file renames should be in this diff.
> 
> Screen Shot 2021-04-07 at 4 32 31 PM

I'll close this pull request and remake the change onto the new merged main

PR #6641, opened 2021-04-01
feat: Goto definition for lu editor

### LINKED ISSUE #6658: Support LU go to definiton
1. go to a imported lu file
```
    [Import Lu file](./a.lu)
```
Once user press ctrl + F12 on this line, the editor should navigate to a.lu file
2. go to a imported intent
```
    [Import Lu file](./a.lu#TextResponse)
```
Once user press ctrl + F12 on this line, the editor should navigate to the line of intent `TextResponse` in a.lu file

cosmicshuai:
closes #6658
1. Navigation to Lu file
luNaviSingleLangFile
2. Navigation to an intent
luNaviSingleLangIntent
3. Navigation in Form Dialog
luNaviSingleLangForm

Multi-Language Navigations:
luNaviMultiLangFile
luNaviMultiLangIntent
luNaviMultiLangForm
luNaviInContextMenu


Composer/packages/lib/code-editor/src/LuEditor.tsx
```diff
@@ -169,6 +174,13 @@ const LuEditor: React.FC<LULSPEditorProps> = (props) => {
         webSocket,
         onConnection: (connection: MessageConnection) => {
           const languageClient = createLanguageClient(formatMessage('LU Language Client'), ['lu'], connection);
+          languageClient.onReady().then(() =>
```
hatpick, line 177: dont repeat yourself, put this logic in a function and reuse it in L213

Composer/packages/tools/language-servers/language-understanding/src/LUServer.ts
```diff
@@ -101,6 +107,36 @@ export class LUServer {
     this.connection.listen();
   }
 
+  protected definitionHandler(params: TextDocumentPositionParams): Location | undefined {
```
hatpick, line 110: this is not in parity with what Lg does, in Lg you also had the lineNumber, is there any reason you dont have it here?

cosmicshuai: In LU editor, the goto definition only needs to handle the navigation of import another LU file or another intent from other files. For example, we only need to extract the target file and section as the lines below.
```
[import add item](./additem.en-us.lu#TextInput_Response_282825)
[import delete item](./deleteitem.en-us.lu)
```
So there is no goto between different lines in the same file.
Also the signature of onNavigateToLuPage is `(luFileId: string, luSectionId?: string | undefined) => void)`, that's why I did not use lineNumber here.

Composer/packages/lib/code-editor/src/LuEditor.tsx
cosmicshuai: simplified this function

Composer/packages/tools/language-servers/language-understanding/src/LUServer.ts
hatpick: I was under the impression that you'll have go to definition for entity as well.

hatpick:
From the gifs you included above, it seems like the user should hold down either cmd key on mac, and ctrl key on windows and then click an import and it would jump to the imported file. Does it also support right click and "go to"? the first interaction is kind of hidden and is more for a pro user vs. regular users, it's not very discoverable.

cosmicshuai:
> From the gifs you included above, it seems like the user should hold down either cmd key on mac, and ctrl key on windows and then click an import and it would jump to the imported file. Does it also support right click and "go to"? the first interaction is kind of hidden and is more for a pro user vs. regular users, it's not very discoverable.

@hatpick 
I was also wondering that, the monaco editor right click menu was somehow disabled. I thought it is designed. If the right click menu is enabled again. User can select 'Goto Definition', and it should work.
I will find out how to reenable the monaco editor right click menu.

hatpick:
> > From the gifs you included above, it seems like the user should hold down either cmd key on mac, and ctrl key on windows and then click an import and it would jump to the imported file. Does it also support right click and "go to"? the first interaction is kind of hidden and is more for a pro user vs. regular users, it's not very discoverable.
> 
> @hatpick
> I was also wondering that, the monaco editor right click menu was somehow disabled.
> The contextmenu is set to false here:
> https://github.com/microsoft/BotFramework-Composer/blob/0e35f8dc5d5418473c37a5de54ea55e5c5193271/Composer/packages/lib/code-editor/src/LuEditor.tsx#L126
> 
> 
> I thought it is designed. If the right click menu is enabled again. User can select 'Goto Definition', and it should work.

Please enable it again, I remember we disabled it as a part of a prototype for a labeler. Is that right @tdurnford ?

tdurnford:
@hatpick @cosmicshuai Yes, we disabled the context menu because we were using cmd + click to show the label; however, that is no longer the case, so enabling it should be fine.

cosmicshuai:
@hatpick @tdurnford 
I re-enabled the context menu and updated the PR description. Could you take a look?
BTW, do you have any clue that, the completion in LU editing is no longer available?
I looked into the network during editing. The initialized capability was right. But all completion requests has been canceled.
image

hatpick:
@cosmicshuai It has something to do with this PR:
https://github.com/microsoft/BotFramework-Composer/pull/5853

Can you follow up with the auhtor?

hatpick (approved):

cosmicshuai:
@hatpick Thanks for the fix!

PR #6645, opened 2021-04-01
feat: new validation pipeline - schema existence validation

### LINKED ISSUE #5996: Advanced diagnostics - Action Policy, SDK schema validation
## Describe the bug

Extend diagnostic data scope to support
1. Action Policy `$policies` should be validated and generates diagnostics data
2. Should validate existence of $kind in sdk.schema. When $kind 'X' was created based on schema v1, after updating to schema v2 where 'X' was removed, there should be an error to alert the potential runtime error in Debug Panel.
    (especially after we supporting install / uninstall an SDK package) (leveraging the `disabled` property, runtime won't break)

## Browser

- [ ] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

yeze322:
## Description

When a $kind is not defined in sdk.schema, raise a diagnostic error in debug panel and disable this Action.

## Task Item
closes #5996 
refs #6526 

## Screenshots
image

image
diagnostics-schema-dsiabled


cwhitten (approved):

Composer/packages/lib/indexers/src/validations/schemaValidation/schemaUtils.ts
```diff
@@ -0,0 +1,93 @@
...
+export const discoverNestedPaths = (data: BaseSchema, schema: JSONSchema7): string[] => {
+  if (isTrigger(schema)) return triggerNesterProperties;
+  if (!schema.properties) return [];
+
+  const nestedPaths: string[] = [];
+
+  const entries = Object.entries(schema.properties);
+  for (const entry of entries) {
+    const [propertyName, propertyDef] = entry;
+    const { type, items } = propertyDef as any;
```
yeze322, line 31: here causes the bug

yeze322 (commented):
bug to be triaged with new creation flow

Composer/packages/lib/indexers/src/validations/schemaValidation/schemaUtils.ts
yeze322: needs to check the schema definition

PR #6660, opened 2021-04-02
feat: homepage feed json

zhixzhan:
## Task Item
#minor


srinaath:
@zhixzhan  lets wait on @garypretty  to review this content before we merge. So Tuesday is probably when we are looking to land this.

boydc2014 (approved):

boydc2014:
> @zhixzhan  lets wait on @garypretty  to review this content before we merge. So Tuesday is probably when we are looking to land this.

Sounds good, i think it is provided by @garypretty , do you want to confirm?

cwhitten:
I'd like to understand the following before we take this:

1. Is this feed fetched every time the home page loads? If so that seems inefficient.
2. Do we know what type of latency we can anticipate when we load this from github? Have we considered publishing the resource to a CDN?
3. What error cases are considered in the implementation? Malformed data, network errors, etc.

boydc2014:
> I'd like to understand the following before we take this:
> 
> 1. Is this feed fetched every time the home page loads? If so that seems inefficient.
> 2. Do we know what type of latency we can anticipate when we load this from github? Have we considered publishing the resource to a CDN?
> 3. What error cases are considered in the implementation? Malformed data, network errors, etc.

For 1, i think the current implementation loads it every time Composer loads, not every time home page load. Is this correct @zhixzhan? 
For 2, we haven't measured much about the latency, manual testing of the home page PR shows it's quite good, it could be there is less load. srinaath helped investigated the CDN usage, the recommendation is it's a bit of over kill given the effort to onboard, @srinaath can you elaborate more on this?  
For 3, the feed defines a certain widgets (news, docs) in the home pages, when any error happens, the widget will be empty. The core functionality like creating\open, will still be functional.

zhixzhan:
> 1. Is this feed fetched every time the home page loads? If so that seems inefficient.

Hi @cwhitten @boydc2014  the feed will be fetched at every composer loads.

PR #6662, opened 2021-04-02
feature: Added provision resource checks before publish

### LINKED ISSUE #6627: [3/29 bug bash] Publishing profile error not clear/progress message suggestion
From Jordan Berger https://fuselabs.visualstudio.com/BotFramework/_workitems/edit/58566/?triage=true
(6 more identical bugs filed for this, see the linked items in the bug above if needed)

Repro steps:

1. Create a new bot from "Open with composer"
2. Immediately try to publish it

It wasn't clear to me that I had to provision a web app within my publishing profile when first trying to publish my bot. The error shown says "Required field `name` or `hostname` is missing from publishing profile", but if I go to my publishing profile, the name field is filled in. The "Publishing 1 bots" notification also just stays active instead of going away; I had to go into the Publish tab to see the error. I think it could be clearer to state that the web app is missing and reset the "Publishing" notification to show an error. The messages in the publishing tab (Build succeeded, uploading zip file, etc.) could be displayed on the ongoing notification to better show progress.

image

image

GeoffCoxMSFT:
## Description

Before publishing, the publishing profile needs to be checked to verify that the provisioning occurred at least once successfully and that all required resources have values from successful provisioning.

To avoid blocking the user unnecessarily , the checks verify the minimum values exist and do not verify values that could be empty for some deployments.

## Task Item

fixes #6627

## Screenshots

N/A - verification errors are displayed through the standard notification and result messages


GeoffCoxMSFT:
@cwhitten - I think this should fix the ABS 'Open in composer' scenario when trying to publish.

extensions/azurePublish/src/node/index.ts
```diff
@@ -673,6 +703,52 @@ export default async (composer: IExtensionRegistration): Promise<void> => {
       return recommendedResources;
     };
 
+    private isProfileProvisioned = (profile: PublishConfig) => {
```
hatpick, line 706: to make sure this function returns boolean, you can add a boolean return type and fix linter complains 😄

extensions/azurePublish/src/node/index.ts
```diff
@@ -673,6 +703,52 @@ export default async (composer: IExtensionRegistration): Promise<void> => {
...
+      //TODO: Post-migration we can check for profile?.tenantId
+      return profile?.resourceGroup && profile?.subscriptionId && profile?.region;
+    };
+
+    // While the provisioning process may return more information for various resources than is checked here,
+    // this tries to verify the minimum settings are present and that cannot be empty strings.
+    private isResourceProvisionedInProfile = (resource: ResourcesItem, profile: PublishConfig): boolean => {
```
hatpick, line 713: same here

extensions/azurePublish/src/node/index.ts
```diff
@@ -505,6 +506,37 @@ export default async (composer: IExtensionRegistration): Promise<void> => {
...
+              'There was a problem publishing {projectName}/{profileName}. The profile has not been provisioned yet.',
+              { projectName: project.name, profileName }
+            )
+          );
+        }
+
+        // verify the publish profile has the required resources configured
+        const resources = await this.getResources(project, user);
+        const missingResourceNames = [];
+        resources
```
hatpick, line 522: nit: reduce

hatpick (dismissed):

extensions/azurePublish/src/node/index.ts
```diff
@@ -673,6 +703,52 @@ export default async (composer: IExtensionRegistration): Promise<void> => {
...
+          // qnaRegion is not checked (empty may be a valid value)
+          return profile?.settings?.qna?.subscriptionKey;
+        case AzureResourceTypes.SERVICE_PLAN:
+          // no settings exist to verify the service plan was created
+          return true;
+        case AzureResourceTypes.AZUREFUNCTIONS:
+        case AzureResourceTypes.WEBAPP:
+          return profile?.hostname;
+        default:
+          throw new Error(`Azure resource type ${resource.key} is not handled.`);
```
hatpick, line 748: formatMessage?

hatpick (approved):

PR #6667, opened 2021-04-02
fix: Orchestrator worker crashes if accessing getLabelResolversAsync concurrently

### LINKED ISSUE #6666: Orchestrator: Warmup can crash when trying to access getLabelResolverAsync concurrently
## Describe the bug
The existing code uses a `promise.all` to access `getLabelResolverAsync`, and can cause the Orchestrator worker thread to crash. This is occasionally seen when using Orchestrator with a previously trained multilanguage bot with multiple blu files.

## Version
main 

## Browser
All

## OS
All

taicchoumsft:
## Description
Remove `promise.all` when calling `getLabelResolversAsync`, it occasionally crashes the Orchestrator worker thread.

Exceptions were also not forwarded correctly, when exceptions occurred during build or warmup of Orchestrator, the process never ends since we never send a message back to the main thread. It's also not possible to send the error object in raw form, the process boundary prevents this.  So extract the `message` and `stack` property only and forward those.

## Task Item
fixes #6666


Composer/packages/server/src/models/bot/process/orchestratorWorker.ts
```diff
@@ -86,26 +86,36 @@ export async function warmUpCache(generatedFolderPath: string, projectId: string
...
-
-      return modelData.model && snapshotData.length
-        ? await Orchestrator.getLabelResolversAsync(modelData.model, '', new Map(snapshotData), false)
-        : new Map<string, LabelResolver>();
-    })
-  );
-
-  cache.set(projectId, new Map([...enMap, ...multilangMap]));
-
+      if (labelResolverMap?.entries) {
```
hatpick, line 111: ```if(labelResolverMap)``` should be enough

Composer/packages/server/src/models/bot/process/orchestratorWorker.ts
```diff
@@ -179,14 +189,10 @@ const handleMessage = async (msg: RequestMsg) => {
       }
     }
   } catch (error) {
-    return { id: msg.id, error };
+    process.send?.({ id: msg.id, error: { message: error?.message, stack: error?.stack } });
```
hatpick, line 192: shouldn't error be empty if there are no error?
```error: error ? { message: error.message, stack: error.stack } : undefined```

taicchoumsft: Thanks, addressed

Composer/packages/server/src/models/bot/process/orchestratorWorker.ts
taicchoumsft: Addressed

hatpick (approved):

PR #6672, opened 2021-04-04
fix: Improve reliability of Orchestrator downloader

taicchoumsft:
## Description
The Orchestrator language model downloader did not properly catch exceptions in `Orchestrator.baseModelGetAsync`. This would lead to the download notification spinning endlessly if the language model download fails in some cases, though it can be dismissed.  

This PR propagates the error messages by forwarding errors with >400 status codes.  This will be caught by the axios promise rejection in the frontend, and a notification failure with the error message will be shown if the download fails.

## Task Item
#minor


Composer/packages/types/src/orchestrator.ts
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+export type IOrchestratorModelRequest = {
```
hatpick, line 4: nit: please remove the I prefix, that's the convention for defining interfaces in other language such as C#, Java, etc. not JS

Composer/packages/types/src/orchestrator.ts
```diff
@@ -0,0 +1,7 @@
...
+  type: 'en_intent' | 'multilingual_intent';
```
hatpick, line 5: nit: type is a reserved keyword, I suggest using ```somethingType``` or ```kind```

Composer/packages/client/src/recoilModel/dispatchers/orchestrator.ts
```diff
@@ -2,60 +2,69 @@
...
+  const resp = await httpClient.post(addr, { modelData: modelRequest });
+
+  // Model has been downloaded before
+  if (resp.status === 201) {
+    return true;
+  }
+
+  notificationStartCallback();
+
+  return await pollUntilDone(async () => (await httpClient.get(resp.data)).data !== 'DOWNLOADING', 5000);
```
hatpick, line 49: shouldn't this be DownloadState.DOWNLOADING?

Composer/packages/client/src/recoilModel/dispatchers/orchestrator.ts
```diff
@@ -64,36 +73,46 @@ export const availableLanguageModels = (recognizerFiles: RecognizerFile[]) => {
...
+
+      if (isUsingOrchestrator) {
+        // Download Model Notification
+        const { addNotification, deleteNotification } = await snapshot.getPromise(dispatcherState);
+        const notification = createNotification(orchestratorDownloadNotificationProps());
+        const botSettings = await snapshot.getPromise(settingsState(projectId));
+
+        try {
+          for (const languageModel of availableLanguageModels(recognizers, botSettings)) {
+            await downloadModel('/orchestrator/download', languageModel, () => {
```
hatpick, line 102: shouldn't you add the notification first, then update it when needed?

hatpick: I'm confused as which notification Line 103 is updating

Composer/packages/server/src/controllers/orchestrator.ts
```diff
@@ -36,28 +39,38 @@ async function getModelList(): Promise<IOrchestratorNLRList> {
 
 let state: DownloadState = DownloadState.STOPPED;
 
-interface DefaultModelRequest {
-  language: 'en' | 'multilang';
-}
-
-function isDefaultModelRequest(arg: any): arg is DefaultModelRequest {
-  return arg !== undefined;
+function isValidModelRequest(arg: any): arg is IOrchestratorModelRequest {
```
hatpick, line 42: use the type instead of any

Composer/packages/server/src/controllers/orchestrator.ts
```diff
@@ -14,6 +15,8 @@ enum DownloadState {
   DOWNLOADING,
 }
 
+let errorMsg: any;
```
hatpick, line 18: why is this variable defined in the global context? I suggest defining it in a closure context.

hatpick (changes requested):

Composer/packages/server/src/controllers/orchestrator.ts
taicchoumsft: This typeguard is there since arg could indeed be anything coming from client.

Composer/packages/types/src/orchestrator.ts
taicchoumsft: Thanks @hatpick , I was using the convention that others used here. Is `interface` ok?

hatpick: with the current type, change ```IOrchestratorModelRequest``` to ```OrchestratorModelRequest``` (removing the prefix I)

Composer/packages/types/src/orchestrator.ts
hatpick: with the current type, change IOrchestratorModelRequest to OrchestratorModelRequest (removing the prefix I)

Composer/packages/client/src/recoilModel/dispatchers/orchestrator.ts
```diff
@@ -2,60 +2,69 @@
...
-        reject(false);
-        return;
-      }
-
-      notificationStartCallback();
-
-      const statusUri = resp.data;
+import { createNotification, updateNotificationInternal } from './notification';
 
+const pollUntilDone = (predicate: () => Promise<boolean>, pollingIntervalMs: number) =>
```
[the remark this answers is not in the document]
taicchoumsft: Typescript does not seem to like async keyword in a type, it requires an implementation..  [Stackoverflow link discussing this](https://stackoverflow.com/questions/38744159/in-typescript-how-to-define-type-of-async-function).

Composer/packages/types/src/orchestrator.ts
hatpick: I meant the key you use here, ```type```, in strict mode it will complain about it but I guess we are not in that mode. It's still a good practice if you replace it with somethingType or kind:
```type: 'en_intent' | 'multilingual_intent';```

Composer/packages/client/src/recoilModel/dispatchers/orchestrator.ts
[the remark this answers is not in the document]
hatpick: I guess it "infer"s the type because of Promise return value.

Composer/packages/types/src/orchestrator.ts
taicchoumsft: Ah ok understood thanks, will change to `kind`.

taicchoumsft: Changed to `kind` here, enclosed the `orchestrator` controller in a class, added all other requested changes.  Thanks @hatpick

Composer/packages/server/src/controllers/orchestrator.ts
taicchoumsft: wrapped the controller in a class to address this

Composer/packages/client/src/recoilModel/dispatchers/orchestrator.ts
taicchoumsft: Thanks @hatpick , I've changed the names so it's clearer - the update happens on the `downloadNotification`.

hatpick (approved):

hatpick (approved):

PR #6678, opened 2021-04-04
feature: Using runtimeKey for webapp vs. functions

### LINKED ISSUE #6527: Publishing/provisioning: enforce choice of web app vs functions based on runtime
Historically, the runtime has been able to be deployed to either Functions or a webapp.

This changes with the new runtime, which is locked to one or the other from creation time.

During provisioning, there SHOULD NOT BE A CHOICE - we should simplify this selection to "publish to Azure" and then add the appropriate hosting resource to the required list.

GeoffCoxMSFT:
## Description

This uses the project's runtime key to determine if the provisioning should require the Azure web app or Azure function resource.

Also includes required workaround for manual tokens in non-electron for the Azure provisioning dialog flow.

## Task Item

closes #6527

## Screenshots

One a single item now appears for publishing to Azure
image


Composer/packages/lib/shared/src/skillsUtils/index.ts
```diff
@@ -101,5 +101,48 @@ export const isSkillHostUpdateRequired = (skillHostEndpoint?: string) => {
   return !skillHostEndpoint || isLocalhostUrl(skillHostEndpoint);
 };
 
+/**
+ * Returns true if the runtime key is adaptive; false otherwise.
+ * Example adaptive runtime keys: 'adaptive-runtime-dotnet-webapp' or 'adaptive-runtime-node-functions'
+ * Example older adaptive runtime keys: 'csharp-azurewebapp-v2'
```
benbrown, line 107: FWIW, this was used for a pretty brief period of time during the early stages of the runtimes. Only people running preview bits would have  projects with the `csharp-azurewebapp-v2` value.

benbrown (dismissed):
Before we merge this, we need to update the migration function that rewrites the settings to modify existing publishing profiles to use this new type.

hatpick (approved):

PR #6684, opened 2021-04-05
fix: ARM token for provisioning is now passed via body.

tonyanziano:
## Description

Some users were seeing a 431 HTTP response when attempting to provision Azure resources.

This applies the same fix from https://github.com/microsoft/BotFramework-Composer/pull/5700 to the provisioning endpoint.

## Task Item

#minor

### DIFF
--- a/Composer/packages/server/src/controllers/provision.ts
+++ b/Composer/packages/server/src/controllers/provision.ts
@@ -33,8 +33,6 @@ export const ProvisionController = {
     }
   },
   provision: async (req: Request, res) => {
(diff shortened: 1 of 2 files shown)


Composer/packages/server/src/controllers/provision.ts
```diff
@@ -48,7 +46,7 @@ export const ProvisionController = {
           // call the method
           const result = await pluginMethod.call(
             null,
-            { ...req.body, accessToken },
+            { ...req.body },
```
tonyanziano, line 49: `req.body.accessToken` now contains the ARM token

cwhitten (approved):

PR #6686, opened 2021-04-05
fix: detect "Get started" dependencies in all projects not just current project

### LINKED ISSUE #6649: BUG: When in a multibot project such as enterprise assistant, the get started feature is does not lock to the root bot, and also does not detect child bot dependencies.
As a result, the project will say it requires QNA, but not LUIS.

If you select one of the skills, it will switch to LUIS, but not QNA.

This should always be locked to the root bot, and include ALL child deps.

benbrown:
## Description

adjust the mechanism used to determine if a project needs LUIS or QNA to support multi-bot projects. This moves code from the header component into a recoil selector.

## Task Item

fixes #6649

### DIFF
--- a/Composer/packages/client/src/components/Header.tsx
+++ b/Composer/packages/client/src/components/Header.tsx
@@ -177,7 +177,7 @@ export const Header = () => {
   // ... if the get started widget moves, this code should too!
   const requiredStuff = useRecoilValue(allRequiredRecognizersSelector);
   const requiresLUIS = requiredStuff.some((p) => p.requiresLUIS);
-  const requiresQNA = requiredStuff.filter((p) => p.requiresQNA)?.length ? true : false;
+  const requiresQNA = requiredStuff.some((p) => p.requiresQNA);
   // ... end of get started stuff
 
   const isShow = useBotControllerBar();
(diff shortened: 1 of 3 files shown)


GeoffCoxMSFT (dismissed):

Composer/packages/client/src/components/Header.tsx
```diff
@@ -176,11 +175,9 @@ export const Header = () => {
   // These are needed to determine if the bot needs LUIS or QNA
   // this data is passed into the GetStarted widget
   // ... if the get started widget moves, this code should too!
-  const dialogs = useRecoilValue(dialogsWithLuProviderSelectorFamily(projectId));
-  const luFiles = useRecoilValue(luFilesSelectorFamily(projectId));
-  const qnaFiles = useRecoilValue(qnaFilesSelectorFamily(projectId));
-  const requiresLUIS = BotIndexer.shouldUseLuis(dialogs, luFiles);
-  const requiresQNA = BotIndexer.shouldUseQnA(dialogs, qnaFiles);
+  const requiredStuff = useRecoilValue(allRequiredRecognizersSelector);
+  const requiresLUIS = requiredStuff.filter((p) => p.requiresLUIS)?.length ? true : false;
```
a-b-r-o-w-n, line 179:
```suggestion
  const requiresLUIS = requiredStuff.some((p) => p.requiresLUIS);
```

Composer/packages/client/src/components/Header.tsx
```diff
@@ -176,11 +175,9 @@ export const Header = () => {
...
+  const requiresQNA = requiredStuff.filter((p) => p.requiresQNA)?.length ? true : false;
```
a-b-r-o-w-n, line 180:
```suggestion
  const requiresQNA = requiredStuff.some((p) => p.requiresQNA);
```

a-b-r-o-w-n (approved):

a-b-r-o-w-n (approved):

PR #6687, opened 2021-04-05
fix: use iconUrl field instead of icon. 

benbrown:
## Description

This change prepares package manager to use the iconUrl field instead of the icon field currently returned by the dialog merge process.

(The icon field is a relative path to the icon, rather than a web accessible URL which is ALSO available in the package)

## Task Item

This fixes https://github.com/microsoft/botframework-components/issues/802


chrimc62:
iconUrl is deprecated from the nuspec docs--you should be using icon instead.

Issue #6047, opened 2021-03-01
Redirect to Power Virtual Agents Topic from Composer Dialog


iMicknl:
## Is your feature request related to a problem? Please describe.

Currently you can use Bot Framework Composer to build your more advanced dialogs for Power Virtual Agents. In many scenario's, Power Virtual Agents contains topics where you would like to redirect to. For example if you have an Adaptive Card with multiple options, to redirect back to a topic modelled in PVA.

## Describe the solution you'd like

A way to utilize `Dialog Management -> Begin a new dialog` to redirect to a topic in PVA.

## Describe alternatives you've considered

Open for suggestions. Modelling all PVA topics/responses in Composer is not a feasible solution.

## Additional context

Composer is used to overcome limitations of PVA, for example the use of HeroCards or Adaptive Cards. I would like to only model a few steps in a dialog in Composer, where the majority is still handled by PVA.

cwhitten:
hello @iMicknl - this is on our short-term roadmap as new functionality as part of the integration. You should see this soon in an upcoming release

cwhitten:
Will ship with the next release. Closing

PR #6691, opened 2021-04-05
feat: allow users to select PVA topics when beginning new dialog

a-b-r-o-w-n:
## Description

- Includes PVA topics in the select dialog dropdown
- Instead of navigating in app, opens the topic in PVA
- Shows display name of dialog instead of dialog id in visual node

## Task Item

refs #6047 

## Screenshots
- TODO


Composer/packages/adaptive-flow/src/adaptive-flow-renderer/widgets/DialogRef.tsx
```diff
@@ -3,23 +3,61 @@
...
+import { NodeRendererContext } from '../../adaptive-flow-editor/contexts/NodeRendererContext';
 
 export interface DialogRefCardProps extends WidgetContainerProps {
   dialog: string | object;
 }
 
-export const DialogRef: WidgetComponent<DialogRefCardProps> = ({ id, onEvent, dialog }) => {
+export const DialogRef: WidgetComponent<DialogRefCardProps> = ({ id, onEvent, dialog, data }) => {
   const { ElementWrapper } = useContext(RendererContext);
+  const { dialogs, topics } = useContext(NodeRendererContext);
```
a-b-r-o-w-n, line 25: @yeze322 Can you take a look at this and let me know your thoughts?

I wanted to do 2 things:
1. Use the dialog's display name instead of its id
2. Differentiate from dialogs and topics

This is your domain so I want to make sure that this approach is fine, or work with you to do something better.

Composer/packages/lib/indexers/src/dialogIndexer.ts
```diff
@@ -223,6 +223,7 @@ function index(files: FileInfo[], botName: string): DialogInfo[] {
           const isRoot = file.relativePath.includes('/') === false; // root dialog should be in root path
           const dialog: DialogInfo = {
             isRoot,
+            isTopic: file.relativePath.startsWith('topics/'),
```
tonyanziano, line 226: I think the answer is yes, but just confirming: We are fine with not indexing the topic if the user decides to muck around with the folder structure right?

a-b-r-o-w-n: We should engage the PVA team on that. My assumption is that the folder structure is necessary when publishing back to PVA, but I don't know for sure.

Composer/packages/server/src/externalContentProvider/powerVirtualAgentsProvider.ts
```diff
@@ -158,7 +160,8 @@ export class PowerVirtualAgentsProvider extends ExternalContentProvider<PowerVir
 
   private getContentUrl(): string {
     const { envId, baseUrl, botId } = this.metadata;
-    return `${baseUrl || getBaseUrl()}/environments/${envId}/bots/${botId}/composer/content`;
+    const query = process.env.COMPOSER_PVA_TOPICS === 'true' ? '?includeTopics=true' : '';
```
tonyanziano, line 163: Where does this `COMPOSER_PVA_TOPICS` variable get set? Are we going to turn this on at some point in production or is there somewhere we set it that I'm not seeing?

a-b-r-o-w-n: It's not set by default at the moment. I figure we'd default it to true whenever the PVA changes make it to prod.

tonyanziano: Sounds good

tonyanziano (commented):
LGTM but going to wait until Ze responds to your comment before approving.

PR #6697, opened 2021-04-05
chore: Adapters: copy the 'type' passed in the adapter config to fully support adapter schemas

carlosscastro:
#minor


beyackle (dismissed):
Makes sense.

a-b-r-o-w-n:
@carlosscastro looks like a test needs to be updated.

PR #6719, opened 2021-04-06
fix: apply publishing profile to new runtime during publish

### LINKED ISSUE #6683: Publishing: the new runtime uses a different form of appsettings.json file.  The publishing tools will need to be updated to map the publishing profile to this new format.

benbrown:
## Description

The new adaptive runtime settings are in a different form than the legacy runtime. This requires the form of the information in the publishing profile to be adjusted slightly.

This PR introduces a function in the @bfc/shared library called applyPublishingProfileToSettings that takes a DialogSettings type and a PublishProfile type and merges them correctly, returning the "ready to run" settings.

## Task Item

Fixes #6683


Composer/packages/lib/shared/src/settings.ts
```diff
@@ -0,0 +1,66 @@
...
+ * Apply a publishing profile to the base settings, resulting in a "ready to run" version of the settings
+ * for the new adaptive runtime. This function transforms the profile settings format to match the new
+ * Adaptive Runtime settings form.
+ * @param settings
+ * @param profile
+ * @returns
+ */
+export const applyPublishingProfileToSettings = (settings: DialogSetting, profile: PublishProfile): DialogSetting => {
+  // ensure the runtimeSettings key exists
+  settings.runtimeSettings = {
```
GeoffCoxMSFT, line 15: Nit: You probably don't need to spread here.  settings.runtimeSettings = settings.runtimeSettings || {};

Composer/packages/types/src/publish.ts
```diff
@@ -170,6 +170,10 @@ export type PublishProfile = {
       endpoint: string;
       region: string;
     };
+    qna?: {
```
GeoffCoxMSFT, line 173: To match: extensions\azurePublish\src\node\schema.ts

I think we need to add:
tenantId
language
luisResource
botName
resourceGroup
subscriptionId
region
settings/qna/endpoint

extensions/azurePublish/yarn.lock
```diff
@@ -1308,6 +1308,7 @@
     "@emotion/core" "^10.0.27"
     "@emotion/styled" "^10.0.27"
     office-ui-fabric-react "7.71.0"
+    react-markdown "^5.0.3"
```
GeoffCoxMSFT, line 1311: I don't think we need react-markdown.  I didn't end up using it in the new copy on the UI as the designs wanted some precise positioning.

benbrown: I think this is appearing because react-markdown is part of the handoff instructions widget.

Composer/packages/lib/shared/src/settings.ts
```diff
@@ -0,0 +1,64 @@
...
+ * Apply a publishing profile to the base settings, resulting in a "ready to run" version of the settings
+ * for the new adaptive runtime. This function transforms the profile settings format to match the new
+ * Adaptive Runtime settings form.
+ * @param settings
+ * @param profile
+ * @returns
+ */
+export const applyPublishingProfileToSettings = (settings: DialogSetting, profile: PublishProfile): DialogSetting => {
+  // ensure the runtimeSettings key exists
+  settings.runtimeSettings = settings.runtimeSettings || {}
```
GeoffCoxMSFT, line 15: Nit: Missing ; and end of line.

GeoffCoxMSFT (approved):

GeoffCoxMSFT (approved):

PR #6729, opened 2021-04-06
fix: during publish, do not copy project into a build folder

### LINKED ISSUE #6531: Bot publishing fails to include ProjectReferenced DLLs in Adaptive Runtime
## Describe the bug

The bot publishing pipeline copies the csproj into a private directory in AppData/Roaming before running dotnet publish.

The problem with this approach is that if the csproj file has a ProjectReference with relative path, the dotnet publish would not be able to pick up the ProjectReference since these references are not copied over. Thus the deployed app would be missing dependencies on production (despite it running perfectly fine locally).

In effect, this would mean any bot developer who wants to include a Project Reference in their csproj would not be able to deploy their application even if <Private>true</Private> is set to true.

NOTE - this doesn't impact csproj/botproj where the dependency is in the form of <PackageReference> that is because dotnet publish can resolve that from the source regardless of its location in the filesystem.

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Update in the Adaptive Runtime csproj to include <ProjectReference> to another project
2. Publish it
3. Call the bot from Emulator
4. Bot fails to be called because the service keeps returning 500 Internal Server Error

## Expected behavior

Bot starts and works.

benbrown:
## Description

This PR *removes* the step that copies the runtime and project files into a temporary build folder, and instead performs the build and publish directly in the folder.

- [x] Do not copy the files
- [x] Ensure that, for JS apps, we do not leave a modified appsettings.json post-publish
- [x] Ensure that manifest files, etc are in the correct location.

## Task Item

Fixes #6531


extensions/azurePublish/src/node/deploy.ts
```diff
@@ -157,12 +170,15 @@ export class BotProjectDeploy {
           .glob('**/*', {
             cwd: source,
             dot: true,
-            ignore: ['**/code.zip'], // , 'node_modules/**/*'
+            ignore: ['**/code.zip', '**/settings/appsettings.json'],
```
benbrown, line 173: this is where the magic happens as we exclude appsettings from the zip...

extensions/azurePublish/src/node/deploy.ts
```diff
@@ -157,12 +170,15 @@ export class BotProjectDeploy {
...
           })
           .on('error', (err) => reject(err))
           .pipe(stream);
 
         stream.on('close', () => resolve());
+
+        // write the merged settings to the deploy artifact
+        archive.append(JSON.stringify(settings, null, 2), { name: 'settings/appsettings.json' });
```
benbrown, line 181: and then write it directly into the zip here!

extensions/azurePublish/src/node/index.ts
```diff
@@ -177,84 +150,6 @@ export default async (composer: IExtensionRegistration): Promise<void> => {
...
     /* These methods implement the publish actions */
     /*******************************************************************************************************************************/
-    /**
-     * Prepare a bot to be built and deployed by copying the runtime and declarative assets into a temporary folder
-     * @param project
-     * @param settings
-     * @param srcTemplate
-     * @param resourcekey
-     */
-    private init = async (project: any, srcTemplate: string, resourcekey: string, runtime: any) => {
```
benbrown, line 187: This section of code is where we set up the temporary folders for handling non-ejected bots.  BYE!

extensions/runtimes/src/index.ts
```diff
@@ -316,13 +316,6 @@ export default async (composer: any): Promise<void> => {
       }
       throw new Error(`Runtime already exists at ${destPath}`);
     },
-    setSkillManifest: async (
```
benbrown, line 319: all of this work is now done as a standard part of the azure deploy

extensions/azurePublish/src/node/deploy.ts
```diff
@@ -114,13 +116,24 @@ export class BotProjectDeploy {
       // this returns a pathToArtifacts where the deployable version lives.
       const pathToArtifacts = await this.runtime.buildDeploy(this.projPath, project, settings, profileName);
 
+      // COPY MANIFESTS TO wwwroot/manifests
```
cwhitten, line 119: @luhan2017 FYI for the skill manifest scenarios you are working on

cwhitten (approved):

cwhitten (approved):

PR #6733, opened 2021-04-07
fix: file selector bug

### LINKED ISSUE #6225: bot is created in wrong folder
## Describe the bug

I went through new creation flow and the bot is put in the wrong folder.

## Version

Nightly
image

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Create bot project
2. Select empty bot
3. Type into location c:\scratch
4. Click on OK button
Bot will not be put in C:\scratch, it will be created wherever the folder was when it opened.

## Expected behavior

I think that if move focus OUT of the folder box before I hit OK it will work correctly, so I think the issue is that the folder 
is not saved before the OK is executed

liweitian:
## Description
root cause: target path is not updated if it is focusing on the input field.

## Task Item
Closes #6225 

## Screenshots
dddd

filepath


Composer/packages/client/src/components/CreationFlow/FileSelector.tsx
```diff
@@ -480,6 +480,13 @@ export const FileSelector: React.FC<FileSelectorProps> = (props) => {
       setCurrentPath(value as string);
     }
   };
+
+  const updatePathPending = (option?: IComboBoxOption, index?: number, value?: string) => {
+    if (!option && value) {
+      setCurrentPath(value as string);
```
hatpick, line 486: I think here is a good opportunity to test the path and if it's invalid show an error or something.

hatpick (changes requested):
I think here is a good opportunity to test the path and if it's invalid show an error or something, and prevent moving forward

liweitian:
> I think here is a good opportunity to test the path and if it's invalid show an error or something, and prevent moving forward

path validation added

hatpick (approved):

PR #6752, opened 2021-04-07
feature: Provisioning partial success handling

### LINKED ISSUE #5959: Partial provision failure & publish profile update
If error hit during provision process, user will get the notification but the publish profile will not be updated.

- Click edit to update existing publish profile
- Go through the wizard to create new resource group or webapp service
- hit errors during provision process
- publish profile will not be updated (subscription, resource group, hostname info etc...)
- once user click the edit , user need to provide all the info again

GeoffCoxMSFT:
## Description

When provisioning if one resource fails, this returns partial success and puts up a warning notification.
The profile is also saved on partial success.
The user will be able to open the profile target and try provisioning again.

## Task Item

closes #5959

## Screenshots
Manually introduced error to show this partial success (uses the standard warning card for the notification)
image

Prod7 profile was saved
image

### DIFF
--- a/Composer/packages/client/src/recoilModel/dispatchers/provision.ts
+++ b/Composer/packages/client/src/recoilModel/dispatchers/provision.ts
@@ -27,6 +27,13 @@ export const provisionDispatcher = () => {
       type: 'success',
     };
   };
+  const getProvisionPartialSuccessNotification = (value: string): CardProps => {
+    return {
+      title: formatMessage('Provision partially completed'),
+      description: formatMessage('{msg}', { msg: value }),
+      type: 'warning',
+    };
+  };
   const getProvisionFailureNotification = (value: string): CardProps => {
     return {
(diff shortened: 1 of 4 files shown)


Composer/packages/client/src/recoilModel/dispatchers/provision.ts
```diff
@@ -115,7 +122,11 @@ export const provisionDispatcher = () => {
         isCleanTimer = false;
       try {
         const response = await httpClient.get(`/provision/${projectId}/status/${targetType}/${targetName}/${jobId}`);
-        if (response.data?.status === 200 && response.data.config && response.data.config != {}) {
+        if (
+          (response.data?.status === 200 || response.data?.status === 206) &&
+          response.data.config &&
+          response.data.config != {}
```
hatpick, line 128: nit: use lodash isEmpty({})

hatpick (approved):

PR #6756, opened 2021-04-07
fix: fixed CosmosDb parameter name in Readme

filipelteixeira:
createCosmosDB to createCosmosDb

## Description

Found a small typo in the project generated readme file, that prevented the Cosmos DB resource to not be created, if used as described. The value is case sensitive, and is written as createCosmosDB instead of createCosmosDb.

I was provisioning the Azure infrastructure for an article I wrote about MBF and found out the Cosmos DB resource was still being created although I used the correct described value set as false.
After looking into the script code, I realized that the value had a typo, and recreated with the correct parameter name. It worked. :)

## Task Item

Didn't found an issue related to it. #minor


filipelteixeira:
@cwhitten added new PR change here to avoid merge problems

cwhitten:
Thank you for the contribution!

PR #6782, opened 2021-04-08
feat: Added selected state to Web Chat log panel and auto-inspection on load

### LINKED ISSUE #6749: Additional Web Chat Integration P1 / P2's
This is a continuation of #6511 to represent the remaining P1 and P2 work items to be done for the Web Chat integration.

### P1's

- [x] Implement a selected state indicator for the log message currently being shown in the inspector pane
- [x] Auto-inspect the latest message in the log pane if there is no message currently inspected

### P2's

- [x] Extend debug panel outside of the authoring canvas -- it should be at the same level as the project tree
- [ ] Make the Web Chat panel float above / respect the debug panel
- [ ] ~~Implement status bar (Needs investigation)~~

tonyanziano:
## Description

This PR adds a selected state to the currently inspected web chat log item, and it also auto-inspects the latest log item if the panel is active and no item is currently being inspected.

## Task Item

Fixes #6749 

## Screenshots

**Selected log item**

selected-log

**Later log item is currently being inspected, but hovering an earlier log item**

selected-log-hovered-second-item


Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WebChatLog/WebChatLogContent.tsx
```diff
@@ -59,9 +64,34 @@ export const WebChatLogContent: React.FC<DebugPanelTabHeaderProps> = ({ isActive
...
+          setWebChatInspectionData(currentProjectId, { item: trafficItem });
+        }
+      }
+    }, 500)
+  ).current;
+
+  const inspectLatestLogMessage = () => {
+    // inspect latest log message if nothing is being inspected
+    if (!inspectionData && currentProjectId) {
+      const sortedTraffic = [...rawWebChatTraffic].sort((t1, t2) => t1.timestamp - t2.timestamp);
```
srinaath, line 83: Is there a reason why the items coming in through the socket are not sorted by default @tonyanziano ?

Composer/packages/client/src/recoilModel/dispatchers/webchat.ts
```diff
@@ -4,14 +4,23 @@
...
 import { ConversationTrafficItem } from '@botframework-composer/types';
 import { useRecoilCallback, CallbackInterface } from 'recoil';
+import { v4 as uuid } from 'uuid';
 
 import { webChatTrafficState, webChatInspectionDataState, isWebChatPanelVisibleState } from '../atoms';
 import { WebChatInspectionData } from '../types';
 
+const addIdToTrafficItem = (item: ConversationTrafficItem) => {
+  if (!item.id) {
+    item.id = uuid();
```
srinaath, line 14: nit: are we opposed to just using the timestamp which is a number as the id?

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WebChatLog/WebChatLogContent.tsx
tonyanziano: Good call @srinaath , let me add the sorting logic to the dispatcher that puts the items in the state

Composer/packages/client/src/recoilModel/dispatchers/webchat.ts
tonyanziano: On occasion, I observed scenarios where several items had the same timestamp

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WebChatLog/WebChatLogContent.tsx
```diff
@@ -42,6 +43,10 @@ const logPane = css`
   box-sizing: border-box;
 `;
 
+const itemIsSelected = (item: ConversationTrafficItem, currentInspectionData: WebChatInspectionData | undefined) => {
```
hatpick, line 46: nit: currentInspectionData?: WebChatInspectionData

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WebChatLog/WebChatLogContent.tsx
```diff
@@ -80,16 +110,37 @@ export const WebChatLogContent: React.FC<DebugPanelTabHeaderProps> = ({ isActive
   );
 
   const renderLogItem = useCallback(
-    (item: ConversationTrafficItem, index: number) => {
+    (item: ConversationTrafficItem, index: number, inspectionData: WebChatInspectionData | undefined) => {
```
hatpick, line 113: same nit :D

Composer/packages/client/src/recoilModel/dispatchers/webchat.ts
hatpick: good call, why is it optional tho?

tonyanziano: On the server the traffic items are instantiated in one spot, and then stamped in the web socket server on their way to the client with whatever "traffic type" they have. I basically made it optional so I could just add it onto the object later without worrying about it at the time of instantiation.

These items are also instantiated in the client at certain points. So i could make it mandatory, but I would have to add the GUID generating code to each of these places. This seemed easier to manage.

hatpick: I'd call it something else, ```id``` is usually non optional 😄

srinaath (approved):

Composer/packages/client/src/recoilModel/dispatchers/webchat.ts
tonyanziano: All the other potential property names I can think of that still preserve the meaning are basically just synonyms of `id`. So I'm just going to make the property mandatory and initialize to empty string on the server.

hatpick (approved):

Issue #6750, opened 2021-04-07
Output pane, style issues


mewa1024:
## Describe the bug

Minor style issues in the Output pane:
1. change name from "Outputs" to "Output"
2. use same font as for all the code/Monaco editors
3. reduce space around output text

## Version

Version:  1.4.0-nightly.231667.d104286
Electron: 8.2.4
Chrome: 80.0.3987.165
NodeJS: 12.13.0
V8: 8.0.426.27-electron.0

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [x] macOS
- [ ] Windows
- [ ] Ubuntu

## To Reproduce

1. Start a bot. 
2. Go to the Output pane.

The font is too large; it should be same as in the Code view. And the space around the text is too large. It should be 16px on the left and right and 8px on the top and bottom:

![image](https://user-images.githubusercontent.com/39317542/113934020-1e172580-97aa-11eb-8f24-cac1a32a347d.png)

## Expected behavior

Figma spec is here: https://www.figma.com/file/HwYEQoCmWbNrIDdCOmGvzN/bot-testing-BRAINS-storm?node-id=1087%3A20841

(To see the distance between objects in Figma, select the first object and hold Opt or Alt key and hover the second object):
![image](https://user-images.githubusercontent.com/39317542/113935532-86b2d200-97ab-11eb-92ec-7be5e3ac804c.png)

PR #6786, opened 2021-04-08
feat: Use websocket to continuously surface runtime log into the UI

### LINKED ISSUE #6511: Support activity and bot state inspection in the embedded Web Chat control
- [x] Capture message / trace / event activities (P0)
- [x] Capture network activity (P0)
- [x] Show selected item in inspector (P0)
- [ ] Fix layout so that Web Chat panel **does not** overlap with inspector pane (P2)
- [x] Expose runtime messages in "Output" tab (P0)
- [ ]~~Add different inspector experience for LUIS traces (P1)~~
- [ ] Selected inspected state in log -- link log item to item shown in inspector pane (P1)
- [ ] On initial load, select most recent activity (P1)

srinaath:
## Description

Instead of polling constantly, using a websocket to deliver the stdout and stderr from a started runtime.

Output Window padding fixes.

## Task Item

Fixes #6511 #6750


Composer/packages/client/src/components/BotRuntimeController/ErrorCallout.tsx
```diff
@@ -81,6 +78,8 @@ fieldsWhiteList.set('code', { visible: false, name: 'code' });
 fieldsWhiteList.set('signal', { visible: false, name: 'signal' });
 fieldsWhiteList.set('stderr', { visible: false, name: 'stderr' });
 
+const DEFAULT_FONT_SETTINGS = getDefaultFontSettings();
```
hatpick, line 81: why is this using the "default" font settings instead font settings?

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/RuntimeOutputLog/RuntimeOutputLog.tsx
```diff
@@ -21,15 +33,73 @@ const createMarkup = (txt: string) => {
...
+  useEffect(() => {
+    const setupLogConnection = async () => {
+      try {
+        const runtimeStreamUrl = await httpClient.get(`/publish/runtimeLogUrl/${projectId}`);
+
+        runtimeTrafficChannel.current = new WebSocket(runtimeStreamUrl.data);
+
+        if (runtimeTrafficChannel.current) {
+          runtimeTrafficChannel.current.onmessage = (event) => {
+            try {
```
hatpick, line 59: why do you need this try within the outer try, the catch block is doing the same thing for both

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/RuntimeOutputLog/RuntimeOutputLog.tsx
```diff
@@ -38,33 +108,31 @@ export const RuntimeOutputLog: React.FC<{ projectId: string }> = ({ projectId })
...
+            wordBreak: 'break-all',
+            whiteSpace: 'pre-wrap',
+            lineHeight: '20px',
+          }}
+          // eslint-disable-next-line react/no-danger
+          dangerouslySetInnerHTML={createMarkup(runtimeData.standardOutput)}
+          data-testid="runtime-standard-output"
+        />
+      )}
+      {botBuildErrors && <ErrorCallout error={botBuildErrors} />}
```
hatpick, line 134: does botBuildErrors ever has no value?

extensions/localPublish/src/WebSocketServer.ts
```diff
@@ -0,0 +1,83 @@
+// Copyright (c) Microsoft Corporation.
```
hatpick, line 1: nit: javascript convention for classes is camelCase naming for the file (in oppose to Java and c# which is PascalCase)

extensions/localPublish/src/WebSocketServer.ts
```diff
@@ -0,0 +1,83 @@
...
+import portfinder from 'portfinder';
+import express, { Request, Response } from 'express';
+import { Server as WSServer } from 'ws';
+
+interface WebSocket {
+  close(): void;
+  send(data: string, cb?: (err?: Error) => void): void;
+}
+
+export class WebSocketServer {
```
hatpick, line 15: I'd give this a better name to reflect on what the purpose of the web socket server is.

extensions/localPublish/src/WebSocketServer.ts
```diff
@@ -0,0 +1,83 @@
...
+          this.servers[projectId] = wsServer;
+        }
+      });
+      this.port = port;
+      return this.port;
+    }
+  }
+
+  public static sendRuntimeLogToSubscribers(projectId: string, standardOutput: string, standardError: string): void {
+    this.sockets[projectId]?.send(
```
hatpick, line 76: dont you wanna know if the projectId is invalid and someone is trying to send a message on its socket?

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/RuntimeOutputLog/RuntimeOutputLog.tsx
srinaath: It has a value only if there are build errors. For ex LUIS key invalid, publish to LUIS failed.

extensions/localPublish/src/WebSocketServer.ts
srinaath: Yep. Changing it to runtimeLogServer

extensions/localPublish/src/WebSocketServer.ts
srinaath: If there is a subscriber for the projectID it should exist. If there is no subscriber there is actually no client (composer UI) for me to send this runtime log for a project.

Composer/packages/client/src/components/BotRuntimeController/ErrorCallout.tsx
srinaath: Removed it. Its using the font of the parent.

extensions/localPublish/package.json
```diff
@@ -13,20 +13,25 @@
     "@botframework-composer/types": "file:../../Composer/packages/types",
     "adm-zip": "^0.4.14",
     "archiver": "^5.0.2",
+    "express": "^4.17.1",
+    "express-serve-static-core": "^0.1.1",
```
tonyanziano, line 17: is this `...-serve-static-core` package needed? I'm not seeing it used anywhere in the diff

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/RuntimeOutputLog/RuntimeOutputLog.tsx
```diff
@@ -21,15 +33,69 @@ const createMarkup = (txt: string) => {
...
+
+    if (!runtimeTrafficChannel.current) {
+      setupLogConnection();
+    }
+
+    return () => {
+      runtimeTrafficChannel.current?.close();
+      runtimeTrafficChannel.current = null;
+    };
+  }, []);
```
tonyanziano, line 98: should this function list `projectId` as a dependency? What happens if you open a second project? Will it have only opened the first, initial socket for project A even though you are now using project B?

extensions/localPublish/src/WebSocketServer.ts
hatpick: I do understand the logic of how it works, but the ```?.``` accessor is showing that there might be a projectId value that leads to an undefined socket. That's a bad state that shouldn't happen but when it happens you definitely wanna know about it to fix it perhaps.

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/RuntimeOutputLog/RuntimeOutputLog.tsx
hatpick: the default value in the recoil state shows an object, meaning it always has a value, when you reset it falls back to that default value right?

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/RuntimeOutputLog/RuntimeOutputLog.tsx
srinaath: So, i would have added if the component destroy wasn't called on project change. Essentially, when we change project the component is destroyed and the socket as well.

extensions/localPublish/package.json
srinaath: Nice catch. I had ported it from server but things work fine without it.

tonyanziano (approved):

PR #6793, opened 2021-04-08
fix: clean up if project creation fails

### LINKED ISSUE #6726: If the Creation process fails, we're not cleaning up properly
## Describe the bug
I'm using a nightly build.  so what's nice about this is that there is some fragility in the Creation flow.  As a result of that, failure cases during creation are easier to produce than when we'll be working against release candidates and GA release builds.

When the creation process fails, we don't clean up the new folder.  So, when you try and create the bot again, the name is taken, though the bot was never actually created during the first attempt.

## Version
Version:  1.4.0-nightly.231667.d104286
Electron: 8.2.4
Chrome: 80.0.3987.165
NodeJS: 12.13.0
V8: 8.0.426.27-electron.0

## Browser
- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS
- [x] macOS
- [ ] Windows
- [ ] Ubuntu

## To Reproduce

1. Using the above nightly, launch composer.  This nightly has the new creation flow enabled by default.
2. From the Composer Home Screen, click on '+' to create a new bot
3. Select a template from the list of templates.  I chose 'Basic Conversational Bot'
4. Click 'Next' to display the "Create a bot project" modal.
5. Name you bot with a legal name.  e.g. basic_x
6. Wait for the NPM error dialog to pop.  
Screen Shot 2021-04-06 at 2 22 43 PM
7.  Dismiss the error dialog
8. Repeat steps 2, 2, 4, & 5 from above
9. Note that in step 5 when you try to name your bot 'basic_x', you'll be informed that bot name is in use

if you open a terminal window and browser the folder where you have Composer create new bots you'll see an empty folder named 'basic_x'.  The existence of this folder prevents the name reuse.

## Expected behavior
Given that the bot creation failed, and in fact failed early on, it would be nice that the failure case here actually cleans up.

benbrown:
## Description

This removes the newly created folder if the creation process fails

## Task Item

Fixes #6726

### DIFF
--- a/Composer/packages/server/src/services/project.ts
+++ b/Composer/packages/server/src/services/project.ts
@@ -562,6 +562,10 @@ export class BotProjectService {
         throw new Error('Could not find root bot');
       }
     } catch (err) {
+      // Clean up failed projects
+      log('Cleaning up failed project at ', locationRef.path);
+      const storage = StorageService.getStorageClient(locationRef.storageId, user);
+      await storage.rmrfDir(locationRef.path);
       BackgroundProcessManager.updateProcess(jobId, 500, err instanceof Error ? err.message : err, err);
       TelemetryService.trackEvent('CreateNewBotProjectCompleted', { template: templateId, status: 500 });
     }


hatpick (approved):

Composer/packages/server/src/services/project.ts
```diff
@@ -562,6 +562,10 @@ export class BotProjectService {
         throw new Error('Could not find root bot');
```
hatpick, line 562: what about here? (asking because eI have no knowledge)

benbrown: yeah, if the root bot can't be found at the end of the creation process, something bad has happened.

PR #6800, opened 2021-04-08
fix:  Add new/existing skill flow

### LINKED ISSUE #6783: Add new skill and add exisitng skill flows are broken

pavolum:
## Description

Add new skills flow was broken due to lingering versions of the old template fetching call in the addSkill modal. It has been replaced with updated fetchTemplate version.

Add existing skill broken due to not passing runtimeLang and runtimeEnv down to generators as well as param passing bug where an object was passed in place of individual attributes. Both of which are fixed in this PR

## Task Item

fixes #6783


Composer/packages/client/src/pages/design/creationModal.tsx
```diff
@@ -66,7 +65,7 @@ export const CreationModal: React.FC<CreationModalProps> = (props) => {
 
   const fetchResources = async () => {
     await fetchStorages();
-    fetchTemplates();
+    fetchTemplatesV2();
```
srinaath, line 68: the v2 is because we haven't removed the legacy code yet correct?

srinaath (approved):

Composer/packages/client/src/pages/design/creationModal.tsx
pavolum: yes correct. Right now we are prioritizing clean up the scenarios that will be tested by the larger audience during next weeks bug bash. Its on our backlog to clean up and remove loads of legacy code prior to release.

PR #6849, opened 2021-04-09
fix: update bf-dialog to latest rc, adjust syntax

### LINKED ISSUE #6845: Update version of bf-dialog used to the latest.
Includes package manager, server-workers, server.

benbrown:
## Description

* Update bf-dialog library to latest.
* Update syntax of call to SchemaMerger to pass full path to generated folder rather than relative path.

This should address some issues reported with Package Manager failing after the bot has started.

## Task Item

Fixes #6845


extensions/packageManager/src/node/index.ts
```diff
@@ -468,7 +468,7 @@ export default async (composer: IExtensionRegistration): Promise<void> => {
 
           // call do a dry run on the dialog merge
           const merger = new SchemaMerger(
-            [manifestFile, '!**/imported/**', '!**/generated/**'],
+            [manifestFile, `!${path.join(currentProject.dir, 'generated')}/**`],
```
a-b-r-o-w-n, line 471: It seems like this could benefit from extracting a method.

a-b-r-o-w-n (approved):

PR #6858, opened 2021-04-12
fix: the format of hover info in LG editor

cosmicshuai:
#minor

change the display info of hover more close to the original content:
image


Composer/packages/tools/language-servers/language-generation/src/LGServer.ts
```diff
@@ -30,6 +30,7 @@ import isEqual from 'lodash/isEqual';
 import { filterTemplateDiagnostics, isValid, lgUtil } from '@bfc/indexers';
 import { MemoryResolver, ResolverResource, LgFile } from '@bfc/shared';
 import { buildInFunctionsMap } from '@bfc/built-in-functions';
+import { LgTemplate } from '@botframework-composer/types';
```
hatpick, line 33: add this dependency to package.json

Composer/packages/tools/language-servers/language-generation/src/LGServer.ts
```diff
@@ -367,6 +373,18 @@ export class LGServer {
...
 
+  private buildHoverTemplateInfo(template: LgTemplate) {
+    let templateName = '';
+    if (template.parameters.length > 0) {
+      templateName = `\\# ${template.name}(${template.parameters.join(', ')})`;
+    } else {
+      templateName = `\\# ${template.name}`;
+    }
+
+    const templateBody = template.body.replace(/-/g, '\\-').replace(/\r?\n/g, '\n\n');
```
hatpick, line 384: this should only replace all the non-escaped ```-```

hatpick: also only replace new lines at the end of each line

hatpick (changes requested):

Composer/packages/tools/language-servers/language-generation/src/LGServer.ts
cosmicshuai: added

Composer/packages/tools/language-servers/language-generation/src/LGServer.ts
cosmicshuai: I found I only need to keep the original content. So I use a code block to warp the template info here.

hatpick:
@cosmicshuai you have broken tests, can you fix them?

cosmicshuai:
@hatpick I just finished the broken tests.

hatpick (approved):

PR #6863, opened 2021-04-12
fix: QnA files multi-lang migration fix

zhixzhan:
## Description

Fix QnA files multi-lang migration failed, and some regression introduced by lazy indexing 

## Task Item

#minor


liweitian:
If we have two languages, add qna pairs in dialogName.en-us.qna, will only generate dialogName-manual.source.en-us.qna. You can only generate dialogName-manual.source.qna, it will be copied to other languages on the second step of qna file migration.

zhixzhan:
> If we have two languages, add qna pairs in dialogName.en-us.qna, will only generate dialogName-manual.source.en-us.qna. 
You can only generate dialogName-manual.source.qna, it will be copied to other languages on the second step of qna file migration.

@liweitian  I think this is expected, because only  in `dialogName.en-us.qna` have content to migrate, in other qna file `dialogName.{locale}.qna` there is nothing to migrate.
 
> And every dialogName.{locale}.qna files need to import the dialogName-manual.source.{locale}.qna.

if there is not such file, there is no need to import it in every language.

boydc2014 (approved):

PR #6881, opened 2021-04-12
fix: adjust mechanism used to create qna host

benbrown:
## Description

The previous mechanism used to create the appservice config when provisioning a QNA maker was not successfully setting the configuration.

This PR changes the mechanism used, and verifiably sets these config options.

#minor

# The issue:

prior to this PR, the QNA provisioning process would create an app service.  However, the "application settings" necessary for that app service to perform its QNA features were not being set.  They were not appearing in the azure portal, and qna publishing related actions returned an error:

`Qnamaker build failed: No Endpoint keys found. If this is a new service, please wait a minimum of 10 minutes for the runtime to be ready.`

Using this version, the app service *IS* configured with these additional settings, and I am able to publish to QNA without receiving this error.


praful-dhone:
Please fix this issue is Composer.exe which is running on Windows as we are kind of stuck here.

cwhitten (approved):

PR #6911, opened 2021-04-13
fix: 6566 Update provision handoff copy

### LINKED ISSUE #6566: Create Copy for Share Resource Request Instructions

The Copy on the Share Resource Request Instructions needs to be determined and created

## Additional context

image

GeoffCoxMSFT:
## Description

Adjusted ProvisionHandoff to match figma
Updated provisioning instructions to match figma

## Task Item

fixes #6566 

## Screenshots

image
image


hatpick:
@JaskeeratBedi-cmk can we avoid using generic exclamation phrases such as "Okay" in the UX and instead use a more specific action verb that semantically matches the purpose of the button?

JaskeeratBedi-cmk:
@hatpick in general we try to have make the CTA verb first. However this modal is really an exception. 
In this modal the user must 
- copy the instruction
- paste it outside of the app ( to share with the azure admin)
- come back to Composer and dismiss this modal

While designing, we explored CTAs like 'Copy', but then user still has to dismiss after they have completed the action..
So ended up with either done or ok. And it seems like Ok is the standard we use when user needs to acknowledge.

hatpick:
> @hatpick in general we try to have make the CTA verb first. However this modal is really an exception.
> In this modal the user must
> 
> * copy the instruction
> * paste it outside of the app ( to share with the azure admin)
> * come back to Composer and dismiss this modal
> 
> While designing, we explored CTAs like 'Copy', but then user still has to dismiss after they have completed the action..
> So ended up with either done or ok. And it seems like Ok is the standard we use when user needs to acknowledge.

In this particular Dialog, it seems like "Okay" basically closes the dialog, unless it records some data indicating that user acknowledge reading and understanding the information above, I think a simple "Close" will do. What do you think?

benbrown (approved):

PR #6926, opened 2021-04-14
fix: homepage header layout

### LINKED ISSUE #6913: Home Screen layout on desktop monitors needs some refinement
## Describe the bug
We should consider adjusting our media queries to better layout the home page on external monitors.  See attached screenshots

## Version
Version:  1.4.0-nightly.234673.12e1a8a
Electron: 8.2.4
Chrome: 80.0.3987.165
NodeJS: 12.13.0
V8: 8.0.426.27-electron.0
## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS
- [x] macOS
- [ ] Windows
- [ ] Ubuntu

## To Reproduce

1. You'll need a system that has an external monitor.  The monitor I am testing against is a 27" external display
2. Launch Composer
3. Make the Composer application full screen

## Expected behavior
I would expect the Title to remain at the very top.
the Recent Bots, Resources, and Latest Visions to be left justified below the title
the What's new panel on the right to be width appropriate

## Screenshots
This is an example of the Home Screen displayed full screen on a 16" macBook

Screen Shot 2021-04-13 at 12 32 58 PM

This is an example of the Home Screen displayed full screen on a 27" external monitor
Screen_Shot_2021-04-13_at_12_34_13_PM

Here are some other media query breakpoints:
Screen Shot 2021-04-13 at 12 44 34 PM

zhixzhan:
## Description

fix homepage header layout, and image setting

## Task Item

close #6913 

## Screenshots

Large monitor

Apr-14-2021 10-01-40


Composer/packages/client/src/pages/home/CardWidget.tsx
```diff
@@ -74,10 +76,29 @@ export const CardWidget: React.FC<CardWidgetProps> = ({
...
     if (aspectRatio < 1.5) {
+      setOddImageSize(true);
+    }
+    caclImageContainerRect();
+  };
+
+  // detect image container dimention, if it's too wide, apply background.
+  // Why not always put a background behind image, because when image size is nearly to container size, we don't want see 1px or 5px black line.
+  const caclImageContainerRect = () => {
+    const containerRect = imageContainerEl.current?.getBoundingClientRect();
```
hatpick, line 87: calling this function will make browser do a reflow, now imagine calling it on every pixel resize change. It's expensive. Is there a better way for doing this?

hatpick (changes requested):

Composer/packages/client/src/pages/home/CardWidget.tsx
srinaath: @zhixzhan This should not a fix in Javascript. Inside a media query you need to do a display none or block on the image background.

srinaath (changes requested):

Composer/packages/client/src/pages/home/CardWidget.tsx
```diff
@@ -74,10 +76,29 @@ export const CardWidget: React.FC<CardWidgetProps> = ({
     const rect = e.currentTarget.getBoundingClientRect();
     const aspectRatio = rect.width / rect.height;
```
srinaath, line 77: @zhixzhan we can use the screen dimensions (media query) to set the width of the element for different breakpoints. I dont think we need to adjust the height.

Composer/packages/client/src/pages/home/styles.ts
```diff
@@ -234,19 +231,26 @@ export const mediaCardItem = {
   imageCover: css`
     ${cardItem.imageCover}
     width: 100%;
+    margin-bottom: 12px;
+    overflow: hidden;
+    display: flex;
     min-height: 95px;
     height: calc(17vw * 95 / 244);
```
srinaath, line 238: What is this calculation mean. How did you arrive at it?

Composer/packages/client/src/pages/home/styles.ts
```diff
@@ -234,19 +231,26 @@ export const mediaCardItem = {
...
     @media (max-width: 1416px) {
       height: calc(20vw * 95 / 244);
     }
-    margin-bottom: 12px;
-    overflow: hidden;
-    display: flex;
-    .image-cover-background {
+    @media (min-width: 2800px) {
       background: #323130;
```
srinaath, line 243: Can use a equivalent color from FluentThem.palette. If not from CommunicationColors, SharedColors

hatpick: NeutralColors.gray160

zhixzhan: thanks, resolved

Composer/packages/client/src/pages/home/styles.ts
zhixzhan: 17vw is image current width. it mean `ImageCurrentWidth / ImageCurrentHeight` should equal to origin aspect ratio `95 / 244`. I wrote this to implement the image can also increase height and keep ratio.

srinaath: Can u create variables for those constants and use them here.

srinaath (dismissed):

boydc2014 (approved):

Issue #6776, opened 2021-04-08
Creating a publishing profile is not resilient to authentication failures


sgellock:
## Describe the bug
If I create a new publishing profile where the target is Azure, and as part of that profile creation flow, I need to authenticate, and my authentication fails, the "Add a publishing profile" modal remains in a loading state, displaying the loading spinner.  

See screenshot section below

## Version
Version:  1.4.0-nightly.232669.4daad5c
Electron: 8.2.4
Chrome: 80.0.3987.165
NodeJS: 12.13.0
V8: 8.0.426.27-electron.0

## Browser
- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [x] macOS
- [ ] Windows
- [ ] Ubuntu

## To Reproduce

1. Ensure you are not logged into Azure.  If you are, completely log out, as this issue requires the publishing flow
2. Launch Composer, create a new bot.  I selected the Empty bot template, though I doesn't matter for the scenario.
2. Once the new bot has been created, notice the nice Rocket Ship menu suggesting I should create a publishing profile.
3. Click the Create a publishing profile suggestion in the Rocket Ship menu to navigate to the Publishing page
4. Click the "Add new" link to launch the "Add a publishing profile" modal and start the publishing profile creation flow.
5. Name your bot, and select a publishing target of Azure, click Next
6. Get a sign-in modal prompt.
7. Close the sign-in modal prompt without adding your credentials
8. You'll receive a second modal sign-in prompt.  Also, close this modal prompt without adding your credentials
9. With the sign-in prompts ignored, notice the Add a publishing profile modal is still loading.

## Expected behavior
Add a publishing profile should display a reasonable message informing me that Composer was not able to create a publishing profile and, if possible, the reason why.

## Screenshots
<img width="1000" alt="Screen Shot 2021-04-08 at 9 24 01 AM" src="https://user-images.githubusercontent.com/733559/114062267-317acd00-984c-11eb-9ebb-e50a4c7d2a85.png">

PR #6927, opened 2021-04-14
fix: Move auth to when provisioning needs it.

### LINKED ISSUE #6562: Change Create Publishing Profile to have logon after determined that you want to create resources in Azure

Currently, the create publishing profile process asks you to logon ahead of deciding if you want to import / create new / generate a request.  Many users wanting to import or generate a request will not necessarily have an Azure account to do this action.  We need to move the logon to be after you chose to create new resources.

## Additional context

image

### LINKED ISSUE #6900: When signed out of provisioning flow (Create Publishing Profile) still says sign out and won't ask me to sign on again
## Describe the bug

When creating a publishing profile, if you sign out then you still get a sign out prompt at the bottom of the screen.  Also, when you are trying to come back and create a resource it won't let you select anything or prompt for a logon again.

## Version

Version:  1.4.0-nightly.234518.8e98823
Electron: 8.2.4
Chrome: 80.0.3987.165
NodeJS: 12.13.0
V8: 8.0.426.27-electron.0

## Browser

- [X] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [X] Windows
- [ ] Ubuntu

## To Reproduce

1. Open or Create a Bot Project
2. Create a new Publishing Profile selecting the Create new resources flow
3. Create a second publishing profile, but on the Add Publishing profile page click on Sign out.  (This will send you back out of the flow)
4. Again add a second publishing profile and you will see this screen saying sign out, but you are signed out and you don't get prompted to logon again
image
5. If you go further and click next with the Create new resource option, you get the following screen that won't let you select anything
image

## Expected behavior

When you are logged out and you do something that requires you to be logged on, then it should prompt.  When I log out, it should not ask me to Sign out again.  Signing out probably shouldn't end the flow either, we should probably just log you out.

GeoffCoxMSFT:
## Description

Turns out that calling getTenants will pop the sign-in dialog when in electron.
I moved the getTenants call to only run when the page that has the dropdown of them loads.
I updated the error message to semantically be about the tenants and improved the error message strings.

## Task Item

fixes #6562
fixes #6900 
fixes: #6776

## Screenshots
 N/A (offline video available)


benbrown (approved):
I would recommend that we disable the fields on the form when we are doing auth and loading in the supporting values via API calls...

PR #6973, opened 2021-04-14
fix: adjust mechanism used to determine if/when luis and qna teaching bubbles should pop out

### LINKED ISSUE #6939: Bug: The teaching bubbas don't appear in the enterprise assistant (seems like it takes too long to load and the state is wrong?)

benbrown:
## Description

There were some strange behaviors in the get started experience when creating a skill due to the way Composer loads content and then progressively parses and indexes it.  

Essentially, since the projects take a while to fully load and mark all of their LU files as empty or not empty, the test for needing to use LUIS or QNA would change over time, causing the get started content to flicker and shift as it loaded.

This PR shifts the mechanism, no longer using the parsed state of the files, opting for a simpler test on the files that is available at test time.  Additionally, this will now wait for the project to be fully loaded before launching the get started experience.
 
## Task Item

Fixes #6939


Composer/packages/client/src/components/Header.tsx
```diff
@@ -163,6 +164,7 @@ export const Header = () => {
   const settings = useRecoilValue(settingsState(projectId));
   const schemas = useRecoilValue(schemasState(projectId));
   const isWebChatPanelVisible = useRecoilValue(isWebChatPanelVisibleState);
+  const everythingLoaded = useRecoilValue(botProjectSpaceLoadedState);
```
srinaath, line 167: maybe naming botProjectSolutionLoaded?

Composer/packages/client/src/recoilModel/selectors/project.ts
```diff
@@ -378,16 +379,37 @@ export const webChatEssentialsSelector = selectorFamily<WebChatEssentials, strin
...
+      .find(({ id }) => getBaseName(id) === luFileId)
+      ?.content.trim()
+      .replace(/^>.*$/g, '')
+      .trim() === ''
+  );
+};
+
 export const allRequiredRecognizersSelector = selector({
   key: 'allRequiredRecognizersSelector',
   get: ({ get }) => {
```
srinaath, line 400: This selector reevaluates for every change in the bot asset @benbrown . Is this GetStarted section required only on bot load first up and not required afterwards?

benbrown: Yes, it should reevaluate if they add LUIS later...

srinaath (approved):

Composer/packages/client/src/recoilModel/selectors/project.ts
srinaath: One thing to consider is reusing the logic we have in the DebugPanel to report the error for missing LUIS key, QNA Key. This functionality looks similar

Issue #6848, opened 2021-04-09
Click on "appsettings.json" in error message not taking me to LUIS field


xiyangdesign:
## Describe the bug

**Click on "app settings.json" not taking me to LUIS field**

<img width="1090" alt="image" src="https://user-images.githubusercontent.com/74994573/114241577-38820800-993e-11eb-932b-9bca68569267.png">

## Browser

- [ ] Electron distribution
- [ ] Chrome
- [X] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [X] macOS
- [ ] Windows
- [ ] Ubuntu

## To Reproduce

1. Go to 'Project settings page' (open a bot that requires a Luis key)
2. Click on 'appsettings.json'

## Expected behavior

Take me to the field to input Luis key

PR #6992, opened 2021-04-15
fix: Route to appsettings correctly

srinaath:
## Description

This PR routes to the correct tab in settings page when LUIS key or region missing error message is clicked

## Task Item

Refs #6848

CC @pavolum


Composer/packages/electron-server/src/main.ts
```diff
@@ -310,7 +310,6 @@ async function run() {
   app.on('before-quit', () => {
     const mainWindow = ElectronWindow.getInstance().browserWindow;
     mainWindow?.webContents.send('session-update', 'session-ended');
-    mainWindow?.webContents.send('conversation-cleanup');
```
cwhitten, line 313: what is this change for?

srinaath: This is actually a remnant from the previous webchat implementation to cleanup all the sockets when electron is closed. It is not required as i close the socket using the destructor in the useEffect (WebchatPanel.tsx). This can be safely removed.

cwhitten (approved):

Issue #6969, opened 2021-04-14
OnChooseIntent: AdaptiveCard template appears broken


taicchoumsft:
## Describe the bug
1) Set up a brand new bot with LUIS/QnA, or Orchestrator/QnA, in the root dialog.
2) Add a `OnChooseIntent` trigger (aka intent disambiguation trigger)
3) Try creating a test condition where this trigger will be hit

Instead of seeing the adaptive card with choices, we get an error:

![image](https://user-images.githubusercontent.com/61705609/114780220-f127bd00-9d2b-11eb-817c-3053166cb30b.png)

Snip of error message, looks like `$schema` property is missing a value:
```
"empty_1.en-us.lg:Unexpected character at Path , line 3, position 15 when parsing 
 {
 "$schema",
 "version": "1.0",
 "type": "AdaptiveCard",
 "speak": "",
 "body": [
 {
 "type": "TextBlock",
 "text": "I'm not sure which one you mean.",
 "weight": "Bolder"
 },
 {
 "type": "TextBlock",
 "text": "Can you pick one ?",
 "separator": "true"
 },
 {
 "type": "Input.ChoiceSet",
 "placeholder": "Placeholder text",
 "id": "userChosenIntent",
 "choices": [
 {
 "title": "Test",
 "value": "luisResult"
 },
 {
 "title": "See an answer from the Knowledge Base",
 "value": "qnaResult"
 },
 {
 "title": "None of the above",
 "value": "none"
 }
 ],
 "style": "expanded",
 "value": "luis"
 },
 {
 "type": "ActionSet",
 "actions": [
 {
 "type": "Action.Submit",
 "title": "Submit",
 "data": {
 "intent": "chooseIntentCardResponse"
 }
 }
 ]
 }
 ]
 }
```

## Version
Nightly 235171
## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

cwhitten:
@taicchoumsft, you're updating this template as part of the orchestrator disambiguation, right? can you include the fix for this in it?

taicchoumsft:
OK thanks @cwhitten ,  I'll take care of this as part of that effort

PR #7004, opened 2021-04-15
fix: Update OnChooseIntent template to support Orchestrator

### LINKED ISSUE #6431: Utterance is not passed to skill if disambiguation kicks in at parent
Filed for tracking (along with conversation)

Before I file it, I want to make sure I got it right.

Scenario:
•	Root bot with Orchestrator
•	Child bot with LUIS
•	Routing works correctly.
•	Without disambiguation child get’s utterance (turn.text?) and is able to trigger on it.
•	Issue: With disambiguation: I suspect that utterance is not passed to skill, so skill cannot trigger on recognition.

Failure case (with disambiguation):

pic1

Repeated without disambiguation:

pic2

taicchoumsft:
## Description
This is a tricky PR to test, so I'll annotate as best I can.  

Orchestrator introduces the ability to disambiguate between intents.  This means `CrossTrainedRecognizerSet (CTRS)` as a parent recognizer must now potentially handle more than 2 ambiguous results. 

We see more than 2 results when `OrchestratorRecognizer` sends its own nested `OnChooseIntent`. 

### Old Flow
1. Store top LUIS result in `dialog.luisResult`, top QnA result in `dialog.qnaResult`
2. If `dialog.luisResult` > 0.9, and `dialog.qnaResult` < 0.5, return LUIS result
3. If `dialog.qnaResult` > 0.9 and `dialog.luisResult` < 0.5, return QnA result
4. Display the two options

### New Flow
1. Allow user to tweak number of results in `turn.maxChoices`
1. Allow user to adjust result threshold in `turn.minThreshold`
1. Saves user utterance in `conversation.lastAmbiguousUtterance` for child dialogs
1. For each recognizer in `turn.recognized.candidates`:
             -  extract their candidates if any,
             -  flatten the list
             - remove all `DeferToQnA` intents 
             - sort results by score (descending)
             - take results above `turn.minThreshold`
             - take a maximum of `turn.maxChoice` elements
1. If there are no results left after the above step, return `Unknown` intent
1. If there is only one result, return it right away
1. If there are more than 1 result left, display up to `turn.maxChoice` elements

### Pros of New Flow
- Is more general, allows CTRS to deal with child recognizers like Orchestrator that also send `ChooseIntent`
- Allows users to pick max number of results to display
- Allows CTRS to deal with more than 2 recognizers in the future

### Cons of New Flow
- Cannot normalize scores between recognizers: i.e. all recognizers assumed to convey confidence at roughly the same scale, since we sort on the returned list

### Areas the new flow does not improve on the old flow
- When adaptive card is displayed, user has to click the choice to return the right response, or type a number, or an error occurs. In the old flow, if the user doesn't return `luisResult` or `qnaResult`, an error occurs.
- when deleting the `OnChooseIntent` trigger, both new and old flow do not clean up all related LG resources in `common.en-us.lg`.  Not sure if this is an oversight or desired behavior.
  
## Task Item
closes #6431, #6969.  Also closes [#737](https://github.com/microsoft/botframework-components/issues/737) in Components.

## Screenshots
image

Multiple intents
image
Multiple intents with QnA
image


cwhitten:
@taicchoumsft what LG resources are being left behind when deleting the template? It seems like an oversight that we should address

Composer/packages/lib/shared/src/constant.ts
```diff
@@ -352,7 +352,7 @@ export const QnALocales = [
 const adaptiveCardJsonBody =
```
ryanlengel, line 352: We should use more specialized names for all of the LG templates, these are pretty generic and can look confusing in a larger bot.

ryanlengel: Even if they just had something about disambiguation prepending it...like:
- OnChooseIntent_PickOne
- OnChooseIntent_GetAnswerReadBack
- etc.

ryanlengel (changes requested):

Composer/packages/lib/shared/src/constant.ts
ryanlengel: And AdaptiveCardJson should be something like ChooseIntentCard

taicchoumsft:
> @taicchoumsft what LG resources are being left behind when deleting the template? It seems like an oversight that we should address



> @taicchoumsft what LG resources are being left behind when deleting the template? It seems like an oversight that we should address

@cwhitten , these resources get left behind in `common.en-us.lg` when deleting the trigger.  
```
# AdaptiveCardJson()
-```
{
      "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
      "version": "1.0",
      "type": "AdaptiveCard",
      "speak": "",
      "body": [
          {
              "type": "TextBlock",
              "text": "${whichOneDidYouMean()}",
              "weight": "Bolder"
          },
          {
              "type": "TextBlock",
              "text": "${pickOne()}",
              "separator": "true"
          },
          {
              "type": "Input.ChoiceSet",
              "placeholder": "Placeholder text",
              "id": "userChosenIntent",
              "choices": [
                           {
                               "title": "${getIntentReadBack()}",
                               "value": "luisResult"
                           },
                           {
                               "title": "${getAnswerReadBack()}",
                               "value": "qnaResult"
                           },
                           {
                               "title": "None of the above",
                               "value": "none"
                           }
             ],
             "style": "expanded",
             "value": "luis"
         },
         {
             "type": "ActionSet",
             "actions": [
                {
                     "type": "Action.Submit",
                     "title": "Submit",
                     "data": {
                   "intent": "chooseIntentCardResponse"
                }
         }
       ]
     }
    ]
}```
# whichOneDidYouMean()
- I'm not sure which one you mean.
- Hmmm, I find that to be ambiguous.

# pickOne()
- Can you pick one ?
- Can you help clarify by choosing one ?

# getAnswerReadBack()
- See an answer from the Knowledge Base

# getIntentReadBack()
- SWITCH : ${toLower(dialog.luisResult.intent)}
- CASE : ${'GetUserProfile'}
  - Start filling in your profile(GetUserProfile intent)
- DEFAULT :
  - ${dialog.luisResult.intent}
 ```

**Comment by taicchoumsft** [comment]
> > @taicchoumsft what LG resources are being left behind when deleting the template? It seems like an oversight that we should address

@cwhitten , looks like LG resources are not cleaned up at all, I've located the bug, will fix it with this PR.   To repro, create a sendReponse and delete it, and the LG resource for that action still remains on disk.

For OnChooseIntent, I've switched to using LG parameters to display the choice list, and this appears to not be handled. I'll also fix this

File: Composer/packages/lib/shared/src/constant.ts (hunk 1 above)
Comment (taicchoumsft) [reply Composer/packages/lib/shared/src/constant.ts]: Thanks for the suggestion @ryanlengel , the parent template has been renamed to `ChooseIntent`, and now roughly looks like this. `designerId` guid is autogenerated to help us track certain things, but can be removed if hardcoding to a template. The `attachment` extension is to signify to users all these lg contents are part of the attachment, and helps Composer clean up resources.

```json

# ChooseIntent_cbxtWX_attachment_card(candidates)
<actual json file for card>

# ChooseIntent_cbxtWX_attachment_whichOneDidYouMean()
- I'm not sure which one you mean.
- Hmmm, I find that to be ambiguous.

# ChooseIntent_cbxtWX_attachment_pickOne()
- Can you pick one ?
- Can you help clarify by choosing one ?

# ChooseIntent_cbxtWX_attachment_getAnswerReadBack()
- See an answer from the Knowledge Base

# ChooseIntent_cbxtWX_attachment_getIntentReadBack(intent)
- SWITCH : ${intent}
- CASE: ${'QnAMatch'}
    - ${ChooseIntent_cbxtWX_attachment_getAnswerReadBack()}
- CASE : ${'GetUserProfile'}
  - Start filling in your profile(GetUserProfile intent)
- DEFAULT :
  - ${intent}

# ChooseIntent_SendActivity_s3bxdY()
- Sure, no worries.

```

cwhitten (approved):

PR #7015, opened 2021-04-15
ci: use github actions eslint formatter

a-b-r-o-w-n:
## Description

Provide inline annotations for lint errors.


Composer/packages/lib/shared/src/path.ts
```diff
@@ -8,7 +8,7 @@ class PathHandler {
   resolve(...params: string[]) {
     const pathToTransform = path.join(...params);
     if (path.isAbsolute(pathToTransform)) {
-      return pathToTransform.replace(/\\/g, '/');
+      return pathToTransform.replace(/\\/g, '/')
```
tonyanziano, line 11: was this just to test the formatting?

a-b-r-o-w-n: Yeah, but the annotations aren't working like I expected. They show up in the workflow details page, but not on the file.

PR #7040, opened 2021-04-16
feat: migrate E2E test with bot creation v2

zhixzhan:
## Description

Migrate E2E test with bot creation v2.

Related to bot migration https://github.com/microsoft/BotFramework-Composer/pull/6526 e2e failed, it's because all e2e test are creating v1 bot, when run E2E test, all page was blocked by 'ask migrate modal'. 

To fix that need a few steps:
1. update all E2E test with bot creation v2. (in this PR)
2. when this PR done and merged, #6526  e2e are suppose to all pass, because test will not create v1 bot anymore, so there is no 'ask migrate modal' block page.
3. if we want add E2E test on migration, we can add a migration.spect.ts to #6526 or let it merged and check in with another PR.
 

In this PR, 
1. Out-of-date test `ToDoBot.spec.ts` is been removed.
2. Some small feature tests are gather into one `**Page.spec.ts`.  
3. Before all spec executed, will create a `TemplateBot_TestBot` and cache it, when run to each spec file, will reuse it by saving as  a new bot.

## Task Item

#minor


srinaath:
@benbrown . For the runtime migration work!

Composer/cypress/support/commands.ts
```diff
@@ -19,6 +19,43 @@ Cypress.Commands.add('createBot', (botId: string, botName?: string) => {
...
+    }
+  };
+
+  cy.request('post', '/api/v2/projects', params).then((res) => {
+    const { jobId } = res.body;
+    // install package can take a long time.
+    cy.wait(20000);
+    pollingRequestBotStatus(jobId, (result) => callback(result));
+  });
+});
```
a-b-r-o-w-n, line 57: This should cache the result of the first creation and the use that bot as a template by copying files and performing and transforms on the files. I know we don't use this api in too many places, but I think it's worth investing in the performance here to not slow the e2e execution down significantly.

benbrown:
This is awesome to see! Thank you for helping us complete the migration story.

Composer/cypress/support/commands.ts
zhixzhan: I made some updates, before all spec start, will create a template bot. when execute each spec, will use it call save as API, and spec will get a new test bot.

Composer/cypress/support/commands.ts
```diff
@@ -19,6 +21,65 @@ Cypress.Commands.add('createBot', (botId: string, botName?: string) => {
...
+
+Cypress.Commands.add('createTemplateBot', (botName: string, callback: (bot: any) => void) => {
+  cy.createBotV2(`TemplateBot_${botName}`, (bot) => {
+    TemplateBotProjectId = bot.id;
+  });
+});
+
+Cypress.Commands.add('createTestBot', (botName: string, callback: (bot: any) => void) => {
+  const name = `TestBot_${botName}`;
```
zhixzhan, line 68: name not prefix with `__Test*`  anymore is to meet our server name validation.  https://github.com/microsoft/BotFramework-Composer/blob/33f63dbda4d8c67d1f0095314999abb6e65b21d9/Composer/packages/lib/shared/src/dialogUtils/validateDialogName.ts#L7

boydc2014 (approved):

PR #7062, opened 2021-04-16
fix: catch dotnet errors during creation and offer download link

### LINKED ISSUE #6821: Catch error with dotnet user-secret init - add instructions about dotnet version
## Describe the bug

Fresh install of Composer on a mac, creating a new project (conversational core) results in the following error:

```
Command failed: dotnet user-secrets init --project djassistantcoremac.csproj Unrecognized command or argument 'init'
```

## Version

1.4.0-nightly.233315

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [x] macOS
- [ ] Windows
- [ ] Ubuntu

## To Reproduce

1. Fresh Install on Mac (.net core 2.1 installed, not 3.1)
2. New Project - Conversational Core
3. Wait for error after a few moments

## Expected behavior

Project created, if there is a pre-req missing - in this case .net core 2.1 was installed not 3.1 then we should detect lack of pre-req and gracefully highlight to the user that creation cannot be performed until it's installed rather than showing a mostly unhelpful error to the developer who may be unable to resolve.

Installing .net core 3.1 resolved the issue, detecting this and warning before creation starts is needed.

## Screenshots

image

### LINKED ISSUE #6930: Command failed: dotnet user-secrets init --project conversational_core1.csproj 'dotnet' is not recognized as an internal or external command, operable program or batch file.
Hi,

while creating BOT using BotFramework-Composer.
Received this error message.
Please help.

image

image

benbrown:
## Description

This applies the same test to the error message that is used during "bot start" that will catch common errors related to the dotnet version and handle them in a special way.

Rather than throwing a generic error, this will pop up a modal with link to download dotnet and further instructions.

## Task Item

Fixes #6821
Fixes #6930

## Screenshots

Screen Shot 2021-04-16 at 4 18 49 PM


pavolum (approved):

Composer/packages/client/src/components/InstallDepModal.tsx
```diff
@@ -0,0 +1,63 @@
...
+          <span>
+            &nbsp;
+            <Link href={props.learnMore} target="_blank">
+              {props.learnMoreText}
+            </Link>
+          </span>
+        )}
+      </Text>
+      <DialogFooter className={dialogFooterClass}>
+        <PrimaryButton data-testid="InstallButton" href={props.link} target="_blank" text={props.linkText} />
```
srinaath, line 53: This testId seems too generic. installDotNet? In addition, if its not required we can remove it

Composer/packages/client/src/components/InstallDepModal.tsx
```diff
@@ -0,0 +1,63 @@
...
+      isOpen={props.isOpen}
+      title={props.title}
+      onDismiss={() => {
+        props.setIsOpen(false);
+      }}
+    >
+      <Text>
+        <span>{props.text}</span>
+        {props.learnMore && (
+          <span>
```
hatpick, line 44: remove this and nbsp and use margin instead

Composer/packages/client/src/components/InstallDepModal.tsx
```diff
@@ -0,0 +1,63 @@
...
+import React from 'react';
+import { Text } from 'office-ui-fabric-react/lib/Text';
+import { Link } from 'office-ui-fabric-react/lib/Link';
+import { mergeStyles } from 'office-ui-fabric-react/lib/Styling';
+
+const dialogFooterClass = mergeStyles({
+  marginTop: '25px',
+});
+
+type InstallDepModalProps = {
```
hatpick, line 20: names are confusing down below for someone else who wants to reuse this, my suggestions:
setIsOpen => onDismiss: () => void; (I assume we are never gonna tell outside that the dialog is open from inside)
text => content of type React.ReactNode so caller can render whatever
link => downloadLink
linkText => downloadLinkText
learnMore => change type to an optional learnMode: {text: string, link: string} cause they always are together and one can't be defined without other.

Composer/packages/client/src/components/InstallDepModal.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { jsx } from '@emotion/core';
+import formatMessage from 'format-message';
+import { DefaultButton, PrimaryButton } from 'office-ui-fabric-react/lib/components/Button';
+import { DialogFooter } from 'office-ui-fabric-react/lib/components/Dialog';
+import React from 'react';
+import { Text } from 'office-ui-fabric-react/lib/Text';
+import { Link } from 'office-ui-fabric-react/lib/Link';
+import { mergeStyles } from 'office-ui-fabric-react/lib/Styling';
+
+const dialogFooterClass = mergeStyles({
```
hatpick, line 16: dont need to use mergeStyles, just directly assign marginTop: 25px to the root under styles.
<DialogFooter styles={{ root: { marginTop: 25 }}} .../>

Composer/packages/client/src/recoilModel/dispatchers/utils/project.ts
```diff
@@ -377,8 +379,16 @@ export const fetchProjectDataById = async (projectId): Promise<{ botFiles: any;
...
-  setError(callbackHelpers, ex);
+export const handleProjectFailure = (callbackHelpers: CallbackInterface, error) => {
+  const isDotnetError = checkIfDotnetVersionMissing({
+    message: error.response?.data?.message ?? error.message ?? '',
+  });
+
+  if (isDotnetError) {
+    callbackHelpers.set(warnAboutDotNetState, true);
+  } else {
+    setError(callbackHelpers, error);
```
hatpick, line 390: missing ```callbackHelpers.set(warnAboutDotNetState, fase);``` here?

Composer/packages/client/src/components/InstallDepModal.tsx
```diff
@@ -0,0 +1,63 @@
+/* eslint-disable react/no-danger */
```
hatpick, line 1: what happened to NodeModal, if it's removed this PR doesn't have that change

Composer/packages/client/src/components/InstallDepModal.tsx
hatpick: isOpen doesn't seem be needed

Composer/packages/client/src/components/InstallDepModal.tsx
benbrown: No. We've talked about this before several times.  An NBSP is the appropriate way to add a space between 2 inline elements. Margin is overkill and inapprorpriate for inline elements.

Composer/packages/client/src/components/InstallDepModal.tsx
benbrown: Updated.

Composer/packages/client/src/components/InstallDepModal.tsx
benbrown: Ah, weird. Dialog Footer doesn't have a root style, only styles for the buttons. I think this may have been done as a workaround for that.  

I can apply the css directly to the element.

Composer/packages/client/src/recoilModel/dispatchers/utils/project.ts
benbrown: It would be unlikely that users would be able to get to _a different error_ before resolving the dotnet one.

But to ensure future code paths.. I'll add it!

Composer/packages/client/src/components/InstallDepModal.tsx
benbrown: Removed that file.

Composer/packages/client/src/pages/home/Home.tsx
```diff
@@ -234,6 +240,16 @@ const Home: React.FC<RouteComponentProps> = () => {
           <WhatsNewsList newsList={feed.whatsNewLinks} />
         </div>
       </div>
+      {warnAboutDotNet && (
+        <InstallDepModal
+          downloadLink={missingDotnetVersionError.link.url}
+          downloadLinkText={formatMessage('Install .NET Core SDK')}
```
GeoffCoxMSFT, line 246: Nit: It seems strange that the learn more link has text and URL, but the download link has URL and hard-coded text.

GeoffCoxMSFT (approved):

cwhitten (approved):

PR #7063, opened 2021-04-16
fix: remodel ProjectTree using ul/li tags

### LINKED ISSUE #6925: Project tree jumps on skill addition in Electron installer
## Describe the bug

The project tree jumps around quite a bit on hovering over a skill.

1.  Selecting a project tree item should have uniform selection border
2.  Navigating between pages should not make the project tree item jump around.
3.  Expanding and contracting a project tree item should not jump

## Version

Electron 1.4.0-nightly.234673.12e1a8a

## Browser

[https://user-images.githubusercontent.com/13004779/114633007-a26b1c00-9c74-11eb-8e4e-caefdd634bdd.mov](https://user-images.githubusercontent.com/13004779/114633007-a26b1c00-9c74-11eb-8e4e-caefdd634bdd.mov)

Happens only on Electron

beyackle:
## Description

This swaps out the old `details`/`summary` based ProjectTree for one built out of `ul` and `li` elements. This is more supported by more browsers, including the Chromium v80 that our Electron build is based on, and is much friendlier to CSS styling. The behavior should be unchanged; clicking the triangles will expand and collapse folders as usual.

## Task Item

fixes #6925 

## Screenshots

image


a-b-r-o-w-n:
@beyackle Can you check the `TopicsList` component to make sure nothing needs to get ported to it?

Composer/packages/client/src/components/ProjectTree/ExpandableNode.tsx
```diff
@@ -53,8 +51,10 @@ export const ExpandableNode = ({
   }
 
   function handleClick(ev: MouseEvent) {
-    if ((ev.target as Element)?.tagName.toLowerCase() === 'summary') {
+    const target = ev.target as Element;
+    if (target.tagName.toLowerCase() === 'li') {
```
pavolum, line 55: can target.tagName be null? If so perhaps through a unary null check operator in there or run a null check on target.tagname

Composer/packages/client/src/components/ProjectTree/ExpandableNode.tsx
```diff
@@ -64,25 +64,20 @@ export const ExpandableNode = ({
...
       {/* eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions, jsx-a11y/no-noninteractive-tabindex */}
-      <summary
-        css={summaryStyle(depth, isActive, isExpanded)}
-        data-testid={'summaryTag'}
-        tabIndex={0}
-        onClick={handleClick}
-        onKeyUp={handleKey}
-      >
+      <li css={listItemStyle(isActive, isExpanded)} data-testid={'summaryTag'}>
         {summary}
```
pavolum, line 78: Can we safely assume that any text in {summary} jsx component is localized?

pavolum (approved):

Composer/packages/client/src/components/ProjectTree/ExpandableNode.tsx
beyackle: Mostly they're going to be user-input names of bots and dialogs, but in the few cases where someone's making a tree for another purpose, that will have been localized already in the source component.

Composer/packages/client/src/components/ProjectTree/ExpandableNode.tsx
beyackle: I thought about that, but if there's an event at all, it's going to have a target, and tagName is a standard way to get at what HTML tag has been clicked on. You can't click on a space where no element is.

Issue #7035, opened 2021-04-16
Core SDK Triggers are missing


tomlm:
## Describe the bug

**OnContinueConversation** trigger is missing from the UI.
**OnCommandActivity** trigger is missing 
**OnCommandResultActivity** is missing.

## Version

![image](https://user-images.githubusercontent.com/17789481/114957258-82bf2980-9e15-11eb-8b0f-79f4cb6cdd96.png)

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Create new projeect
2. go to trigger menu, try and find the missing triggers.

## Expected behavior

All triggers in the .schema files to be surfaced.

## Additional context

x

yeze322:
I'm using nightly-236088, **OnCommandActivity** and **OnCommandResultActivity** appear under the trigger wizard:
![image](https://user-images.githubusercontent.com/8528761/115004063-01c96600-9ed9-11eb-925d-1eed24d482b4.png).


`OnContinueConversation` doesn't contain an uischema definition in dotnet repo. Requires an update on sdk runtime side.

cc @cwhitten

Issue #7034, opened 2021-04-16
Core SDK Actions are missing from the UI


tomlm:
## Describe the bug

These actions are missing from the UI:

DialogManagement:
* CancelDialog
* GetConversationReference
* ContinueConversation
* ContinueConversationLater

Looping
* GotoAction

Debugging
* DebugBreak

## Expected behavior

Those actions to be there.

clearab:
@EricDahlvang I know you've been interested in this one!

yeze322:
They don't have uischema defined, needs to be updated in dotnet runtime repo.

Refs another thread of Trigger kinds: https://github.com/microsoft/BotFramework-Composer/issues/7035#issuecomment-821047705

Needs to have another wave of SDK release. @cwhitten

yeze322:
@tomlm Tom, I noticed `CancelDialog` and `CancelAllDialogs` have the same schema definition: [CancelDialog.schema](https://github.com/microsoft/botbuilder-dotnet/blob/main/libraries/Microsoft.Bot.Builder.Dialogs.Adaptive/Schemas/Actions/Microsoft.CancelDialog.schema) vs [CancelAllDialogs.schema](https://github.com/microsoft/botbuilder-dotnet/blob/main/libraries/Microsoft.Bot.Builder.Dialogs.Adaptive/Schemas/Actions/Microsoft.CancelAllDialogs.schema)

However, in the underlying implementation they are different from each other, where `CancelDialog.cancellAll` equals `false` and `CancelAllDialogs.cancellAll` equals `true`.

PR #7070, opened 2021-04-17
feat: Add SDK Actions missing in UI Schema to Adaptive Flow Edge Menu

tdurnford:
## Description
feat: Add SDK Actions missing in UI Schema to Adaptive Flow Edge Menu

## Task Item
#minor

## Screenshots
image


Composer/packages/extension-client/src/hooks/useMenuConfig.ts
```diff
@@ -29,6 +33,21 @@ export function useMenuConfig(): { menuSchema: MenuUISchema; forceDisabledAction
...
+    Object.entries(sdkDefinitions).forEach(([$kind, action]) => {
+      if (
+        $kind.startsWith('Microsoft') &&
+        $kind !== SDKKinds.AdaptiveDialog &&
+        !implementedMenuSchema[$kind] &&
+        actionImplementsDialog(action)
+      ) {
+        implementedMenuSchema[$kind] = {
+          label: getFallbackLabel($kind),
+          submenu: ['Other'],
```
tdurnford, line 46: Should we localize this? If so, we need to add the format message package.

a-b-r-o-w-n: @yeze322 would know more here but I think the menu is currently not localized or localized elsewhere.

cwhitten:
@srinaath @tdurnford let's remember to bring this up in the authoring standup tomorrow for discussion on naming/grouping of these

srinaath:
@cwhitten that sounds good. Will get it in the standup tomorrow.

cwhitten:
@srinaath was this discussed in the authoring standup? Was there an outcome?

garypretty:
For any actions that we do think should be added to Composer today, that have an obvious home, we should add them to the right menu if possible.  A couple of questions....

- Is there someone pushing for these (or a subset of these) actions to be added in R13?  I would like to understand if there is a pressing need here, or if we potentially have time and do this in R14.
- Can anybody tell me how 'Cancel all dialogs' differs from 'Cancel all active dialogs' (I would suggest Cancel all Dialogs should go on the Dialog menu, but having both of these seems confusing).
- If we are going to add GoToAction then I think the dialog menu probably makes sense for that too.

In summary, I would like it if we can avoid adding another menu to the list if possible, as it is already pretty long.  If we really do need these in R13 though, then I am happy with the 'Other' option if we don't have a suitable other place for things to live.

cwhitten:
@garypretty the following issues from @tomlm spawned this exercise.

https://github.com/microsoft/BotFramework-Composer/issues/7035
https://github.com/microsoft/BotFramework-Composer/issues/7034

The expectation I believe is that if the component ships with the SDK, it needs to also ship with Composer. I believe this is fine for most of these showing up, but some need custom UX in order to be usable and to show them without it would confuse folks. 

RE: component behavior--'m not sure on this. It may require a discussion with @Stevenic and @tomlm

tdurnford:
@garypretty There are two actions in the SDK Schema - `Microsoft.CancelAllDialogs` and `Microsoft.CancelDialog` - that have the same title and description. I'm honestly not sure how they're different and why the same title. @tomlm Can you provide an explanation? 

https://github.com/microsoft/BotFramework-Composer/blob/c0640064cdb9393edecc7c62203b104c72bea69c/Composer/packages/server/schemas/sdk.schema#L1047
https://github.com/microsoft/BotFramework-Composer/blob/c0640064cdb9393edecc7c62203b104c72bea69c/Composer/packages/server/schemas/sdk.schema#L1114

Issue #5857, opened 2021-02-22
User experience - New look and feel for Composer-Home


mareekuh:
**Composer Welcome / Start Page** 

Refresh of Composer Home to align with our Build release. 

References:

- Visual Studio - http://dotnetmirror.com/Resources/Uploaded-Files/fe0c4c6e-4fcc-49b6-bb35-376b009317b2/visual-studio-2017-Start-Page.png  
- Visual Studio Code - 2019-02-28-14_44_29-Welcome-Visual-Studio-Code.png (1912×1040) (qlikfix.com) 
- [See also (Gary's email/doc): ](https://nam06.safelinks.protection.outlook.com/ap/w-59584e83/?url=https%3A%2F%2Fmicrosoft.sharepoint.com%2F%3Aw%3A%2Ft%2FConversationalAI785%2FEQ95DgvK6eRBn0fK_mMZMAMBNuzGFM9oGUN6K0JUqUoYnQ%3Fe%3DIDwQi4&data=04%7C01%7Cmariekei%40microsoft.com%7C09e7a3a263cf4969308c08d8d68e782d%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C637495254882607416%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C1000&sdata=C2vmNkGB1rfGgfii%2FLEPMeoVBw08z95Dz8%2By5dtaKPA%3D&reserved=0)

Content 
- Recent bots list 
- Featured videos 
-  Tutorial

Functionality/tasks
- New bot 
- Open existing bot 
- Open/install Emulator

Access to:
- What’s new 
- Release notes
- Composer docs (Learn / Help) - Short list (maybe 3?) of links (title & brief sub title), blog posts, preview features etc.  
- Composer GitHub repo 
-  Bot Builder Community?  (Given our likely push for package building?) 
-  Stack Overflow 

Related products?
- PVA?
- Health bot?
- LUIS?
- QnA?
- AI Studio?

**Remove from current design:**
- examples list – 

**Future idea**
- Bundle Emulator
- Pin your favorite projects to the home screen

xiyangdesign:
[...]

PR #7140, opened 2021-04-20
fix: Recent bots sort and update styles

### LINKED ISSUE #7110: Home Page MRU list has confusing sort order visuals and behavior
## Describe the bug
The sort order for the MRU list of previously opened bots has misleading UI.  See screenshot for visuals.  

I'm not exactly sure what drives the MRU list sort order.  It's not recently opened.  Which is should be.  This list should be sorted by most recently opened.  This is the way office products work.  This is the way Visual Studio works.

also, there are headers for this list that indicate the list can be sorted by other columns.  This isn't the case.  

Lastly, the Name column has a sort icon displayed, and this isn't accurate, as the list isn't sorted by name. 

## Version
Electron: 8.2.4
Chrome: 80.0.3987.165
NodeJS: 12.13.0
V8: 8.0.426.27-electron.0
## Browser
- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS
- [x] macOS
- [ ] Windows
- [ ] Ubuntu

## To Reproduce

1. launch Composer
2. Note the MRU header indicates that list is sorted by name.  It actually isn't.
3. Click on the headers of the MRU list, note they indicate they should be sortable, though they are not.
4. Open a previous bot project that is not the first project in the list. 
5. Click the Home icon in the left hand nav to return to the home page.  Note the bot project you opened is not at the top of the MRU list, though is is the Most Recently Used project

## Expected behavior
I would expect the MRU list to behave as other Microsoft MRU lists, e.g. Word, Visual Studio, etc.

I would also expect the headers to not be sortable, and therefore not clickable

I would expect the up arrow icon for the Name column to go away, as we don't sort by Name.

## Screenshots

Screen Shot 2021-04-19 at 12 25 28 PM

zhixzhan:
## Description
1. fix recent bots sort 
2. update homepage padding styles

## Task Item
close #7110 
refs #5857 

## Screenshots
image


xiyangdesign:
@zhixzhan 
Could you help make the **table header non-clickable** visually as well? i.e. **no grey color when hovered over** as in the current build (see below). Thanks!

image

tonyanziano (commented):
Looks good to me. Will approve once Xi's comment is addressed.

zhixzhan:
@xiyangdesign @tonyanziano updated

beyackle:
Let's get this merged so we can close the outstanding issue.

PR #7169, opened 2021-04-20
chore: Updating required node version

GeoffCoxMSFT:
## Description

Need to update the readme to indicate we need node 14 for building composer.

## Task Item

#minor

## Screenshots

N/A

### DIFF
--- a/Composer/README.md
+++ b/Composer/README.md
@@ -3,13 +3,15 @@
 [![Total alerts](https://img.shields.io/lgtm/alerts/g/microsoft/BotFramework-Composer.svg?logo=lgtm&logoWidth=18)](https://lgtm.com/projects/g/microsoft/BotFramework-Composer/alerts/)
 
 # Composer
+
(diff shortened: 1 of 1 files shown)


Composer/README.md
```diff
@@ -3,13 +3,15 @@
...
+
 The web app that can edit bots in OBI format, and can use Bot Launcher to run bot.
 
 ### Instructions
 
 Prerequisite:
-* node > 12
-* yarn         // npm install -g yarn
+
+- node > 14
```
srinaath, line 13: Our required is only [BotFramework-Composer/package.json at main · microsoft/BotFramework-Composer (github.com)](https://github.com/microsoft/BotFramework-Composer/blob/main/Composer/package.json#L24) node 12. A constraint imposed by bf packages all which require only node 12.

cwhitten:
@GeoffCoxMSFT why is node 14 required?

Issue #6989, opened 2021-04-14
ProvisionComposer.js needs to be ARM Templates


cleemullins:
When creating a new Publishing Profile, there is a new hand-crafted replacement for "AZ CLI". This is perhaps not the best idea I've seen today. 

![image](https://user-images.githubusercontent.com/1165321/114793289-e1fe3a80-9d3e-11eb-9a44-16de692779dc.png)

This needs to be removed, and replaced with AZ CLI. This should simple be ARM templates, so that we work with everything across the Microsoft Ecosystem. 

Alternately, we need to review with the Azure Security Team, and examine other implications of this. The AZ CLI team will have significant interest in Microsoft shipping an alternate implementation of AZ CLI.

clearab:
I was pretty confused when I saw this command as well, along with the publishing profile JSON being specific to our format, rather than just re-using the existing publishing profile formats that are available in xml.

cwhitten:
Dewain can explain the scenario around this Share Resource Request dialog, but `provisionComposer.js` was not intended to live long, or be used outside of Composer and was replaced internally with the Azure Publish extension.

benbrown:
This script is a wrapper around ARM templates. I discussed this with Dewain this morning - I think it will be possible to replace this with direct use of the ARM template and some additional documentation.

clearab:
Would that mean we could drop this from the code that is scaffolded out for folks? That .js file shows up in the /Scripts folder, and I think is the source of the requirement for installing node in the provisioning process?

sgellock:
I've talked this through with @benbrown and I'm punting this out of R13.  What we have, though not ideal, meets the bar, given the remaining calendar time

PR #7180, opened 2021-04-20
fix: expose provisionComposer script in easily downloadable way

### LINKED ISSUE #6903: Need a place for an Admin to download the scripts folder
## Describe the bug

When doing the Admin Handoff pattern for provisioning we are asking the Admin to run a command that is in the scripts folder, but the admin doesn't have a scripts folder or access to the project files that the developer created.  We need to be able to provide a download link for the Admin to get the contents of the scripts folder.

## Version

Version:  1.4.0-nightly.234518.8e98823
Electron: 8.2.4
Chrome: 80.0.3987.165
NodeJS: 12.13.0
V8: 8.0.426.27-electron.0

## Browser

- [X] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [X] Windows
- [ ] Ubuntu

benbrown:
## Description

In order to make the provisioning script available for non-composer users, this PR introduces a *zipped copy* of the script and its dependent ARM templates, along with a readme file.  

Composer's provisioning UI will point users to this content as part of the instructions.

## Task Item

Related to #6989 
Fixes #6903


pavolum (approved):
Nice document, seems to cover all the bases to make a complex process rather simple. Curious, where is this surfaced? Is this what is linked to from within Composer?

benbrown:
@pavolum  yes in the handoff part of the provisioning flow, there is an aka.ms link that will point here.

PR #7201, opened 2021-04-21
fix: fix LuEditor block entity in skill dialog

### LINKED ISSUE #7151: Pre-built entities not supported when adding a remote skill
## Describe the bug

Adding a remote skill that uses pre-built entity an error is presented indicating that pre-built entities are not supported.

image

## Version

Version:  1.4.0-nightly.237101.f01cb52
Electron: 8.2.4
Chrome: 80.0.3987.165
NodeJS: 12.13.0
V8: 8.0.426.27-electron.0

## Expected behavior

Pre-built entities should be supported.

cc @darrenj

VanyLaw:
## Description

fix orchestrator install error during adding remote skill
fix luEditor blocked entities in add remote skill dialog. only show error in property editor with orchestrator recognizer.

## Task Item

close #7151


boydc2014:
Please resolve the conflict

hatpick:
@garypretty should the LU editor in this dialog show toolbar (insert + define entity) as well as floating labeling menu if it's LUIS and not orchestrator?

PR #7243, opened 2021-04-21
fix: adjust package manager feeds

### LINKED ISSUE #6854: Default package manager feeds
The split between the default feeds is confusing. We need to change the default feeds to just "NuGet" and "npm", and sort the search results by # of downloads. Having the extra search feeds is going to discourage use of community packages, and discourage their creation.

### LINKED ISSUE #7043: Components that contain only code and no schema do not show as installed in Package Manager
## Describe the bug

If a package contains only code (e.g. middleware etc) and no schema, then the Package, once installed through Package Manager, does not appear as installed and cannot be uninstalled / upgraded via Composer. @benbrown indicated this is because we are using schema to determine which packages are installed.

We need to look at another way of identifying which packages are installed, potentially through the package tag 'msbot-component'.

## To Reproduce

Install the Bot.Builder.Community.Components.Handoff.LivePerson package from the community.

## Expected behavior

The package should show as installed and be able to be uninstalled.

### LINKED ISSUE #7092: Package Manager gives false impression that only a limited number of packages are available
## Describe the bug

Package Manager loads an initial list of packages (try switching to the Community packages filter), but only loads a finite number.  Scrolling down the list does not extend the list by loading any additional packages.  Therefore users need to know exactly what they need to search for to find other packages. e.g. Bot.Builder.Community.Components.Handoff.LivePerson, which is currently not shown in the initial list.

## Version

## Expected behavior

Package Manager should load additional packages as the user scrolls down the list.

### LINKED ISSUE #7112: Orchestrator package not showing up in package manager
## Describe the bug

Orchestrator package doesn't show up in Package Manager - in the default npm feed. If you switch to community then it does show up but it's not a community package?

image

image

## Version

1.4.0-nightly.236490

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Go to package manager
2. Orchestrator not shown
3. Switch to community feed

## Expected behavior

- Shows up in main feed

benbrown:
## Description

1. Removes the "community" feeds - now each language has 1 feed, sorted by downloads
2. includes any component tagged msbot-component in installed list, in addition to using presence of schema
3. shows up to 100 components at a time

**NOTE** Testers will need to DELETE their extension settings file to remove the "community" feeds

## Task Item

fixes #7043 
fixes #6854
fixes #7092 
fixes #7112 

## Screenshots

Screen Shot 2021-04-21 at 5 36 48 PM
Screen Shot 2021-04-21 at 5 36 39 PM


extensions/packageManager/src/node/index.ts
```diff
@@ -14,18 +14,8 @@ import { FeedFactory } from './feeds/feedFactory';
...
-  // additionally, our schemamerge command only returns the top level components found, even though
-  // it does properly discover and include the schema from this dependent package.
-  // without this special case, composer does not see orchestrator as being installed even though it is.
-  // in the future this should be resolved in the schemamerger library by causing the includesSchema property to be passed up to all parent libraries
-  return c.includesSchema || c.name.toLowerCase() === 'microsoft.bot.components.orchestrator';
-};
-
 const isAdaptiveComponent = (c) => {
-  return hasSchema(c) || c.includesExports;
+  return c.includesSchema || c.keywords?.indexOf('msbot-component') >= 0 || c.includesExports;
```
hatpick, line 18: this is a nit:
```typescript
c.includesSchema || c.keywords?.includes('msbot-component') || c.includesExports;
```

hatpick (approved):

extensions/packageManager/src/node/index.ts
beyackle: It's fine as-is. I'd rather have type-safety and clarity than saving 4 bytes.

beyackle (approved):

extensions/packageManager/src/node/index.ts
hatpick: @beyackle what are you talking about?

pavolum (dismissed):

extensions/packageManager/src/node/index.ts
```diff
@@ -14,8 +14,12 @@ import { FeedFactory } from './feeds/feedFactory';
 
 const API_ROOT = '/api';
 
+const hasSchema = (c) => {
+  return c.includesSchema || c.keywords?.indexOf('msbot-component') >= 0;
```
beyackle, line 18: I like this refactoring, though if `c.keywords` is an array, then I'd prefer `.includes` to the `.indexOf` thing.

extensions/packageManager/src/node/index.ts
beyackle: Never mind - I completely misread your suggestion and I do actually agree with it.

benbrown: LOL OK will do!

hatpick (approved):

beyackle (approved):

PR #7254, opened 2021-04-22
fix: add isRemote check to Library list

### LINKED ISSUE #7159: Selecting a remote skill causes unexpected behavior in Package Manager
## Describe the bug

Added a remote Skill in Composer.  I had the remote skill selected in the left hand nav in the designer. When visiting Package Manager, you then get a message about needing to eject your runtime to use Package Manager.

image

## Version

Version:  1.4.0-nightly.237101.f01cb52
Electron: 8.2.4
Chrome: 80.0.3987.165
NodeJS: 12.13.0
V8: 8.0.426.27-electron.0

## To Reproduce

1. Add remote skill
2. Select the remote skill in left hand nav
3. Navigate to package manager

## Expected behavior

Given that this is a remote skill, I had expected to see package manager use the context of the root bot.

beyackle:
## Description

This solves an issue on the package manager page where selecting a remote skill would incorrectly bring up the "please eject your runtime" message (there is no runtime to eject, in that case).

[Draft until I can verify this works]

## Task Item

closes #7159


beyackle:
Discarded in favor of an alternate solution.

PR #7274, opened 2021-04-22
fix: enable creation workers when running electron in dev mode

hatpick:
## Description

Allow creation flow when developing and using electron in dev mode.

## Task Item

#minor


a-b-r-o-w-n (approved):
:100:

PR #7296, opened 2021-04-23
fix: Data loss when cutting / copying & pasting response

### LINKED ISSUE #7045: Data loss when cutting / copying & pasting response
## Describe the bug

Attempting to cut / paste (or copy / paste) a a 'Send a Response' action results in loss of the data within the action.

Cut Paste Error

## Version

Version:  1.4.0-nightly.236088.ef8eb11
Electron: 8.2.4
Chrome: 80.0.3987.165
NodeJS: 12.13.0
V8: 8.0.426.27-electron.0

lei9444:
### Root cause
The clipboard need to get the lg and lu content when cutting/copying. But composer today use the focusPath to get the focused dialog id. But the focusPath will been discarded and some navigation apis don't update the focusPath. So the cutting event can't get the lg templates or lu intents after some navigation event.

### fix
use designPageLocation to make sure the read apis get the correct focused dialog id.

## Task Item
Closes #7045


hatpick (approved):

hatpick:
nice @lei9444 🚀

PR #7326, opened 2021-04-23
fix: refactor way we get COMPOSER_VERSION

a-b-r-o-w-n:
## Description

Fixes an issue where the preload.js script could not be loaded due to `app` being undefined in the renderer process.

### DIFF
--- a/Composer/packages/client/config/env.js
+++ b/Composer/packages/client/config/env.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const fs = require('fs');
+const fs = require('fs-extra');
 const path = require('path');
 const { execSync } = require('child_process');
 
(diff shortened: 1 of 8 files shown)


cwhitten (approved):

Composer/packages/server/src/utility/getVersion.ts
```diff
@@ -0,0 +1,17 @@
...
+
+import path from 'path';
+
+import fs from 'fs-extra';
+
+const isProduction = process.env.NODE_ENV === 'production';
+
+export function getVersion(): string {
+  try {
+    const version = fs.readJSONSync(path.join(__dirname, '../../../electron-server/package.json')).version;
```
hatpick, line 12: does this file exist in packed electron?

Composer/packages/client/config/env.js
```diff
@@ -49,6 +49,14 @@ function getGitSha() {
   }
 }
 
+function getComposerVersion() {
+  try {
+    return fs.readJSONSync(path.join(__dirname, '../../electron-server/package.json')).version;
```
hatpick, line 54: same question

Composer/packages/server/src/utility/getVersion.ts
a-b-r-o-w-n: No, but the env variable gets set in the electron start up sequence. So in that case, this will just error and return `unknown`.

Right now, there is not a lazy evaluation for getting default values using the `setEnvDefault` function.

PR #7350, opened 2021-04-23
chore: bump to 2.0.0

cwhitten:
#minor


cwhitten:
Applying an override to test the 1.4 -> 2.0 upgrade installers

PR #7368, opened 2021-04-26
fix: Catch common QNA maker error about limits

### LINKED ISSUE #7217: Runtime error shown when building a Composer project using QnAMaker resource which has exceeded KB/Azure Search limits
## Describe the bug

Composer shows the following qnamaker build runtime error during the process of starting a Bot.

image

Further analysis shows that under the covers qnamaker is failing due to an azure search max index limitation. An Azure search index is created per KB, as per [these docs](https://docs.microsoft.com/en-us/azure/search/search-limits-quotas-capacity#index-limits) a free tier can have up to 3, the next tier can have up to 15.

Therefore a customer using the free tier will hit this issue when they add a fourth KB.

## Version

1.4.0-nightly.237504

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. I can provide a qnamaker key that repros

## Expected behavior

The full warning from QnAMaker is shown enabling a developer to understand what the issue is (not Composer!) and resolve.

## Screenshots

image

benbrown:
## Description

Catches an error from QNA that indicates that the search service may need to be upgraded.

## Task Item

Fixes #7217

## Screenshots

Screen Shot 2021-04-26 at 12 04 24 PM

### DIFF
--- a/Composer/packages/server/src/models/bot/builder.ts
+++ b/Composer/packages/server/src/models/bot/builder.ts
@@ -121,7 +121,16 @@ export class Builder {
       await this.runQnaBuild(interruptionQnaFiles);
       await this.runOrchestratorBuild(orchestratorBuildFiles, emptyFiles);
     } catch (error) {
-      throw new Error(error.message ?? error.text ?? 'Error publishing to LUIS or QNA.');
+      // handle this special error case where QnA Maker returns this uninformative error.
+      // in their portal, it is accompanied by a message about the search service limits
+      // (A free search is limited to 3 indexes, which can be used up quickly)
+      if (error.text === 'Qnamaker build failed: Runtime error.') {
+        throw new Error(
+          'QnA Maker build failed: This error may indicate that your Search service requires an upgrade. For information about the Search service limits, see here: https://docs.microsoft.com/en-us/azure/search/search-limits-quotas-capacity'
+        );
+      } else {
+        throw new Error(error.message ?? error.text ?? 'Error publishing to LUIS or QNA.');
+      }
     }
   };
 


beyackle (approved):

Composer/packages/server/src/models/bot/builder.ts
```diff
@@ -121,7 +121,16 @@ export class Builder {
       await this.runQnaBuild(interruptionQnaFiles);
       await this.runOrchestratorBuild(orchestratorBuildFiles, emptyFiles);
     } catch (error) {
-      throw new Error(error.message ?? error.text ?? 'Error publishing to LUIS or QNA.');
+      // handle this special error case where QnA Maker returns this uninformative error.
+      // in their portal, it is accompanied by a message about the search service limits
+      // (A free search is limited to 3 indexes, which can be used up quickly)
+      if (error.text === 'Qnamaker build failed: Runtime error.') {
```
hatpick, line 127: is the error always going to be in English?

benbrown: The error from Qnamaker?

Or are you saying we should formatMessage this error?

hatpick: former.

I see that the runQnaBuild is sending a ```culture``` that is basically a locale

benbrown: That is a good question! I don't know how to change that setting...

PR #7400, opened 2021-04-27
fix: plugin page router

zhixzhan:
## Description
fix unable to navigate to plugin page, introduced in #7330.

fixed by adding back route config in no skill case.

## Task Item
refs #7330

## Screenshots

when select on root bot

image

when select on skill bot

image


zhixzhan:
Hi @beyackle , can you help take a look on this ?

boydc2014 (approved):

boydc2014:
I'm taking this to unblock, looks like it's the right fix.

PR #7407, opened 2021-04-27
fix: use feature-specific tags rather than msbot-component

### LINKED ISSUE #7403: Can't start bot after installing any packages
## Describe the bug

Can't start my bot after install any packages like Microsoft.Bot.Components.Welcome dialog. 
It complains couldn't load dll, but actually those packages aren't even including dlls. 
Something must be wrong about the project setup.  

image

## Version

lastest main 4/27

## Browser

- [ ] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

## To Reproduce

1. Create an empty bot
2. Install Microsoft.Bot.Compontents.Welcome or HelpAndCancel, or Graph
3. Start bot

benbrown:
## Description

Part of the package install process is to write a line to the appsettings.components array for code-driven components.

Since these components may or may not have schema, we need to use the package TAGS to determine if they should be included.

A previous PR used the `msbot-component` tag, but this proved to be too broad as it includes the declarative-only components that SHOULD NOT be added to this list.

This PR changes the logic to look for one of a series of feature-specific tags.

## Task Item

Fixes #7403

### DIFF
--- a/extensions/packageManager/src/node/index.ts
+++ b/extensions/packageManager/src/node/index.ts
@@ -15,7 +15,15 @@ import { FeedFactory } from './feeds/feedFactory';
 const API_ROOT = '/api';
 
 const hasSchema = (c) => {
-  return c.includesSchema || c.keywords?.includes('msbot-component');
+  return (
+    c.includesSchema ||
+    c.keywords?.includes('msbot-action') ||
+    c.keywords?.includes('msbot-trigger') ||
+    c.keywords?.includes('msbot-adapter') ||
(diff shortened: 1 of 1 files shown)


benbrown:
Wait on merging this while we make sure this is the right mechanism to use to determine the treatment of certain packages

cwhitten (approved):

PR #7430, opened 2021-04-27
fix: Add a warning about the luis authoring key 1000/month limit 

### LINKED ISSUE #7146: Authoring key incorrectly used for endpoint calls
## Describe the bug

When selecting a LUIS key in Composer this is purely retrieving the authoring key and not enabling me to select an endpoint key. This has been raised a few times and the (old) composer behaviour of using Authoring key for endpoint (prediction) calls is not correct. Whilst it works, it has a limited quota and will quickly run out of quota meaning the developers Bot stops working which is hard to diagnose and root of many a developer issues today. 1000 calls to the key are allowed per month, with LUIS being run on every turn these days it doesn't take long to hit it. 

## Version

1.4.0-nightly.237101

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Create new empty bot
2. Add LUIS intent/trigger
3. From rocket ship menu select an existing LUIS resource
4. Observe that LUIS authoring key is set and endpoint key is not.

## Expected behavior

Selection of an endpoint key as well as authoring. Ensuring we give the developer a recommended, best practice configuration out of the box not something setting them up for failure.

benbrown:
## Description

By default, a local bot will be set up with ONLY a LUIS authoring key. The authoring key can be used up to 1000 times a month as an endpoint key, which works for local testing.

However, if this limit is reached, the bot will stop working and throw LUIS errors.

This PR introduces a warning thaat appears on the bot settings page if a LUIS authoring key is set without an endpoint key.

The warning tells the user to publish to Azure if this limit is reached, and links to information about the key limits.

## Task Item

Fixes #7146

## Screenshots

Screen Shot 2021-04-27 at 3 13 43 PM

### DIFF
--- a/Composer/packages/client/src/pages/botProject/RootBotExternalService.tsx
+++ b/Composer/packages/client/src/pages/botProject/RootBotExternalService.tsx
@@ -420,7 +420,7 @@ export const RootBotExternalService: React.FC<RootBotExternalServiceProps> = (pr
         {localRootLuisKey && !rootLuisEndpointKey && (
           <MessageBar messageBarType={MessageBarType.info} styles={{ root: { width: '75%', marginTop: 20 } }}>
             {formatMessage.rich(
-              'Your bot is configured with only a LUIS authoring key, which has a limit of 1,000 calls per month. If your bot hits this limit, publish it to Azure using a<a>publishing profile</a> to continue testing.<a2>Learn more</a2>',
+              'Your bot is configured with only a LUIS authoring key, which has a limit of 1,000 calls per month. If your bot hits this limit, publish it to Azure using a <a>publishing profile</a> to continue testing.<a2>Learn more</a2>',
               {
                 a: ({ children }) => (
                   <Link key="luis-endpoint-key-info" href={linkToPublishProfile}>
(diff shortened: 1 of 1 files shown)


Composer/packages/client/src/pages/botProject/RootBotExternalService.tsx
```diff
@@ -412,6 +417,29 @@ export const RootBotExternalService: React.FC<RootBotExternalServiceProps> = (pr
             setDisplayManageLuis(true);
           }}
         />
+        {localRootLuisKey && !rootLuisEndpointKey && (
+          <MessageBar messageBarType={MessageBarType.info} styles={{ root: { width: '75%', marginTop: 20 } }}>
+            {formatMessage.rich(
+              'Your bot is configured with only a LUIS authoring key, which has a limit of 1,000 calls per month. If your bot hits this limit, publish it to Azure using a<a>publishing profile</a> to continue testing.<a2>Learn more</a2>',
```
hatpick, line 423: ```to Azure using a<a>publishing```
add a space between a and <a>

cwhitten (approved):

PR #7431, opened 2021-04-27
fix: luis provisioning to return location

### LINKED ISSUE #7288: Provisioning modal | Asks for Luis region even when luis has already been added..
Even though I have provisioned new Luis & Q&A services when I come back to add more resources to this profile, I get asked to specify the region for luis. This seems odd, because I have already provided this information..

( attached screenshots)

Screen Shot 2021-04-22 at 4 27 05 PM
Screen Shot 2021-04-22 at 4 27 11 PM

GeoffCoxMSFT:
## Description

The azure provisioning was not returning the location field expected in the provisioning result.

## Task Item

fixes #7288

## Screenshots

image

### DIFF
--- a/extensions/azurePublish/src/node/azureResourceManager/azureResourceManager.ts
+++ b/extensions/azurePublish/src/node/azureResourceManager/azureResourceManager.ts
@@ -133,7 +133,9 @@ export class AzureResourceMananger {
    * Deploy luis authoring resource
    * @param config
    */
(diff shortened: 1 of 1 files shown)


benbrown (approved):

benbrown:
@GeoffCoxMSFT the end result of this, if i read correctly, is that the region used to create LUIS is included in the publishing profile?

Issue #7230, opened 2021-04-21
Open Azure Bot resource (ABS > Composer) dialog too small


emivers8:
## Describe the bug

Implementation looks like this:
![image](https://user-images.githubusercontent.com/55601267/115608044-9ac1fd80-a29a-11eb-8fec-7b0cb37f137b.png)

Should be this:
![image](https://user-images.githubusercontent.com/55601267/115608119-af9e9100-a29a-11eb-8694-a2bab8f127c0.png)

PR #7432, opened 2021-04-27
fix: add skills flow refactored to desired R13 spec

### LINKED ISSUE #7115: "Create a skill in your bot" should be renamed to "Add a new skill"
## Describe the bug
When you have an existing bot project open and you go through the flow to create a new skill, the title of the dialog we display is "Create a skill in your bot". We don't create anything in a bot.  We are actually adding a new skill to a bot projects.  

Using Visual Studio as a guild, that use very simple language which would equate to "Add a new skill" or if we love the verb create, calling it "Create a new skill" would work as well.

## Version
Electron: 8.2.4
Chrome: 80.0.3987.165
NodeJS: 12.13.0
V8: 8.0.426.27-electron.0

## Browser
- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS
- [x] macOS
- [ ] Windows
- [ ] Ubuntu

## To Reproduce

1. Launch Composer
2. Open an existing bot project
3. From the main authoring screen, click the "+ Add" menu item at the top of the project window
4. Click the "Create a new skill" menu item, this will launch the dialog of I referring to
5. Note the title of that modal.

## Expected behavior
A more simplified title that more accurately represents the action and purpose of the modal

## Screenshots
Screen Shot 2021-04-19 at 1 43 30 PM

pavolum:
## Description

This PR makes the UI changes to match the desired add skills UI for R13 which is described in the figma that is linked from the below issue.

## Task Item

fixes #7115 

## Screenshots

**Toolbar drop down change**
image

**New Choice Modal**
image

**If you select 'add existing'**
image

**If you select 'add blank'**
image

### DIFF
--- a/Composer/packages/client/src/components/CreationFlow/v2/AddBotModal.tsx
+++ b/Composer/packages/client/src/components/CreationFlow/v2/AddBotModal.tsx
@@ -45,7 +45,7 @@ export const AddBotModal: React.FC<AddBotModalProps> = (props) => {
   return (
     <DialogWrapper
       isOpen
-      dialogType={DialogTypes.CreateFlow}
+      dialogType={DialogTypes.DesignFlow}
       subText={formatMessage('Which bot would you like to add to your project')}
       title={formatMessage('Add a bot')}
       onDismiss={props.onDismiss}
--- a/Composer/packages/client/src/components/CreationFlow/v2/AddBotModal.tsx
+++ b/Composer/packages/client/src/components/CreationFlow/v2/AddBotModal.tsx
@@ -1,0 +1,60 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import React, { useState } from 'react';
+import { DialogTypes, DialogWrapper } from '@bfc/ui-shared';
+import formatMessage from 'format-message';
+import { ChoiceGroup, IChoiceGroupOption } from 'office-ui-fabric-react/lib/ChoiceGroup';
+import { DialogFooter } from 'office-ui-fabric-react/lib/Dialog';
+import { DefaultButton, PrimaryButton } from 'office-ui-fabric-react/lib/Button';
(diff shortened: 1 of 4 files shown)


hatpick:
@emivers8 do you want the same sizing mentioned here #7230 for the "add a bot" dialog (second screenshot above)

benbrown (approved):

clearab:
@pavolum that should be "empty" not "blank" bot.

sgellock:
This should not be either.  I"ve raised this already, that we will provide any template that is not a multibot template.  so it's not just Empty.

clearab:
I was just responding to the flow above that skips the template selection and jumps straight to creating an empty bot. If it instead goes to the template selection screen with a filtered set of templates it should probably say just "bot", but however y'all would like this to work.

sgellock:
I know.  I'm saying why is this the flow?  it completely removes my ability to add anything other than an empty bot to an existing project?  why would we scope to this?  that's my question

mewa1024:
sorry--I thought the discussion from last week landed on this: the only new bot you could create while adding the bot was the empty aka blank bot for now.

mewa1024:
okay, new example of the flow with the template picker shown (but we will filter out the multibots):

image

not sure if we also need to take out the language picker (C#/Node)?

emivers8:
As long as the node templates also hide any multi-bots, I think it's fine to stay? Right now I don't think there are any node multi-bot templates.

sgellock:
we don't yet have multi bot templates for node, but we will.  so we need to apply the same filtering in both languages, and given we use the same package to support both languages, we should just get this for free.  @benbrown

mewa1024:
So a C# root bot can have a Node skill (or vice versa)?

sgellock:
yes.  each bot has it's own language specific runtime.

PR #7436, opened 2021-04-27
fix: Prevent object from making line-height invalid

### LINKED ISSUE #7300: Set properties panel height is not fit
## Describe the bug

Set properties panel height is not fit 
ex : defined properties  "dialog.username = @username" but in the visual canvas "g" is not display properly. 

## Version

Version:  1.4.0-nightly.238265.5106247
Electron: 8.2.4
Chrome: 80.0.3987.165
NodeJS: 12.13.0
V8: 8.0.426.27-electron.0

## Browser

- [X] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [X ] Windows
- [ ] Ubuntu

## To Reproduce

Add the properties in the Set Properties panel ... "g" is not visible properly.

## Expected behavior

All the letters should be visible

## Screenshots

prop

GeoffCoxMSFT:
## Description

Updated to protect the case where the set properties card in the adaptive flow is sending an object as the height for the body of the card template.  

## Task Item

fixes #7300 

## Screenshots

N/A


Composer/packages/lib/ui-shared/src/styled/styledComponents.tsx
```diff
@@ -41,7 +41,7 @@ export const BorderedDiv = styled.div<DivProps>(
 export const SingleLineDiv = styled.div<DivProps>`
   ${StandardFontCSS};
   ${TruncatedCSS};
-  line-height: ${(height) => (height ? height + 'px' : undefined)};
+  line-height: ${(height) => (height && typeof height === 'number' ? height + 'px' : undefined)};
```
hatpick, line 44: the fix is to destructure height, that function signature is basically (props: DivProps) => { styles here }

DivProps has children + height and width according.

hatpick (changes requested):

Composer/packages/lib/ui-shared/src/styled/styledComponents.tsx
GeoffCoxMSFT: When debugging sometimes the height comes in as a number and sometimes it comes in as an object that doesn't have any height within it.

hatpick: I dont think it can be number:
https://emotion.sh/docs/styled

PR #7463, opened 2021-04-28
fix: detect bots vs skills in Most Recently Used list

### LINKED ISSUE #7438: Home page: MRU should not list bots that were created or opened as part of a multi-bot project
## Describe the bug

The Home page MRU lists bots that were created or opened as part of a multibot project. This might lead a user to think we recommend editing those bot separately, when we don't. We would like the MRU to show only root bots or nonroot bots that were explicitly opened by choosing the Open control and navigating to the nonroot bot.

## Version

BotFramework-Composer-2.0.0-nightly.239895.e3e1b3e-mac.zip

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [x] macOS
- [ ] Windows
- [ ] Ubuntu

## To Reproduce

1. Create a new bot and choose the Enterprise Assistant template.
2. When it is created, you will see this in your MRU:
image

## Expected behavior

Expected is to see only the root bot listed:

image

beyackle:
## Description

This adds better detection to the most-recently-used bots list; importing an existing skill or creating a new one from within the Add menu in Design view will no longer make those skills pop into the list.

## Task Item

fixes #7438


Composer/packages/server/src/services/project.ts
```diff
@@ -462,6 +462,7 @@ export class BotProjectService {
       schemaUrl,
       runtimeType,
       runtimeLanguage,
+      isRoot: creatingRootBot,
```
tdurnford, line 465:
```suggestion
      isRoot: creatingRootBot = true,
```

Composer/packages/server/src/services/project.ts
```diff
@@ -562,7 +563,7 @@ export class BotProjectService {
         const id = await BotProjectService.openProject(
           { storageId: rootBot?.storageId, path: rootBot.path },
           user,
-          true
+          creatingRootBot ?? true
```
tdurnford, line 566: You should just give `creatingRootBot` a default value, so you don't have to worry about it being `undefined`. This is difficult to understand. 

```suggestion
          creatingRootBot
```

tdurnford (changes requested):

Composer/packages/server/src/services/project.ts
beyackle: Good idea. I'll add that in.

srinaath:
Ran through the scenarios locally. They work well.

tdurnford (approved):

PR #7475, opened 2021-04-29
fix: data race writing on setting file

### LINKED ISSUE #7153: SkillDialog action automatically added when connecting to a Skill cannot select Skill endpoint
## Describe the bug

When connecting a Skill to an basic (empty) bot a trigger is added for the Skill. Within that trigger is a SkillDialog custom action. When testing you get a `Value cannot be null. (Parameter uriString)` error which appears to be because the Skill Endpoint selection isn't selected automatically (a bug) and cannot be manually selected.

## Version

1.4.0-nightly.237101

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Connect a skill to a Bot (e.g. https://djtest6.azurewebsites.net/manifests/djbfskill1-2-1-manifest.json)
2. Observe that the Skill endpoint setting on SkillDialog action in newly created trigger isn't set
3. Skill isn't invoked

## Screenshots

image

zhixzhan:
## Description

Fix data race writing on setting file.

1.  when import packages write setting file, get latest setting before set
2. add default value `skillConfiguration` in setting to prevent every `get` in settingManager produce a `set`.

## Task Item
close #7153


extensions/packageManager/src/node/index.ts
```diff
@@ -386,7 +386,7 @@ export default async (composer: IExtensionRegistration): Promise<void> => {
                 newlyInstalledPlugin &&
                 !currentProject.settings.runtimeSettings?.components?.find((p) => p.name === newlyInstalledPlugin.name)
               ) {
-                const newSettings = currentProject.settings;
+                const newSettings = await currentProject.getEnvSettings();
```
zhixzhan, line 389: During package fetch, settings can be expired, get latest before modify it.

Composer/packages/server/src/models/bot/botProject.ts
```diff
@@ -242,9 +242,9 @@ export class BotProject implements IBotProject {
       // if endpointKey has not been set, migrate old key to new key
       if (!settings.qna.endpointKey) {
         settings.qna.endpointKey = settings.qna.endpointkey;
+        delete settings.qna.endpointkey;
+        await this.updateEnvSettings(settings);
       }
-      delete settings.qna.endpointkey;
-      await this.updateEnvSettings(settings);
     }
```
zhixzhan, line 248: Reduce unnecessary write on settings

Composer/packages/server/src/models/settings/defaultSettingManager.ts
```diff
@@ -121,6 +121,10 @@ export class DefaultSettingManager extends FileSettingManager {
         maxImbalanceRatio: -1,
       },
       skill: {},
+      skillConfiguration: {
+        isSkill: false,
```
tdurnford, line 125: I think we removed the isSkill property

tdurnford (approved):

Composer/packages/server/src/models/settings/defaultSettingManager.ts
```diff
@@ -121,6 +121,9 @@ export class DefaultSettingManager extends FileSettingManager {
         maxImbalanceRatio: -1,
       },
       skill: {},
+      skillConfiguration: {
+        allowedCallers: ['*'],
```
cwhitten, line 125: i don't know why this is here. this isn't an appropriate default parameter

cwhitten (changes requested):

Composer/packages/server/src/models/settings/defaultSettingManager.ts
zhixzhan: I see,  `allowedCallers`  has been removed now

tdurnford (approved):

cwhitten (approved):

Issue #7476, opened 2021-04-29
Composer overwrite manually added languages files 


boydc2014:
## Describe the bug
When adding a new language in composer, composer will create language files (lg, lu, qna) files based on current language's content. When there is already a file with the expected name in disk, composer will override the file. 

This will disrupt make some workflow, in which content authors translated the language files outside composer, and copied back to composer, then add a language in composer. 

## Version
latest main

## Browser

- [X] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [X] Windows
- [ ] Ubuntu

## To Reproduce

1. Create a bot
2. Created a translated lg file like a.fr-fr.lg with French
3. Add fr-fr in Composer
4. See a.fr-fr.lg been overwrite

PR #7478, opened 2021-04-29
fix: Composer overwrite manually added languages files

lei9444:
### Root Cause
If the languages in settings doesn't contain one language, but the bot has some related language files(lg, lu, qna). The old files will be overwrote when adding the language in composer. 

### Fix in this PR
filter the copied files if composer already has these files.

### Next Step 
Check the missing parts (templates, intents) when copying the files from default language files. 

## Task Item
refs #7476


Composer/packages/client/src/recoilModel/dispatchers/multilang.ts
```diff
@@ -22,6 +23,17 @@ import {
   botDisplayNameState,
 } from './../atoms/botState';
 
+type FileWithId = { id: string };
+
+const filterFiles = (targetFiles: FileWithId[], currentFiles: FileWithId[]) => {
```
hatpick, line 28: I think this one needs a better name, from my understanding it returns only new files, so if the name reflects that it would be easier to understand for other devs:
```getNewlyAddedFiles``` (or something like that)

Also please add a comment to the function explaining what it does.

File: Composer/packages/client/src/recoilModel/dispatchers/multilang.ts [hunk 2]
```diff
@@ -94,15 +106,15 @@ export const multilangDispatcher = () => {
       // copy files from default language
       set(lgFilesSelectorFamily(projectId), (oldLgFiles) => {
         const addedLgFiles = copyLanguageResources(oldLgFiles, defaultLang, languages);
-        return [...oldLgFiles, ...addedLgFiles];
+        return [...oldLgFiles, ...(filterFiles(addedLgFiles, oldLgFiles) as LgFile[])];
```
hatpick, line 109: while you are at it, can you rename ```oldLgFiles``` to ```prevLgFiles``` to match the other namings below?

zhixzhan:
Hi @hatpick I made some refinement on this branch, since @lei9444 is off from today.

hatpick (approved):

cwhitten:
I'm experimenting with this change on a special branch.

Issue #6808, opened 2021-04-08
Home Page "Save as" menu option is confusing


sgellock:
## Describe the bug
We seem to be mixing metaphors on the new home page.  I get creating new bots, or opening an existing bot from this page.  What I don't understand is what is the role of the "Save as" menu option?

What am I saving as?

## Version
Version:  1.4.0-nightly.232669.4daad5c
Electron: 8.2.4
Chrome: 80.0.3987.165
NodeJS: 12.13.0
V8: 8.0.426.27-electron.0

## Browser
- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS
- [x] macOS
- [ ] Windows
- [ ] Ubuntu

## To Reproduce

1. Open or create a new bot
2. Click on home icon in the left hand nav so you can return to the home page.
3. Click the "Save as" menu option
4. Notice you get a Create a bot project model

## Expected behavior
I understand the need for a Save As, or even a rename.  I just find it odd to be on the home page.

emivers8:
FYI design incoming for this and other project settings updates @pavolum.

mewa1024:
The "Save As" button appearing on the home page is part of larger inconsistency: for actions that one would normally perform on a file/project, we sometimes perform at the bot project level and sometimes at the bot-*within*-a-project level.

### Current behavior

- "**New**" (on home page): creates a new **bot project**. The user selects the location of the bot project folder.
- "**Open**" (on home page): opens a **bot within a project**. The user must select the location of the bot, not the bot project folder. (If the bot is the root bot of a project, it in effect opens the project. If it is a skill within a project, it opens only that bot.)
- "**Delete a bot**" (on the project settings page): deletes a **bot within a project.**
- "**Save As**" (on the home page): duplicates the **root bot** only. Does not duplicate the bot project. 
(_**Problems**: The location of the Save As button on project-wide part of the experience would imply that it saves the project, but it does not. There is no control within Composer to duplicate a bot project.)_

### New behavior
- "**New**" (on Home): _no change_
- "**Open**" (on Home): _no change_
- "**Delete**" (on Project Settings): Split Delete action:
   1.  On Project Settings, above the page content, cchange "**Delete**" to "**Delete project**" and change the behavior so it deletes the entire bot project, rather than a bot in the project.
[...]

PR #7482, opened 2021-04-29
chore: Temporarily removes save as button for R13

tdurnford:
## Description
Since we don't have a clear definition on what the behavior should be for copying both parent bots and skills nor a clear solution on how to implement the behavior, we are disabling the 'Save as' button on the home page in R13 with the intention of adding the feature back in a future release.

## Task Item
ref #6808

## Screenshots
image


hatpick (approved):
I'd use comment blocks like /* commented code */ vs. //

PR #7500, opened 2021-04-29
fix: allow dashes for pva bot names

### LINKED ISSUE #7486: [PVA] relax naming constraints for bots imported from PVA
PVA bot names do not need to be as constrained as they are from BF created bots. The only ask is to allow `-`.

a-b-r-o-w-n:
## Description

The naming constraints for BF created bots do not apply for PVA bots. Dashes are allowed.

Also converts invalid characters to dashes upon import so there is no error.

## Task Item

fixes #7486

### DIFF
--- a/Composer/packages/client/src/constants.ts
+++ b/Composer/packages/client/src/constants.ts
@@ -421,7 +421,7 @@ export const nameRegex = /^[a-zA-Z0-9-_]+$/;
 export const nameRegexV2 = /^[a-zA-Z0-9_]+$/;
 
 export const invalidNameCharRegex = /[^a-zA-Z0-9-_]/g;
-export const invalidNameCharRegexV2 = /[^a-zA-Z0-9-_]/g;
+export const invalidNameCharRegexV2 = /[^a-zA-Z0-9_]/g;
 
 export const authConfig = {
   // for web login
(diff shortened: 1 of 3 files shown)


Composer/packages/client/src/constants.ts
```diff
@@ -420,7 +420,8 @@ export const nameRegex = /^[a-zA-Z0-9-_]+$/;
 
 export const nameRegexV2 = /^[a-zA-Z0-9_]+$/;
 
-export const invalidNameCharRegex = /[^a-z^A-Z^0-9^_]/g;
+export const invalidNameCharRegex = /[^a-zA-Z0-9-_]/g;
+export const invalidNameCharRegexV2 = /[^a-zA-Z0-9-_]/g;
```
beyackle, line 424: These look identical, and I don't see where you use the V2 one.

beyackle (changes requested):

Composer/packages/client/src/constants.ts
a-b-r-o-w-n: That's a good catch. I just copied but forgot to update. I need to remove the - from the non V2 one. As for not being used-- the original variable isn't used anywhere, but I thought I'd add the V2 one to differentiate it in the future.

beyackle (approved):

PR #7554, opened 2021-05-03
fix: get tenant specific arm tokens for connections 

### LINKED ISSUE #7536: BUG: On connections tab of settings, auth should use tenantId from the publishing profile!
Related to #6963 

Publishing profiles now contain a tenantId.  We should use this when retrieving auth to work with the ABS channels on the connections tab.  Otherwise, this will not work for users with alternative tenants or personal accounts.

benbrown:
## Description

Applies the tenantId from the publishing profile to the auth process, thus requesting a token for the appropriate tenant.

If not present in the publishing profile, falls back to token auth as this indicates a legacy (pre-tenant support in Composer) publishing profile.

## Task Item

fixes #7536


Composer/packages/client/src/pages/botProject/adapters/ABSChannels.tsx
```diff
@@ -122,27 +123,40 @@ export const ABSChannels: React.FC<RuntimeSettingsProps> = (props) => {
...
+          }
+          newtoken = getTokenFromCache('accessToken');
+        } else {
+          try {
+            // if tenantId is present, use this to retrieve the arm token.
+            // absence of a tenantId indicates this was a legacy (pre-tenant support) provisioning profile
+            if (config.tenantId) {
+              newtoken = await AuthClient.getARMTokenForTenant(config.tenantId);
+            } else {
+              newtoken = await AuthClient.getAccessToken(armScopes);
```
a-b-r-o-w-n, line 152: This is still a broken path (which may be fine for this release). I think we could follow the same pattern in [Publish.tsx](https://github.com/microsoft/BotFramework-Composer/blob/main/Composer/packages/client/src/pages/publish/Publish.tsx#L303) and show a error notification which prompts the user to edit their configuration.

Composer/packages/client/src/pages/botProject/adapters/ABSChannels.tsx
```diff
@@ -321,16 +335,27 @@ export const ABSChannels: React.FC<RuntimeSettingsProps> = (props) => {
...
+          setErrorMessage(error.message || error.toString());
+          setCurrentResource(undefined);
+        }
       }
-      newtoken = getTokenFromCache('accessToken');
-    } else {
-      newtoken = await AuthClient.getAccessToken(armScopes);
+      setToken(newtoken);
     }
-    setToken(newtoken);
```
a-b-r-o-w-n, line 333: We should consolidate the auth code into a single method that can be called here and above.

a-b-r-o-w-n (changes requested):

Composer/packages/client/src/pages/botProject/adapters/ABSChannels.tsx
benbrown: But wouldn't it only be broken if the user had *successfully* created a publishing profile in the past using the default tenant used when accessing the getAccessToken method.... right?

a-b-r-o-w-n: This is just broken for users with MSA accounts I think. Just saying we have an opportunity to address that here and should consider doing it because is addresses the root cause of the original issue.

Composer/packages/client/src/pages/botProject/adapters/ABSChannels.tsx
benbrown: done

Composer/packages/client/src/pages/botProject/adapters/ABSChannels.tsx
benbrown: I added these additional checks to try to fetch a tenantId from the tenant list, etc.

a-b-r-o-w-n (approved):

PR #7563, opened 2021-05-03
fix: Update provisionScript.zip with small syntax changes

### LINKED ISSUE #7539: Update provisionComposer.js for handoff
The provisionComposer script used for the admin handoff could be tweaked for its new use case.  This pertains to the *zipped copy of the provision script* distributed for admin handoff that lives on Github.

MUST DO:
- [x] if specified, resourceGroup is used for name of several items, resulting in mismatch in the final publishing profile.   (We either need to fix this, or remove the resourceGroup parameter from the instructions, as this can cause broken profiles)
- [x] and add tenantId (recently added in Composer)

NICE TO HAVE:
- [x] remove accessToken field (not needed)
- [x] luisResource gets set even if luis is not created (no impact, but confusing)

benbrown:
## Description

This is a slightly tough PR since the delta is inside a binary...

This changes the provisionComposer.js script used for the 'handoff to azure admin' scenario in the following ways:

* correct names of resources in the resulting publish profile if resource group is specified
* remove access token field
* remove luisResource field if luis not created
* add tenantId field

## Task Item

Fixes #7539


beyackle:
image
For reference, this is the only change inside the zip file.

beyackle (approved):

Issue #7587, opened 2021-05-04
Fit & finish - UI Copy | Create/connect to a skill incl. orchestrator


mareekuh:
Fit & finish - UI Copy | Create/connect to a skill incl. orchestrator

White cells = existing copy
Yellow/blue cells = new copy

Link to 'live' copy on ' Add a skill ' tab: 
https://microsoft.sharepoint.com/:x:/r/teams/ConversationalAI785/_layouts/15/Doc.aspx?sourcedoc=%7B6B1FACBD-B5FE-4AD2-A35B-995635C1696B%7D&file=glossary.xlsx&action=edit&mobileredirect=true&wdPreviousSession=36539654-a42b-4b87-be1b-de96507b4422&wdOrigin=TEAMS-ELECTRON.teams.undefined&cid=676cf454-61e0-4182-bd27-39044bb51cea

![Enable orchestrator](https://user-images.githubusercontent.com/48503434/116964136-a54ba200-ac5f-11eb-8d7c-db76b99b98a2.png)
![ Create dialog](https://user-images.githubusercontent.com/48503434/116964138-a67ccf00-ac5f-11eb-8480-50782619fb70.png)
![Connect to skill](https://user-images.githubusercontent.com/48503434/116964139-a67ccf00-ac5f-11eb-9e6f-9580f23a703a.png)
![Which bots allowed](https://user-images.githubusercontent.com/48503434/116964141-a7156580-ac5f-11eb-8d5c-ba6703180bf5.png)
![Export as skill](https://user-images.githubusercontent.com/48503434/116964144-a7156580-ac5f-11eb-888e-40ed178f11ea.png)
![add menu - skill](https://user-images.githubusercontent.com/48503434/116964146-a7adfc00-ac5f-11eb-834d-5ae1a12bdc9e.png)

sgellock:
Shiproom - approving as today is fit and finish day.  must be merged by EOD today

sgellock:
Shiproom - approved.  @beyackle can you get this merged ASAP please?  thank you

PR #7639, opened 2021-05-05
fix: rephrase Enable Orchestrator dialog

beyackle:
## Description

This changes the wording of the "Enable Orchestrator" dialog that pops up when you add a second bot to a project. Note that the ref'd ticket also has what appears to be a table of recognizer types and descriptions; I was unable to find these anywhere in the code, so I'm leaving this as a "refs" and not a "closes" until we can address that too.

## Task Item

refs #7587

## Screenshots

image

### DIFF
--- a/Composer/packages/client/src/constants.tsx
+++ b/Composer/packages/client/src/constants.tsx
@@ -199,8 +199,10 @@ export const BotStatusesCopy = {
 export const DialogCreationCopy = {
   get CREATE_OPTIONS() {
     return {
-      title: formatMessage('Open your Azure Bot resource'),
-      subText: formatMessage('Do you want to create a new bot, or connect your Azure Bot resource to an existing bot?'),
+      title: formatMessage('Your new Azure Bot is available in Composer'),
+      subText: formatMessage(
+        'The Azure Bot created in Azure Bot Services contains bot resources that can be used as the basis for a new bot, or to add or replace resources of an existing bot.'
+      ),
     };
   },
   get CREATE_NEW_BOT() {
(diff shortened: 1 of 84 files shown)


Composer/packages/client/src/constants.tsx
```diff
@@ -358,14 +367,11 @@ export const selectIntentDialog = {
...
-    return formatMessage('Enable Orchestrator Recognizer');
+    return formatMessage('Enable Orchestrator');
   },
   get subText() {
-    return formatMessage('Enable Orchestrator as the recognizer for routing to other skills');
-  },
-  get content() {
     return formatMessage(
-      'Multi-bot projects work best with the Orchestrator recognizer set at the dispatching dialog (typically the root dialog). Orchestrator helps identify and dispatch user intents from the root dialog to the respective skill that handles the intent. Orchestrator does not support entity extraction. If you plan to combine entity extraction and routing at the root dialog, use LUIS instead.'
+      'A bot that consists of multiple bots or connects to skills (multi-bot project) needs Orchestrator to detect and route user input to the appropriate bot or skill.'
```
taicchoumsft, line 374: @vivekkshankar , when we did the second pass on this text, we were trying to convey that we **recommend** Orchestrator for this scenario (after all, the user can say no to using it), but the third pass here has changed back to **need** orchestrator... Is this right?

beyackle: FYI - This is on line 136 on the "Add a skill" tab of the shared spreadsheet. I'm taking all of my changes from there.

taicchoumsft: Yeah, sorry I should've put that comment in that spreadsheet. I'll do that instead.

cwhitten (approved):

PR #7640, opened 2021-05-05
fix: Orchestrator Node Install for Skill Flow

### LINKED ISSUE #7635: Orchestrator Skill Dialog pops when using Node runtime, but won't install correctly
## Describe the bug
Orchestrator skill dialog prompts after creating a new skill, local or remote.   But we currently only support installing the Orchestrator C# package - if user clicks on `Continue` when using NodeJs, the Orchestrator installer fails silently, since it tries to install the nuget package in a Node bot (there is no Node installer)

## Version
Main 7b94964

## To Reproduce
1. Create a new NodeJs bot
2. Add a remote or local skill
3. See Orchestrator dialog prompt, click Continue
4. Error message pops up shortly after, saying install failed

## Expected behavior
~Don't show prompt for Orchestrator skill dialog unless bot is using new adaptive .net webapp runtime.~
Show this dialog and Install Orchestrator package for Adaptive C# and NodeJS runtimes.

taicchoumsft:
## Description
~Add an additional check to render the EnableOrchestrator dialog during Skill flows only if we're using the new adaptive dotnet webapp runtime.~

This PR has been updated to not hide the orchestrator skill dialog for node, but instead fix the Orchestrator package install for node.

Render the `EnableOrchestrator` dialog only if we're using the adaptive runtime, and add the Orchestrator Recognizer install for Node when user clicks Continue on this dialog.

## Task Item
fixes #7635


hatpick (approved):

boydc2014:
@taicchoumsft there seems to be conflicts must be resolved

boydc2014 (approved):

PR #7645, opened 2021-05-05
fix: Added cleanup logic to app on quit

### LINKED ISSUE #7521: Composer doesn't kill bots on shutdown
## Describe the bug

Composer is not consistently killing running bots on shutdown.  Gif of the behavior is below.  Oddly, you can also see that once Composer shuts down, a second bot process appears in task manager, which I cannot explain either.

Bot process in this case is ContosoInsurance.

## Version

Version:  2.0.0-nightly.240870.919a971
Electron: 8.2.4
Chrome: 80.0.3987.165
NodeJS: 12.13.0
V8: 8.0.426.27-electron.0

## OS

Windows 10

## To Reproduce

1. Open Composer, start the bot
2. Open task manager and find the bot process
3. Kill Composer with the x in the top right
4. Process doesn't end.

## Expected behavior

Composer should correctly kill the process when shutting down.

## Screenshots

Composer Not Killing Bots on Shutdown

tonyanziano:
## Description

If a bot is running, and you close Composer, the process running the bot will be orphaned. This PR adds logic that handles the quit and sends a signal to the client to stop all the currently running bots.

## Task Item

fixes #7521


Composer/packages/client/src/App.tsx
```diff
@@ -32,6 +34,7 @@ export const App: React.FC = () => {
     checkNodeVersion();
     fetchExtensions();
     fetchFeatureFlags();
+    performAppCleanupOnQuit();
```
tdurnford, line 37: Shouldn't this be invoke when electron sends the `'cleanup'` event? How is this supposed to be triggered when the user is running it in the browser?

```js
const { ipcRenderer } = window;

useEffect(() => {
      performAppCleanupOnQuit();
  });
}, []);
```

tonyanziano: It is essentially a no-op for the browser.

My understanding is that since the server (Node process) that runs the app in the browser is responsible for spawning the bot processes, the processes should shut down when the user stops the server. The user exiting the browser tab shouldn't necessarily stop the server.

tonyanziano: Oh I see now. I was moving files around and I forgot to add the ipc code that I previously had. Yes, it should have a listener like that.

hatpick (commented):
Here are the steps:
1. User closes app
2. Electron main emits ```before-quit```
3. an ipc call is made to renderer
4. ipc handler on renderer does a bunch of async tasks and then makes an ajax call back to server who sits on main side
5. Server handler does a bunch of async calls to stop other processes

(meanwhile electron main is quitting)

What's the guarantee that all above are successful? Does electron wait for ```before-quit``` ipc/async/ajax calls to finish before it moves on to quit the app?

tonyanziano:
@hatpick I was thinking about this as well. After looking through the docs, I don't think there's a direct synchronous way for the electron process to wait on the client. (You can do this from client -> main, but not the other way.)

Maybe what I can do is in the "before quit" handler, I can prevent the quit from happening, and set up a one-time listener that is invoked once the client sends back the "all-clear." This handler would then manually call `app.quit()`;

hatpick:
@tonyanziano your proposal is what I was thinking about. Mac terminal does the same thing, for example when it's running a node process and you want to close the terminal all together, it shows a popup asking "are you sure you want to terminate" and then on terminate it gracefully shuts down the running process then closes the terminal window.

tonyanziano:
Ok let me take a stab at it.

tonyanziano:
Soroush and I talked offline, and we have a good idea on how to guarantee / improve the application cleanup process but it's a little bit out of scope for this P0 fix.

For now we think that the current implementation in this PR should be sufficient, and we will talk about building on top of it in R14.

hatpick (approved):

PR #7683, opened 2021-05-06
fix: exclude items from the package manager list that do not have names

### LINKED ISSUE #7680: Package manager listing components that aren't packages
## Describe the bug

If you add a local component (a component added directly to your app code, rather than through a package) to your bot project, Package Manager will display blank lines for them.

## Version

daily

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Create an empty bot
2. Open your bot project in Visual Studio
3. Add the multiply component as outlined here: https://github.com/microsoft/BotBuilder-Samples/tree/ComposerSamples/composer-samples/csharp_dotnetcore/packages/MultiplyDialog
  4. Alternatively, download the attached .zip and open the solution there. 
4. Open your bot in Composer
5. Go to Package Manager > Installed tab
6. Observe blank lines

## Expected behavior

Only packages are enumerated in the list of installed packages.

## Screenshots

image

## Additional context

[CreateComponents (2).zip](https://github.com/microsoft/BotFramework-Composer/files/6436794/CreateComponents.2.zip)

benbrown:
Fixes #7680  -

The schema merge process returns locally developed custom actions as if they were installed components.  These entries, however, lack all metadata including the name and version.

This PR filters items without both a name and a version from appearing in the package manager UI.

### DIFF
--- a/extensions/packageManager/src/node/index.ts
+++ b/extensions/packageManager/src/node/index.ts
@@ -28,7 +28,7 @@ const hasSchema = (c) => {
 };
 
 const isAdaptiveComponent = (c) => {
-  return hasSchema(c) || c.includesExports;
+  return c.name && c.version && (hasSchema(c) || c.includesExports);
 };
 
 const readFileAsync = async (path, encoding) => {


beyackle (approved):

extensions/packageManager/src/node/index.ts
```diff
@@ -28,7 +28,7 @@ const hasSchema = (c) => {
 };
 
 const isAdaptiveComponent = (c) => {
-  return hasSchema(c) || c.includesExports;
+  return c.name && c.version && (hasSchema(c) || c.includesExports);
```
GeoffCoxMSFT, line 31: Nit - you could do c?.name to handle c being undefined since c doesn't have any type, but since this is internal you can protect it from the caller.

cwhitten (approved):

Issue #7157, opened 2021-04-20
dotnet build errror when starting a bot project (because it is being used by another process)


darrenj:
## Describe the bug

I've seen this a number of times now when switching between different bot projects. Bots start to report errors when starting due to the .exe being used by another process. bot project has only been launched as part of using Composer, VS hasn't been opened against the csproj or anything outside composer.

Digging around in task manager it seems djbfskill6 had a process running in the background, Composer perhaps isn't terminating these processes correctly in some instances?

```
Error occurred building the bot

Message:
Command failed: dotnet build djbfskill1.csproj
Command:
dotnet build djbfskill1.csproj
Stack Trace:
Error: Command failed: dotnet build djbfskill1.csproj

    at ChildProcess.exithandler (child_process.js:308:12)
    at ChildProcess.emit (events.js:210:5)
    at maybeClose (internal/child_process.js:1021:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:283:5)
Output:
Microsoft (R) Build Engine version 16.9.0+57a23d249 for .NET
Copyright (C) Microsoft Corporation. All rights reserved.

  Determining projects to restore...
C:\Users\darrenj\djbfskill1\djbfskill1\djbfskill1.csproj : warning NU1701: Package 'Microsoft.Azure.KeyVault.Core 1.0.0' was restored using '.NETFramework,Version=v4.6.1, .NETFramework,Version=v4.6.2, .NETFramework,Version=v4.7, .NETFramework,Version=v4.7.1, .NETFramework,Version=v4.7.2, .NETFramework,Version=v4.8' instead of the project target framework '.NETCoreApp,Version=v3.1'. This package may not be fully compatible with your project.
  All projects are up-to-date for restore.
C:\Users\darrenj\djbfskill1\djbfskill1\djbfskill1.csproj : warning NU1701: Package 'Microsoft.Azure.KeyVault.Core 1.0.0' was restored using '.NETFramework,Version=v4.6.1, .NETFramework,Version=v4.6.2, .NETFramework,Version=v4.7, .NETFramework,Version=v4.7.1, .NETFramework,Version=v4.7.2, .NETFramework,Version=v4.8' instead of the project target framework '.NETCoreApp,Version=v3.1'. This package may not be fully compatible with your project.
[...]

PR #7689, opened 2021-05-06
fix: (Un) Installing a package now stops the currently running bot beforehand

### LINKED ISSUE #7498: Installing a package while a bot is running results in multi-project build errors
## Describe the bug
If you install a package while a bot is running the underlying process isn't fully killed. As a result Composer gets into a bad state where the `Restart bot` button becomes `Start bot` but there's already an process running. 

## Version
Version:  2.0.0-nightly.241057.db66154
Electron: 8.2.4
Chrome: 80.0.3987.165
NodeJS: 12.13.0
V8: 8.0.426.27-electron.0

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Create or open a bot and start it
2. Install a package (I did the Teams package)
3. Note how the `Restart bot` button becomes `Start bot` after the package is installed
3. Click the `Start bot` button

## Expected behavior
The button should either be `Start bot` where the underlying process is actually killed, or it should be `Restart bot` assuming restart rebuilds with any delta in the packages

## Additional context

Error occurred building the bot

Message:
Command failed: dotnet build TeamsTest.csproj
Command:
dotnet build TeamsTest.csproj
Stack Trace:
Error: Command failed: dotnet build TeamsTest.csproj

    at ChildProcess.exithandler (child_process.js:308:12)
    at ChildProcess.emit (events.js:210:5)
    at maybeClose (internal/child_process.js:1021:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:283:5)
Output:
Microsoft (R) Build Engine version 16.9.0+57a23d249 for .NET
Copyright (C) Microsoft Corporation. All rights reserved.

tonyanziano:
## Description

The package manager was not stopping the bot if it was running which was causing processes to be orphaned. These orphaned processes were blocking certain actions within the app -- like starting a bot -- from accessing the assets because they were busy.

This PR exposes a `stopBot()` API to the shell API that allows extensions to stop bots. The package manager now uses this API.

@srinaath also found a bug in the bot controller that was showing "Starting bot(s)..." when they were being stopped explicitly. This PR also fixes that bug.

## Task Item

Fixes #7498, #7157

### DIFF
--- a/Composer/packages/types/src/shell.ts
+++ b/Composer/packages/types/src/shell.ts
@@ -147,6 +147,7 @@ export type ProjectContextApi = {
   updateDialogSchema: (_: DialogSchemaFile) => Promise<void>;
   createTrigger: (id: string, formData, autoSelected?: boolean) => void;
   createQnATrigger: (id: string) => void;
+  stopBot: (projectId?: string) => Promise<void>;
   updateSkill: (skillId: string, skillsData: { skill: Skill; selectedEndpointIndex: number }) => Promise<void>;
   updateRecognizer: (projectId: string, dialogId: string, kind: LuProviderType) => void;
 };
--- a/Composer/packages/types/src/shell.ts
+++ b/Composer/packages/types/src/shell.ts
@@ -147,7 +147,7 @@ export type ProjectContextApi = {
   updateDialogSchema: (_: DialogSchemaFile) => Promise<void>;
   createTrigger: (id: string, formData, autoSelected?: boolean) => void;
   createQnATrigger: (id: string) => void;
-  stopBot: (projectId?: string) => void;
+  stopBot: (projectId: string) => void;
   updateSkill: (skillId: string, skillsData: { skill: Skill; selectedEndpointIndex: number }) => Promise<void>;
   updateRecognizer: (projectId: string, dialogId: string, kind: LuProviderType) => void;
 };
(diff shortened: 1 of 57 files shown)


Composer/packages/types/src/shell.ts
```diff
@@ -147,6 +147,7 @@ export type ProjectContextApi = {
   updateDialogSchema: (_: DialogSchemaFile) => Promise<void>;
   createTrigger: (id: string, formData, autoSelected?: boolean) => void;
   createQnATrigger: (id: string) => void;
+  stopBot: (projectId?: string) => void;
```
srinaath, line 150: Can we remove the conditional projectId since we are passing currentProjectID to the function @tonyanziano ?

tonyanziano: sure

benbrown (commented):
Seems like stopBot should be awaited before calling the installComponentApi?

tonyanziano:
@srinaath care to comment on why you removed the await on `stopBot()`?

srinaath:
@benbrown the idea behind not awaiting stopBot is we want the hook to be like a fire and forget model. We would use recoil to get the status of the bot inside the component rather than awaiting inside the component. Is there a reason to stop before installing?

srinaath:
The problem we had was resetting of the bot state (due to reloadProject) orphaned the process. So the solution we implemented was to stop the bot to kill the process which could happen pre/post install of component. Once the project is reloaded the "Start bot" would spawn a new process.

PR #7720, opened 2021-05-10
fix: breadcrumb & show code button collision

### LINKED ISSUE #7705: Breadcrumb & toggle code button collision
## Describe the bug

Depending on the length of the breadcrumb items and width of the composer window, the following can happen:
Screen Shot 2021-05-07 at 11 48 05 AM

## Version

Latest on main

## Browser

All browsers and electron.

## OS

All platforms.

### LINKED ISSUE #7762: "Show Code" label overlaps on long names
## Describe the bug

The "Show Code" label can overlap the dialog > trigger name.

## Version

Daily

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Create a dialog with a long name
2. Add a trigger with a long name
3. Select your new trigger
4. Horizontally resize your Composer app
5. Observe error

## Expected behavior

Names do not overlap.

## Screenshots

image

zhixzhan:
## Description
Fix breadcrumb & show code button collision

## Task Item
close #7705 
close #7762

## Screenshots

image


Composer/packages/client/src/pages/design/styles.ts
```diff
@@ -103,9 +103,23 @@ export const formEditor = css`
...
 `;
 
+export const visualPanelHeaderContainer = css`
+  height: 65px;
+`;
+
+export const visualPanelHeaderLeft = css`
+  overflow: hidden;
+  float: left;
+  width: calc(100% - 100px);
```
hatpick, line 113: why dont you use ```flex``` so the browser itself decides how much width give to each side. With this width and float, we will still have the issue if ```show code``` needs more than 100px to render in a certain locale.

hatpick (changes requested):

Composer/packages/client/src/pages/design/styles.ts
zhixzhan: @hatpick Make sense, updated now.

hatpick:
@mewa1024 can you chime in here? We are looking for an acceptable UX when the screen size is not wide enough to fit a long breadcrumb path AND show/hide code button.

Here's what I propose:
1. If it has enough space, it shows it without issues.
2. When there are not enough space, it gives show/hide code whatever space it needs, and allows horizontal scrolling on the breadcrumb.
https://user-images.githubusercontent.com/1483492/117774474-659c3180-b1ee-11eb-8ba6-7270f2778fa7.mp4

Composer/packages/client/src/pages/design/styles.ts
```diff
@@ -103,9 +103,18 @@ export const formEditor = css`
...
 `;
 
+export const visualPanelHeaderContainer = css`
+  display: flex;
+  justify-content: space-between;
+  height: 65px;
+`;
+
+export const visualPanelHeaderShowCodeButton = css`
+  line-height: 65px;
```
hatpick, line 113: instead of this, add align-items: center to the container on L106

mewa1024:
Can we follow this behavior in Fluent when there is not enough space?
https://developer.microsoft.com/en-us/fluentui#/controls/web/breadcrumb


**MIddle items are collapsed into a ... menu:**

   image


**The ... item expands to show all the in between items which are clickable:**

   image

zhixzhan:
@hatpick @srinaath To implement above behavior in fluentui, seems the only way is to use javascript  calculate the size of breadcrumb element, do you think it make sense, or have better suggestions?

srinaath:
@zhixzhan, So the example with no maxDisplayedItems we should be using that option. Essentially, the way u would do something like this, pseudo coding here
```
<div style="display: flex;"> 
    <div style="width: 80%">
          <Breadcrumb ariaLabel="Breadcrumb with no maxDisplayedItems" 
           overflowAriaLabel="More links" />
      </div>   
      <div style="width: 20%">
         Show Code
       </div>
</div>
```
So, the breadcrumb starts shrinking and always occupies only a particular width in percentage. Does that work?

hatpick:
@mewa1024 I agree with @zhixzhan, if we simply say always truncate the middle items and keep first and last item in the view, we can still have the following issue based on the size of the window and length of texts:
FIRST ITEM > ... > LAS [BUTTON]
FIRST ITEM > .. [BUTTON]
FIR [BUTTON]
etc.
Even if we do calculations as @zhixzhan suggested, these situations can happen.
My suggestion is doing horizontal scrolling + Breadcrumb capabilities of moving items in overflow menu, so in case any of the above happens, user can still scroll horizontally to find breadcrumb first and last item, as well as the overflow menu in between.

Thoughts?

mewa1024:
I like @srinaath's suggestion.

zhixzhan:
@srinaath @hatpick @mewa1024 PR updated, according to the codepen example, thanks for your help.

hatpick (approved):

PR #7722, opened 2021-05-10
fix: incorrect expression validation for dialog

### LINKED ISSUE #7718: Incorrect validation on OutputFormat blocks bot to start
When supporting customer on https://github.com/microsoft/BotFramework-Composer/issues/7312 i found that, the latest nightly's validation on outputFormat is wrong, causing a correct term fail to start bot

image

lei9444:
## Description
**root cause**
some dialog filed schema use expression (the type is string) and normal types, but validator use the type string as return type directly, so some fields like the outputformat in numberInput is judged as string, but the return type is number in fact.

**fix**
  case:
    json filed type is string;
    the $role is expression;
    there is no oneOf;

    we don't know which return type is correct,
    so we think the expression support all return types.

  case:
    the $role is expression;
    the oneOf has some normal types and there is no expression in oneOf

    the return type is one of the types in oneOf filed.

  case:
    the $role is expression;
    the oneOf has some normal types and there is an expression role in oneOf

    the filed maybe a string expression and the expression return type is one of the types in oneOf filed besides the type with expression role.

## Task Item
closes #7718


Composer/packages/lib/indexers/src/validations/expressionValidation/types.ts
```diff
@@ -11,6 +11,7 @@ export const StringMapExpressionType = {
   object: ReturnType.Object,
   array: ReturnType.Array,
   integer: ReturnType.Number,
+  all: 31,
```
boydc2014, line 14: How about not hard-coding this value?

boydc2014 (dismissed):
LGTM, please hold on merge till ship-room's triage

Composer/packages/lib/indexers/src/validations/expressionValidation/utils.ts
```diff
@@ -41,22 +41,54 @@ export function findRequiredProperties(schema: any): { [key: string]: boolean }
   return required;
 }
 
-export function findTypes(schema: any): number[] {
+/*
+  case:
+    json filed type is string;
```
tdurnford, line 46:
```suggestion
    json field type is string;
```

Composer/packages/lib/indexers/src/validations/expressionValidation/types.ts
```diff
@@ -11,6 +11,7 @@ export const StringMapExpressionType = {
   object: ReturnType.Object,
   array: ReturnType.Array,
   integer: ReturnType.Number,
+  all: ReturnType.Number + ReturnType.String + ReturnType.Boolean + ReturnType.Object + ReturnType.Array,
```
hatpick, line 14: semantically, this means more like an ```any```, what do you think?

Composer/packages/lib/indexers/src/validations/expressionValidation/utils.ts
```diff
@@ -41,22 +41,54 @@ export function findRequiredProperties(schema: any): { [key: string]: boolean }
...
+
+    the return type is one of the types in oneOf filed.
+
+  case:
+    the $role is expression;
+    the oneOf has some normal types and there is a expression role in oneOf
+
+    the filed maybe a string expression and the expression return type is one of the types in oneOf filed besides the type with expression role.
+*/
+export function findTypes(schema: JSONSchema7): number[] {
```
hatpick, line 65: can you make this an arrow function, and great explanation above, but can you add a jsdoc style comment right above the function explaining the input and output? Currently I have no idea why is the function is returning an array of numbers and what these numbers represent. (can type ```number``` be replaced by ```StringMapExpressionType```)

hatpick: I do understand what it does, adding comments helps other devs coming across this code understand it quicker too. :)

hatpick (approved):

PR #7756, opened 2021-05-11
DO NOT MERGE fix: exclude node_modules from deploy asset

### LINKED ISSUE #7743: A JS bot that includes native package does not run when deployed to Azure
## Describe the bug

Create a JS Function bot that includes Orchestrator, LUIS, and QnA.  The bot will run locally.  However, when deployed it does not function through Web Chat.

It does not appear to be specific to Functions though.  Josh G will be providing details on required changes.

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1) Create a JS Bot
2) Install Orchestrator and and LUIS intent and QnA KB
3) Provision
4) Publish
5) Test via Web Chat on Azure.

benbrown:
## Description

This is the composer side of the fix for #7743 -- 

ultimately, to get orchestrator's binaries properly installed we need to:

* exclude the node_modules folder from the deploy artifact
* add a .deployment file to the generator to instruct kudu to do the npm install step

## Task Item

fixes #7743

### DIFF
--- a/Composer/packages/client/__tests__/components/skill.test.tsx
+++ b/Composer/packages/client/__tests__/components/skill.test.tsx
@@ -16,36 +16,54 @@ jest.mock('../../src//utils/httpUtil');
 
 jest.mock('../../src/components/Modal/dialogStyle', () => ({}));
 
-let recoilInitState;
 const projectId = '123a.234';
 
-describe('Skill page', () => {
-  beforeEach(() => {
-    recoilInitState = ({ set }) => {
-      set(currentProjectIdState, projectId);
-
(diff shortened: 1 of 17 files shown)


GeoffCoxMSFT (approved):

srinaath:
If node_modules is removed from deployment do we still need this here @benbrown  https://github.com/microsoft/BotFramework-Composer/blob/main/extensions/runtimes/src/index.ts#L260

PR #7782, opened 2021-05-13
fix: add inner scrollbar to selection area

### LINKED ISSUE #7780: [UI bug] [Skills] Select all checkbox not visible in select dialog and trigger screen
image

**Expected UI:** 
- Replace circular checkbox with standard checkbox control
- Place select all checkbox in-line as shown in mock
- Select all by default

image

alanlong9278:
## Description
move scroll bar to selection area, remove outer scroll bar & `stick select all` button
image

## Task Item

closes #7780


hatpick (commented):
Why is scrollbars are showing on the header of that detailslist as well?

VanyLaw:
> Why is scrollbars are showing on the header of that detailslist as well?

I think it's by design. And this PR is only fix the select All button non-shown issue.

luhan2017 (approved):

hatpick:
> > Why is scrollbars are showing on the header of that detailslist as well?
> 
> Because we used the DetailList component of the Fluent UI package, and it's by design of the DetailList Component. If we want to just scroll the list content, we need to implement the component by ourselves, but it will cause huge change, so we decide to change in next release.

for your reference, so you dont implement it from scratch in R14:
https://codepen.io/hatpick/pen/yLMORGQ

cwhitten (approved):

Issue #7788, opened 2021-05-13
Composer becomes unresponsive after creating and testing a bot


xiyangdesign:
## Describe the bug
Composer is having huge delay to any actions (e.g. click on buttons, resizing window, etc.) Web chat pane in the debugging window seems to be logging a lot. This is after I created a CoreWithLanguageUnderstanding bot, added a trigger and tested it. 

I ran into this issue yesterday as well, without doing anything. Reinstalling Composer didn't help. (Restarting my laptop solved the issue)

See video below

https://user-images.githubusercontent.com/74994573/118157350-4b6d7980-b3cf-11eb-8184-67f85491e59a.mov

## Version
![image](https://user-images.githubusercontent.com/74994573/118156643-74d9d580-b3ce-11eb-853e-eaff3c24ecb8.png)

## Browser

- [X] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [X] macOS
- [ ] Windows
- [ ] Ubuntu

## To Reproduce

_Note: I ran into this issue yesterday as well, without doing anything. Reinstalling Composer didn't help. (Restarting my laptop solved the issue)_
1. Create a bot using CoreWithLanguageUnderstanding template
2. Create a trigger, added an action: ask for user name
3. Start bot
4. Open webchat
5. In the Webchat panel, click "restart conversation" a couple times quickly
6. Error occurs
7. Restart Composer sees another error (see screenshot)

![image](https://user-images.githubusercontent.com/74994573/118157100-fd587600-b3ce-11eb-91f0-b76bb77925f1.png)

emivers8:
repro of #7561

darrenj:
Reproed on Windows too.

cwhitten:
@darrenj were your repro steps exactly as @xiyangdesign 's?

```
1. Create a bot using CoreWithLanguageUnderstanding template
2. Create a trigger, added an action: ask for user name
3. Start bot
4. Open webchat
5. In the Webchat panel, click "restart conversation" a couple times quickly
6. Error occurs
7. Restart Composer sees another error (see screenshot)
```

sgellock:
Shiproom - approved.  @srinaath can you merge today, ASAP please.

srinaath:
Will do @sgellock

cwhitten:
Throttle PR merged, issue to be left open for additional investigation

PR #7824, opened 2021-05-14
fix: Throttle restart conversation

srinaath:
## Description
This PR addresses a scenario where "Restart Conversation" was hit quite a few times in under a second. In certain situations, it freezes Composer. This is a partial fix to avoid Composer getting into this state by a) throttling the button (1s) b) Disabling it while the bot is being restarted c) Disabling the button until the asynchronous operation of restarting the conversation completes.

The second half of the fix would be in a separate PR. This would involve performing cleanup tasks on electron before-quit event. The cleanup tasks include code written to manually close the directline server, runtime log server and the express server. This could be leveraged for closing dangling bot runtimes as well.

## Task Item
Refs #7788

### DIFF
--- a/Composer/packages/client/src/components/WebChat/WebChatPanel.tsx
+++ b/Composer/packages/client/src/components/WebChat/WebChatPanel.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-import React, { useMemo, useEffect, useState, useRef } from 'react';
+import React, { useMemo, useEffect, useState, useRef, useCallback } from 'react';
 import { useRecoilValue } from 'recoil';
 import {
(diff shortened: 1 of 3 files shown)


Composer/packages/client/src/components/WebChat/WebChatPanel.tsx
```diff
@@ -134,6 +136,14 @@ export const WebChatPanel: React.FC<WebChatPanelProps> = ({
     }
   }, [botUrl]);
 
+  useEffect(() => {
+    if (botStatus === BotStatus.connected) {
```
tdurnford, line 140: This can be done in a single line
```
setIsRestartButtonDisabled(botStatus === BotStatus.connected);
```

tonyanziano (approved):

cwhitten (approved):

PR #7842, opened 2021-05-15
fix: Always use intermediate lg template for text and speak modalities

### LINKED ISSUE #7841: Response editor dereferences/expands LG templates in the UI, making it difficult to know if the template was saved
## Describe the bug

If I create an LG template via the Bot Responses page, and then use a reference to that template in a response editor, the response editor renders it as the first response in the template. This is confusing.

## Version

Version:  2.0.0-nightly.245345.48d2167
Electron: 8.2.4
Chrome: 80.0.3987.165
NodeJS: 12.13.0
V8: 8.0.426.27-electron.0

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [x] macOS
- [ ] Windows
- [ ] Ubuntu

## To Reproduce

1. Create a bot. Add a **Send a response** node to a trigger and enter some text response.
2. Go to the **Bot responses** page.
3. Go to the dialog with the **Send a response** node you just created and add a template. I was following the tutorial in our docs, so I used this:

```
# DescribeWeather(weather)
- It is "${dialog.weather}" in ${user.postalcode} and the temperature is ${dialog.fahrenheit}&deg;F or ${dialog.celsius}&deg;C. Have a nice day.
```

3. Go back to the response editor. Replace the existing text response with a reference to the template you created:
imageimage

4. When you hit Enter, the reference appears in the table of text responses and the first response in the template is shown in the node on the canvas. image

5. Navigate to another page.
6. Go back to this node, and you will see that in the list of text responses, the template reference has been replaced by its first response variation, in this example, the text is "**It is "${dialog.weather}" in ${user.postalcode} and the temperature is ${dialog.fahrenheit}&deg;F or ${dialog.celsius}&deg;C. Have a nice day.**": 
image

This is confusing because it looks like the template reference was expanded or, if the first response the template is very similar to the text response you copied over, it looks like the change was not saved.

Looking at the code editor shows that the reference was actually saved: 
    image image
but a new user might not know to look there.

## Expected behavior

I'd expect the text response to show the reference, like this, especially since that's how it appeared when I first entered it (step 4 above, before navigating away from the page):
image

## Additional context

When I went through this flow, I thought the template reference was not saved, so I copied the reference again in the response editor. That created this error:
image
also in here:
image

and also removed my template from the Bot Responses page so when I clicked on the link under "Location", nothing happened.

tdurnford:
## Description
Updates the `useStringArray` hook to always use an intermediate template for the text and speak modalities in the LG Response Editor.

## Task Item
Closes #7841

## Screenshots
May-15-2021 07-34-42

### DIFF
--- a/Composer/packages/lib/code-editor/src/lg/hooks/useStringArray.ts
+++ b/Composer/packages/lib/code-editor/src/lg/hooks/useStringArray.ts
@@ -76,9 +76,6 @@ export const useStringArray = <T extends ArrayBasedStructuredResponseItem>(
         setTemplateId(id);
         onUpdateResponseTemplate({ [kind]: { kind, value: [], valueType: 'direct' } });
         onRemoveTemplate(id);
-      } else if (fixedNewItems.length === 1 && !/\r?\n/g.test(fixedNewItems[0].value) && lgOption?.templateId) {
-        onUpdateResponseTemplate({ [kind]: { kind, value: [fixedNewItems[0].value], valueType: 'direct' } });
-        onTemplateChange(id, '');
       } else {
         setTemplateId(id);
         onUpdateResponseTemplate({ [kind]: { kind, value: [`\${${id}()}`], valueType: 'template' } });


hatpick:
@mewa1024 what do you think?

mewa1024:
> @mewa1024 what do you think?

this makes sense to me. And the user wanted to replace the intermediary template, they could edit the activity template directly, replacing the template for that mode? like this:
image

image

?

hatpick:
any template in front of Text or Speak will get dereferenced, no matter how it got there, manually typed by user or generated by the editor. Does that answer your question? @mewa1024

tdurnford:
@mewa1024 Yes, if the user wanted to change `SendActivitvity_123abc_text` to `TemplateCreatedInBotResponses`, they could do that in the code editor; however, it they switched back to the response editor, it would expand `TemplateCreatedInBotResponses` and they wouldn't see `${TemplateCreatedInBotResponses}` as item in the list.

mewa1024:
> @mewa1024 Yes, if the user wanted to change `SendActivitvity_123abc_text` to `TemplateCreatedInBotResponses`, they could do that in the code editor; however, it they switched back to the response editor, it would expand `TemplateCreatedInBotResponses` and they wouldn't see `${TemplateCreatedInBotResponses}` as an item in the list. The would see the response variations in `TemplateCreatedInBotResponses`.

Excellent. one last question, if they switched back to the response editors and then edited one of the responses, that edit would be saved in `TemplateCreatedInBotResponses`?

tdurnford:
@mewa Yes, if the `Text = ${TemplateCreatedInBotResponses()}, than any changes made in the response editor would appear under TemplateCreatedInBotResponses.

hatpick (approved):

PR #7857, opened 2021-05-18
fix: rewrite l10n:check to catch more errors and fix some l10n issues

beyackle:
## Description

As part of the L10n pipeline rewrite, I've rewritten the `l10n:check` script to use `formatMessage` itself to verify that the JSON files are readable. Many of the ways they aren't can be fixed automatically (which is what those `replace` calls are for), but some will still need manual inspection. I have plans to integrate that into our pipeline; for now, though, this change needs to happen first.

## Task Item

#minor, but fixes a potential for an app-crashing bug if someone using Swedish, Portuguese, or Chinese-Simplified locales happens across the wrong part of Composer.

## Screenshots

image
This is an example of when the app would crash; the `0/2` is part of an ICU string that was formatted wrong.


Composer/scripts/l10nCheck.js
```diff
@@ -6,40 +6,75 @@
...
+let totalCount = 0;
+
+formatMessage.setup({ missingTranslation: 'ignore' });
+
+for (const localePath of localePaths) {
+  let errorCount = 0;
+  console.log(`Checking within ${localePath}:`);
+  const dir = fs.opendirSync(localePath);
+  for (;;) {
+    const dirent = dir.readSync();
```
a-b-r-o-w-n, line 22: Consider using `fs.readdir` instead.

```ts
for (const dirent of fs.readdirSync(localePath) { ... }
```

a-b-r-o-w-n (changes requested):

Composer/scripts/l10nCheck.js
beyackle: 👍 Good idea. I'll fix that.

a-b-r-o-w-n (approved):

PR #7867, opened 2021-05-19

GeoffCoxMSFT:


beyackle:
This was me checking if a script is working, and it looks like it does. We won't need to run it until other stuff merges, though.

PR #7869, opened 2021-05-20
fix: add file path support in add skill modal

### LINKED ISSUE #7864: Load skill manifests from any location
In composer 2.x a user can load a manifest from an http or https endpoint only, but this should be the only way. 

Manifests describe the skill operations but they don't necessarily need to be exposed on an http endpoint, we do this in samples for convenience but a skill provider may choose not to do this and send the manifest and related files to the consumer offband. In this case, I should be able to load the manifest from the file system if I need to. 

image

Also, I think manifests and related lu files should be stored locally with the bot project once the skill is added and only be rechecked when the developer chooses to (sort of like a decision to update nuget packages). Right now, composer tries to retrieve the manifest every time I open the project and that shouldn't be necessary once a skill has been added and a copy of the manifest and lu files are stored with the botproj.

image

VanyLaw:
## Description

support local file as manifest in adding skill scenario.

## Task Item

close #7864


luhan2017:
@VanyLaw I made this PR to draft now, Let's follow Vivek's design to better support the load manifest from local scenario.

Composer/packages/server/src/locales/en-US.json
```diff
@@ -4121,8 +4127,8 @@
   "url_22a5f3b8": {
     "message": "URL"
   },
-  "url_should_start_with_http_or_https_c34632bb": {
-    "message": "URL should start with http:// or https://"
+  "url_should_start_with_http_or_https_or_file_path_o_2b060341": {
+    "message": "URL should start with http:// or https:// or file path of your system"
```
tdurnford, line 4131: @vivekkshankar Can you recommend a better error message?

VanyLaw:
close it and fix #7864 in #8146

PR #7887, opened 2021-05-21
feat: Docker publish

cdonke:
## Description

Creation of a new publish profile to create Docker Images and publish to Local Docker, ACR, Docker Hub or custom Docker registry.
image

image

With logic to query for existing images and fetch tags - Available to Local Docker, ACR and Docker Hub:
image

Algo, allows to create a new tag, validating if exists:
image

When targetting ACR, it looks for available ACR on Subscription:
image
(username and secret must be provided)

At the end, a summary is provided:
image


cdonke:
Unit tests broke because of timeout (30min)...
can I increase it?

PR #7987, opened 2021-06-01
chore: upgrade json-ptr to 2.2

beyackle:
## Description

This clears up some warnings that were preventing us from merging any more changes.

## Task Item

#minor


beyackle:
redoing; something got stuck here

PR #8000, opened 2021-06-03
fix: shortcuts for paste in canvas does not work

### LINKED ISSUE #7958: Copy Paste in canvas does not work and throws
## Describe the bug

Copy does not seem copying anything, nothing seems happening when I selected an item in the canvas and then clicked Copy.
When Ctrl+V, an error pops up, and seems like even when the clip board is empty, the error still pops up

## Version

2.0

## Browser

- [X] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [X] Windows
- [ ] Ubuntu

## To Reproduce

image

image

lei9444:
## Description
**Root cause**
composer doesn't support ctrl+v today, so if we use this shortcut will throw error for undefined selected actions in canvas.

**fix**
add ctrl+v behaviors. 

1. If some actions are selected, paste it to the end of the last selected action.
2. if there is no selected action, paste it to the end of the current dialog 

## Task Item
closes #7958

## Screenshots
paste

### DIFF
--- a/Composer/packages/adaptive-flow/src/adaptive-flow-editor/hooks/useEditorEventApi.ts
+++ b/Composer/packages/adaptive-flow/src/adaptive-flow-editor/hooks/useEditorEventApi.ts
@@ -216,15 +216,35 @@ export const useEditorEventApi = (
         break;
       case NodeEventTypes.PasteSelection:
         handler = () => {
-          const currentSelectedId = selectedIds[0];
-          if (currentSelectedId.endsWith('+')) {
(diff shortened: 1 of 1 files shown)


hatpick (approved):

srinaath:
@lei9444 have we made sure Ctrl + v and Cmd + v work fine for both platforms?

lei9444:
Hi, @srinaath I have tested in mac with cmd-v and win10 with ctrl-v, all worked well.

srinaath (approved):

boydc2014 (approved):

PR #8005, opened 2021-06-03
fix: resolving getting started teaching callout not rendering

### LINKED ISSUE #7734: Teaching bubble for LUIS isn't popping up
## Describe the bug

I created a new enterprise assistant bot and the teaching bub didn't pop! The rocket menu was open, so I clicked on the LUIS item, set that up, and then the QnA bub and the start bot bub DID pop as expected, it was just the first LUIS one that didn't fire.

When I subsequently created a bot with language, it triggered the LUIS bub as expected.

(bub = teaching bubble)

## Version

Version:  2.0.0-nightly.243997.543d0b4
Electron: 8.2.4
Chrome: 80.0.3987.165
NodeJS: 12.13.0
V8: 8.0.426.27-electron.0
macOS

pavolum:
## Description

Issue
- Hash routing does not work with multi bot templates due to hash lost in redirects.
- We should use recoil state for things like detecting if panes are open or not

Solution
 -Move teachingbubble render state to recoil and respective calls to manage that new state value         

## Task Item

fixes #7734


pavolum:
NOTE: This code change was already approved by @benbrown  when it was merged into the R14_staged branch.
This is a new PR merging into main now that it is open for R14 changes.

Composer/packages/client/src/components/GetStarted/GetStartedNextSteps.tsx
```diff
@@ -402,6 +402,7 @@ export const GetStartedNextSteps: React.FC<GetStartedProps> = (props) => {
             target="#luis"
             onDismiss={() => {
```
tdurnford, line 403: Since there are two `onDismiss` actions with the same code, you should move this logic into a callback.

Composer/packages/client/src/components/Header.tsx
```diff
@@ -191,18 +188,16 @@ export const Header = () => {
...
+    setShowGetStartedTeachingBubble(false);
     setShowGetStarted(newvalue);
   };
 
   // pop out get started if #getstarted is in the URL
   useEffect(() => {
-    if (location.hash === '#getstarted') {
-      setShowGetStartedTeachingBubble(true);
+    if (showGetStartedTeachingBubble) {
       setShowGetStarted(true);
```
tdurnford, line 198: Are there two states to track `showGetStartedTeachingBubble`? Would it be simpler to just use the recoil state as the source of truth instead of maintaining two states?

tdurnford (changes requested):

Composer/packages/client/src/components/Header.tsx
pavolum: There are not two states for `showGetStartedTeachingBubble`. The recoil state value  is the source of truth for the teaching bubble within the getStarted pane.

The confusion is from the other local state value (showTeachingBubble) this is a local state value that controls a different teaching bubble (the teaching bubble that points the user to the start bot feature in the header).

To avoid confusion I have renamed this local state variable so we have `showStartBotTeachingBubble` and `showGetStartedTeachingBubble` respectively.

Composer/packages/client/src/components/GetStarted/GetStartedNextSteps.tsx
tdurnford: Ahh missed that difference.

tdurnford (approved):

PR #8006, opened 2021-06-03
feat: Added UX to link log items and WC activities

### LINKED ISSUE #7983: Link selected bot activities between Web Chat and Log Panel
**P0's**
- [x] Selecting an activity displayed in the Web Chat panel should inspect the activity in the log panel
- [x] Inspecting an activity in the log panel should select an activity in the Web Chat panel

This should resemble how the Emulator behaves.

tonyanziano:
## Description

This PR adds several features to the Web Chat / Web Chat Log UX:

- visual "inspected" state for activities in Web Chat that are being inspected in the log
- selecting a log item scrolls the corresponding Web Chat activity into view
- focusing a (message) activity inside of Web Chat now scrolls the corresponding log item into view and inspects it

## Task Item

Fixes #7983 

## Screenshots

webchat-log-scrolling

### DIFF
--- a/Composer/packages/client/src/components/WebChat/ActivityInspectionListener.tsx
+++ b/Composer/packages/client/src/components/WebChat/ActivityInspectionListener.tsx
@@ -1,0 +1,28 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { FC, useEffect } from 'react';
+import { hooks } from 'botframework-webchat';
+import { useRecoilValue } from 'recoil';
+
+import { rootBotProjectIdSelector, webChatInspectionDataState } from '../../recoilModel';
+
+const { useScrollTo } = hooks;
+
+export const ActivityInspectionListener: FC<{}> = () => {
+  const currentProjectId = useRecoilValue(rootBotProjectIdSelector);
+  const webChatInspectionData = useRecoilValue(webChatInspectionDataState(currentProjectId ?? ''));
+  const scrollToActivity = useScrollTo();
+
+  useEffect(() => {
+    if (
+      webChatInspectionData?.item.trafficType === 'activity' &&
+      webChatInspectionData.item.activity.type === 'message' // TODO: loosen this restriction?
+    ) {
+      scrollToActivity({ activityID: webChatInspectionData.item.activity.id }, { behavior: 'smooth' });
+    }
+  }, [webChatInspectionData]);
+
+  // only listen for when a traffic item is inspected from the log; do not render anything
+  return null;
+};
--- a/Composer/packages/client/src/components/WebChat/ActivityInspectionListener.tsx
+++ b/Composer/packages/client/src/components/WebChat/ActivityInspectionListener.tsx
@@ -17,7 +17,7 @@ export const ActivityInspectionListener: FC<{}> = () => {
   useEffect(() => {
     if (
       webChatInspectionData?.item.trafficType === 'activity' &&
-      webChatInspectionData.item.activity.type === 'message' // TODO: loosen this restriction?
+      webChatInspectionData.item.activity.type === 'message'
     ) {
       scrollToActivity({ activityID: webChatInspectionData.item.activity.id }, { behavior: 'smooth' });
     }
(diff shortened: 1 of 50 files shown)


Composer/packages/client/src/components/WebChat/TranscriptFocusListener.tsx
```diff
@@ -0,0 +1,46 @@
...
+  debugPanelExpansionState,
+  dispatcherState,
+  rootBotProjectIdSelector,
+  webChatTrafficState,
+} from '../../recoilModel';
+import { WebChatInspectorTabKey } from '../../pages/design/DebugPanel/TabExtensions/types';
+
+const { useObserveTranscriptFocus } = hooks;
+
+export const TranscriptFocusListener: FC<{}> = () => {
```
hatpick, line 20: can this be a hook instead of a faceless component?

Composer/packages/client/src/components/WebChat/ActivityInspectionListener.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+import { FC, useEffect } from 'react';
+import { hooks } from 'botframework-webchat';
+import { useRecoilValue } from 'recoil';
+
+import { rootBotProjectIdSelector, webChatInspectionDataState } from '../../recoilModel';
+
+const { useScrollTo } = hooks;
+
+export const ActivityInspectionListener: FC<{}> = () => {
```
hatpick, line 12: can this be a hook instead?

Composer/packages/client/src/components/WebChat/TranscriptFocusListener.tsx
tonyanziano: Good suggestion, let me give that a shot.

tonyanziano: I don't think I can make them hooks, because they have to be within the Web Chat `<Composer />` component to have access to the APIs.

So I could make a component called `<WebChatHooksContainer />` or something and have that be a faceless component with all the hooks inside of it.

a-b-r-o-w-n: Could the hooks just be invoked in the `WebChatComposer` component?

tonyanziano: Ahhh, that might work. Let me try that.

hatpick: yeah, that's what I meant, it can be hooks used in the exact same place they are rn.

tonyanziano: [WebChatComposer is created as a `React.memo`](https://github.com/microsoft/BotFramework-Composer/blob/main/Composer/packages/client/src/components/WebChat/WebChatComposer.tsx#L84) which hooks doesn't seem to like when I call them inside of the component. I think it was written as a memo to make it re-render only when certain props change.

Is there a way to make hooks work with a memo, or would I have to rewrite the component? In the latter case, I don't think it's worth rewriting a heavily used component for a small refactor.

tonyanziano: image

This is what I get when I try to use them in `<WebChatComposer />`

hatpick: That's an error out of WebChat, probably due to using useObserveTranscriptFocus not within a Context.Provider provided by <Composer> from WebChat 

If you still want to keep them as hooks, you probably need to wrap them within a faceless component and put it as a child of <Composer>, having them as hooks might be better for unit tests, but at this point I'm fine with either way: separate faceless components, or two hooks wrapped in a faceless component

tonyanziano: Yup, went ahead with a faceless component wrapping both of the hooks

hatpick (approved):

PR #8013, opened 2021-06-04
chore: add duplicated skill warning in add dialog

### LINKED ISSUE #7827: error pop up for duplicated skill
## Describe the bug

Add duplicated skill will pop up following :
image

We'd better pop up other info. @vivekkshankar . This looks like a composer bug instead of error notification to user.

## Browser

- [X] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [X] Windows
- [ ] Ubuntu

alanlong9278:
## Description
Add duplicated skill warning in add dialog

image

## Task Item

closes #7827


Composer/packages/client/src/components/AddRemoteSkillModal/CreateSkillModal.tsx
```diff
@@ -53,14 +54,19 @@ export interface CreateSkillModalProps {
...
   const { manifestUrl: _, ...errors } = formDataErrors;
 
   if (!manifestUrl) {
     setFormDataErrors({ ...errors, manifestUrl: formatMessage('Please input a manifest URL') });
   } else if (!urlRegex.test(manifestUrl)) {
     setFormDataErrors({ ...errors, manifestUrl: formatMessage('URL should start with http:// or https://') });
+  } else if (skills.includes(manifestUrl)) {
+    setFormDataErrors({
+      ...errors,
+      manifestUrl: formatMessage('The bot is already part of the Bot Project'),
```
tdurnford, line 68: Can we change the message to something like “This skill has already been added to this project.” @vivekkshankar can you weigh in?

luhan2017: This is proposed by Vivek. 
“This skill has already been added”

@alanlong9278 let's talk about the local scenario, can we do the same for adding local bot as a skill, do we need extra design here?

alanlong9278: > This is proposed by Vivek.
> “This skill has already been added”
> 
> @alanlong9278 let's talk about the local scenario, can we do the same for adding local bot as a skill, do we need extra design here?

Yes，It will pop the err dialog when adding duplicated local bot as skill now. So we need a new design for adding  duplicated local bot. @luhan2017 
duplicated
 duplicated local bot

PR #8017, opened 2021-06-04
feat: add ability to test local template through composer behind hidden  feature flag

### LINKED ISSUE #6538: Allow for local template testing
We need to add an easy way where internal template developers can test local generators through Composer.

- [x] Create feature flag called SHOW_LOCAL_TEMPLATE_UI
- [x] When the above feature flag is set to true have a UI element appear in the select template view where developers can add a path to their local generator

pavolum:
## Description

- Added feature flag for advanced template options that when enabled allows users to test local generators 
- Added local generator option in template selection list

## Task Item

fixes #6538 

## Screenshots

image


Composer/packages/lib/shared/src/featureFlagUtils/index.ts
```diff
@@ -14,4 +14,13 @@ export const getDefaultFeatureFlags = (): FeatureFlagMap => ({
     isHidden: false,
     enabled: false,
   },
+  ADVANCED_TEMPLATE_OPTIONS: {
+    displayName: formatMessage('Advanced Template Options'),
```
hatpick, line 18: nit: The title is not reflective if the flag

not nit: "Advanced template options"

Composer/packages/client/src/recoilModel/atoms/appState.ts
```diff
@@ -364,3 +364,8 @@ export const warnAboutFunctionsState = atom<boolean>({
   key: getFullyQualifiedKey('warnAboutFunctionsState'),
   default: false,
 });
+
+export const localTemplatePathState = atom<string>({
```
hatpick, line 368: This is a temporary state that its life is bound to the creation view, why does it need to be an atom instead of a local component state that gets passed around as prop?

Composer/packages/client/src/components/CreationFlow/v2/TemplateDetailView.tsx
```diff
@@ -44,25 +53,73 @@ type TemplateDetailViewProps = {
...
+  const validatePath = async (path): Promise<string> => {
+    const response = await httpClient.get(`/storages/validate/${encodeURI(path)}`);
+    return response.data.errorMsg;
+  };
+
+  const renderLocalTemplateForm = () => (
+    <Fragment>
+      <Text>
+        {formatMessage.rich(
+          `To create a bot from your own <templateDocLink>Bot Framework Template</templateDocLink> you need to add a path to your local templates index.js file. <templateDocLink>Learn more</templateDocLink>`,
```
hatpick, line 77: nit: what's the point of having two links with the same source right next to each other, is the first of necessary?

hatpick (changes requested):

Composer/packages/client/src/components/CreationFlow/v2/CreateBot.tsx
```diff
@@ -27,6 +27,7 @@ import { csharpFeedKey, nodeFeedKey } from '@botframework-composer/types';
 import { useRecoilState, useRecoilValue } from 'recoil';
 
 import msftIcon from '../../../images/msftIcon.svg';
+import addIcon from '../../../images/addIcon.svg';
```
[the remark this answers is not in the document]
pavolum: I have made this change

Composer/packages/server/src/controllers/asset.ts
```diff
@@ -55,6 +56,26 @@ export async function getProjTemplatesV2(req: any, res: any) {
           packageVersion: qnaTemplateVersion,
         },
       });
+      if (FeatureFlagService.getFeatureFlagValue('ADVANCED_TEMPLATE_OPTIONS')) {
+        templates.push({
```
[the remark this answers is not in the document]
pavolum: Resolved

Composer/packages/lib/shared/src/featureFlagUtils/index.ts
pavolum: I have made the caps change but have not changed the string value, in R14 this will be a hidden feature flag so the text will never be in the UI for this release, if/when it is a visible feature flag then we will take another pass on the copy.

Composer/packages/client/src/components/CreationFlow/v2/TemplateDetailView.tsx
pavolum: Removed

Composer/packages/client/src/recoilModel/atoms/appState.ts
pavolum: I generally lean towards central state even if it is bound to the creation view as it results in less files touched with prop passing and consistency. However you are right in that most of Composer code does favor local state in these scenarios so I have snapped to that standard.

hatpick: The rule of thumb is that UI state belongs to the components and app state belongs to the recoil atoms. UI state gets garbage collected when the component goes away from screen, while app state hangs around for the life of the app, which in this case is totally unnecessary. You can easily imagine that every component brings its UI state to the recoil atoms and now we have memory leaks all over.

Composer/packages/client/__tests__/components/CreationFlow/CreateOptions/index.test.tsx
```diff
@@ -33,7 +34,9 @@ describe('<CreateOptions/>', () => {
     return renderWithRecoil(
       <CreateOptions
         fetchReadMe={handleFetchReadMeMock}
+        localTemplatePath={''}
         path="create"
+        setLocalTemplatePath={setLocalTemplatePathMock}
```
hatpick, line 39: this is a callback, name it accordingly: onUpdateLocalTemplatePath

Composer/packages/client/src/components/CreationFlow/CreateBot.tsx
```diff
@@ -274,7 +301,17 @@ export function CreateBot(props: CreateBotProps) {
...
           <div css={templateDetailContainer} data-is-scrollable="true">
-            {fetchReadMePending ? <LoadingSpinner /> : <TemplateDetailView readMe={readMe} template={getTemplate()} />}
+            {fetchReadMePending ? (
+              <LoadingSpinner />
+            ) : (
+              <TemplateDetailView
+                localTemplatePath={localTemplatePath}
+                readMe={readMe}
+                setLocalTemplatePath={setLocalTemplatePath}
+                setLocalTemplatePathValid={setLocalTemplatePathValid}
```
hatpick, line 311: onValidateLocalTemplatePath

Composer/packages/client/__tests__/components/CreationFlow/CreateOptions/index.test.tsx
hatpick: There are other similar cases I didn't comment on but please fix.

Composer/packages/client/src/components/CreationFlow/CreateBot.tsx
pavolum: Handled

Composer/packages/client/__tests__/components/CreationFlow/CreateOptions/index.test.tsx
pavolum: Handled

hatpick (approved):

a-b-r-o-w-n (approved):

PR #8018, opened 2021-06-04
feat: allow selection of generator version

### LINKED ISSUE #8016: Allow for testing old versions of a published first party template
We need to add a way where internal template developers ca publish preview versions of template generators which Composer can pick up for broader internal testing without affecting latest stable versions of that template.
- [ ] When the advanced template feature flag is set to true a version dropdown will be present for each template in the template view
  - [ ] version dropdown will be populated with all versions and when selected that version is instantiated in the creation flow

pavolum:
## Description

Add a version selector in the template selection screen that shows all published versions of the selected template and allowsd the user to use whichever version they like to create their bot

## Task Item

fixes #8016

## Screenshots
New version view if hidden FF is enabled
image

On version dropdown click
image


Composer/packages/client/src/components/CreationFlow/CreationFlow.tsx
```diff
@@ -125,9 +127,11 @@ const CreationFlow: React.FC<CreationFlowProps> = () => {
   };
 
   const handleCreateNew = async (formData, templateId: string, qnaKbUrls?: string[]) => {
-    const templateVersion = templateProjects.find((template: BotTemplate) => {
-      return template.id == templateId;
-    })?.package?.packageVersion;
+    const templateVersion = selectedTemplateVersion
```
srinaath, line 130: ```selectedTemplateVersion ??  templateProjects.find...```

srinaath:
@pavolum  could you attach snapshots to the PR as changes affect UI?

Composer/packages/client/src/components/CreationFlow/TemplateDetailView.tsx
```diff
@@ -122,9 +153,13 @@ export const TemplateDetailView: React.FC<TemplateDetailViewProps> = (props) =>
   return (
     <div>
       <div css={templateTitleContainer(isLocalTemplate)}>
-        {renderTemplateIcon()}
-        <span css={templateTitle(isLocalTemplate)}>{props.template?.name}</span>
-        {!isLocalTemplate && <span css={templateVersion}>{props.template?.package?.packageVersion}</span>}
+        <Stack horizontal>
+          <Stack.Item>{renderTemplateIcon()}</Stack.Item>
+          <Stack.Item>
+            <span css={templateTitle(isLocalTemplate)}>{props.template?.name}</span>
```
srinaath, line 159: This needs to conditionally rendered if props.template could be undefined.

Composer/packages/server/src/models/asset/assetManager.ts
```diff
@@ -272,36 +272,43 @@ export class AssetManager {
...
+              const displayName = this.getPackageDisplayName(name);
+              const templateToReturn = {
+                id: name,
+                name: displayName,
+                description: description,
+                package: {
+                  packageName: name,
+                  packageSource: 'npm',
+                  packageVersion: version,
+                  availableVersions: availableVersions,
```
srinaath, line 289: nit: use shorthand

Composer/packages/server/src/models/asset/assetManager.ts
```diff
@@ -334,4 +341,18 @@ export class AssetManager {
 
     return await res.text();
   }
+
+  public async getNpmPackageVersions(packageName: string): Promise<string[]> {
+    const registryUrl = `https://registry.npmjs.org/${packageName}`;
+    const response = await fetch(registryUrl);
+    const data = await response.json();
+    const versionsMap = data?.versions || [];
```
srinaath, line 349: data?.versions ?? []

Composer/packages/server/src/models/asset/assetManager.ts
```diff
@@ -334,4 +341,18 @@ export class AssetManager {
 
     return await res.text();
   }
+
+  public async getNpmPackageVersions(packageName: string): Promise<string[]> {
+    const registryUrl = `https://registry.npmjs.org/${packageName}`;
+    const response = await fetch(registryUrl);
```
srinaath, line 347: How are you handling errors? I dont see a try catch wrapper at the calling site as well.

Composer/packages/server/src/models/asset/assetManager.ts
```diff
@@ -272,36 +272,43 @@ export class AssetManager {
...
-              packageVersion: version,
-            },
-          } as BotTemplate;
-          if (isArray(keywords)) {
-            if (keywords.includes('bf-dotnet-functions') || keywords.includes('bf-dotnet-webapp')) {
-              templateToReturn.dotnetSupport = {
-                functionsSupported: keywords.includes('bf-dotnet-functions'),
-                webAppSupported: keywords.includes('bf-dotnet-webapp'),
-              };
+        const result: BotTemplate[] = await Promise.all(
```
srinaath, line 275: Is the behavior here to bail out even if you are not able to get one of the feed contents? Shouldt we be able to continue processing just the valid feeds?

srinaath (changes requested):

Composer/packages/server/src/models/asset/assetManager.ts
srinaath: In addition, we need error handling around these API calls.

Composer/packages/server/src/models/asset/__tests__/assetManager.test.ts
```diff
@@ -140,6 +142,7 @@ describe('assetManager', () => {
             packageName: 'generator-conversational-core',
             packageSource: 'npm',
             packageVersion: '1.0.3',
+            availableVersions: [''],
```
srinaath, line 145: Since we are adding the functionality to process multiple versions of packages. Would it better to write more tests asserting around multiple package versions?

Composer/packages/client/src/components/CreationFlow/TemplateDetailView.tsx
```diff
@@ -58,6 +64,31 @@ type TemplateDetailViewProps = {
...
+  const { setSelectedTemplateVersion } = useRecoilValue(dispatcherState);
+  const selectedTemplateVersion = useRecoilValue(selectedTemplateVersionState);
+
+  useEffect(() => {
+    props.template?.package?.packageVersion && setSelectedTemplateVersion(props.template.package.packageVersion);
+  }, [props.template]);
+
+  const renderVersionButton = () => {
+    const availableVersions = props.template?.package?.availableVersions || ([] as string[]);
+    availableVersions;
```
srinaath, line 76: can be removed.

Composer/packages/client/src/components/CreationFlow/TemplateDetailView.tsx
```diff
@@ -58,6 +64,31 @@ type TemplateDetailViewProps = {
...
+    const versionOptions = {
```
srinaath, line 77: Do you feel there would be value in memoizing this with a dependancy on props.template? The render for versionButton does not need change for every time the component rerenders right?

Composer/packages/server/src/models/asset/assetManager.ts
pavolum: The root caller has error handling, this is only called from `getProjTemplates` method which handles any thrown error. Its unlikely that this will be called elsewhere in the future but just in case I have added a try catch to the function itself so it isnt assumed that you need to add one to any future callers.

Composer/packages/server/src/models/asset/assetManager.ts
pavolum: - It is currently set up to skip feeds which you cannot get content for. If a feed is empty or the call to the feed throws an error then the calling function (`getCustomFeedTemplates`) skips that feed and jumps to the next one for processing.
- I have added additional error handling

Composer/packages/client/src/components/CreationFlow/TemplateDetailView.tsx
pavolum: Yeah that makes sense, I have made this change

Composer/packages/server/src/models/asset/__tests__/assetManager.test.ts
pavolum: Good point. I have added a unit test for the `getNpmPackageVersions` function. I think that should suffice seeing as this is a function and feature that will only be available to internal devs who set the hidden feature flag. If we make this feature GA then at that point we can write a more robust test suite that tests version selection in the UI all the way through creation execution.

pavolum:
> @pavolum could you attach snapshots to the PR as changes affect UI?

Yes I will add some now. Note that this feature is behind a hidden feature flag so the UI isnt up to shipping standards but has been approved by design for internal use which for the time being is the future of this feature

srinaath:
@pavolum Looking at the snapshot do you see the need to have a maxheight and scroll on the callout for versions? It seems it would span the height of the Composer container.

pavolum:
> @pavolum Looking at the snapshot do you see the need to have a maxheight and scroll on the callout for versions? It seems it would span the height of the Composer container.

image

srinaath (approved):

PR #8023, opened 2021-06-07
chore: move constants into their own directory

beyackle:
## Description

This is a minor reorganization in support of some wider changes; there will be more constants showing up (being collected from across Composer) and sticking everything into a single file would've gotten unwieldy.

## Task Item

#minor


beyackle:
superseded by new PR to include the color constants (#8034)

PR #8039, opened 2021-06-09
chore: introduce tests for ManageService component

### LINKED ISSUE #7902: Test coverage: /Manage*
Screen Shot 2021-05-24 at 8 52 25 AM

benbrown:
## Description

Adds a suite of unit test cases for the base ManageService component (which is in turn used to create the ManageLuis, ManageQNA and ManageSpeech components)

Tests so far:

* Test that the component renders with the correctly customized ui text
* Test that the "select an existing resource" path works as expected
* Test that the "Create a new resource" path works as expected
* Test that the "Create a new resource" path works when a tier selection is provided
* Test that the cancel button calls the right callback
* Tests that the data is sent back to the calling component when user has completed session
* Test that the "createService" prop is called (this is where the other components inject their specific details)

This involved creating simple mocks for a variety of Azure ARM APIs. 

Still to come:

- [x] test ManageLuis directly
- [x] test ManageSpeech directly
- [x] test ManageQNA directly (hardest)

## Task Item

Fixes #7902


Composer/packages/client/src/components/ManageLuis/__tests__/ManageLuis.test.tsx
```diff
@@ -0,0 +1,378 @@
...
+import { renderWithRecoil } from '../../../../__tests__/testUtils';
+import { ManageLuis } from '../ManageLuis';
+
+const serviceName = 'Language Understanding';
+const DOWN_ARROW = { keyCode: 40 };
+
+// token creds
+jest.mock('@azure/ms-rest-js');
+
+jest.mock('@azure/arm-resources', () => ({
```
hatpick, line 16: you can reuse these mocks instead of repeating them:
https://jestjs.io/docs/manual-mocks

pavolum:
Component tests are generally in `client/__tests__/components` and in this PR they are siblings of the tested components dir

Composer/packages/client/src/components/ManageLuis/__tests__/ManageLuis.test.tsx
benbrown: I moved most of these to the __mocks__ folder so they can be shared. Thanks for the tip! The items i did not move are either different for the different tests or not mocking node modules.

hatpick (approved):

Issue #8029, opened 2021-06-08
HaTs telemetry items 


cwhitten:
- [x] Mechanism for displaying a banner/notification in Composer to show to the users a prompt to complete a survey

- [x] Ability to dynamically generate the survey link parameters populated with values from composer usage (for example Version) & (survey link itself would be stable and hardcoded)

- [x] Ability to store some of the characteristics required for the trigger mechanism in local composer storage&nbsp;(e.g.&nbsp;count of sessions)

beyackle:
This is nearly done - #8078 will address all of the issues at once because it was easier to just group them all together into a batch instead of trying to shuttle between fixing the cards or the URL generation separately.

PR #8040, opened 2021-06-09
feature: add HaTS stuff to localStorage

beyackle:
## Description

This addresses the issue of storing things on the client so we know when to display the HaTS survey; it will keep track, in local storage, of the number of unique calendar days that a user uses Composer. If this exceeds 5, it sets a flag in the Recoil store that can be read in a later step (to be written) to see if we should display the survey. Once the survey has been taken, we can use a similar mechanism to store the latest date the survey was taken and make sure we aren't showing the link to a user who's taken it in the last 90 days.

## Task Item

refs #8029

## Screenshots
Not really a UI change, but here are the (currently) two new values in localStorage.
image

### DIFF
--- a/Composer/packages/client/src/App.tsx
+++ b/Composer/packages/client/src/App.tsx
@@ -9,6 +9,7 @@ import { Announcement } from './components/AppComponents/Announcement';
 import { MainContainer } from './components/AppComponents/MainContainer';
 import { dispatcherState, userSettingsState } from './recoilModel';
 import { loadLocale } from './utils/fileUtil';
+import { ClientStorage } from './utils/storage';
 import { useInitializeLogger } from './telemetry/useInitializeLogger';
 import { setupIcons } from './setupIcons';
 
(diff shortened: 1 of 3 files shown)


Composer/packages/client/src/App.tsx
```diff
@@ -38,6 +45,19 @@ export const App: React.FC = () => {
     ipcRenderer?.on('cleanup', (_event) => {
       performAppCleanupOnQuit();
     });
+
+    let days = surveyStorage.get('days', 0);
```
tdurnford, line 49: All of this logic should be handled in the `setSurveyEligibility` dispatcher. You should just call the dispatcher on mount and let it determine how and if it should update update the value.

Composer/package.json
```diff
@@ -20,7 +20,9 @@
     "url-parse": "^1.5.1",
     "underscore": "^1.12.1",
     "dns-packet": "^1.3.4",
-    "json-ptr": "^2.1"
+    "json-ptr": "^2.1",
```
tdurnford, line 23: Are these dependencies necessary? They seem out of scope for this PR.

tdurnford (changes requested):

Composer/packages/client/src/App.tsx
beyackle: Good idea. I'll fix that.

Composer/package.json
beyackle: They were causing a security alert, but I've now handled that in a separate PR which has merged already (so these can/will go away on a rebase).

Composer/packages/client/src/constants.tsx
```diff
@@ -525,3 +525,11 @@ export const defaultTeamsManifest: TeamsManifest = {
 
 export const defaultBotPort = 3979;
 export const defaultBotEndpoint = `http://localhost:${defaultBotPort}/api/messages`;
+
+const DAYS = 1000 * 60 * 60 * 24; // ms per day
```
tdurnford, line 529:
```suggestion
const MS_PER_DAY = 1000 * 60 * 60 * 24; 
```

tdurnford (approved):

Composer/packages/client/src/constants.tsx
beyackle: I thought about that, but I like the way it makes "90 * DAYS" legible a few lines farther down.

PR #8046, opened 2021-06-09
feat: Add dynamic schema for build forms in form dialogs

### LINKED ISSUE #5586: Support .uischema for form dialogs schema building
The schema returned from bf-generate will have default values for describing the fields for each property added to the schema.  These strings will need to be supported for localizing in the client.

- [ ] Investigate these options:
-  whether to have additions to the SDK schema and UI schema for these types
- if a path modification system could work with the en-US.json in Composer
- if a language specific UI schema should be returned from bf-generate

hatpick:
## Description

Provides dynamic display of properties based on the schema returned from the backend form dialogs

## Task Item

Fixes #5586

## Screenshots
(little bit old but shows most of the new stuff)

https://user-images.githubusercontent.com/1483492/121443422-51a93400-c942-11eb-9bf8-0b76ecef66a3.mp4


hatpick:
FYI @chrimc62

GeoffCoxMSFT (approved):

Issue #7632, opened 2021-05-04
Componentize the azure provisioning wizard


GeoffCoxMSFT:
The azurePublish extension dialog is a complete wreck.  It needs to be refactored into proper react components and then have steps plugged into a general wizard harness to get the step navigation and standardized header/footer.

Issue #7631, opened 2021-05-04
Create shared azure resource UI components


GeoffCoxMSFT:
Tenant, Subscription, Resource Group, Resource, Region, Tier inputs and dropdowns should be in a shared location for the azurePublish extension and other composer UI for creation and config.

VamsiModem:
### Tasks

1. Create a new combo box component with below features.
      - Filter the options as the user is typing
      - Show loading status when there is some asynchronous action. 
      - Allow creation when the option is not available
2. Create the below shared components, with all the data retrieval and validation encapsulated.   
      - TenantPicker
      - SubscriptionPicker
      - ResourceGroupPicker
      - RegionPicker
      - ResourceNameTextField
3. Sync up with the UX team on coming up with a design for 'Create new' option for resource group picker.

PR #8070, opened 2021-06-14
feat: Added implementation for ResourceConfigurationStep

VamsiModem:
## Description

Added implementation & validation for resource configuration step.
      1. Added 'Create new' option for searchable dropdown.
      2. Validation for new resource group
      3. Persist the configuration to local  storage.

## Task Item
refs #7632 
refs #7631 

## Screenshots

image

### DIFF
--- a/extensions/azurePublishNew/src/components/AzureProvisionWizard.tsx
+++ b/extensions/azurePublishNew/src/components/AzureProvisionWizard.tsx
@@ -4,9 +4,10 @@
 import React, { useState } from 'react';
 import styled from '@emotion/styled';
 import { RecoilRoot } from 'recoil';
-import { usePublishApi } from '@bfc/extension-client';
+import { usePublishApi, useLocalStorage } from '@bfc/extension-client';
 
 import { ProvisionAction } from '../types';
+import { resourceConfigurationState } from '../recoilModel/atoms/resourceConfigurationState';
 
 import { ChooseProvisionAction } from './ChooseProvisionAction';
 import { CreateResourcesWizard } from './provisioningWizards/CreateResourcesWizard';
@@ -24,7 +25,8 @@ const Root = styled.div`
 export const AzureProvisionWizard = () => {
   const [provisionAction, setProvisionAction] = useState<ProvisionAction>('create');
   const [activeStepIndex, setActiveStepIndex] = useState<number>(0);
-  const { setTitle } = usePublishApi();
+  const { setTitle, getName } = usePublishApi();
+  const { getItem } = useLocalStorage();
 
   const handleStepChange = (index, step) => {
     setActiveStepIndex(index);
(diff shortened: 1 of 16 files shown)


extensions/azurePublishNew/src/components/AzureProvisionWizard.tsx
```diff
@@ -43,7 +45,14 @@ export const AzureProvisionWizard = () => {
   }, [provisionAction]);
 
   return (
-    <RecoilRoot>
+    <RecoilRoot
```
GeoffCoxMSFT, line 48: The current extension has initialization of merging incoming state with defaults.  We should replicate that logic here when initializing the recoil state.  It might be good to factor it into a hook and then use that hook to get the initial state for the recoil root.

extensions/azurePublishNew/src/components/provisioningWizards/CreateResourcesWizard.tsx
```diff
@@ -47,6 +49,7 @@ export const CreateResourcesWizard = React.memo((props: Props) => {
           <ResourceConfigurationStep onResourceConfigurationChange={setIsValidResourceConfiguration} />
         ),
         navigationState: { canGoNext: isValidResourceConfiguration },
+        onBack: () => persistResourceConfiguration(),
```
GeoffCoxMSFT, line 52: It seems strange to be persisting resource configuration during onBack.  Is this how you keep state between the sub-wizards?  My assumption was that recoil would be holding the current state for all the steps while the wizard was running, the user could freely move forward and back, and when the wizard complete it would save the state for provisioning to pick up.

extensions/azurePublishNew/src/recoilModel/atoms/resourceConfigurationState.ts
```diff
@@ -14,6 +14,8 @@ export const resourceConfigurationState = atom<ResourceConfigurationState>({
     tenantId: '',
     deployLocation: '',
     luisRegion: '',
+    isNewResourceGroup: false,
+    hostName: '',
```
GeoffCoxMSFT, line 18: You don't have to have a single state atom for all the things across the wizard.  Recoil is built for distributed state, so you could have them as individual atoms.  This way you don't get re-renders when one control updates a single value.

extensions/azurePublishNew/src/recoilModel/dispatchers/resourceConfigurationDispatcher.ts
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
```
GeoffCoxMSFT, line 2: Consider trying to separate the overall resource configuration (aka publish profile config) from the wizard.  The dispatcher could abstract the load and save of the publish profile as a whole, then the individual pieces could be in separate recoil atoms used during the wizard, and at the end you save back to the profile.  Kind of if the profile was a file on disk and the in memory was a bunch of variables.

GeoffCoxMSFT (approved):

extensions/azurePublishNew/src/components/provisioningWizards/CreateResourcesWizard.tsx
VamsiModem: We use recoil to keep state between wizard steps, local storage is used to stash the wizard state when exited from the extension and restore it, when the extension loads the next time.

PR #8114, opened 2021-06-16
feat: introduce azure auth status in header, abstract tenant selection from components

### LINKED ISSUE #7626: Put token auth behind a facade for auth (with tenant ID)
Extensions and UI components should not be aware that the user is prompted for tokens in the web app scenario.  This needs to be behind a facade that provide either token dialog or azure sign in.

benbrown:
## Description

This PR introduces a new "global" auth indicator for Azure. In addition, it hoists the tenant selection and manual token logic into its own component, and removes it from the various components where this had been repeatedly implemented.

## Task Item

Fixes #7626 

## Screenshots

Screen Shot 2021-06-25 at 10 30 45 AM


Composer/packages/client/src/components/Auth/AuthDialog.tsx
```diff
@@ -9,17 +9,21 @@ import { DialogFooter } from 'office-ui-fabric-react/lib/Dialog';
...
 export interface AuthDialogProps {
   needGraph: boolean;
   onDismiss: () => void;
-  next: () => void;
+  next?: () => void;
 }
 export const AuthDialog: React.FC<AuthDialogProps> = (props) => {
-  const [graphToken, setGraphToken] = useState('');
+  const { setPrimaryToken, setGraphToken } = useRecoilValue(dispatcherState);
+
```
pavolum, line 25: Why is there a local state and recoil state graph token. Cant we just use the recoils state graph token and remove the local one that responds to it?

pavolum: Oh nvm I see, Its beause you do not set the recoil state one until the continue button is clicked. disregard this comment.

Composer/packages/client/src/recoilModel/dispatchers/auth.ts
```diff
@@ -0,0 +1,156 @@
...
+  const setShowTenantDialog = useRecoilCallback((callbackHelpers: CallbackInterface) => (show: boolean) => {
+    callbackHelpers.set(showTenantDialogState, show);
+  });
+
+  const refreshLoginStatus = useRecoilCallback((callbackHelpers: CallbackInterface) => async () => {
+    setPrimaryToken(getTokenFromCache('accessToken'));
+    setGraphToken(getTokenFromCache('graphToken'));
+  });
+
+  const resetCreds = () => {
```
pavolum, line 92: Should reset creds also clear creds from cache/storage?

Composer/packages/client/src/recoilModel/atoms/appState.ts
```diff
@@ -369,3 +369,33 @@ export const showGetStartedTeachingBubbleState = atom<boolean>({
   key: getFullyQualifiedKey('showGetStartedTeachingBubbleState'),
   default: false,
 });
+
+export const showAuthDialogState = atom<boolean>({
+  key: getFullyQualifiedKey('showAuthDialog'),
+  default: false,
+});
+
+export const primaryTokenState = atom<string>({
```
pavolum, line 378: Shouldn't we check storage/cache when initializing state

Composer/packages/client/src/recoilModel/atoms/appState.ts
```diff
@@ -369,3 +369,33 @@ export const showGetStartedTeachingBubbleState = atom<boolean>({
   key: getFullyQualifiedKey('showGetStartedTeachingBubbleState'),
   default: false,
 });
+
```
GeoffCoxMSFT, line 372: Consider a new module called authState for these atoms.

Composer/packages/client/src/recoilModel/atoms/appState.ts
```diff
@@ -369,3 +369,33 @@ export const showGetStartedTeachingBubbleState = atom<boolean>({
...
+  key: getFullyQualifiedKey('graphToken'),
+  default: '',
+});
+
+export const availableTenantsState = atom<AzureTenant[]>({
+  key: getFullyQualifiedKey('availableTenants'),
+  default: [],
+});
+
+export const currentTenantState = atom<string>({
```
GeoffCoxMSFT, line 393: consider currentTenantIdState as you are storing a string rather than the AzureTenant structure.  You could also have a selector to give you back the currentAzureTenant.

Composer/packages/client/src/recoilModel/atoms/appState.ts
```diff
@@ -369,3 +369,33 @@ export const showGetStartedTeachingBubbleState = atom<boolean>({
...
+export const showAuthDialogState = atom<boolean>({
+  key: getFullyQualifiedKey('showAuthDialog'),
+  default: false,
+});
+
+export const primaryTokenState = atom<string>({
+  key: getFullyQualifiedKey('primaryToken'),
+  default: '',
+});
+
```
GeoffCoxMSFT, line 382: As I mentioned in the last meeting, it would be great to have the expiration date/time of the token to reference.  You could have a selector for isTokenExpired (or isTokenValid) as a convenience.

Composer/packages/client/src/recoilModel/dispatchers/auth.ts
```diff
@@ -0,0 +1,156 @@
...
+  showTenantDialogState,
+  currentTenantState,
+} from '../atoms/appState';
+import { AuthClient } from '../../utils/authClient';
+import storage from '../../utils/storage';
+
+import { addNotificationInternal, createNotification } from './notification';
+
+export const authDispatcher = () => {
+  const setShowAuthDialog = useRecoilCallback((callbackHelpers: CallbackInterface) => (show: boolean) => {
```
GeoffCoxMSFT, line 29: Tiny Nit: You can destructure the set here rather than importing CallbackInterface type.

Composer/packages/client/src/recoilModel/dispatchers/auth.ts
```diff
@@ -0,0 +1,156 @@
...
+  });
+
+  const setCurrentTenant = useRecoilCallback(
+    (callbackHelpers: CallbackInterface) => async (tenant: string, notify = true) => {
+      callbackHelpers.set(currentTenantState, tenant);
+      setTenantId(tenant);
+      if (tenant) {
+        // get arm token for tenant
+        try {
+          const token = await AuthClient.getARMTokenForTenant(tenant);
```
GeoffCoxMSFT, line 51: This might put up some UI right? This logic down in the model feels a little too low level vs. setting state that triggers someone to run this code.

Composer/packages/client/src/recoilModel/dispatchers/auth.ts
```diff
@@ -0,0 +1,156 @@
...
+          if (token) {
```
GeoffCoxMSFT, line 52: If the token is undefined or empty do you want to set the primary token to empty anyway (vs. leaving the existing value in there)?

Composer/packages/client/src/recoilModel/dispatchers/auth.ts
```diff
@@ -0,0 +1,156 @@
...
+          });
+          addNotificationInternal(callbackHelpers, notification);
+          // clear out app state
+          resetCreds();
+        }
+      }
+    }
+  );
+
+  const setAvailableTenants = useRecoilCallback((callbackHelpers: CallbackInterface) => (tenants: AzureTenant[]) => {
```
GeoffCoxMSFT, line 80: Nit: You don't have to put this in the dispatcher unless you see logic being added here later, components can use recoil directly.

Composer/packages/client/src/components/Auth/AuthCard.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+export const AuthCard: React.FC = () => {
+  const token = useRecoilValue(primaryTokenState);
+  const [authCardVisible, setAuthCardVisible] = useState<boolean>(false);
+  const { refreshLoginStatus, requireUserLogin, logoutUser, setShowAuthDialog, setShowTenantDialog } = useRecoilValue(
+    dispatcherState
+  );
+  const showAuthDialog = useRecoilValue(showAuthDialogState);
+  const showTenantDialog = useRecoilValue(showTenantDialogState);
+  const [currentUser, setCurrentUser] = useState<any>(undefined);
```
a-b-r-o-w-n, line 60: Why wouldn't this also live in the recoil state?

benbrown: the current user? I agree! I am going to move it there.

Composer/packages/client/src/recoilModel/atoms/appState.ts
benbrown: done!

Composer/packages/client/src/components/Auth/AuthCard.tsx
benbrown: done! i also refactored the state a bit so that we have nice accessors like `currentUser` and `isAuthenticated` that are also available via the extension client.

Composer/packages/client/src/recoilModel/atoms/appState.ts
benbrown: the expiration timestamp is now part of the currentUser object!

Composer/packages/client/src/recoilModel/dispatchers/auth.ts
benbrown: yes, it does pop up "native" UI from the electron server. But this acts like a REST call.  We discussed this in Teams.

Composer/packages/client/src/recoilModel/dispatchers/auth.ts
benbrown: good call.  getArmTokenForTenant will throw if it fails, so the catch -- which contains a call to reset the token -- should fire.  But I added an else just in case to throw if token is undefined and make sure!

Composer/packages/client/src/recoilModel/atoms/appState.ts
benbrown: done! good call.

Composer/packages/client/src/components/AddRemoteSkillModal/CreateSkillModal.tsx
```diff
@@ -375,17 +372,6 @@ export const CreateSkillModal: React.FC<CreateSkillModalProps> = (props) => {
           </Fragment>
         )}
       </DialogWrapper>
-      {showAuthDialog && (
```
GeoffCoxMSFT, line 378: This no longer shows the auth dialog.  Does it need to require authentication through the new auth?

Composer/packages/client/src/components/ManageService/ManageService.tsx
```diff
@@ -213,33 +153,7 @@ const ManageService: React.FC<ManageServiceProps> = (props: ManageServiceProps)
...
-            )
-          );
-        }
-      }
-    } else {
-      setUserProvidedTokens(false);
-    }
-    setCurrentStep('subscription');
-  };
+  }, [isAuthenticated]);
```
GeoffCoxMSFT, line 156: You reference currentUser, does that need to be in the dependencies?

Composer/packages/client/src/pages/publish/Publish.tsx
```diff
@@ -288,8 +282,8 @@ const Publish: React.FC<RouteComponentProps<{ projectId: string; targetName?: st
       if (target && isPublishingToAzure(target)) {
         const { tenantId } = JSON.parse(target.configuration);
 
-        if (userShouldProvideTokens()) {
-          token = getTokenFromCache('accessToken');
+        if (userShouldProvideTokens() && currentUser) {
```
GeoffCoxMSFT, line 285: Non-blocking issue: I'm hoping that we can remove the call differences so that userShouldProvideTokens isn't called anywhere.  The require login call can inspect if we are in electron or not and bring up the right login prompt or token dialog.

Composer/packages/client/src/plugins/api.ts
```diff
@@ -25,7 +25,7 @@ interface AuthAPI {
 
 interface PublishAPI {
   getPublishConfig?: () => PublishConfig | void;
-  startProvision?: (config: any) => void;
+  startProvision?: (config: any, arm: string, graph: string) => void;
```
GeoffCoxMSFT, line 28: To not break existing plugins that aren't authored in composer, you either need to make arm and graph optional or add a new method (startProvision2).

Composer/packages/client/src/recoilModel/DispatcherWrapper.tsx
```diff
@@ -86,7 +86,8 @@ const InitDispatcher = ({ onLoad }) => {
   //so use the useLayoutEffect here
   useLayoutEffect(() => {
     setDispatcher(dispatcherRef.current);
-    prepareAxiosWithRecoil(currentDispatcherState);
+    prepareAxiosWithRecoil();
+    currentDispatcherState; // ToDo: deprecate this
```
GeoffCoxMSFT, line 90: Consider providing more details about when we will deprecate this.

Composer/packages/client/src/pages/publish/Publish.tsx
benbrown: good point

Composer/packages/client/src/utils/authClient.ts
```diff
@@ -123,14 +125,20 @@ async function getARMTokenForTenant(tenantId: string): Promise<string> {
...
+  // do we have a valid token in the cache for this tenant?
+  if (getTokenFromCache(`token-${tenantId}`)) {
+    return getTokenFromCache(`token-${tenantId}`);
   } else {
-    const { accessToken = '' } = await result.json();
-    return accessToken;
+    const result = await fetch(`/api/auth/getARMTokenForTenant?tenantId=${tenantId}`, options);
+    if (result.status >= 400) {
+      const data = await result.json();
+      throw Error(data.error?.diagnostics?.description || 'get ARM token failure');
```
GeoffCoxMSFT, line 135: Nit: most of the time we throw new Error, but in other places (less often) we have throw Error

Composer/packages/client/src/recoilModel/DispatcherWrapper.tsx
benbrown: There is a wider swathe of code that dates to a previous authentication scheme related that never shipped. I was going to open another ticket to track a separate "clean up old unused auth code" task....

I *think* I can probably go back and remove that parameter up the chain

Composer/packages/extension-client/src/hooks/usePublishApi.ts
```diff
@@ -58,6 +58,13 @@ export function usePublishApi() {
   function userShouldProvideTokens(): boolean {
     return window[ComposerGlobalName].userShouldProvideTokens();
   }
+  function getPrimaryToken(): boolean {
```
GeoffCoxMSFT, line 61: Does this ensure the token is not expired?

Composer/packages/client/src/utils/authClient.ts
benbrown: good catch throw new Error is the right way i think!

Composer/packages/extension-client/src/hooks/usePublishApi.ts
GeoffCoxMSFT, line 61: Should this be a string?

Composer/packages/extension-client/src/hooks/usePublishApi.ts
```diff
@@ -58,6 +58,13 @@ export function usePublishApi() {
...
+    return window[ComposerGlobalName].getPrimaryToken();
+  }
+  function requireUserLogin(tenantId: string, options?: { graph: boolean }): boolean {
```
GeoffCoxMSFT, line 64: Nit: require feels like it should throw if the user doesn't get logged in.  requestUserLogin would cause me to expect a value returned.

Composer/packages/types/src/auth.ts
```diff
@@ -30,3 +30,12 @@ export type WebAuthParameters = {
   /** List of OAuth scopes that will be granted once the user has authenticated. */
   scopes?: string[];
 };
+
+export type CurrentUser = {
+  token: string; // aad token
+  graph: string | null; // graph token
+  email?: string;
+  name?: string;
+  expiration?: number;
```
GeoffCoxMSFT, line 39: Can we expose this in the auth API that plugins use?

benbrown: this is available from `useAuthApi()` in the plugin context.

Composer/packages/client/src/components/AddRemoteSkillModal/CreateSkillModal.tsx
benbrown: that authentication now happens automatically when the user gets to the publish profile dialog. the auth dialog is part of the global component now, so doesn't have to be invoked explicitly here.

Composer/packages/client/src/components/ManageService/ManageService.tsx
benbrown: isAuthenticated and currentUser are set in tandem.  if isAuthenticated is true, currentUser is set. So I thought it made more sense semantically to have it fire when the authentication status changed rather than changes to the user model...

Composer/packages/client/src/pages/publish/Publish.tsx
benbrown: there are not a ton of places this is called, and many of them are probably not needed anymore.  we should do this separately imo.

Composer/packages/extension-client/src/hooks/usePublishApi.ts
benbrown: this is not being used anymore, i removed it.

benbrown: removed

Composer/packages/extension-client/src/hooks/usePublishApi.ts
benbrown: yes we should commune on the best name for this method, which ... causes the login process to begin.

Composer/packages/client/src/components/Auth/AuthCard.tsx
```diff
@@ -0,0 +1,164 @@
...
+            root: { width: 320, zIndex: zIndices.authContainer, backgroundColor: NeutralColors.white },
+          }}
+          target={`#persona`}
+          onDismiss={toggleAuthCardVisibility}
+        >
+          {isAuthenticated ? (
+            <Stack tokens={{ childrenGap: 10, padding: 20 }}>
+              <Stack.Item align="end">
+                <Link styles={styles.logoutLink as ILinkStyles} onClick={logout}>
+                  Sign out
```
hatpick, line 109: formatMessage

Composer/packages/client/src/components/Auth/AuthCard.tsx
```diff
@@ -0,0 +1,164 @@
...
+    requireUserLogin();
+    toggleAuthCardVisibility();
+  };
+
+  return (
+    <Fragment>
+      {/* this is the icon that appears at the top of the header bar */}
+      <Persona
+        hidePersonaDetails
+        id={'persona'}
```
hatpick, line 87: since it's on navbar and it's always showing, this id needs to be unique across the app. can we make sure of that? probably not, so it's better to use a ref for callout target 😄

Composer/packages/client/src/components/Auth/AuthDialog.tsx
```diff
@@ -3,23 +3,47 @@
...
+      paddingTop: '14px',
+      paddingBottom: '11px',
+    },
+    subText: {
+      fontSize: FontSizes.size14,
+    },
+  },
+  modal: {
+    main: {
+      // maxWidth: '416px !important',
```
hatpick, line 37: remove

Composer/packages/client/src/components/Auth/TenantDialog.tsx
```diff
@@ -0,0 +1,56 @@
...
+      isBlocking
+      isOpen
+      dialogType={DialogTypes.CreateFlow}
+      subText={formatMessage(
+        'The directory you choose will impact subscription, resource group, and region filters that are available when you select or create Azure resources. You can change this later.'
+      )}
+      title={formatMessage('Select directory')}
+      onDismiss={props.onDismiss}
+    >
+      <ChoiceGroup
```
hatpick, line 35: isn't this supposed to show the selected tenant?

Composer/packages/client/src/pages/botProject/PublishProfieWrapperDialog.tsx
```diff
@@ -18,25 +16,23 @@ import { PublishProfileDialog } from './create-publish-profile/PublishProfileDia
...
   const [showPublishProfileDialog, setShowPublishProfileDialog] = useState(false);
-  const [showAuthDialog, setShowAuthDialog] = useState(false);
 
   const [currentPublishProfile, setCurrentPublishProfile] = useState<{ index: number; item: PublishTarget } | null>(
     null
   );
 
   useEffect(() => {
-    isShowAuthDialog(true) ? setShowAuthDialog(true) : setShowPublishProfileDialog(true);
+    setShowPublishProfileDialog(true);
```
hatpick, line 35: every time this component it re-rendered you are showing the publish profile dialog, is that what you want?

Composer/packages/client/src/pages/botProject/adapters/ABSChannels.tsx
```diff
@@ -162,26 +125,18 @@ export const ABSChannels: React.FC<RuntimeSettingsProps> = (props) => {
...
-          });
-        }
+        // NOTE: if config.tenantId is missing (as might be the case with a pre-1.4 composer publish profile)
+        // this will cause the user to be prompted for the correct tenant when they login.
+        // if they choose the WRONG tenant, an error will appear.
+        requireUserLogin(config.tenantId);
+
+        setCurrentResource({
+          microsoftAppId: config?.settings?.MicrosoftAppId,
+          resourceName: config.botName || config.name,
```
hatpick, line 135: if config can be undefined (Line 134 thinks so) this and the lines below with config.something will throw

Composer/packages/client/src/recoilModel/atoms/authState.ts
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+import { AzureTenant, CurrentUser } from '@botframework-composer/types';
+import { atom } from 'recoil';
+
+const getFullyQualifiedKey = (value: string) => {
```
hatpick, line 6: why can't you use the existing util function that does this used in other places?

Composer/packages/client/src/recoilModel/dispatchers/auth.ts
```diff
@@ -0,0 +1,202 @@
...
+  requiresGraphState,
+} from '../atoms/authState';
+import { AuthClient } from '../../utils/authClient';
+import storage from '../../utils/storage';
+import { graphScopes } from '../../constants';
+
+import { addNotificationInternal, createNotification } from './notification';
+
+export type UserLoginOptions = {
+  graph?: boolean;
```
hatpick, line 33: nit: requireGraph

Composer/packages/client/src/components/Auth/TenantDialog.tsx
benbrown: meaning pre-populate the previously chosen tenant?

Composer/packages/client/src/pages/botProject/PublishProfieWrapperDialog.tsx
benbrown: yes ... in fact, that whole check can probably be removed. It originally functioned to hide this modal until the token auth UI completed -- however, the token auth ui now happens later in the flow *inside the publishprofiledialog*

Composer/packages/client/src/pages/botProject/adapters/ABSChannels.tsx
benbrown: it should be set. unless the json parse failed, which shouldn't happen, but I added a try/catch just in case.

Composer/packages/client/src/recoilModel/atoms/authState.ts
benbrown: .... this is copypasta from the other xState files, where it is defined.... I THINK to give those variables some different namespace? I had not done this, so I updated it here... but ..we might want to consider consolidating this.

Composer/packages/client/src/components/Auth/TenantDialog.tsx
hatpick: that's right

Composer/packages/client/src/components/Auth/AuthCard.tsx
benbrown: I'm not sure how to achieve this! I'll ping you in teams.

Composer/packages/client/src/components/Auth/TenantDialog.tsx
benbrown: I will check with design when I can -- typically, we only show the tenant selection when we do not know or have a previous tenant. However, in the case where user is changing tenants we may want to display the current selection.

benbrown: ok this has been updated with feedback from @emivers8  - the current and/or default tenant will be pre-selected.

benbrown:
@hatpick @a-b-r-o-w-n @GeoffCoxMSFT - I haven't had any reports of issues from small amount of internal crew bashing. Would like to get this approved and ready to merge. Any further comments?

srinaath (approved):

srinaath:
@benbrown The cursor on the AuthCard is a crosshair instead of a pointer. In addition, is there no selected state on the button ?
Screen Shot 2021-07-07 at 9 25 30 AM

Composer/packages/client/src/components/Auth/AuthCard.tsx
```diff
@@ -0,0 +1,168 @@
...
+              <Stack.Item>
+                <Stack horizontal tokens={{ childrenGap: 20 }}>
+                  <Stack.Item>
+                    <Persona hidePersonaDetails initialsColor={NeutralColors.gray160} text={currentUser?.name} />
+                  </Stack.Item>
+                  <Stack.Item>
+                    <div style={styles.username}>{currentUser.name}</div>
+                    <div style={styles.email}>{currentUser.email}</div>
+                    <Link
+                      href={'https://myaccount.microsoft.com/'}
```
srinaath, line 125: does this need to be a aka.ms link?

Composer/packages/client/__tests__/components/Adapters/ABSChannels.test.tsx
```diff
@@ -7,7 +7,7 @@ import { act, fireEvent } from '@botframework-composer/test-utils';
 import { renderWithRecoil } from '../../testUtils/renderWithRecoil';
 import ABSChannels from '../../../src/pages/botProject/adapters/ABSChannels';
 import { botDisplayNameState, settingsState } from '../../../src/recoilModel';
-import * as authUtils from '../../../src/utils/auth';
+// import * as authUtils from '../../../src/utils/auth';
```
hatpick, line 10: remove

hatpick (approved):

Composer/packages/client/src/components/Auth/AuthCard.tsx
benbrown: > @benbrown I was playing with the build of this branch. Functionality looks great!
> 
> One thing I did see was the cursor on the AuthCard is a crosshair instead of a pointer. In addition, is there no selected state on the button ?
> Screen Shot 2021-07-07 at 9 25 30 AM

Fixed the cursor here.  There is not an active state as far as I know. @emivers8 care to chime in?

benbrown: No active state needed.

PR #8145, opened 2021-06-21
fix: Update the skillHostEndpoint when settings.skill has value

### LINKED ISSUE #7856: Composer doesn't add the skillendpoint to appsettings if you don't start the bot
## Describe the bug
Composer doesn't add the `skillHostEndpoint` to app settings until you start the bot.

This might be a P1 since it's critical, but not its importance is less clear. 

## Version
Version:  2.0.0-nightly.246099.8fa93
Electron: 8.2.4
Chrome: 80.0.3987.165
NodeJS: 12.13.0
V8: 8.0.426.27-electron.0

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Create a skill bot, provision, and make a manifest
2. Create an dotnet wa parent bot, and provision
3. Add the skill as a remote skill
4. Publish the bot

Note: Do not start the bot before publishing

## Expected behavior
This should work, however 

## Screenshots
image

## Additional context
If you start the bot locally the runtime will fill in the `skillHostEndpoint` property and everything works as expected. The odds are low that customers hit this, but it's very opaque if someone does.

luhan2017:
## Description

Previously, the logic to reset skillHostEndpoint in the publish process is:
- when skillHostEndpoint has value, reset it to https://${hostname}.azurewebsites.net/api/skills
This will cause an issue that when the bot has never been started from local, it will not be set in the publish process. which will cause a runtime error. 

This fix is to Set skillHostEndpoint when settings.skill has value. 

Closes #7856

### DIFF
--- a/extensions/azurePublish/src/node/deploy.ts
+++ b/extensions/azurePublish/src/node/deploy.ts
@@ -64,7 +64,9 @@ export class BotProjectDeploy {
       }
 
       // Update skill host endpoint
-      settings.skillHostEndpoint = `https://${hostname}.azurewebsites.net/api/skills`;
+      if (Object.keys(settings?.skill).length > 0) {
+        settings.skillHostEndpoint = `https://${hostname}.azurewebsites.net/api/skills`;
+      }
 
       // STEP 1: CLEAN UP PREVIOUS BUILDS
       // cleanup any previous build
--- a/extensions/azurePublish/src/node/deploy.ts
+++ b/extensions/azurePublish/src/node/deploy.ts
@@ -64,9 +64,7 @@ export class BotProjectDeploy {
       }
 
       // Update skill host endpoint
-      if (settings.skillHostEndpoint) {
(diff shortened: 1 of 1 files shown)


extensions/azurePublish/src/node/deploy.ts
```diff
@@ -64,7 +64,7 @@ export class BotProjectDeploy {
       }
 
       // Update skill host endpoint
-      if (settings.skillHostEndpoint) {
+      if (Object.keys(settings?.skill).length > 0) {
```
hatpick, line 67: Object.keys on undefined value will throw

hatpick (changes requested):

extensions/azurePublish/src/node/deploy.ts
luhan2017: Good catch, thanks :)

cwhitten (approved):

PR #8146, opened 2021-06-21
feat: connect to local manifest & copy manifest.json and lu files to current bot

### LINKED ISSUE #7852: Add skill via manifest fails to load remote LU
## Describe the bug

While connecting to a remote skill, the step that is supposed to import the remote LU files fails an error that says "get remote file fail" and requires me to put in my own LU for the skill trigger

Screen Shot 2021-05-17 at 3 45 00 PM

## Version

2.0.0-test

### LINKED ISSUE #7864: Load skill manifests from any location

alanlong9278:
## Description

 - connect to local manifest @VanyLaw 
 
 - copy manifest.json and lu files to current bot
 1. from http url
 files

2. from local zip
skillZip

## Task Item

close #7864
close #7852


cwhitten:
@alanlong9278 does this support the format that PVA will export? It is a .zip format with both the manifest and the language data inside of it.

cc @luhan2017 @gabog

alanlong9278:
> @alanlong9278 does this support the format that PVA will export? It is a .zip format with both the manifest and the language data inside of it.
> 
> cc @luhan2017 @gabog

@cwhitten We will support zip format, I changed  this pr into draft

luhan2017:
> > @alanlong9278 does this support the format that PVA will export? It is a .zip format with both the manifest and the language data inside of it.
> > cc @luhan2017 @gabog
> 
> @cwhitten We will support zip format, I changed this pr into draft

Not yet, we just got the requirement yesterday, working on that.

Composer/packages/client/src/components/AddRemoteSkillModal/CreateSkillModal.tsx
```diff
@@ -180,21 +196,54 @@ export const CreateSkillModal: React.FC<CreateSkillModalProps> = (props) => {
...
+        zipContent[fPath] = await files[fPath].async('string');
+        // eslint-disable-next-line no-useless-escape
+        if (fPath.match(/\.([^\.]+)$/)?.[1] === 'json' && isManifestJson(zipContent[fPath])) {
+          manifestFiles.push(files[fPath]);
+          setManifestDirPath(fPath.substr(0, fPath.lastIndexOf('/') + 1));
+        }
+      }
+
+      // update content for detail panel and show it
+      if (manifestFiles.length > 1) {
```
luhan2017, line 228: image

Add a condition for no manifest.json?

Composer/packages/client/src/components/AddRemoteSkillModal/CreateSkillModal.tsx
```diff
@@ -73,7 +79,46 @@ export const validateManifestUrl = ({ formData, formDataErrors, setFormDataError
...
+      // eslint-disable-next-line no-useless-escape
+      if (fPath.match(/\.([^\.]+)$/)?.[1] === 'json' && isManifestJson(zipContent[fPath])) {
+        manifestFiles.push(files[fPath]);
+        result.path = fPath.substr(0, fPath.lastIndexOf('/') + 1);
+      }
+    }
+
+    // update content for detail panel and show it
+    if (manifestFiles.length > 1) {
+      result.error = { manifestUrl: formatMessage('zip folder has multiply manifest json') };
```
EricDahlvang, line 102:
```suggestion
      result.error = { manifestUrl: formatMessage('zip folder has multiple manifest json') };
```

Composer/packages/client/src/components/AddRemoteSkillModal/CreateSkillModal.tsx
```diff
@@ -73,7 +79,46 @@ export const validateManifestUrl = ({ formData, formDataErrors, setFormDataError
...
+      zipContent[fPath] = await files[fPath].async('string');
+      // eslint-disable-next-line no-useless-escape
+      if (fPath.match(/\.([^\.]+)$/)?.[1] === 'json' && isManifestJson(zipContent[fPath])) {
+        manifestFiles.push(files[fPath]);
+        result.path = fPath.substr(0, fPath.lastIndexOf('/') + 1);
+      }
+    }
+
+    // update content for detail panel and show it
+    if (manifestFiles.length > 1) {
```
EricDahlvang, line 101: The .zip file will also contain the .lu file.

luhan2017:
@alanlong9278 can we add the telemetry data to tell if the manifest is http-based or load from zip content?

alanlong9278:
> @alanlong9278 can we add the telemetry data to tell if the manifest is http-based or load from zip content?

Sure

luhan2017 (approved):

Composer/packages/client/src/components/AddRemoteSkillModal/CreateSkillModal.tsx
luhan2017: @EricDahlvang Here we only check if manifest.json is existed and valid, for lu files, we will check it later and show a warning if it is not accessible like the following, which means, if there is no lu files in .zip, user can still add the manifest but they need to configure their own trigger phrases. 

image

Composer/packages/client/src/components/AddRemoteSkillModal/BrowserModal.tsx
```diff
@@ -0,0 +1,42 @@
...
+    if (zipFile) {
+      // create zip instance
+      const jszip = new JSZip();
+      jszip
+        .loadAsync(zipFile)
+        .then((zip) => {
+          props.onUpdate(zipFile.name, zip.files);
+        })
+        .catch((error) => {
+          console.log(error);
```
lei9444, line 27: Don't use console.log here.

Composer/packages/client/src/components/AddRemoteSkillModal/CreateSkillModal.tsx
```diff
@@ -82,12 +127,14 @@ export const getSkillManifest = async (projectId: string, manifestUrl: string, s
     });
     setSkillManifest(data);
   } catch (error) {
+    console.log(error);
```
lei9444, line 130: We should remove this one.

Composer/packages/client/src/components/AddRemoteSkillModal/CreateSkillModal.tsx
```diff
@@ -310,14 +383,18 @@ export const CreateSkillModal: React.FC<CreateSkillModalProps> = (props) => {
...
               <div style={{ width: '50%' }}>
-                <TextField
-                  required
-                  errorMessage={formDataErrors.manifestUrl}
-                  label={formatMessage('Skill Manifest URL')}
-                  placeholder={formatMessage('Ask the skill owner for the URL and provide your bot’s App ID')}
-                  value={formData.manifestUrl || ''}
-                  onChange={handleManifestUrlChange}
-                />
+                <div style={{ display: 'flex' }}>
```
lei9444, line 386: can we use @emotion/core here?

Composer/packages/server/src/controllers/project.ts
```diff
@@ -351,6 +352,36 @@ async function getSkill(req: Request, res: Response) {
...
+  }
+}
+
+async function removeSkillFiles(req: Request, res: Response) {
+  const projectId = req.params.projectId;
+  const user = await ExtensionContext.getUserFromRequest(req);
+
+  const currentProject = await BotProjectService.getProjectById(projectId, user);
+  if (currentProject !== undefined) {
+    console.log(req.params.name);
```
lei9444, line 377: can we remove it?

PR #8141, opened 2021-06-20
feat: Debug panel feature updates

srinaath:
## Description
Fit and Finish items for DebugPanel

## Task Item
Fixes #7984, #8032

## Screenshots

Screen Shot 2021-06-20 at 3 50 56 PM
Screen Shot 2021-06-20 at 3 51 06 PM

PR #8151, opened 2021-06-21
feat: Add "Watch" tab to the debug panel

### LINKED ISSUE #7978: Add "Watch" as new tab to debugging pane
## Summary
Tracking state throughout a conversation in Web Chat (or Emulator) is possible, but difficult. The user has to find the "Bot State" log item, select it, and then dig into the inspector pane to find the property they are looking for. Elevating bot state to its own tab and allowing users to select the custom properties thy want to follow will 

1. reduce the clicking needed to find a property
2. enable a persistent view into the properties of interest

### P0s
- [x] Add tab that filters the Web Chat logs to Bot State (follow same layout: logs on left, inspector in the right side)
- [x] Add filter for properties to watch **_- update: instead of filter, using a button to add a new row to the table that opens the property picker._**
- [x] Visual presentation of the user-defined properties only 
- [ ] Abillty to autoexpand to the properties/scope selected in the filter 

## P1
- [ ] ~~Ability to move to previous/next state~~ - **update: _removed per discussion june 9_. may reconsider after R14 based on user feedback.**

## P2
- [ ] ~~Ability to click on messages in Web Chat and select the state for that turn~~ **update: _removed. per discussion june 9, w are making an assumption that users are only interested in the latest state, not in the history of what's in memory.**

## Design mockups

(to come)

tonyanziano:
## Description

This PR adds the **Watch** tab to the debugging panel.

The Watch tab allows users to observe arbitrary variables stored in the bot's different memory scopes while interacting with the bot.

## Task Item

Fixes #7978

## Screenshots

**Bot not yet started**

image

**Talking to bot**

image

image

**Validation**

image


Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
```diff
@@ -0,0 +1,326 @@
...
+
+const addIcon: IIconProps = {
+  iconName: 'Add',
+};
+
+const removeIcon: IIconProps = {
+  iconName: 'Cancel',
+};
+
+const NameColumnKey = 'column1';
```
hatpick, line 103: use a more descriptive name? :D

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
```diff
@@ -0,0 +1,326 @@
...
+        return botTrace?.value[segment];
+      }
+      // if we are not on the root, try accessing the next value of the desired property
+      if (typeof accumulator === 'object') {
+        return accumulator[segment];
+      } else {
+        return undefined;
+      }
+    },
+    undefined
```
hatpick, line 144: ?

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
```diff
@@ -0,0 +1,326 @@
...
+  );
+  return value;
+};
+
+export const WatchTabContent: React.FC<DebugPanelTabHeaderProps> = ({ isActive }) => {
```
hatpick, line 149: why is the tab knows it's active or not, isn't the parent deciding to show it or not?

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
```diff
@@ -0,0 +1,326 @@
...
+  const rawWebChatTraffic = useRecoilValue(webChatTrafficState(currentProjectId ?? ''));
+  const [watchedVars, setWatchedVars] = useState<Record<string, string>>({});
+  const [pickerErrorMessages, setPickerErrorMessages] = useState<Record<string, string>>({});
+  const [selectedVars, setSelectedVars] = useState<IObjectWithKey[]>();
+  const [memoryVariablesPayload, setMemoryVariablesPayload] = useState<WatchDataPayload>({
+    kind: 'property',
+    data: { properties: [] },
+  });
+
+  const watchedVarsSelection = useMemo(
```
hatpick, line 160: this is practically a ref

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
```diff
@@ -0,0 +1,326 @@
...
+    ) as ConversationActivityTrafficItem[];
+    if (botStateTraffic.length) {
+      return botStateTraffic[botStateTraffic.length - 1];
+    }
+  }, [rawWebChatTraffic]);
+
+  const onSelectPath = useCallback(
+    (variableId: string, path: string) => {
+      const watchedVar = Object.values(watchedVars).find((varPath) => varPath === path);
+      if (watchedVar) {
```
hatpick, line 198: have we considered a UX in which we dont show the selected vars in the list OR we disable them?

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
```diff
@@ -0,0 +1,326 @@
...
+        delete pickerErrorMessages[variableId];
+        setPickerErrorMessages({ ...pickerErrorMessages });
+      }
+    },
+    [pickerErrorMessages, watchedVars]
+  );
+
+  // we need to refresh the details list every time a new bot state comes in
+  const refreshedWatchedVars = useMemo(() => {
+    return Object.entries(watchedVars).map(([key, value]) => {
```
hatpick, line 219: use {...watchedVars} ?

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
```diff
@@ -0,0 +1,326 @@
...
+      return {
+        key,
+        value,
+      };
+    });
+  }, [mostRecentBotState, pickerErrorMessages, watchedVars]);
+
+  const renderRow = useCallback((props?: IDetailsRowProps) => {
+    if (props) {
```
hatpick, line 228: nit: condition ? this : that

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
```diff
@@ -0,0 +1,326 @@
...
+      return <DetailsRow {...props} styles={rowStyles} />;
+    }
+    return null;
+  }, []);
+
+  const renderColumn = useCallback(
```
hatpick, line 234: I think it's much better if you add each column renderer to the column itself.

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
```diff
@@ -0,0 +1,326 @@
...
+        } else if (column.key === ValueColumnKey) {
+          // render the value display
+          if (mostRecentBotState) {
+            const value = getValueFromBotTraceScope(item.value, mostRecentBotState?.activity);
+            if (value !== null && typeof value === 'object') {
+              // render monaco view
+              return <WatchTabObjectValue value={value} />;
+            } else if (value === undefined) {
+              // render undefined indicator
+              return <span css={undefinedValue}>undefined</span>;
```
hatpick, line 258: formatMessage?

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
```diff
@@ -0,0 +1,326 @@
...
+
+  const onClickAdd = useCallback(() => {
+    setWatchedVars({
+      ...watchedVars,
+      [uuidv4()]: '',
+    });
+  }, [watchedVars]);
+
+  const onClickRemove = () => {
+    const updated = produce(watchedVars, (draftState) => {
```
hatpick, line 283: this is an overkill to bring in immer here, just delete and spread maybe instead?

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
```diff
@@ -0,0 +1,326 @@
...
+        selectedVars.map((item: IObjectWithKey) => {
+          delete draftState[item.key as string];
+        });
+      }
+    });
+    setWatchedVars(updated);
+  };
+
+  const removeIsDisabled = useMemo(() => {
+    return selectedVars === undefined || selectedVars.length === 0;
```
hatpick, line 294: !selectedVars?.length

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/config.ts
```diff
@@ -0,0 +1,16 @@
...
+import formatMessage from 'format-message';
+
+import { TabExtensionConfig, WatchTabKey } from '../types';
+
+import { WatchTabHeader } from './WatchTabHeader';
+import { WatchTabContent } from './WatchTabContent';
+
+export const WatchTabConfig: TabExtensionConfig = {
+  key: WatchTabKey,
+  description: () => formatMessage('Watch tab'),
```
hatpick, line 13: this wont work when locale changes, convert the whole thing to a function

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/index.ts
```diff
@@ -0,0 +1,4 @@
+// Copyright (c) Microsoft Corporation.
```
hatpick, line 1: avoid barrel files for the sake of webpack tree shaking

Composer/packages/client/src/pages/design/DebugPanel/WatchVariableSelector/WatchVariablePicker.tsx
```diff
@@ -0,0 +1,251 @@
...
+  return {
+    emptyMessage: formatMessage('No properties found'),
+    searchPlaceholder: formatMessage('Add a property'),
+  };
+};
+
+const textFieldStyles = (errorMessage?: string): Partial<ITextFieldStyles> => ({
+  field: {
+    fontFamily: DEFAULT_FONT_SETTINGS.fontFamily,
+    fontSize: 12,
```
hatpick, line 41: use fluenttheme font size

Composer/packages/client/src/pages/design/DebugPanel/WatchVariableSelector/WatchVariablePicker.tsx
```diff
@@ -0,0 +1,251 @@
...
+`;
+
+const redErrorMessage = css`
+  color: ${SharedColors.red20};
+`;
+
+export const WatchVariablePicker = React.memo((props: WatchVariablePickerProps) => {
+  const { errorMessage, payload, variableId, path, onSelectPath } = props;
+  const [query, setQuery] = useState(path);
+  const inputBoxElement = useRef<ITextField | null>(null);
```
hatpick, line 69: inputBoxElementRef makes it clear that it's a ref

Composer/packages/client/src/pages/design/DebugPanel/WatchVariableSelector/WatchVariablePicker.tsx
```diff
@@ -0,0 +1,251 @@
...
+  const pickerContainerElement = useRef<null | HTMLDivElement>(null);
```
hatpick, line 70: same here

Composer/packages/client/src/pages/design/DebugPanel/WatchVariableSelector/WatchVariablePicker.tsx
```diff
@@ -0,0 +1,251 @@
...
+      } else {
+        setItems(menuItems);
+      }
+    }, 500),
+    [menuItems, flatPropertyListItems, noSearchResultMenuItem, query]
+  );
+
+  useEffect(() => {
+    handleDebouncedSearch();
+  }, [menuItems, flatPropertyListItems, noSearchResultMenuItem, query]);
```
hatpick, line 186: handleDebouncedSearch as dep should be enough

Composer/packages/client/src/pages/design/DebugPanel/WatchVariableSelector/WatchVariablePicker.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+  const onTextBoxFocus = (event: FocusEvent<HTMLInputElement>) => {
+    onShowContextualMenu(event);
+  };
+
+  const onTextBoxKeyDown = useCallback(
+    (event: KeyboardEvent<HTMLInputElement>) => {
+      // enter
+      if (event.keyCode == 13) {
```
hatpick, line 195: event.key === 'Enter', keyCodes are deprecated

Composer/packages/client/src/pages/design/DebugPanel/WatchVariableSelector/WatchVariablePicker.tsx
```diff
@@ -0,0 +1,251 @@
...
+        items={items}
+        shouldFocusOnMount={false}
+        styles={{
+          root: {
+            maxHeight: '200px',
+            overflowY: 'auto',
+            width: '240px',
+          },
+        }}
+        target={pickerContainerElement}
```
hatpick, line 245: pickerContainerElement.current

Composer/packages/client/src/pages/design/DebugPanel/WatchVariableSelector/utils/components/PickerContextualMenuItem.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+import { PropertyItem, PropertyTreeItem } from './PropertyTreeItem';
+
+const defaultTreeItemHeight = 36;
+
+const labelContainerStyle: IStackStyles = {
+  root: { whiteSpace: 'nowrap', overflow: 'hidden', textOverflow: 'ellipsis', height: defaultTreeItemHeight },
+};
+
+export const GetPickerContextualMenuItem = (query: string, propertyTreeExpanded: Record<string, boolean>) => (
```
hatpick, line 18: camelCase naming: getPickerContextualMenuItem

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
tonyanziano: This is basically the equivalent of:

```
const segments = ['user.outer.inner.someValue'].split('.')
return botTrace?.value?.[segments[0]]?.[segments[1]]?.[segments[2]]?....[segments[n]]
```

Do you know of a better way to do this with a dynamic number of properties to try and access? This basically works its way from the root of the object down to the desired property.

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
tonyanziano: This is the pattern that the debug panel tabs use so I am following it. I do not know the reasoning behind the tabs knowing if they are active or not, but I do not think rewriting it should be part of this PR.

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
tonyanziano: 1. What would be a scenario in which we don't show the selected variables in the list?

2. There are no designs at the moment for disabled variables

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
tonyanziano: I suppose I could do something like:

```
let returnVal = botTrace?.value;
for const segment in segments.split('.') {
  returnVal = returnVal?.[segment]
}
return returnVal
```

tonyanziano: ^^ Just tried this, it works and looks way cleaner.

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
hatpick: when a variable is selected and they re-select it, it shows an error, which I think is too much, I think it should be silent about it, or if we consider it an "error" we can easily prevent it by not showing the selected variables in the select list or show them as disabled with a selected UI treatment (an explicit label or color etc.)

tonyanziano: Ahh I see. It might be the wording / naming of the code that is confusing you here.

"onSelectPath" is actually the "Submit" action of the picker. So let's say you have a row in the table that is already watching `user.myValue`. If you then add another row in the table, type in `user.myValue` and hit "Enter" you will see this validation error underneath the duplicate row and the duplicate **will not** be watched.

I spoke to @mewa1024 about this and we felt that this was a better UX than silently failing or doing a no-op.

tonyanziano: Here's an example:

image

hatpick: @mewa1024 @tonyanziano thanks for explaining the use case, it wasn't far from what I imagined. 👍🏼 

Chrome/edge dev tools have no issues with devs adding the same variable multiple times, I thought we may wanna consider following that.

tonyanziano: Ahhh, interesting! I wonder what VS / VS Code do in this situation.

tonyanziano: VS Code lets you do this without any feedback at all:

image

hatpick: another point I wanted to mention is captured in this screenshot above, "not available", if user enters something random that doesn't exist or they think it does but it doesn't, "not available" is a great indicator vs. undefined (if that's the case)
@mewa1024

mewa1024: agree with all of the above. some other context about the red error message: I think Tony and I landed on that instead of just hiding the variable from the list because the user could type in something freeform (since it's a combobox-ish input rather than a pure dropdown input). 

About the "not available" versus "undefined", "not available" sounds good to me (I'll update the mocks for future reference.)

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
srinaath: Sweet. I had a similar requirement as well. Used lodash get
https://github.com/microsoft/BotFramework-Composer/pull/8141/files#diff-c949aaa883b99af0c85135f197dfbdf385d73097d1f5ba2511fbe540a4f39a49R60.

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
srinaath: The active property is sent to all the tabs but only 1 uses it and is controlled at the parent level. The reason for isActive property passed through to one of the tabs is for implementing the unread notification indicator. 
https://github.com/microsoft/BotFramework-Composer/blob/main/Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WebChatLog/WebChatLogItemHeader.tsx#L23

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
```diff
@@ -0,0 +1,326 @@
...
+  );
+  return value;
+};
+
+export const WatchTabContent: React.FC<DebugPanelTabHeaderProps> = ({ isActive }) => {
+  const currentProjectId = useRecoilValue(rootBotProjectIdSelector);
+  const rawWebChatTraffic = useRecoilValue(webChatTrafficState(currentProjectId ?? ''));
+  const [watchedVars, setWatchedVars] = useState<Record<string, string>>({});
+  const [pickerErrorMessages, setPickerErrorMessages] = useState<Record<string, string>>({});
+  const [selectedVars, setSelectedVars] = useState<IObjectWithKey[]>();
```
srinaath, line 154: Can we use the expanded version itself? 
selectedVariables
watchedVariables

Composer/packages/client/src/recoilModel/dispatchers/webchat.ts
```diff
@@ -5,14 +5,20 @@
...
+  watchedVariablesState,
+} from '../atoms';
 import { WebChatInspectionData } from '../types';
 
 export const webChatLogDispatcher = () => {
   const clearWebChatLogs = useRecoilCallback((callbackHelpers: CallbackInterface) => (projectId: string) => {
     const { set } = callbackHelpers;
     set(webChatTrafficState(projectId), []);
     set(webChatInspectionDataState(projectId), undefined); // clear the inspection panel
+    set(watchedVariablesState(projectId), []); // TODO: might not want to do this depending on how annoying it is for the user -- do you want to wipe variables when you restart convo?
```
srinaath, line 21: @tonyanziano We definitely dont want to wipe out the variable list on bot restart like you said following the VSCode model

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
```diff
@@ -0,0 +1,326 @@
...
+              payload={memoryVariablesPayload}
+              variableId={item.key}
+              onSelectPath={onSelectPath}
+            />
+          );
+        } else if (column.key === ValueColumnKey) {
+          // render the value display
+          if (mostRecentBotState) {
+            const value = getValueFromBotTraceScope(item.value, mostRecentBotState?.activity);
+            if (value !== null && typeof value === 'object') {
```
srinaath, line 253: how are we rendering the value `null`

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
hatpick: @mewa1024 just to clarify, if a variable is indeed available but has no value at watch time, it would be undefined, but if it's not available (doesn't exist in the list, like user thinks it's there but it's not) it shows unavailable.

mewa1024: ah, thanks for the clarification. in the JSON for the bot, would an undefined thing look like this:

`"propertyname" : "undefined"`

if so, seems like it should be shown in monospaced font, like this:
image

rather than this:
image

Thoughts?

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
mewa1024: @sorgh and I chatted about this offline and here is an example of a table that is watching variables that return an object, "undefined", "null", or are not available:
image

"unavailable" in Segoe gray 130 (#605E5C); "undefined" and "null"  appear the same as the other values (same monospace font families and color coding).

Will this work?

srinaath: That should work for us. Also, unrelated, boolean would just be rendered as true or false without the string interpolation "true".

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
tonyanziano: Awesome, I'll do that instead.

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
tonyanziano: it needs to be of type `{ key, value }[]` so this won't work

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
tonyanziano: I think our editor only allows you to set a value to `null` if it is not set, if it is actually `undefined`, then I believe Monaco just omits the value from the display:

So here's what it would look like:

image


If we had a `prop4: undefined` it would not even show up.

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/config.ts
tonyanziano: I'm not familiar with how the strings for each locale are generated. What needs to be converted to a function here?

Composer/packages/client/src/pages/design/DebugPanel/WatchVariablePicker/WatchVariablePicker.tsx
```diff
@@ -0,0 +1,263 @@
...
+  useEffect(() => setItems(menuItems), [menuItems]);
+
+  const handleDebouncedSearch: () => void = useCallback(
+    debounce(() => {
+      if (query) {
+        const searchableItems = flatPropertyListItems;
+
+        const predicate = getFilterPredicate(query);
+
+        const filteredItems = searchableItems.filter(predicate);
```
beyackle, line 176: `.filter()` makes a new array anyway, so there's no need to reassign flatPropertyListItems to searchableItems in here. I think just `const filteredItems = flatPropertyListItems.filter(predicate)` ought to work.

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
```diff
@@ -0,0 +1,325 @@
...
+  height: 16px;
+  line-height: 16px;
+`;
+
+const watchTableStyles: Partial<IDetailsListStyles> = {
+  root: {
+    maxHeight: `calc(100% - ${toolbarHeight}px)`,
+  },
+  contentWrapper: {
+    overflowY: 'auto' as 'auto',
```
hatpick, line 70: nit: as ```any```

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
```diff
@@ -0,0 +1,325 @@
...
+          let variables = await getMemoryVariables(currentProjectId, { signal: abortController.signal });
+          // we don't want to show variables that are already being watched
+          variables = variables.filter((v) => !watched.find((watchedV) => watchedV === v));
+
+          setMemoryVariablesPayload({ kind: 'property', data: { properties: variables } });
+        } catch (e) {
+          // error can be due to abort
+        }
+      })();
+    }
```
hatpick, line 157: effect should return the following:
```
return () => {
  abortController.abort();
};
```

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
```diff
@@ -0,0 +1,325 @@
...
+        // no bot trace available
+        return <span css={unavailbleValue}>{formatMessage('unavailable')}</span>;
+      }
+    },
+    [mostRecentBotState, watchedVariables]
+  );
+
+  // TODO: update to office-ui-fabric-react@7.170.x to gain access to "flexGrow" column property to distribute proprotional column widths
+  // (name column takes up 1/3 of space and value column takes up the remaining 2/3)
+  const watchTableColumns: IColumn[] = [
```
hatpick, line 212: nit: this can be memoized

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
```diff
@@ -0,0 +1,325 @@
...
+      setWatchedVariables(currentProjectId, updatedCommitted);
+      setUncommittedWatchedVariables(updatedUncommitted);
+    }
+  }, [currentProjectId, selectedVariables, setWatchedVariables, setUncommittedWatchedVariables]);
+
+  const removeIsDisabled = useMemo(() => {
+    return !selectedVariables?.length;
+  }, [selectedVariables]);
+
+  function onRenderDetailsHeader(props, defaultRender) {
```
hatpick, line 271: be consistent with arrow vs. traditional function definition

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
```diff
@@ -0,0 +1,325 @@
...
+    );
+  }
+
+  if (!isActive) {
+    return null;
+  }
+
+  return (
+    <Stack verticalFill>
+      <Stack.Item
```
hatpick, line 288: use CommandBar? (it has accessibility + points vs. current implementation)

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabObjectValue.tsx
```diff
@@ -0,0 +1,58 @@
...
+  border: none;
+`;
+
+const objectCellStyle = (numLinesOfJson: number) => css`
+  height: ${numLinesOfJson * 18}px;
+  width: 360px;
+`;
+
+type WatchTabObjectValueProps = {
+  value: object;
```
hatpick, line 23: any is probably a better choice

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabObjectValue.tsx
```diff
@@ -0,0 +1,58 @@
...
+    return objectCellStyle(numLinesOfJson);
+  }, [value]);
+
+  return (
+    <div css={objectCell}>
+      <JsonEditor
+        editorSettings={{
+          fadedWhenReadOnly: false,
+          fontSettings: {
+            fontFamily: DEFAULT_FONT_SETTINGS.fontFamily,
```
hatpick, line 40: should this be using font settings from user settings or default?

Composer/packages/client/src/pages/design/DebugPanel/WatchVariablePicker/WatchVariablePicker.tsx
```diff
@@ -0,0 +1,261 @@
...
+const getStrings = () => {
+  return {
+    emptyMessage: formatMessage('No properties found'),
+    searchPlaceholder: formatMessage('Add a property'),
+  };
+};
+
+const textFieldStyles = (errorMessage?: string): Partial<ITextFieldStyles> => ({
+  field: {
+    fontFamily: DEFAULT_FONT_SETTINGS.fontFamily,
```
hatpick, line 40: same comment

Composer/packages/client/src/pages/design/DebugPanel/WatchVariablePicker/WatchVariablePicker.tsx
```diff
@@ -0,0 +1,261 @@
...
+        }
+      }
+    },
+    [currentProjectId, variableId, query, watchedVariables]
+  );
+
+  const onDismiss = useCallback(() => {
+    setPropertyTreeExpanded({});
+    onHideContextualMenu();
+  }, []);
```
hatpick, line 218: onHideContextualMenu to deps

Composer/packages/client/src/pages/design/DebugPanel/WatchVariablePicker/WatchVariablePicker.tsx
```diff
@@ -0,0 +1,261 @@
...
+    return (item: IContextualMenuItem) =>
+      item.data.node.children.length === 0 && item.secondaryText?.toLowerCase().indexOf(q.toLowerCase()) !== -1;
+  }, []);
+
+  const menuItems = useMemo(getContextualMenuItems, [payload, propertyTreeExpanded]);
+
+  useEffect(() => setItems(menuItems), [menuItems]);
+
+  const handleDebouncedSearch: () => void = useCallback(
+    debounce(() => {
```
hatpick, line 170: does this work? The debounced function gets redefined everytime any of the deps including query changes.

Composer/packages/client/src/pages/design/DebugPanel/WatchVariablePicker/WatchVariablePicker.tsx
```diff
@@ -0,0 +1,261 @@
...
+        setItems(filteredItems);
+      } else {
+        setItems(menuItems);
+      }
+    }, 500),
+    [menuItems, flatPropertyListItems, noSearchResultMenuItem, query]
+  );
+
+  useEffect(() => {
+    handleDebouncedSearch();
```
hatpick, line 189: adding to the comment above, this effect is pretty confusing

Composer/packages/client/src/pages/design/DebugPanel/WatchVariablePicker/utils/components/PickerContextualMenuItem.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+import React from 'react';
+import { IContextualMenuItemProps } from 'office-ui-fabric-react/lib/ContextualMenu';
+import { IStackStyles, Stack } from 'office-ui-fabric-react/lib/Stack';
+import { Text } from 'office-ui-fabric-react/lib/Text';
+import { NeutralColors } from '@uifabric/fluent-theme';
+
+import { PropertyItem, PropertyTreeItem } from './PropertyTreeItem';
+
+const defaultTreeItemHeight = 36;
```
hatpick, line 12: this is redefined in the next file below too, maybe reuse from a constants.ts?

srinaath (dismissed):
@tonyanziano Im running the branch with its latest commits. The following are the issues im seeing.
1. The "Watch Variables" seems to be persisted across bots. The watch variables should be a bot state should wipe out when opening a new bot
2. ~~Collapsing and opening the Debug Panel  seems to wipe off variables.~~ FIXED
3. New variables added to the memory scope are showing up in the "Send a Response" variable picker but not in the variable picker inside the "Watch" Tab. Im assuming some sort of memoization issue.
4. ~~Boolean variables are showing up as a string. For ex "user.greeted".
Let me know if you are seeing the same bugs as well.~~ -- false alarm

Composer/packages/client/src/pages/design/DebugPanel/WatchVariablePicker/WatchVariablePicker.tsx
tonyanziano: looking into it

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabObjectValue.tsx
tonyanziano: this should probably use the user settings

Composer/packages/client/src/pages/design/DebugPanel/WatchVariablePicker/WatchVariablePicker.tsx
tonyanziano: If the user settings font affects the form editors (all inputs in the app) then it should probably use the user settings font, otherwise I think it should use the default because it is a text input and not a code editor

tonyanziano:
I have addressed all of Soroush's comments and also adjusted some styles that looked off from the specs. Here is the updated tab with new "not available" message and placeholder text:

watch-tab-empty

@srinaath I still can't repro bug 4 as seen in the screenshot, but I can confirm bugs 1-3 and am working to fix these.

srinaath:
@tonyanziano Ignore bug 4. I just realized a bot logic error which was converting the boolean to string to on the variable. So basically it was casting to "true". So we are good on that front.
Screen Shot 2021-06-28 at 3 29 02 PM

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabObjectValue.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { useRecoilValue } from 'recoil';
+
+import { userSettingsState } from '../../../../../recoilModel';
+
+const editorStyles = css`
+  border: none;
+`;
+
+const objectCellStyle = (numLinesOfJson: number) => css`
+  height: ${numLinesOfJson * 18}px;
```
srinaath, line 17: @tonyanziano On a huge JSON like the dialog object we would need to cap the height of the cell. Basically, allowing scroll to inside the editor. Can we confirm the max height of the value cell @mewa1024 
Screen Shot 2021-06-28 at 3 47 06 PM

tonyanziano: Great point!

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
```diff
@@ -0,0 +1,336 @@
...
+
+const watchTableLayout: DetailsListLayoutMode = DetailsListLayoutMode.justified;
+
+// Returns the specified property from the bot state trace if it exists.
+// Ex. getValueFromBotTraceMemory('user.address.city', trace)
+export const getValueFromBotTraceMemory = (
+  valuePath: string,
+  botTrace: Activity
+): { value: any; propertyIsAvailable: boolean } => {
+  const pathSegments = valuePath.split('.');
```
srinaath, line 107: When we are trying to access memory scope properties. For ex `dialog` or `conversation` as whole. The valuePath.split('.') just leads to one element and gets popped up. If the goal is to check if the path exists or not can we just not use 
```javascript
const valueOfProp = get(botTrace?.value, valuePath, undefined)
if(valueOfProp === undefined) then the propertyIsAvailable is false?
```

tonyanziano: I'm using `propertyIsAvailable` to distinguish between an `undefined` value and an `unavailable` value because the design has different treatments for each.

`undefined` meaning that the property could exist on the parent key, but it is not set
`unavailable` meaning that the property **could not and does not exist** because the property's parent path is not set

That being said, as far as I know, we do not offer a way for users to set a property to undefined in the form editor, because `"someProperty": undefined` is not actually valid JSON even though it is valid JavaScript.

If we treat `undefined` values as `unavailable` then we don't need both treatments, but that might be confusing to users because standard dev tools treat them differently:

image

So I guess this all depends on what the design is: Do we want to call all undefined / unset values `unavailable`?

@mewa1024

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabObjectValue.tsx
mewa1024: What about 80px max height? That would show 4 full lines, plus the top of a 5th line, if any. (Are we assuming that if it's an enormous JSON object the user would copy and paste into a code editor to view the whole thing?)

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
mewa1024: yes, the design was trying to match what users would expect from "Watch" in dev tools, namely, different treatments for "not available" and "undefined". (@sorgh, any opinions? I think you mentioned the not available/undefined difference earlier.)

tonyanziano: related but separate:

to Srinaath's point, there is actually a bug when trying to watch the root memory scopes themselves `user`, `conversation`, etc. and I am trying to fix that

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabObjectValue.tsx
srinaath: Yep.

Composer/packages/client/src/pages/design/DebugPanel/TabExtensions/WatchTab/WatchTabContent.tsx
hatpick: > yes, the design was trying to match what users would expect from "Watch" in dev tools, namely, different treatments for "not available" and "undefined". (@sorgh, any opinions? I think you mentioned the not available/undefined difference earlier.)

undefined is an accepted value, not available is for the gibberish user enters freely in the picker like:
user.name is defined by the user and until it's set, it's undefined
soroush.is.awesome is not available but it's true

tonyanziano: so would the distinction then be:

`undefined` is anything not set in a **defined memory scope** (user, conversation, customUserScope, etc.)
`not available` is anything outside one of these scopes

tonyanziano:
All of the bugs / issues have been fixed and styles updated to appear more like the updated designs:

watch-tab-input-padding
watch-tab-values
watch-tab-empty-state

mewa1024:
@tonyanziano minor thing, can the table be slightly narrower so there is 16px of space on either side? where the pink lines are:
image

tonyanziano:
@mewa1024 Fixed spacing:

table-padding-1

table-padding-2

hatpick (approved):

PR #8078, opened 2021-06-14
feature: show survey card on Home page

beyackle:
## Description

This adds the logic to show a notification on the home screen prompting the user to take a survey. This also adds some new notification types and a fairly generic way to track a user's usage of various features inside the browser's own local storage, which will be useful for other possible business cases (like extending the conditions surveys pop up in, or cosmetic things like Achievements).

## Task Item

closes #8029

## Screenshots

image

PR #8155, opened 2021-06-21
feature: add new survey notification to design page

### LINKED ISSUE #8029: HaTs telemetry items

beyackle:
## Description

[This is a redo of #8078, which was closed by mistake]

This adds the logic to show a notification on the home screen prompting the user to take a survey. This also adds some new notification types and a fairly generic way to track a user's usage of various features inside the browser's own local storage, which will be useful for other possible business cases (like extending the conditions surveys pop up in, or cosmetic things like Achievements).

## Task Item

refs #8029
[will close when this merges into main]

## Screenshots

image


Composer/packages/electron-server/src/main.ts
```diff
@@ -295,11 +295,16 @@ async function run() {
...
     const mainWindow = getMainWindow();
+    const machineId = await getMachineId();
+
     mainWindow?.webContents.send('session-update', 'session-started');
 
     if (process.env.COMPOSER_DEV_TOOLS) {
       mainWindow?.webContents.openDevTools();
     }
+
+    log(`Machine ID is ${machineId}`);
```
srinaath, line 306: Is this log required?

Composer/packages/client/src/components/Notifications/useSurveyNotification.tsx
```diff
@@ -0,0 +1,89 @@
...
+const buildUrl = (info: MachineInfo) => {
+  // User OS
+  // hashed machineId
+  // composer version
+  // maybe include subscription ID; wait for global sign-in feature
+  // session ID  (global telemetry GUID)
+  const version = process.env.COMPOSER_VERSION;
+
+  const parameters = {
+    Source: 'Composer',
```
srinaath, line 23: camelCase why capitalized?

Composer/packages/client/src/components/Notifications/useSurveyNotification.tsx
```diff
@@ -0,0 +1,89 @@
...
+    os: info?.os || 'Unknown',
+    machineId: info?.id,
+    version,
+  };
+
+  return (
+    `${SURVEY_URL_BASE}?` +
+    Object.keys(parameters)
```
srinaath, line 31: nit: can use `query-string` stringify method

Composer/packages/client/src/App.tsx
```diff
@@ -42,6 +43,10 @@ export const App: React.FC = () => {
     ipcRenderer?.on('cleanup', (_event) => {
       performAppCleanupOnQuit();
     });
+
+    ipcRenderer?.on('machine-info', (_event, info) => {
+      setMachineInfo(info);
```
srinaath, line 48: It looks like you are using the machine info only inside the useSurveyNotification hook. If so, why do we need to create a recoil state for it. You can just use the `ipcRenderer.on` inside the hook. Ex `useInitializeLoger.ts`

srinaath (changes requested):

Composer/packages/electron-server/src/main.ts
beyackle: It's not required, but it's useful - this goes into the debug log which users won't see unless they're running via the console anyway.

Composer/packages/client/src/components/Notifications/useSurveyNotification.tsx
beyackle: I know these are usually lowercase, but uppercase S is the spec I was given.

Composer/packages/client/src/components/Notifications/useSurveyNotification.tsx
beyackle: Good idea. I'll fix that.

Composer/packages/client/src/App.tsx
beyackle: I was trying to make this work last week, but it runs into annoying race conditions where the message handler isn't initialized before the message gets sent and so nothing gets through. Using the Recoil store lets us put the machine info down somewhere whenever we receive it and then count on it being available when we need it.

Composer/packages/client/src/components/Notifications/NotificationCard.tsx
```diff
@@ -36,7 +37,7 @@ const cardContainer = (show: boolean, ref?: HTMLDivElement | null) => () => {
     border-left: 4px solid #0078d4;
     background: white;
     box-shadow: 0 6.4px 14.4px 0 rgba(0, 0, 0, 0.132), 0 1.2px 3.6px 0 rgba(0, 0, 0, 0.108);
-    min-width: 340px;
+    width: 340px;
```
tdurnford, line 40: Why did you change this from `min-width` to `width`?

tdurnford, line 40: Why did you change this from `min-width` to `width`?

Composer/packages/client/src/App.tsx
srinaath: Makes sense to usethe recoil state unless the hook can be registered in app.tsx

Composer/packages/client/src/components/Notifications/NotificationCard.tsx
```diff
@@ -112,13 +113,20 @@ const cardDescription = css`
...
-const linkButton = css`
-  color: #0078d4;
-  float: right;
-  font-size: 12px;
-  height: auto;
-  margin: 4px 0 4px 8px;
-`;
+const linkButton = {
+  root: {
+    padding: '0',
```
tdurnford, line 118: `padding` and `border` should be `0px`. Also, why are you setting the border styles? `ActionButton`s don't have borders by default.

Composer/packages/client/src/components/Notifications/NotificationCard.tsx
```diff
@@ -148,27 +156,58 @@ export type NotificationProps = {
...
-        ))}
+        <Stack horizontal horizontalAlign="space-between">
+          <Stack {...stackProps}>
+            {leftLinkList.map(
+              (link) => link != null && <Stack.Item key={link.label}>{makeLinkLabel(link)}</Stack.Item>
+            )}
+          </Stack>
+          <Stack {...stackProps}>
+            {rightLinkList.map(
+              (link) => link != null && <Stack.Item key={link.label}>{makeLinkLabel(link)}</Stack.Item>
```
tdurnford, line 207: Why would a link be null?

Composer/packages/electron-server/src/main.ts
tdurnford: Logs are only visible from the terminal, so I think it's fine.

Composer/packages/types/src/shell.ts
```diff
@@ -67,12 +67,16 @@ export type DisabledMenuActions = {
...
 
 export type Notification = {
-  type: 'info' | 'warning' | 'error' | 'pending' | 'success' | 'question' | 'congratulation';
+  type: 'info' | 'warning' | 'error' | 'pending' | 'success' | 'question' | 'congratulation' | 'custom';
   title: string;
   description?: string;
   retentionTime?: number;
   link?: NotificationLink;
-  links?: NotificationLink[];
+  links?: (NotificationLink | null | undefined)[];
```
tdurnford, line 75: Why are we adding null or undefined? Seems very unnecessary.

Composer/packages/client/src/App.tsx
tdurnford: Can you explain the race condition a bit more. I'm also confused why we can't just initialize the survey here when the event is received? If the message handler doesn't get initialized before the message is sent, we would never get the message, right?

tdurnford (changes requested):

Composer/packages/client/src/components/Notifications/useSurveyNotification.tsx
```diff
@@ -0,0 +1,85 @@
...
+    Source: 'Composer',
+    os: info?.os || 'Unknown',
+    machineId: info?.id,
+    version,
+  };
+
+  return `${SURVEY_URL_BASE}?${querystring.stringify(parameters)}`;
+};
+
+export function useSurveyNotification() {
```
tdurnford, line 33: Arrow function

Composer/packages/client/src/components/Notifications/useSurveyNotification.tsx
```diff
@@ -0,0 +1,85 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
```
tdurnford, line 3: This should be a ts instead of a tsx file since it is a hook and it does not have any react components

Composer/packages/client/src/components/Notifications/NotificationCard.tsx
beyackle: I was using that to indicate a gap, to make the design work, but you're right that this is unnecessary now. I'll fix it.

Composer/packages/types/src/shell.ts
beyackle: It is, now that I've split them into left and right groups.

Composer/packages/client/src/components/Notifications/NotificationCard.tsx
beyackle: 0 is equivalent to 0 in any unit; there's no need to specify pixels there. And there was a 1 pixel border there that this eliminates, just to align to the designs better.

Composer/packages/client/src/components/Notifications/NotificationCard.tsx
beyackle: This is in order to keep the width of a card consistent when the text inside needs to wrap to multiple lines. They fit into the sidebar that way.
image

Composer/packages/electron-server/src/utility/machineId.ts
```diff
@@ -13,17 +13,20 @@ import { readTextFileSync, writeJsonFileSync } from './fs';
 
 export const persistedFilePath = path.join(app.getPath('userData'), 'persisted.json');
 
+function truncate(str: string): string {
```
tdurnford, line 16: Arrow function

Composer/packages/lib/code-editor/src/components/toolbar/__tests__/ToolbarButtonMenu.test.tsx
```diff
@@ -35,7 +35,7 @@ const templatePayload: TemplateRefPayload = {
 const propertiesPayload: PropertyRefPayload = {
   kind: 'property',
   data: {
```
tdurnford, line 37: Why are these tests being changed? They're not related to this work at all.

beyackle: They're being changed because they were failing mysteriously. I feel like the word "this" is too common to be used as a name and search target in tests like this without false positives happening.

Composer/packages/electron-server/src/utility/machineId.ts
beyackle: I don't see what that would get us here. This isn't inside a component that gets re-rendered, this isn't exported anywhere, it's just a simple utility function that belongs to this one file.

tdurnford: The other functions in this file use arrow functions. Best to be consistent. More of a nit.

PR #8160, opened 2021-06-22
feat: import QnA from portal

### LINKED ISSUE #7981: Importing an existing QNA Maker KB
Initial thoughts on the workflow 

1. An update to the Create a QnA Maker KB modal to provide a 3rd option (in addition to import from URL and start from scratch), to import from an existing KB. 

2. If the user chooses this option, they would be prompted to select a subscription / resource group / qna Maker resource (Reuse existing views from the Settings page for the QNA KB action and build on it to add a view to list the Kbs and download the content when choosing a KB) (this might prompt the user to authenticate)

3. Once we have the QnA subscription, we can get a list of KBs associated with it with a single API call (see Cognitive Services APIs Reference (microsoft.com)).

4. The user selects a KB and then we can import the existing content in .qna format with a single call -> e.g. https://westus.api.cognitive.microsoft.com/qnamaker/v4.0/knowledgebases/a39be546-b70a-4d04-91e9-72faf031fd42/Prod/qna?qnaformat=true.

The QNA portal API’s are available to fetch KBs in a subscription and pull the content as .qna format and persist them locally. 
In R14 the timeline is pretty tight to venture into any scenarios involving syncing between remote portal kb and local kb. The current understanding is the content in Composer would always trump what is in the portal as it would just write the local file contents into the remote portal without accounting for lastAccessed timestamp etc.

### LINKED ISSUE #8194: Be more visible (user friendly) for QnA action
Be more visible (user friendly) for QnA action 

Scenario 1: 
Move to the QNA authoring all-up. If there is no focus on the dialog item (left panel). The KB authoring 3 dot will not show up. 
image
After choosing the dialog first and then the KB authoring 3 dot will be shown up as following. 
image

Expect to see the 3 dot no matter the focus is when mouse hover to the KB name.

Scenario 2: 
If this is no KB available , we will show "create kb"  button as below. 
image

However if I created one KB for a locale and then switch  to other locale or change the composer authoring locale .
Even if there is no KB available for this locale, we do not show "create kb". 
image
image

Expect to the the "create KB" button for each locale if there is no KB under such locale. 

Discussed with @zhixzhan for the root cause and we need @vivekkshankar 's comments on this. 

including @boydc2014 as well for more comments.

### LINKED ISSUE #8212: User unable to add LUIS or QnA resource from composer
## Describe the bug

 customer is using bot framework composer 2.0.0, when he tries to add LUIS for the project, he gets the login page, after successfully login.

he could see the azure directory and subscription, however the LUIS resource never displays like following:

while i dont have this issue in my lab, not sure if this is related with the sponsorship sub?

Please see ICM issue:
https://portal.microsofticm.com/imp/v3/incidents/details/247570921/home

## Version

2.0.0

## Browser

- [ ] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

## To Reproduce

See ICM

zhixzhan:
## Description

Import QnA knowledge base from QnA Maker portal.

## Task Item
close #7981 
close #8194 
close #8212 

#### Sign in (web application)
image

#### Sign in (electron)
image

#### Replace QnA from QnA Portal
test


Composer/packages/client/src/components/QnA/utilities.ts
```diff
@@ -0,0 +1,14 @@
...
+export const QnALanguageToLocale = (language: string): string => {
+  switch (language) {
+    case 'Chinese_Simplified':
+      return 'zh-cn';
+    case 'English':
+      return 'en-us';
+    case 'French':
+      return 'fr-fr';
+  }
+  return language;
```
zhixzhan, line 13: QnAMaker portal API returns `language` of KB, but we need` locale` to do filter. Need to find a better way than hard code

zhixzhan:
image

if selected resource did not match current `QNA_SUBSCRIPTION_KEY`, will not be able to fetch KB content.

image

Composer/packages/server/src/models/utilities/parser.ts
```diff
@@ -64,3 +70,18 @@ export async function parseQnAContent(url: string, multiTurn: boolean) {
...
   return await importQnAFromUrl(builder, url, subscriptionKey, multiTurn);
 }
+
+// subscriptionKey from selection.
+export async function importQnAContentFromQnAMakerPortal(endpoint: string, kbId: string, subscriptionKey: string) {
+  const cognitiveServicesCredentials = new CognitiveServicesCredentials(subscriptionKey);
+  const resourceClient = new QnAMakerClient(cognitiveServicesCredentials, endpoint);
+
+  // 'Test' or 'Prod'
+  const result = await resourceClient.knowledgebase.download(kbId, 'Test');
```
srinaath, line 80: @boydc2014 Thoughts here? https://docs.microsoft.com/en-us/javascript/api/@azure/cognitiveservices-qnamaker/knowledgebase?view=azure-node-latest#download_string__Models_EnvironmentType__KnowledgebaseDownloadOptionalParams__ServiceCallback_QnADocumentsDTO__

It looks like we can have the test Index and then we can promote the KB to the prod index.

Composer/packages/client/src/components/QnA/utilities.ts
srinaath: Aah. This is a bit tricky. We do have have a locale file here https://github.com/microsoft/BotFramework-Composer/blob/main/Composer/packages/client/src/locales.ts and its in memory for us to parse back the locale from the language.

Composer/packages/client/src/components/QnA/CreateQnAFromUrl.tsx
```diff
@@ -0,0 +1,173 @@
...
+  const validFormDataName = validateName(qnaFiles);
+
+  const onChangeNameField = (value: string | undefined) => {
+    updateNameField(value);
+    onUpdateInitialName?.(value ?? '');
+    updateNameError(value);
+  };
+
+  const onChangeUrlsField = (value: string | undefined, index: number) => {
+    const urls = cloneDeep(formData.urls);
```
srinaath, line 72: why do we need cloneDeep here instead of spread?

Composer/packages/client/src/components/ManageService/ManageService.tsx
```diff
@@ -270,14 +270,18 @@ const ManageService: React.FC<ManageServiceProps> = (props: ManageServiceProps)
...
-            name: name,
-            resourceGroup: resourceGroup,
-            region: accounts[account].location || '',
-            key: keys?.key1 || '',
-          });
+        try {
+          const keys = await cognitiveServicesManagementClient.accounts.listKeys(resourceGroup, name);
+          if (keys?.key1) {
+            keyList.push({
+              name: name,
```
srinaath, line 277: very minor nit. Can use shorthand. {name, resourceGroup...}

Composer/packages/client/src/pages/knowledge-base/table-view.tsx
```diff
@@ -456,7 +471,7 @@ const TableView: React.FC<TableViewProps> = (props) => {
                       {
                         key: 'update',
                         iconProps: { iconName: 'Download' },
-                        name: formatMessage('Import new URL and overwrite'),
+                        name: formatMessage('Replace content'),
```
srinaath, line 474: Screen Shot 2021-06-24 at 4 51 58 PM
Is the UI in the figma outdated?

Composer/packages/server/src/models/utilities/parser.ts
zhixzhan: I noticed that if a KB have not been published, only 'Test' can download, so here set 'Test' to default to cover more case.

Composer/packages/client/src/components/QnA/CreateQnAFromQnAMaker.tsx
```diff
@@ -0,0 +1,94 @@
...
+    <Fragment>
+      <Stack>
+        <div>
+          <Text styles={knowledgeBaseStyle}>
+            {formatMessage('Import content from an existing KB on the QnA maker portal')}
+          </Text>
+          <p>
+            <span css={subText}>
+              {formatMessage(
+                'Select this option when you want to Import content from an existing KB on the QnA maker portal. The imported content will be mapped to the target locale you select.'
```
srinaath, line 57: when you want to `i`mport content..

Composer/packages/client/src/components/QnA/CreateQnAFromQnAMaker.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  useEffect(() => {
+    const disabled = hasErrors || !formData.name;
+    onChange(formData, disabled);
+  }, [formData]);
+
+  return (
+    <Fragment>
+      <Stack>
+        <div>
```
srinaath, line 50: why a Stack and a div with no styling inside?

Composer/packages/client/src/pages/knowledge-base/table-view.tsx
zhixzhan: Hi @srinaath did you find the latest one
image

image

it's a little different than the archived one.
image

Composer/packages/client/src/components/QnA/utilities.ts
zhixzhan: I found the language/locale mapping of bf-lu publishing, 

https://github.com/microsoft/botframework-cli/blob/132e0cd1f5c393edc35fdc57f971b6cb8da94445/packages/lu/src/parser/utils/enums/localeToQnALanguageMap.js

And use it created a inverted map

image

finally list out all related locale KBs. @srinaath @boydc2014

srinaath: Good find @zhixzhan. Looks like bf-cli faced the same issue and built this file :)

Composer/packages/client/src/components/QnA/CreateQnAModal.tsx
```diff
@@ -0,0 +1,732 @@
...
+    }
+    onSubmit(formData);
+    setInitialName('');
+    TelemetryClient.track('AddNewKnowledgeBaseCompleted', { source: formData.urls?.length ? 'url' : 'none' });
+  };
+
+  const onSubmitImportKB = async () => {
+    if (key && token && selectedKb && formData) {
+      // TODO: add to all matched language or ask user for specific locale.
+      const createdOnLocales = locales.filter((item) => localeToLanguage(item) === selectedKb.language);
```
zhixzhan, line 688: Update to auto match locales according to language of selected KB. @srinaath @boydc2014

boydc2014 (approved):

PR #8168, opened 2021-06-22
fix: Publish bots one at a time to prevent LUIS 429 errors

### LINKED ISSUE #7807: Publishing three bots at once puts user at risk of hitting 429 error on LUIS authoring key
When publishing three bots at once (root and skills), they all start at the same time, which puts the user at risk of hitting 429 errors from luis authoring, and it is not obvious that you just need to restart the publish to resolve it. 

Ideally, multiple publish processes started at the same time would have some delay added in between them, or if the 429 error is hit, there would be an automatic retry or a message suggesting they retry publishing

taicchoumsft:
## Description
Publish bots to Azure one at a time (similar to local publish starting bots one at a time) to prevent LUIS rate exceeded errors.  This sometimes happens when we use a low TPS authoring key, or when multiple bots are using the same key. 

In the long run we should use a js ['semaphore'](https://github.com/abrkn/semaphore.js) or [rate limiter](https://github.com/nfriedly/express-rate-limit) on the sensitive publish and build controller endpoints, or turn the `lubuild.Builder` into a service.

## Task Item
closes #7807


taicchoumsft:
@boydc2014 , @srinaath , here's our luBuilder buildcommand in `builder.ts`: 

```
    const buildResult = await this.luBuilder.build(luContents, config.authoringKey, config.botName, {
      endpoint: authoringEndpoint,
      suffix: config.suffix,
      keptVersionCount: 10,
      isStaging: false,
      region: config.region,
      directVersionPublish: directVersionPublish,
    });
```

We don't appear to use `luisAPITPS`, `retryCount` or `retryDuration` for `lubuild` (defaults to [5, 1, 1000](https://github.com/microsoft/botframework-cli/blob/main/packages/lu/src/parser/lubuild/builder.ts#L189) respectively). 

Should we also tweak this options?  Maybe lowering `luisAPITPS` will be enough to prevent this issue entirely.

srinaath:
https://docs.microsoft.com/en-us/azure/cognitive-services/luis/luis-limits#authoring-resource-limits
So looking at the LuisAPITPS in `bf-lu` it looks it chunks the luis content with default being 5 before sending it for publishing. So, when three bots are hitting the same service at the same time it is definitely possible that 15 transactions are done in an instant which leads to the error. Have you tried tweaking those options on a set of 3-4 bots in a botproject  @taicchoumsft?  

We are actually doing a lubuild even on bot start but we dont encounter the error as bots are started synchronously which aligns with what you have done in this PR.

I think for R14, the preference would be
1) Tweak the parameters. Ideal, as we can still continue parallel builds. With retry mechanism we may achieve the robustness in builds
2) Synchronous solution that you have built. Deployment will be slower but its a hit we could take as the deployment in general is a slow process.

The ideal solution of course like you said is a service for lubuild on Composer side that processes the queue to prevent overloads.

taicchoumsft:
> https://docs.microsoft.com/en-us/azure/cognitive-services/luis/luis-limits#authoring-resource-limits
> So looking at the LuisAPITPS in `bf-lu` it looks it chunks the luis content with default being 5 before sending it for publishing. When three bots are hitting the same service at the same time it is definitely possible that 15 transactions are done in an instant which leads to the error. Have you tried tweaking those options on a set of 3-4 bots in a botproject @taicchoumsft?
> 
> We are actually doing a lubuild even on bot start but we dont encounter the error as bots are started synchronously which aligns with what you have done in this PR.
> 
> I think for R14, the preference would be
> Option 1) Tweak the parameters. Ideal, as we can still continue parallel builds. With retry mechanism we may achieve the robustness in builds
> Option 2) Synchronous solution that you have built. Deployment will be slower but its a hit we could take as the deployment in general is a slow process.
> 
> The ideal solution of course like you said is a service for lubuild on Composer side that processes the queue to prevent overloads.

Thanks for the feedback @srinaath .  Yes sounds good, I'll experiment more to see if these luBuild tweaks will work first.

boydc2014:
@srinaath @taicchoumsft 

The two options are both valid approaches as we've discussed in last chat. I will prefer the second which makes it sequential over the option than tweaking the parameters, because of the follow reasons
* There is certain limitation of tweaking the parameter, the smallest value you can give to LUISTPS is 1, which means if you have >5  projects, your totally TPS will still be >5 and will potentially hit an error. 
* There is no good way to tweaking the parameter, you probably have to MAXTPS(5)/COUNT_OF_PROJECTS, which is not guaranteed to work and it requires you plugin through the data needed.

taicchoumsft:
> @srinaath @taicchoumsft
> 
> The two options are both valid approaches as we've discussed in last chat. I will prefer the second which makes it sequential over the option than tweaking the parameters, because of the follow reasons
> 
> * There is certain limitation of tweaking the parameter, the smallest value you can give to LUISTPS is 1, which means if you have >5  projects, your totally TPS will still be >5 and will potentially hit an error.
> * There is no good way to tweak the parameter, you probably have to do MAXTPS(5)/COUNT_OF_PROJECTS, which is not guaranteed to work and it requires plugging through the extra information needed, touching probably many interfaces.

Thanks @boydc2014! Talked to @srinaath yesterday, we agreed take this change first (let me know if I missed something @srinaath) , instead of tweaking the lu configs, and later go in to add the LUbuilder service at a later date when/if needed.

boydc2014 (approved):

srinaath:
@taicchoumsft Can we get this PR merged so that @lauren-mills can check if multiple bot publish does not error it in the nightly tomorrow?

taicchoumsft:
> @taicchoumsft Can we get this PR merged so that @lauren-mills can check if multiple bot publish does not error it in the nightly tomorrow?

Thanks @srinaath , just merged.

PR #8173, opened 2021-06-22
feat: Added HandOffResourcesToAdmin step

VamsiModem:
## Description
1. Added HandOffToAdmin step
2. Fixed height issues in the wizard
3. Fixed selection issue in the choose resource list component.

## Task Item
refs #7632

### DIFF
--- a/extensions/azurePublishNew/src/components/provisioningWizards/HandOffToAdminWizard.tsx
+++ b/extensions/azurePublishNew/src/components/provisioningWizards/HandOffToAdminWizard.tsx
@@ -4,26 +4,76 @@
 import * as React from 'react';
 import { usePublishApi } from '@bfc/extension-client';
 import formatMessage from 'format-message';
+import { useRecoilValue } from 'recoil';
+import { ProvisionHandoff } from '@bfc/ui-shared';
 
 import { WizardStep, Wizard } from '../shared/wizard';
+import { useDispatcher } from '../../hooks/useDispatcher';
+import { enabledHandOffResourcesState } from '../../recoilModel/atoms/handOffToAdminState';
+import {
+  subscriptionState,
+  hostNameState,
+  resourceGroupState,
+  deployLocationState,
+} from '../../recoilModel/atoms/resourceConfigurationState';
 
 import { HandOffInstructionsStep } from './steps/HandOffInstructionsStep';
 import { WizardFooter } from './footers/WizardFooter';
+import { ChooseResourcesStep } from './steps/ChooseResourcesStep';
 
 type Props = {
-  onStepChange?: (stepIndex: number, stepId: string) => void;
+  onStepChange?: (stepIndex: number, step: WizardStep) => void;
   stepIndex?: number;
+};
+
+const urls = {
+  createNewResources: 'https://aka.ms/composer-publish-bot#create-new-azure-resources',
 };
 
 export const HandOffToAdminWizard = React.memo((props: Props) => {
   const { onStepChange } = props;
   const [steps, setSteps] = React.useState<WizardStep[]>([]);
-  const { setTitle, onBack, closeDialog: onCancel } = usePublishApi();
+  const { onBack, closeDialog: onCancel } = usePublishApi();
+  const resources = useRecoilValue(enabledHandOffResourcesState);
+  const subscriptionId = useRecoilValue(subscriptionState);
+  const hostName = useRecoilValue(hostNameState);
+  const deployRegion = useRecoilValue(deployLocationState);
+  const { name: resourceGroupName } = useRecoilValue(resourceGroupState);
+  const { setEnabledHandOffResources } = useDispatcher();
+  const [showHandOff, setShowHandOff] = React.useState<boolean>(false);
 
-  const setDialogTitle = (step: WizardStep) => {
-    step && setTitle({ title: step.title, subText: step.subTitle });
-    return <></>;
-  };
+  const handOffInstructions = React.useMemo(() => {
+    let instructions = '';
+    if (showHandOff) {
+      const createLuisResource = resources?.filter((r) => r.key === 'luisPrediction').length > 0;
+      const createLuisAuthoringResource = resources?.filter((r) => r.key === 'luisAuthoring').length > 0;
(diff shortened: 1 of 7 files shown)


extensions/azurePublishNew/src/components/provisioningWizards/HandOffToAdminWizard.tsx
```diff
@@ -4,26 +4,76 @@
...
+      const createAppInsights = resources?.filter((r) => r.key === 'applicationInsights').length > 0;
+      const createQnAResource = resources?.filter((r) => r.key === 'qna').length > 0;
+
+      const provisionComposer = `node provisionComposer.js --subscriptionId ${
+        subscriptionId || '<YOUR SUBSCRIPTION ID>'
+      } --name ${hostName || '<RESOURCE NAME>'} --appPassword=<16 CHAR PASSWORD> --location=${
+        deployRegion || 'westus'
+      } --resourceGroup=${
+        resourceGroupName || '<RESOURCE GROUP NAME>'
+      } --createLuisResource=${createLuisResource} --createLuisAuthoringResource=${createLuisAuthoringResource} --createCosmosDb=${createCosmosDb} --createStorage=${createStorage} --createAppInsights=${createAppInsights} --createQnAResource=${createQnAResource}`;
```
hatpick, line 61: You can encapsulate this long logic in a stand alone hook:

useSomething(resources: type) => command:string

extensions/azurePublishNew/src/components/provisioningWizards/HandOffToAdminWizard.tsx
```diff
@@ -35,16 +85,54 @@ export const HandOffToAdminWizard = React.memo((props: Props) => {
...
+        onStepChange={(index, step) => onStepChange(index, step)}
+      />
+      {showHandOff && (
+        <ProvisionHandoff
+          developerInstructions={formatMessage(
+            'If Azure resources and subscription are managed by others, use the following information to request creation of the resources that you need to build and run your bot.'
+          )}
+          handoffInstructions={handOffInstructions}
+          hidden={!showHandOff}
+          learnMoreLink="https://aka.ms/how-to-complete-provision-handoff"
```
hatpick, line 126: add this to your urls object up, then it would make sense to have it as an object.

extensions/azurePublishNew/src/components/shared/ChooseResourcesList.tsx
```diff
@@ -155,7 +155,7 @@ export const ChooseResourcesList = (props: Props) => {
 
   return (
     <FocusZone direction={FocusZoneDirection.vertical}>
-      <List items={items} onRenderCell={renderItem} />
+      <List items={[...items]} onRenderCell={renderItem} />
```
hatpick, line 158: why are you doing this? it would cause lots of unnecessary re-renders, you already are sending a filtered array to this component which guarantees a new reference every time.

Also, in azureProvisionDialog.tsx L1030 where you use this component, the return value should be null instead of Fragment if you dont want to render anything.

extensions/azurePublishNew/src/recoilModel/atoms/handOffToAdminState.ts
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { atom } from 'recoil';
+
+import { ResourcesItem } from '../../types';
+
+export const enabledHandOffResourcesState = atom<ResourcesItem[] | undefined>({
+  key: 'hor_optional_resources',
```
hatpick, line 9: probably wanna avoid using ```hor```, call it `handoff_whatever`

extensions/azurePublishNew/src/recoilModel/atoms/handOffToAdminState.ts
```diff
@@ -0,0 +1,16 @@
...
+  default: undefined,
+});
+
+export const requiredHandOffResourcesState = atom<ResourcesItem[]>({
+  key: 'hor_required_resources',
```
hatpick, line 14: same here

hatpick (changes requested):

extensions/azurePublishNew/src/components/shared/ChooseResourcesList.tsx
VamsiModem: ChooseResourceList was not re-rendering on selection change even though the newly selected values are coming in. please suggest if there is a better way.

extensions/azurePublishNew/src/components/provisioningWizards/HandOffToAdminWizard.tsx
VamsiModem: good idea!!

extensions/azurePublishNew/src/components/shared/ChooseResourcesList.tsx
GeoffCoxMSFT: Yeah the Fluent controls don't re-render unless the items array changes.  It doesn't care about each item so you have to either do this or slice()

extensions/azurePublishNew/src/recoilModel/atoms/handOffToAdminState.ts
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { atom } from 'recoil';
+
+import { ResourcesItem } from '../../types';
+
+export const enabledHandOffResourcesState = atom<ResourcesItem[] | undefined>({
+  key: 'handoff_optional_resources',
```
GeoffCoxMSFT, line 9: nite: 'handoffOptionalResources' would be the more canonical key string format.

GeoffCoxMSFT (approved):

extensions/azurePublishNew/src/recoilModel/atoms/handOffToAdminState.ts
VamsiModem: I agree, will change it in the next push.

PR #8174, opened 2021-06-22
feat: Implement provision servies for azure resources

### LINKED ISSUE #7629: Revamp azure provisioning server algorithm
The provisioning server algorithm needs to be revamped to separate provisioning of resources by type rather than being omniscient.

natalgar:
## Description

Setting up provision services for each Azure Resource
Checking in the structure before the implementations

## Task Item

closes #7629


extensions/azurePublishNew/src/node/getResources.ts
```diff
@@ -0,0 +1,54 @@
...
+import { botRegistrationDefinition } from './azureResources/botChannel';
+import { luisAuthoringDefinition } from './azureResources/luisAuthoring';
+import { appInsightsDefinition } from './azureResources/appInsights';
+import { cosmosDbDefinition } from './azureResources/cosmosDb';
+import { luisPredictionDefinition } from './azureResources/luisPrediction';
+import { blobStorageDefinition } from './azureResources/blobStorage';
+import { qnaDefinition } from './azureResources/qna';
+import { servicePlanDefinition } from './azureResources/servicePlan';
+
+export const AZURE_HOSTING_GROUP_NAME = 'Azure Hosting';
```
GeoffCoxMSFT, line 24: FYI - I changed this to 'App Services' per request by UX design

extensions/azurePublishNew/src/node/provisionService.ts
```diff
@@ -0,0 +1,92 @@
...
+import { getCosmosDbProvisionService } from './azureResources/cosmosDb';
+import { getLuisAuthoringProvisionService } from './azureResources/luisAuthoring';
+import { getLuisPredictionProvisionService } from './azureResources/luisPrediction';
+import { getBlogStorageProvisionService } from './azureResources/blobStorage';
+import { getQnAProvisionService } from './azureResources/qna';
+import { getAppServiceProvisionService } from './azureResources/servicePlan';
+
+// bot project => candidate resources => select & configure resources => order & provision
+
+export type ProvisionWorkingSet = Record<string, object>;
```
GeoffCoxMSFT, line 19: Consider moving the types that are used types.ts files.   /node/types.ts would have types that are used across all node, node/azureResources/types.ts would have types used only across the individual resources (if there are any of those).  The types are compiled away, so the const methods would be put into this module.  This way when resources only depend on types, they don't take an additional module dependency.

extensions/azurePublishNew/src/node/provisionService.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export type ResourceProvisionService = {
+  getDependencies: () => string[];
+  getRecommendationForProject: (project: IBotProject) => 'required' | 'optional' | 'invalid';
```
GeoffCoxMSFT, line 23: Consider 'none' or 'notAllowed'  - invalid implies something is wrong with user input or state.

extensions/azurePublishNew/src/node/provisionService.ts
```diff
@@ -0,0 +1,92 @@
...
+  required: boolean;
+};
+
+type ProvisionConfig = {
+  key: string;
+};
+
+type ProvisionCredentials = {
+  token: string;
+  graphToken: string;
```
GeoffCoxMSFT, line 45: graphToken should be optional

extensions/azurePublishNew/src/node/provisionService.ts
```diff
@@ -0,0 +1,92 @@
...
+  subscriptionId: string;
```
GeoffCoxMSFT, line 46: I don't think the subscription is part of the credentials.  More likely it would be tenantId.  I would add subscriptionId to the config of the resources that need it, or if everyone needs it add it to the ProvisionConfig.

extensions/azurePublishNew/src/node/provisionService.ts
```diff
@@ -0,0 +1,92 @@
...
+    appInsights: getWebAppProvisionService(),
+    luisAuthoring: getLuisAuthoringProvisionService(),
+    luisPrediction: getLuisPredictionProvisionService(),
+    blobStorage: getBlogStorageProvisionService(),
+    qna: getQnAProvisionService(),
+    servicePlan: getAppServiceProvisionService(),
+  };
+};
+
+export const setUpProvisionService = (creds: ProvisionCredentials) => {
```
GeoffCoxMSFT, line 71: I can see two ways to get the token, graphToken to the services.  

1. You could pass it to each service in getProvisionServices and have the services hold it via closure.  This is good because you can't create services without tokens, you don't have to check if you have a token all the time, and the token doesn't affect every method.  The downside is that tokens expire and you have to re-create the services whenever there is a new token.
2. You could pass the credentials to each provision service method.  This reduces the chance that any service will be holding onto an expired token, but is super tedious to author and call the methods.

extensions/azurePublishNew/src/node/provisionService.ts
```diff
@@ -0,0 +1,92 @@
...
+    const provisionServices = getProvisionServices();
+
+    const workingSet: Record<string, object> = {};
+    selectedResources.forEach((resourceConfig) => {
+      const service = provisionServices[resourceConfig.key];
+      if (service) {
+        service.provision(resourceConfig, workingSet);
+      }
+    });
+  };
```
GeoffCoxMSFT, line 89: This method doesn't return anything so the provision method is defined and then deleted.

extensions/azurePublishNew/src/node/provisionService.ts
```diff
@@ -0,0 +1,92 @@
...
+export const setUpProvisionService = (creds: ProvisionCredentials) => {
+  const token = creds.token;
+  const subscriptionId = creds.subscriptionId;
+  const graphToken = creds.graphToken;
+
+  const provision = (config: ProvisionConfig): void => {
+    // config => sorted resource config
+    const selectedResources: ResourceConfig[] = [];
+
+    const provisionServices = getProvisionServices();
```
GeoffCoxMSFT, line 80: If this returns a Record<> then you'll need to figure out the ordering here.

extensions/azurePublishNew/src/node/getResources.ts
```diff
@@ -0,0 +1,54 @@
...
+  appInsightsDefinition,
+  luisAuthoringDefinition,
+  luisPredictionDefinition,
+  blobStorageDefinition,
+  qnaDefinition,
+  servicePlanDefinition,
+];
+
+export const getResources = (project: IBotProject): GetResourcesResult[] => {
+  const provisionServices: Record<string, ResourceProvisionService> = getProvisionServices();
```
GeoffCoxMSFT, line 43: The plug-in interface is an instance where the getresources and provision methods have a shared lifeltime. Consider that we don't want to create the provision services here and in provision.  I might pass the list of services as a parameter.  You might end up with a provisioning.ts that has the getResources and provision methods and a constants.ts that contains the list of definitions and services.  Keeping those constants together can be good because they should be updated in lock step.

extensions/azurePublishNew/src/node/provisionService.ts
GeoffCoxMSFT: Another reason for this is you'll avoid with a circular dependency with the resources depending on this for types and this depending on the resources for the get methods.

extensions/azurePublishNew/src/node/azureResources/webApp.ts
```diff
@@ -0,0 +1,56 @@
...
+
+import {
+  ProvisionMethod,
+  ProvisionWorkingSet,
+  ResourceConfig,
+  ResourceDefinition,
+  ResourceProvisionService,
+} from '../provisionService';
+import { AZURE_HOSTING_GROUP_NAME } from '../getResources';
+import { parseRuntimeKey } from '../../../../../Composer/packages/lib/shared';
```
GeoffCoxMSFT, line 12: This should be an @bfc/shared reference.

extensions/azurePublishNew/src/node/azureResources/webApp.ts
```diff
@@ -0,0 +1,56 @@
...
+
+type WebAppConfig = ResourceConfig & {
+  key: 'webApp';
+  name: string;
+  location: string;
+  resourceGroupName: string;
+};
+
+export type WebAppResult = {
```
GeoffCoxMSFT, line 21: Nice! I really like how you got type safety around the results and then use the working set Record<>

extensions/azurePublishNew/src/node/azureResources/webApp.ts
```diff
@@ -0,0 +1,56 @@
...
+  key: 'webApp';
+  endpoint: string;
+  hostname: string;
+};
+
+export const webAppResourceDefinition: ResourceDefinition = {
+  key: 'webApp',
+  description:
+    'App Service Web Apps lets you quickly build, deploy, and scale enterprise-grade web, mobile, and API apps running on any platform. Hosting for your bot.',
+  text: AZURE_HOSTING_GROUP_NAME,
```
GeoffCoxMSFT, line 31: I don't think the text and the group name will be the same.  Even if they are the same string, the text shouldn't reference the group name constant so they can vary independently.

extensions/azurePublishNew/src/node/azureResources/webApp.ts
```diff
@@ -0,0 +1,56 @@
...
+  tier: 'S1 Standard',
+  group: AZURE_HOSTING_GROUP_NAME,
+};
+
+const getWebAppProvisionMethod = (): ProvisionMethod => {
```
GeoffCoxMSFT, line 36: Interesting to break this out from the get service method.  I assume because they could be many lines of code?

extensions/azurePublishNew/src/node/azureResources/webApp.ts
```diff
@@ -0,0 +1,56 @@
...
+};
+
+export const getWebAppProvisionService = (): ResourceProvisionService => {
+  return {
+    getDependencies: () => ['appRegistration', 'servicePlan'],
+    getRecommendationForProject: (project) => {
+      const { runtimeType } = parseRuntimeKey(project.settings?.runtime?.key);
+      return runtimeType !== 'functions' ? 'required' : 'invalid';
+    },
+    provision: getWebAppProvisionMethod(),
```
GeoffCoxMSFT, line 54: You don't have to make a method call to return a method.  Define getWebAppProvisionMethod and then do:

`provision: getWebAppProvisionMethod,`

otherwise you are making an extra method call when you don't have to.

extensions/azurePublishNew/src/node/getResources.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  return availableResources.map((availableResource) => {
+    const service = provisionServices[availableResource.key];
+    if (service.getRecommendationForProject(project) !== 'invalid') {
+      return {
+        ...availableResource,
+        required: service.getRecommendationForProject(project) === 'required',
+      };
+    }
```
GeoffCoxMSFT, line 52: map will return undefined in the invalid case, leaving a bunch of undefined in your array.  Consider using reduce, or creating an empty array and doing forEach with pushing onto that result array.

extensions/azurePublishNew/src/node/azureResources/blobStorage.ts
```diff
@@ -0,0 +1,38 @@
...
+  ResourceDefinition,
+  ResourceProvisionService,
+} from '../provisionService';
+
+export const blobStorageDefinition: ResourceDefinition = {
+  key: 'blobStorage',
+  description:
+    'Azure blob storage provides scalable cloud storage, backup and recovery solutions for any data, including bot transcript logs.',
+  text: 'Azure Blob Storage',
+  tier: 'Standard_LRS',
```
GeoffCoxMSFT, line 17: Are the tiers specific IDs or are they display strings?  If they are IDs then the UI can render the right display name and description, if they are names then I guess we should return something readable.  Consider having a set of constants for these to make sure they are consistent (just like the group name)

extensions/azurePublishNew/src/node/azureResources/blobStorage.ts
```diff
@@ -0,0 +1,38 @@
...
+    return {
+      ...workingSet,
+      blobStorage: provisionResult,
+    };
+  };
+};
+
+export const getBlogStorageProvisionService = (): ResourceProvisionService => {
+  return {
+    getDependencies: () => ['appRegistration'],
```
GeoffCoxMSFT, line 34: Does blob store depend on app registration?

extensions/azurePublishNew/src/node/azureResources/luisPrediction.ts
```diff
@@ -0,0 +1,35 @@
...
+
+const getLuisPredictionProvisionMethod = (): ProvisionMethod => {
+  return <TConfig>(config: TConfig, workingSet: ProvisionWorkingSet): ProvisionWorkingSet => {
+    return workingSet;
+  };
+};
+
+export const getLuisPredictionProvisionService = (): ResourceProvisionService => {
+  return {
+    getDependencies: () => ['appRegistration'],
```
GeoffCoxMSFT, line 29: LUIS probably doesn't depend on app registration right?

extensions/azurePublishNew/src/node/azureResources/webApp.ts
natalgar: good point. this is how it was but I've updated it so that `text` is not tied to that constant anymore

extensions/azurePublishNew/src/node/azureResources/blobStorage.ts
natalgar: these are display names. set them up in constants 👍

extensions/azurePublishNew/src/node/getResources.ts
natalgar: using reduce

extensions/azurePublishNew/src/node/azureResources/webApp.ts
natalgar: yep!

extensions/azurePublishNew/src/node/provisionService.ts
natalgar: going with approach 1

extensions/azurePublishNew/src/node/provisionService.ts
natalgar: added to ProvisionConfig

GeoffCoxMSFT (approved):

PR #8197, opened 2021-06-25
fix: integration test to fill out dropdown

pavolum:
## Description

Fix broken integration test in main. 

## Task Item

#minor


pavolum:
The question is whether or not the design calls for a default value for runtime type in the dropdowin in the creation flow when the bot is not imported.



With your change there is no default value and the user has to select function/webapp prior to being allowed to move forward. If that is the desire then we simply need to add those click actions to the creation integration test (that fix is in this PR)



OR does the design call for that having a default value. If so the test wont need the steps in this PR and instead we just need to change the logic a bit to add a default value in non import scenarios

tonyanziano (approved):

Issue #7630, opened 2021-05-04
Revamp azure publishing server algorithm


GeoffCoxMSFT:
The publishing server algorithm needs to be revamped to separate provisioning of resources by type rather than being omniscient. This should also fix the flaky zip upload by encapsulating that into one of the service layers.

PR #8198, opened 2021-06-25
chore: Refactored to pull plugin logic  together

GeoffCoxMSFT:
## Description

Refactored a bit to centralize the plug-in logic and encapsulate resources knowledge to each resource.

## Task Item

#minor

related to #7630

## Screenshots

N/A


extensions/azurePublishNew/src/node/availableResources.ts
```diff
@@ -0,0 +1,72 @@
...
+import { blobStorageDefinition, getBlogStorageProvisionService } from './azureResources/blobStorage';
+import { botRegistrationDefinition, getBotChannelProvisionService } from './azureResources/botChannel';
+import { cosmosDbDefinition, getCosmosDbProvisionService } from './azureResources/cosmosDb';
+import { getLuisAuthoringProvisionService, luisAuthoringDefinition } from './azureResources/luisAuthoring';
+import { getLuisPredictionProvisionService, luisPredictionDefinition } from './azureResources/luisPrediction';
+import { getQnAProvisionService, qnaDefinition } from './azureResources/qna';
+import { getAppServiceProvisionService, servicePlanDefinition } from './azureResources/servicePlan';
+import { getWebAppProvisionService, webAppResourceDefinition } from './azureResources/webApp';
+
+export const availableResources: ResourceDefinition[] = [
```
GeoffCoxMSFT, line 17: I put the available resources and the creation of the services together because if you register one you want to register the other.

extensions/azurePublishNew/src/node/azureResources/appInsights.ts
```diff
@@ -1,10 +1,19 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-import { ProvisionMethod, ProvisionWorkingSet, ResourceProvisionService } from '../types';
+import { ProvisionMethod, ProvisionWorkingSet, ResourceDefinition, ResourceProvisionService } from '../types';
 
+import { PAY_AS_YOU_GO_TIER, AZURE_HOSTING_GROUP_NAME } from './constants';
 import { AppInsightsConfigNew } from './types';
 
+export const appInsightsDefinition: ResourceDefinition = {
```
GeoffCoxMSFT, line 9: I put the individual resource definitions in the same module as their provision method because they should be paired.  Callers will always need to know about both the definition and the service that provisions it.

extensions/azurePublishNew/src/node/azureResources/constants.ts
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation.
```
GeoffCoxMSFT, line 1: I moved these constants into a constants file as they are used across the resources.  There are some grouping improvements we can do later rather than having a set of individual strings.

extensions/azurePublishNew/src/node/backgroundProcessManager.ts
```diff
@@ -0,0 +1,78 @@
+// Copyright (c) Microsoft Corporation.
```
GeoffCoxMSFT, line 1: I added this just for now to satisfy the provision method in the plug-in.  We need to refactor this to be more functional which I can do if/when I take on the telemetry and comm back to the client.

extensions/azurePublishNew/src/node/index.ts
```diff
@@ -1,51 +1,91 @@
...
+import { BackgroundProcessManager } from './backgroundProcessManager';
+import { GetResourcesResult, ProvisionConfig, ResourceDefinition, ResourceProvisionService } from './types';
 
-function initialize(registration: IExtensionRegistration) {
-  const plugin1 = {
+/**
+ * Creates the azure publishing plug-in for this extension.
+ * @returns The plug-in with properties and methods.
+ */
+const createAzurePublishPlugin = (): PublishPlugin<ProvisionConfig> => {
```
GeoffCoxMSFT, line 14: I created a factory function here rather than a class.  The goal is that each method implementation should be short and straightforward if we have architected and implemented the resources and services correctly.

natalgar (approved):

extensions/azurePublishNew/src/node/backgroundProcessManager.ts
```diff
@@ -0,0 +1,78 @@
...
+// Licensed under the MIT License.
+import { ProcessStatus } from '@botframework-composer/types';
+import { v4 as uuid } from 'uuid';
+
+export interface ProcessList {
+  [key: string]: ProcessStatus;
+}
+
+export class BackgroundProcessManager {
```
hatpick, line 10: a class with a bunch of statics sounds like a good use for a function

extensions/azurePublishNew/src/node/backgroundProcessManager.ts
```diff
@@ -0,0 +1,78 @@
...
+        } else {
+          return 0;
+        }
+      })[0];
+    } else {
+      return null;
+    }
+  }
+
+  static updateProcess(id: string, status: number, message: string, config = undefined): string {
```
hatpick, line 65: I'd make the second arg a partial of ProcessStatus (or a subtype of it) instead of a list of args

hatpick: this method should throw if the passed id does not exist, in case an update happens after a delete.

extensions/azurePublishNew/src/node/backgroundProcessManager.ts
```diff
@@ -0,0 +1,78 @@
...
+  static getStatus(id: string): ProcessStatus {
+    return this.processes[id];
+  }
+
+  /* Find a status by projectid/processname without knowing the id
+   * for example, find status of "default" publish target, which is localpublish for which there can only be 1 running process
+   * or find most recent publish to a given publish target
+   */
+  static getStatusByName(projectId: string, processName: string): ProcessStatus {
+    const matching = [];
```
hatpick, line 43: following can be replaced by filter and sort. Time is of type Date and can be sorted like b.time - a.time

extensions/azurePublishNew/src/node/backgroundProcessManager.ts
```diff
@@ -0,0 +1,78 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+import { ProcessStatus } from '@botframework-composer/types';
+import { v4 as uuid } from 'uuid';
+
+export interface ProcessList {
```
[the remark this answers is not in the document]
GeoffCoxMSFT: I ported this over, but will rewrite in next PR.

extensions/azurePublishNew/src/node/azureResources/appInsights.ts
```diff
@@ -1,10 +1,19 @@
...
+  key: 'appInsights',
+  description: 'Application Insights allows you to monitor and analyze usage and performance of your bot.',
```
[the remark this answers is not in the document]
GeoffCoxMSFT: Extensions need some localization love all up.  We use format message, but we don't pull them into en-US.json yet.

extensions/azurePublishNew/src/node/index.ts
```diff
@@ -1,51 +1,91 @@
...
+    const jobId = BackgroundProcessManager.startProcess(202, project.id, config.key, 'Creating Azure resources...');
+    setUpProvisionService(config);
+    return BackgroundProcessManager.getStatus(jobId);
+  };
+
+  const publish = async (_config, _project, metadata /*_user */) => {
+    const response = {
+      status: 202,
+      result: {
+        time: new Date().toString(),
```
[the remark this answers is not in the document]
GeoffCoxMSFT: One of the results defined for the plug in is Date and the other is string :|

extensions/azurePublishNew/src/node/backgroundProcessManager.ts
GeoffCoxMSFT: yep - I ported this over but haven't rewritten yet.

hatpick (approved):

PR #8201, opened 2021-06-28
feat: implement provisioning for app registration, web app, service plan

natalgar:
## Description

implement provisioning for app registration, web app, service plan

## Task Item

refs #7629


extensions/azurePublishNew/src/node/availableResources.ts
```diff
@@ -40,19 +41,17 @@ export const getProvisionServices = (config: ProvisionConfig): Record<string, Re
...
 
 export const setUpProvisionService = (config: ProvisionConfig) => {
   const provisionServices = getProvisionServices(config);
-
-  const provision = (): void => {
-    const selectedResources: ResourceConfig[] = [];
-
-    const provisionServices = getProvisionServices(config);
-
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
GeoffCoxMSFT, line 49: You don't need this eslint comment.

extensions/azurePublishNew/src/node/availableResources.ts
```diff
@@ -32,6 +32,7 @@ export const getProvisionServices = (config: ProvisionConfig): Record<string, Re
   return {
     appRegistration: getAppRegistrationProvisionService(config),
     webApp: getWebAppProvisionService(config),
+    servicePlan: getAppServiceProvisionService(config),
```
GeoffCoxMSFT, line 35: FYI. With an associative array, the order isn't guaranteed to be preserved.

extensions/azurePublishNew/src/node/availableResources.ts
natalgar: removed

extensions/azurePublishNew/src/node/types.ts
```diff
@@ -9,6 +9,7 @@ export type ResourceProvisionService = {
   getDependencies: () => string[];
   getRecommendationForProject: (project: IBotProject) => 'required' | 'optional' | 'notAllowed';
   provision: ProvisionMethod;
+  canPollStatus: boolean;
```
GeoffCoxMSFT, line 12: I don't understand this boolean.  I was thinking that the provision method would take the status update callback (which is what I am going to try and define next). Can you tell me more about this boolean?

extensions/azurePublishNew/src/node/availableResources.ts
natalgar: right! I was just organizing

extensions/azurePublishNew/src/node/types.ts
natalgar: Some of the SDK create/update methods return a special `LROPoller` object with methods for interacting with long running Azure operations (response 202). I thought it would be cool to use that. 
But some (such as App Reg, Bot Channel Registration, and App Insights) won't have that - those will just take the onUploadProgress/onDownloadProgress callbacks.

GeoffCoxMSFT (approved):

Issue #8066, opened 2021-06-11
Update provisioning and deployment algorithms with additional telemetry


GeoffCoxMSFT:
Once the service layer is in place, we will have the atomic blocks to call from the provisioning and deployment algorithms.
We should add telemetry around each action the algorithms take including duration.

GeoffCoxMSFT:
@natalgar - I'll take this to see what I can do in the existing extension for R14.  I think we'll have better telemetry in the new extension as you factor things.

GeoffCoxMSFT:
The extension on the node side only has the logger, not the telemetry client.  We'll need to add the telemetry client to be usable on the node side of extensions before we can add server-side telemetry to provisioning and publishing.  This was a P0 for R13, but we fixed the issue that was causing insufficient error information to be returned, so now this is a lower priority need.

PR #8203, opened 2021-06-28
feat: Add process status tracker

GeoffCoxMSFT:
## Description

This replaces the backgroundProcessManager class with a functional processStatusTracker component.
Wrote new unit tests for the tracker.

## Task Item

#minor
related to #8066

## Screenshots

N/A


natalgar (approved):
nice!

extensions/azurePublishNew/src/node/processStatusTracker.ts
```diff
@@ -0,0 +1,92 @@
...
+    return statusById[id];
+  };
+
+  /**
+   * Gets the first, most recent matching process name.
+   */
+  const getByProcessName = (processName: string) => {
+    const matches = Object.values(statusById)
+      .filter((status) => status.processName === processName)
+      .sort((a, b) => b.time.valueOf() - a.time.valueOf());
```
hatpick, line 49: getTime instead of valueOf, valueOf belongs to Object, getTime belongs to Date.

extensions/azurePublishNew/src/node/processStatusTracker.ts
```diff
@@ -0,0 +1,92 @@
...
+    return matches?.[0];
+  };
+
+  /**
+   * Gets the first, most recent matching process name.
+   */
+  const getByProjectId = (projectId: string) => {
```
hatpick, line 56: name doesn't reflect the purpose

extensions/azurePublishNew/src/node/processStatusTracker.ts
```diff
@@ -0,0 +1,92 @@
...
+   * Updates the status of the process with the specified ID.
+   */
+  const update = (id: string, update: ProcessStatusUpdate) => {
+    if (statusById[id]) {
+      statusById[id].status = update.status;
+      statusById[id].message = update.message;
+      statusById[id].log.push(update.message);
+      statusById[id].config = update.config || statusById[id].config;
+    } else {
+      console.warn(`processTracker.update called for process ${id} that does not exist.`);
```
hatpick, line 73: I think this should throw

extensions/azurePublishNew/src/node/processStatusTracker.ts
```diff
@@ -0,0 +1,92 @@
...
+    const matches = Object.values(statusById)
+      .filter((status) => status.projectId === projectId)
+      .sort((a, b) => b.time.valueOf() - a.time.valueOf());
+    return matches?.[0];
+  };
+
+  /**
+   * Updates the status of the process with the specified ID.
+   */
+  const update = (id: string, update: ProcessStatusUpdate) => {
```
hatpick, line 66: update as arg name is the same as the function name, maybe use a different name?

hatpick: isn't the update obj partial? should it check if it has any value before it overrides existing values with potential undefined?

extensions/azurePublishNew/src/node/processStatusTracker.ts
```diff
@@ -0,0 +1,92 @@
...
+    if (statusById[id]) {
+      statusById[id].status = update.status;
+      statusById[id].message = update.message;
+      statusById[id].log.push(update.message);
```
hatpick, line 70: e.g., it may push undefined to logs

extensions/azurePublishNew/src/node/processStatusTracker.ts
```diff
@@ -0,0 +1,92 @@
...
+    } else {
+      console.warn(`processTracker.update called for process ${id} that does not exist.`);
+    }
+  };
+
+  /**
+   * Stops tracking the status of the process with the specified ID.
+   */
+  const stop = (id: string) => {
+    delete statusById[id];
```
hatpick, line 81: I think it should throw here if it doesn't exist

extensions/azurePublishNew/src/node/processStatusTracker.ts
```diff
@@ -0,0 +1,92 @@
...
+  processName: string;
+};
+
+type ProcessStatusUpdate = Pick<Partial<ProcessStatus>, 'status' | 'message' | 'config'>;
+
+/**
+ * Creates a tracker of ProcessStatus.
+ * The tracker avoids throwing errors to prevent status tracking from halting the process itself.
+ */
+export const createProcessStatusTracker = () => {
```
hatpick, line 17: it seems like most of the time there is only one process in this tracker. Should it be exporting one single tracker?
```export const tracker = createProcessStatusTracker();```

File: extensions/azurePublishNew/src/node/index.ts [hunk 8]
```diff
@@ -29,21 +38,64 @@ const createAzurePublishPlugin = (): PublishPlugin<ProvisionConfig> => {
...
+        id,
+        log: log.map((item) => `---\n${JSON.stringify(item, null, 2)}\n---\n`).join('\n'),
+        message,
+        status,
+        time: time.toString(),
       },
     };
   };
 
+  const getStatus = async (config: PublishConfig, project: IBotProject) => {
```
hatpick, line 57: what's the difference between getStatus and getProvisionStatus?

GeoffCoxMSFT: getStatus is for the publishing - different return types.

extensions/azurePublishNew/src/node/processStatusTracker.ts
GeoffCoxMSFT: I don't want to restrict it to a single tracker within the module itself.

hatpick (approved):

PR #8209, opened 2021-06-29
fix: Persist Orchestrator Model settings in appsettings

### LINKED ISSUE #7946: Orchestrator:  Persist model settings data when Orchestrator is used
## Describe the bug
We need to persist the Orchestrator model data in Composer.

Right now, we always use the recommended(default) Orchestrator language models, but the defaults and behavior can change - we don't want to have users program to a model, and suddenly change the model when the default recommended model gets updated.

taicchoumsft:
## Description
Persist the model name in `appsettings.json`, which would look like this:
```
  "orchestrator": {
    "model": {
      "en_intent": "pretrained.20210521.microsoft.dte.01.06.int.en.onnx",
      "multilingual_intent": "pretrained.20210608.microsoft.dte.01.06.int.unicoder_multilingual.onnx"
    }
  },
```
These settings are only written when orchestrator is used for the first time.  The default model names  will be picked by checking against the Orchestrator models API.  The user can later alter these models by changing the model name in appsettings and the bot will use the new model.  If an invalid model is chosen, the bot will fail to start with a warning that the model is invalid. 

We only persist the model names in appsettings.  The `generated/orchestrator.settings.json` and orchestrator recognizer dialog file logic are the same.

## Task Item
closes #7946


Composer/packages/types/src/orchestrator.ts
```diff
@@ -1,8 +1,10 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+export type ModelTypes = 'en_intent' | 'multilingual_intent';
+
 export type OrchestratorModelRequest = {
-  kind: 'en_intent' | 'multilingual_intent';
+  kind: ModelTypes;
```
srinaath, line 7: Singular `ModelType`

srinaath (changes requested):

Composer/packages/types/src/orchestrator.ts
```diff
@@ -11,3 +13,17 @@ export enum DownloadState {
   ALREADYDOWNLOADED = 'ALREADYDOWNLOADED',
   DOWNLOADING = 'DOWNLOADING',
 }
+
+export interface IOrchestratorNLRList {
```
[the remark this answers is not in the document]
taicchoumsft: thanks, fixed!

Composer/packages/types/src/orchestrator.ts
taicchoumsft: fixed!

Composer/packages/client/src/recoilModel/dispatchers/orchestrator.ts
```diff
@@ -49,11 +56,15 @@ export const downloadModel = async (
   return await pollUntilDone(async () => (await httpClient.get(resp.data)).data !== DownloadState.DOWNLOADING, 5000);
 };
 
-export const availableLanguageModels = (recognizerFiles: RecognizerFile[], botSettings?: DialogSetting) => {
+export const availableLanguageModels = async (
```
hatpick, line 59: can you rename this to ```getAvailableLanguageModels```

taicchoumsft: thanks @hatpick , renamed.

hatpick (approved):

PR #8210, opened 2021-06-29
fix: race condition on finding port while starting all bots

### LINKED ISSUE #7765: Starting multiple bots causes port collisions
## Describe the bug
When I `Start all bots` with  1 parent and 4 skills 2 skills I have multiple skills running on the same port

## Version
Version:  2.0.0-nightly.244572.8406071
Electron: 8.2.4
Chrome: 80.0.3987.165
NodeJS: 12.13.0
V8: 8.0.426.27-electron.0

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

1. Create a parent bot
2. Add 4 skill bots
3. Start all bots

## Expected behavior
Each skill should start on its own port

## Screenshots
image

liweitian:
## Description
should set running bots' port to avoid the conflict during finding a port for each request.
## Task Item
Closes #7765

## Screenshots
test

### DIFF
--- a/extensions/localPublish/src/index.ts
+++ b/extensions/localPublish/src/index.ts
@@ -114,6 +114,16 @@ class LocalPublisher implements PublishPlugin<PublishConfig> {
     );
   };
 
+  private isPortUsed = (port: number) => {
+    for (const key in LocalPublisher.runningBots) {
+      const bot = LocalPublisher.runningBots[key];
+      if (bot?.port === port) {
+        return true;
+      }
+    }
+    return false;
+  };
+
   private publishAsync = async (botId: string, version: string, fullSettings: DialogSetting, project: any, user) => {
     try {
       let port;
@@ -126,7 +136,10 @@ class LocalPublisher implements PublishPlugin<PublishConfig> {
       if (!port) {
         // Portfinder is the stablest amongst npm libraries for finding ports. https://github.com/http-party/node-portfinder/issues/61. It does not support supplying an array of ports to pick from as we can have a race conidtion when starting multiple bots at the same time. As a result, getting the max port number out of the range and starting the range from the max.
         const maxPort = max(map(LocalPublisher.runningBots, 'port')) ?? 3979;
-        port = await portfinder.getPortPromise({ port: maxPort + 1, stopPort: 6000 });
+        do {
+          port = await portfinder.getPortPromise({ port: maxPort + 1, stopPort: 6000 });
+        } while (this.isPortUsed(port));
+
         const updatedBotData: RunningBot = {
           ...LocalPublisher.runningBots[botId],
           port,
--- a/extensions/localPublish/src/index.ts
+++ b/extensions/localPublish/src/index.ts
@@ -127,6 +127,11 @@ class LocalPublisher implements PublishPlugin<PublishConfig> {
         // Portfinder is the stablest amongst npm libraries for finding ports. https://github.com/http-party/node-portfinder/issues/61. It does not support supplying an array of ports to pick from as we can have a race conidtion when starting multiple bots at the same time. As a result, getting the max port number out of the range and starting the range from the max.
         const maxPort = max(map(LocalPublisher.runningBots, 'port')) ?? 3979;
         port = await portfinder.getPortPromise({ port: maxPort + 1, stopPort: 6000 });
+        const updatedBotData: RunningBot = {
+          ...LocalPublisher.runningBots[botId],
+          port,
+        };
+        LocalPublisher.runningBots[botId] = updatedBotData;
       }
 
       const runtime = this.composer.getRuntimeByProject(project);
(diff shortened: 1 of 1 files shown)


extensions/localPublish/src/index.ts
```diff
@@ -56,7 +54,7 @@ class LocalPublisher implements PublishPlugin<PublishConfig> {
   public description = 'Publish bot to local runtime';
   static runningBots: { [key: string]: RunningBot } = {};
   private composer: IExtensionRegistration;
-
+  private lastUsedPort = 3979;
```
boydc2014, line 57: Basically we are using an atomic variable to sync all the threads there, seems like a less breaking change, than making changes in how those tasks are scheduled. 

Looks good to me, what do you think @srinaath

boydc2014 (approved):

extensions/localPublish/src/index.ts
```diff
@@ -124,9 +122,19 @@ class LocalPublisher implements PublishPlugin<PublishConfig> {
         await this.stopBot(botId);
       }
       if (!port) {
-        // Portfinder is the stablest amongst npm libraries for finding ports. https://github.com/http-party/node-portfinder/issues/61. It does not support supplying an array of ports to pick from as we can have a race conidtion when starting multiple bots at the same time. As a result, getting the max port number out of the range and starting the range from the max.
-        const maxPort = max(map(LocalPublisher.runningBots, 'port')) ?? 3979;
```
srinaath, line 128: @boydc2014 How different is this solution from max(map(LocalPublisher.runningBots, 'port')). Essentially, instead of lastUsedPort it was using the LocalPublisher.runningBots to get all ports of bots that are running/starting. In this way it was avoiding the conflicts.

srinaath: @liweitian The core issue is lets say we have 3 bots [bot1, bot2, bot3]. 

1. Bot1 is started.  it runs await runtime.build(runtimePath, project, fullSettings, port); before setting the port of the bot in the LocalPublisher object (this.setBot does this). Can you update the LocalPublisher.runningBots[projectId].port immediately after portfinder finds the port rather than assigning it after the runtime.build. In this way, the next bot that starts ignores that port as max(map(LocalPublisher.runningBots, 'port')) + 1 would be the new port.

boydc2014: The difference is max(max...) runs before the port been find. For example, two co-routines both get the maxPort = 3980, and both looking for that, and both get a "pass" and both trying to start that port, and ops conflict happened. 

While the approach here is to let each coroutine find port first (which is async and can return same result), ok, after finding the target port, check again, if this port is already been used. Because the post part of the port finding is sync by the nature of node js, so only co-routine will declare it gets the port, and the other co-routine will have to find port again. That's how the conflict can be avoided.

liweitian: I made some changes on the code. After testing it again, the find port operation does not causes a race condition. It is probably not an multi-threaded IO operation but the runtime.build or some code after that causes the race condition. If we set the running bots' port right after the find port method. The conflict can be solved. And only setting the port without changing status won't cause any error on the FE. So it is better to use it to replace the lastUsedPort.

boydc2014 (approved):

extensions/localPublish/src/index.ts
boydc2014: I'm guessing, maybe actually the async operation is always correct, but it's just mean the async part in node is done, probably doesn't count the os part (which is the port eventually get used).

boydc2014: image

I think it's the upper layer that is not waiting, so our solution is probably still have issues

srinaath: @boydc2014 Good catch. What if we just wrap the code to find a port which is currently inside publishAsync line 119 to 132. We do not ever want to await on runtime.build. But we can await until we find the port. In this way, Localpublisher.runningBots[projectId].port is set before the next bot is called

```javascript
publish = () => {
const port = await getPort()
 this.publishAsync(port)
}

const getPort = async () => // await portfinder.find() returns the port
```
Thoughts? @boydc2014 @liweitian

boydc2014: That sounds good.

srinaath (changes requested):
@liweitian Could you take a stab at implementing the propsed solution?

boydc2014:
> @liweitian Could you take a stab at implementing the propsed solution?

After further investigation, this is so far the best solution given the FE is actually starting bots "concurrently". More details are sent to you via Teams.

boydc2014 (approved):

srinaath:
Tried with a root bot and 4 skills a few times and it works. We can use this as a fix for R14. @liweitian Lets use this ticket in R15 to handle the elegant solution of fetching ports upfront and assigning them to the bots before starting.

PR #8218, opened 2021-06-29
feat: provision bot channel

natalgar:
## Description

bot channel provisioning

## Task Item

refs #7629


extensions/azurePublishNew/src/node/azureResources/botChannel.ts
```diff
@@ -15,29 +36,49 @@ export const botRegistrationDefinition: ResourceDefinition = {
...
+    const endpoint = `https://${hostname ?? config.hostname + '.azurewebsites.net'}/api/messages`;
 
-  return {
-    ...workingSet,
-    botChannelResult: provisionResult,
+    try {
+      await azureBotService.bots.create(config.resourceGroupName, config.displayName, {
+        properties: {
+          displayName: config.displayName,
+          endpoint: endpoint ?? '',
```
GeoffCoxMSFT, line 52: There isn't any way for the endpoint to be undefined.  Consider checking above when constructing it.

extensions/azurePublishNew/src/node/azureResources/botChannel.ts
```diff
@@ -15,29 +36,49 @@ export const botRegistrationDefinition: ResourceDefinition = {
...
-  const { appId } = appRegistrationResult.appId;
-  const { endpoint } = webAppResult.endpoint;
-  const { hostname } = webAppResult.hostname;
+const botChannelProvisionMethod = (provisionConfig: ProvisionConfig): ProvisionMethod => {
+  const tokenCredentials = new TokenCredentials(provisionConfig.accessToken);
+  const azureBotService = new AzureBotService(tokenCredentials, provisionConfig.subscriptionId);
 
-  const provisionResult: BotChannelResult = { botName: '' };
+  return async (config: BotServiceConfig, workingSet: ProvisionWorkingSet): Promise<ProvisionWorkingSet> => {
+    const appId = workingSet.appRegistration?.appId;
```
GeoffCoxMSFT, line 44: Do you want to throw if the working set doesn't contain what this service needs?

GeoffCoxMSFT:
Looks like you have an empty provisionService.ts file that shouldn't be added.

extensions/azurePublishNew/src/node/azureResources/botChannel.ts
natalgar: set up a new function for checking required fields

hatpick:
I think validating the args for service calls is the job of the callers (handlers) not the service layer. Thoughts? @GeoffCoxMSFT @natalgar

natalgar:
> I think validating the args for service calls is the responsibility of the callers (handlers) not the service layer. Thoughts? @GeoffCoxMSFT @natalgar

From OOP I tend to think of the "tell, don't ask" principle. So here I did see it as a responsibility of the service layer to validate its input.

hatpick:
> > I think validating the args for service calls is the responsibility of the callers (handlers) not the service layer. Thoughts? @GeoffCoxMSFT @natalgar
> 
> From OOP I tend to think of the "tell, don't ask" principle. So here I did see it as a responsibility of the service layer to validate its input.

Look at it from SRP and open/closed point of view. In order to extend the validation you have to modify the service layer code. Also it's doing both service calls and validations.

GeoffCoxMSFT:
I think if the Azure SDK is going to check the input and throw then we can rely on that and don't have to check the field and throw our own custom error.  If it is a field that we use or where we make a call and it will succeed but not do what we want we should check it.

GeoffCoxMSFT (dismissed):

GeoffCoxMSFT (approved):

natalgar:
replacing this PR with #8321

PR #8227, opened 2021-06-30
fix: Quotes in node spawn arguments make the runtime fail

### LINKED ISSUE #8205: Cannot debug a bot with MsAppId and Pass
## Describe the bug
Create a bot. Add an msappid and msapppassword. Start the bot. Try running it in Emulator or Integrated Webchat in Composer. It throws a 401 error.

## Version
Current Nightly on Composer. Composer 2.0 works totally fine.

## Browser
Both electron and web

lei9444:
## Description
**root cause**
The system shell will strip off quotes around parameters, so the program gets the unquoted value at the end. Spawning a process bypasses this step as it bypasses the shell, so the program gets those literal quotes as part of the parameter and doesn't know how to handle them appropriately.

**fix**
use { shell: true }. This will pass the spawn request through the shell

## Task Item
fixes #8205

### DIFF
--- a/extensions/localPublish/src/index.ts
+++ b/extensions/localPublish/src/index.ts
@@ -281,7 +281,7 @@ class LocalPublisher implements PublishPlugin<PublishConfig> {
           cwd: botDir,
           stdio: ['ignore', 'pipe', 'pipe'],
           detached: !isWin, // detach in non-windows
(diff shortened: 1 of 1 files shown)


boydc2014 (approved):

srinaath (approved):

srinaath:
Great explanation @lei9444.

srinaath:
Solid debugging @lei9444!

PR #8241, opened 2021-07-01
chore: orchestrator-core changed to @microsoft/orchestrator-core

taicchoumsft:
## Description
`orchestrator-core` changed to `@microsoft/orchestrator-core` for 4.14, the electron-builder config is updated to reflect this so that the mac/linux electron builds can start. Windows build does not have problems even without this fix because for some reason electron-builder's binary detection only works correctly in Windows for this lib.  

manual nightly build with this change here for verification: <https://fuselabs.visualstudio.com/Composer/_build/results?buildId=255945&view=results>

#minor

### DIFF
--- a/Composer/packages/electron-server/electron-builder-config.json
+++ b/Composer/packages/electron-server/electron-builder-config.json
@@ -15,15 +15,10 @@
     "build/templates",
     "resources/composerIcon_1024x1024.png",
     "resources/ms_logo.svg",
-    "**/node_modules/orchestrator-core/**/*",
+    "**/node_modules/@microsoft/orchestrator-core/**/*",
     "**/node_modules/@bfc/server-workers/**/*"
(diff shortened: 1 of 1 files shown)


Composer/packages/electron-server/electron-builder-config.json
```diff
@@ -15,15 +15,10 @@
     "build/templates",
     "resources/composerIcon_1024x1024.png",
     "resources/ms_logo.svg",
-    "**/node_modules/orchestrator-core/**/*",
+    "**/node_modules/@microsoft/orchestrator-core/**/*",
```
taicchoumsft, line 18: this is the only change - the rest of changes appear to be whitespace changes by linter upon checkin.

boydc2014 (approved):

PR #8261, opened 2021-07-05
fix: Ensure manifest.json not found in zip displays error

### LINKED ISSUE #8260: PVA as a Skill .zip manifest parsing issues
## Describe the bug

If a .json file is not found in the manifest, an exception should be displayed accordingly.

Checking the schema version should check if the allowed versions are in the manifest string, no need to check the entire string for a match.

EricDahlvang:
## Description

1) When a .zip is parsed, but the .json manifest is not found, display the exception
2) When checking schema compat, check just the version and not the entire schema path

## Task Item

fixes #8260
closes https://github.com/microsoft/BotFramework-Composer/issues/8260

### DIFF
--- a/Composer/packages/lib/shared/src/skillsUtils/index.ts
+++ b/Composer/packages/lib/shared/src/skillsUtils/index.ts
@@ -149,11 +149,13 @@ export const parseRuntimeKey = (
 
 export const isManifestJson = (content) => {
   try {
-    const urls = [
-      'https://schemas.botframework.com/schemas/skills/v2.1/skill-manifest.json',
-      'https://schemas.botframework.com/schemas/skills/v2.2/skill-manifest.json',
+    const versions = [
+      '/v2.0/',
+      '/v2.1/',
+      '/v2.2/',
     ];
-    return urls.includes(JSON.parse(content).$schema);
+    const schema = JSON.parse(content).$schema;
+    return versions.some(v => schema.includes(v));
   } catch (e) {
     return false;
   }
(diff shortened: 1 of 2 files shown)


luhan2017 (approved):

Composer/packages/lib/shared/src/skillsUtils/index.ts
```diff
@@ -149,11 +149,9 @@ export const parseRuntimeKey = (
 
 export const isManifestJson = (content) => {
   try {
-    const urls = [
-      'https://schemas.botframework.com/schemas/skills/v2.1/skill-manifest.json',
-      'https://schemas.botframework.com/schemas/skills/v2.2/skill-manifest.json',
-    ];
-    return urls.includes(JSON.parse(content).$schema);
+    const versions = ['/v2.0/', '/v2.1/', '/v2.2/'];
```
beyackle, line 152: Is there any reason we should worry about a `$schema` field matching these shorter version strings but being something other than a longer URL (like the ones that used to be here)?

EricDahlvang: While the entire path is more correct, this seems less restrictive. I had a mistake in a manifest schema path, while the rest of the manifest was correct. I do not feel strongly about this though. If you object, we can skip this change.

Maybe we should be checking the $version, instead of the #schema path.  Considering, our previous .schema had no version in the $id path: https://schemas.botframework.com/schemas/skills/v2.0/skill-manifest.json 

```
"$id": "https://schemas.botframework.com/schemas/skills/skill-manifest.json",
```

EricDahlvang: @beyackle adjusted to endsWith '{v}/skill-manifest.json' and remove v2.0, considering the above comment

hatpick (approved):

PR #8263, opened 2021-07-06
fix: Add suggestion for imported entities in Lu editing

### LINKED ISSUE #8215: LU intellisense does not support imported definitions
## Describe the bug

When you have an LU file that includes other LU files, the intellisense in the code view of LU does not work.  This shows up in form generation, but will also be an issue for any both that factors into multiple LU files.

## Version

Version:  2.0.0-nightly.254897.cdf76b6
Electron: 8.2.4
Chrome: 80.0.3987.165
NodeJS: 12.13.0
V8: 8.0.426.27-electron.0

## To Reproduce

1. Create an empty bot 'testbot'
2. Open VS code and define an LU file language-understanding/en-us/definitions.en-us.lu with the content `@ ml foo`.
3. Open language-understanding/en-us/testbot.en-us.lu and add 
```
[definitions](definitions.en-us.lu)

# someIntent
- {@
```
4. If you open the testbot.en-us.lu file and do a code view you will not get intellisense after the @ or the ability to insert an entity.

## Expected behavior

Intellisense should work on the LU file even if you use imports.

cosmicshuai:
closes #8215 
In this PR, Lu editor would be capable to suggest entities defined in a imported lu file.
For example:
There is 3 entities defined in common.en-us.lu
```
@ ml myname
@ ml yourname
@ ml testExpr
```

In current lu file, it could suggest both imported entities and entities defined in the current file:
image

### DIFF
--- a/Composer/packages/tools/language-servers/language-understanding/src/LUServer.ts
+++ b/Composer/packages/tools/language-servers/language-understanding/src/LUServer.ts
@@ -31,6 +31,8 @@ import { LUOption, LUDocument, generateDiagnostic, convertDiagnostics, createFol
 // define init methods call from client
 const LABELEXPERIENCEREQUEST = 'labelingExperienceRequest';
 const InitializeDocumentsMethodName = 'initializeDocuments';
+const SIMPLEINTENTSECTION = 'simpleIntentSection';
+const NEWENTITYSECTION = 'newEntitySection';
 
 export class LUServer {
   protected workspaceRoot: URI | undefined;
@@ -40,7 +42,7 @@ export class LUServer {
   private luParser = new LuParser();
   private _curFileId = '';
   private _curProjectId = '';
-  private _importedEntitties: string[] = [];
+  private _importedEntities: string[] = [];
 
   constructor(
     protected readonly connection: IConnection,
@@ -231,7 +233,7 @@ export class LUServer {
       this._curFileId = id;
       this._curProjectId = projectId || '';
       const { intents: sections, diagnostics, imports } = await this.luParser.parse(content, id, luFeatures);
-      this._importedEntitties = await this.findAllImportedEntities(imports, importResolver, luFeatures);
+      this._importedEntities = await this.findAllImportedEntities(imports, importResolver, luFeatures);
       return { sections, diagnostics, content };
     };
     const luDocument: LUDocument = {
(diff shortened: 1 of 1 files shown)


Composer/packages/tools/language-servers/language-understanding/src/LUServer.ts
```diff
@@ -243,6 +245,38 @@ export class LUServer {
...
+          for (const entity of section.Entities) {
+            if (entity.SectionType === 'newEntitySection') {
+              result.push(entity.Name);
+            }
+          }
+        }
+      }
+    }
+
+    return Promise.resolve(uniq(result));
```
taicchoumsft, line 277: >Promise.resolve [](http://example.com/codeflow?start=11&length=15)

minor question: is `Promise.resolve` needed here?

Composer/packages/tools/language-servers/language-understanding/src/LUServer.ts
```diff
@@ -39,6 +40,7 @@ export class LUServer {
   private luParser = new LuParser();
   private _curFileId = '';
   private _curProjectId = '';
+  private _importedEntitties: string[] = [];
```
taicchoumsft, line 43: >_importedEntitties [](http://example.com/codeflow?start=10&length=18)

minor spelling error: `_importedEntities`

Composer/packages/tools/language-servers/language-understanding/src/LUServer.ts
```diff
@@ -243,6 +245,38 @@ export class LUServer {
...
+      try {
+        content = (await importResolver('.', importFile.id)).content;
+      } catch (error) {
+        // ignore if file not exist
+      }
+
+      const { resource } = await this.luParser.parse(content, importFile.id, luFeatures);
+      const sections = resource.Sections;
+      for (const section of sections) {
+        if (section.SectionType === 'newEntitySection') {
```
taicchoumsft, line 265: >newEntitySection [](http://example.com/codeflow?start=37&length=16)

minor: looks like these strings are also defined here, maybe we can use them? 

<[https://github.com/microsoft/botframework-cli/blob/7e5fa93174a86f61467d6fc3c867d0cf36d8d8c3/packages/lu/src/parser/utils/enums/lusectiontypes.js](https://github.com/microsoft/botframework-cli/blob/7e5fa93174a86f61467d6fc3c867d0cf36d8d8c3/packages/lu/src/parser/utils/enums/lusectiontypes.js)
>

Composer/packages/tools/language-servers/language-understanding/src/LUServer.ts
```diff
@@ -243,6 +245,38 @@ export class LUServer {
...
+    let content = '';
+    const result: string[] = [];
+    for (const importFile of imports) {
+      try {
+        content = (await importResolver('.', importFile.id)).content;
+      } catch (error) {
+        // ignore if file not exist
+      }
+
+      const { resource } = await this.luParser.parse(content, importFile.id, luFeatures);
```
taicchoumsft, line 262: > await this.luParser.parse [](http://example.com/codeflow?start=26&length=26)

maybe it is safer to put this inside the try block as well?  According to [here](https://github.com/microsoft/botframework-cli/blob/7e5fa93174a86f61467d6fc3c867d0cf36d8d8c3/packages/lu/src/parser/lufile/luParser.js#L193)
This will be `undefined` if content is ''.  Then the destructuring will probably throw an error

Composer/packages/tools/language-servers/language-understanding/src/LUServer.ts
cosmicshuai: removed

Composer/packages/tools/language-servers/language-understanding/src/LUServer.ts
cosmicshuai: fixed

Composer/packages/tools/language-servers/language-understanding/src/LUServer.ts
cosmicshuai: It is exported outside bf-lu package. I defined some constants for these string values in LuServer.

Composer/packages/tools/language-servers/language-understanding/src/LUServer.ts
cosmicshuai: added try catch block

hatpick (approved):

PR #8265, opened 2021-07-06
feat: app registration provisioning

natalgar:
## Description

app registration provisioning

## Task Item

refs #7629


extensions/azurePublishNew/src/node/availableResources.ts
```diff
@@ -19,6 +19,13 @@ import { getLuisPredictionProvisionService, luisPredictionDefinition } from './a
 import { getQnAProvisionService, qnaDefinition } from './azureResources/qna';
 import { getAppServiceProvisionService, servicePlanDefinition } from './azureResources/servicePlan';
 import { getWebAppProvisionService, webAppResourceDefinition } from './azureResources/webApp';
+import { ProvisionConfig2 } from './provisioning';
```
GeoffCoxMSFT, line 22: ProvisionConfig2 is the required input to the provisioning system from the UI.  The current ProvisionConfig should probably be renamed (ProvisionServiceConfig?) and ProvisionConfig2 should be ProvisionConfig.

extensions/azurePublishNew/src/node/constants.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+export const AzureResourceTypes = {
+  APP_REGISTRATION: 'appRegistration',
```
GeoffCoxMSFT, line 5: IMHO, I don't know why we keep using UPPER_CASE like we're a C++ team writing header files 😜

GeoffCoxMSFT (approved):

extensions/azurePublishNew/src/node/constants.ts
hatpick: I agree with Geoff + I think this should be a union type.

GeoffCoxMSFT (approved):

PR #8269, opened 2021-07-06
feat: Provision app insights

natalgar:
## Description

Provision app insights

## Task Item

refs #7629


extensions/azurePublishNew/src/node/availableResources.ts
```diff
@@ -27,6 +27,15 @@ export type AppRegistrationResourceConfig = ResourceConfig & {
   appName: string;
 };
 
+type ApplicationType = 'web' | 'other';
+
+export type AppInsightsResourceConfig = ResourceConfig & {
```
GeoffCoxMSFT, line 32: As per previous reviews, the config types should be in the modules that own them.  That way availableResource.ts has dependencies on individual resource files, but the individual files don't also depend on availableResource.ts (i.e. a circular dependency).

extensions/azurePublishNew/src/node/azureResources/appInsights.ts
```diff
@@ -1,34 +1,138 @@
...
+import { ApplicationInsightsManagementClient } from '@azure/arm-appinsights';
+import { AzureBotService } from '@azure/arm-botservice';
 
-import { PAY_AS_YOU_GO_TIER, AZURE_HOSTING_GROUP_NAME } from './constants';
-import { AppInsightsConfigNew } from './types';
+import { ProvisionServiceConfig, ProvisionWorkingSet, ResourceDefinition, ResourceProvisionService } from '../types';
+import { AzureResourceTypes } from '../constants';
+import { AppInsightsResourceConfig } from '../availableResources';
+import {
+  createCustomizeError,
```
GeoffCoxMSFT, line 12: This method bothers me - I don't understand why we need to manipulate and stringify errors with "owned props".

GeoffCoxMSFT (approved):

extensions/azurePublishNew/src/node/availableResources.ts
natalgar: fixed

GeoffCoxMSFT (approved):

PR #8273, opened 2021-07-06
feature: add events for opt in/out of telemetry

### LINKED ISSUE #8245: Add telemetry event for user opting in or out of telemetry.
## Describe the feature

Log a new telemetry event when someone opts in or out of telemetry when presented within the initial modal dialog. 

Event should be called **TelemetryOptInOut** and add a custom property should be added to the event called **enabled** -  which is set to true if they opt in and false if they opt out. 

If the user opts in then we can send any cached events (handoff / session start), along with this new event, and all can contain all of the global properties we currently send.

If the user opts out, we should only send this single event (drop the cached ones) and the event should not contain the Id for the user.

beyackle:
## Description

This adds a `TelemetryOptInOut` event. When this event is triggered to disable telemetry, it gets emitted as a bare event with no context (by resetting the client's telemetry settings), and when it's triggered to enable, it will re-initialize the telemetry client's settings.

## Task Item

closes #8245


Composer/packages/client/src/telemetry/AppInsightsClient.ts
```diff
@@ -29,7 +30,7 @@ export default class AppInsightsClient {
   }
 
   public static trackEvent(name: string, properties: LogData) {
-    if (this._telemetrySettings?.allowDataCollection) {
+    if (alwaysTrackEvents.includes(name as TelemetryEventName) || this._telemetrySettings?.allowDataCollection) {
```
tdurnford, line 33: I think this should be a separate condition that just logs the event straight to the server. There might be other events in the queue that we want to log.

Composer/packages/server/src/services/telemetry.ts
```diff
@@ -43,10 +57,16 @@ if (instrumentationKey) {
   AppInsights.defaultClient.addTelemetryProcessor((envelope: AppInsights.Contracts.Envelope): boolean => {
     const { sessionId, telemetry, composerVersion, userId, architecture, cpus, memory } = getTelemetryContext();
 
+    const data = envelope.data as AppInsights.Contracts.Data<AppInsights.Contracts.RequestData>;
+
+    if (alwaysTrackEvents.includes(data.baseData.name as TelemetryEventName) && !telemetry?.allowDataCollection) {
+      stripPII(data);
+      return true;
```
tdurnford, line 64: I like the thought process here by returning true before the other properties get set; however, we do need some of them to be set like data.baseData.properties.toolName

Composer/packages/server/src/constants.ts
```diff
@@ -17,4 +17,4 @@ export enum ClaimNames {
 
 export const APPINSIGHTS_INSTRUMENTATIONKEY = process.env.APPINSIGHTS_INSTRUMENTATIONKEY;
 
-export const piiProperties = [];
+export const piiProperties = ['userId', 'rootProjectId', 'projectId'];
```
tdurnford, line 20: @garypretty Are there any other PII properties that we need to strip out for `TelemetryOptInOut` events? Probably be good to include session id as well.

Composer/packages/server/src/services/telemetry.ts
```diff
@@ -43,10 +57,16 @@ if (instrumentationKey) {
   AppInsights.defaultClient.addTelemetryProcessor((envelope: AppInsights.Contracts.Envelope): boolean => {
     const { sessionId, telemetry, composerVersion, userId, architecture, cpus, memory } = getTelemetryContext();
 
+    const data = envelope.data as AppInsights.Contracts.Data<AppInsights.Contracts.RequestData>;
+
+    if (alwaysTrackEvents.includes(data.baseData.name as TelemetryEventName) && !telemetry?.allowDataCollection) {
```
tdurnford, line 62: We might run into a race condition here if the client hasn't updated the telemetry settings before the event gets logged. Maybe we make a requirement that all items in the alwaysTrackEvents array must include the `enabled` prop that we use to check if we should strip pii.

tdurnford (changes requested):

Composer/packages/types/src/telemetry.ts
```diff
@@ -227,7 +227,9 @@ type PropertyEditorEvents = {
   RecognizerChanged: { recognizer: string };
 };
 
-type OtherEvents = {};
+type OtherEvents = {
+  TestEvent: undefined;
```
tdurnford, line 231: Make sure this gets removed

Composer/packages/server/src/services/telemetry.ts
```diff
@@ -59,11 +53,6 @@ if (instrumentationKey) {
 
     const data = envelope.data as AppInsights.Contracts.Data<AppInsights.Contracts.RequestData>;
 
-    if (alwaysTrackEvents.includes(data.baseData.name as TelemetryEventName) && !telemetry?.allowDataCollection) {
```
tdurnford: Where is the PII being stripped?

Composer/packages/client/src/telemetry/AppInsightsClient.ts
```diff
@@ -29,14 +29,16 @@ export default class AppInsightsClient {
     }
   }
 
-  public static trackEvent(name: string, properties: LogData) {
-    if (alwaysTrackEvents.includes(name as TelemetryEventName) || this._telemetrySettings?.allowDataCollection) {
+  public static trackEvent(name: TelemetryEventName, properties: LogData) {
```
tdurnford, line 32: Make sure this builds correctly. I tried this on my last telemetry PR, and there were a ton of type errors. Might be better to just leave it as a string for the moment.

Composer/packages/client/src/telemetry/AppInsightsClient.ts
```diff
@@ -29,14 +29,16 @@ export default class AppInsightsClient {
...
+    if (alwaysTrackEvents.includes(name)) {
```
tdurnford, line 33: I think this condition should be last. If telemetry is enabled, it's fine to add the event to the queue and let it propagate through the normal path.

beyackle: Good idea - will do.

Composer/packages/client/src/telemetry/AppInsightsClient.ts
beyackle: This does build - the only messing around with types I had to do was to add TestEvent to the "Other" set, since that's what we use in unit tests. If we don't to make a new type just for testing, we could make the unit tests use some other existing event type.

Composer/packages/server/src/services/telemetry.ts
beyackle: I'll double-check this. I don't actually see it offhand, so I'll make sure it gets done somewhere along the line.

Composer/packages/types/src/telemetry.ts
beyackle: This is here for type-correctness (unless we want to do without the "TestEvent" in the unit tests; I could go either way on this).

beyackle: I'm removing TestEvent; I'm just using real event types in the unit tests now.

Composer/packages/types/src/telemetry.ts
```diff
@@ -294,3 +295,13 @@ export type TelemetryClient = {
...
+
+/**
+ * These events are ones which are always tracked regardless of whether tracking is opted
+ * into or not (i.e. those having to do with turning telemetry itself on or off). If
+ * the user opts out, we will strip all PII from the event before sending it.
+ */
+export const alwaysTrackEvents: TelemetryEventName[] = ['TelemetryOptInOut'];
+
+/** Names of the properties containing PII Data */
+export const piiProperties = ['userId', 'rootProjectId', 'projectId'];
```
srinaath, line 307: RootProjectId and projectId are just session GUIDS. They shouldn't be classified as PII

tdurnford: @garypretty Should we consider project ids PII?

Composer/packages/client/src/telemetry/AppInsightsClient.ts
tdurnford: Nah, that's fine.

Composer/packages/client/src/telemetry/AppInsightsClient.ts
```diff
@@ -6,14 +6,24 @@ import {
...
+  piiProperties,
 } from '@bfc/shared';
 import chunk from 'lodash/chunk';
+import fromPairs from 'lodash/fromPairs';
 
 import httpClient from '../utils/httpUtil';
 
 const BATCH_SIZE = 20;
+
+const blankPIIObject = fromPairs(piiProperties.map((p) => [p, null]));
```
tdurnford, line 21: You don't need to remove PII in the client. We handle that on the server since that's were most of the PII is set.

Composer/packages/server/src/constants.ts
```diff
@@ -16,5 +16,3 @@ export enum ClaimNames {
 }
 
 export const APPINSIGHTS_INSTRUMENTATIONKEY = process.env.APPINSIGHTS_INSTRUMENTATIONKEY;
-
-export const piiProperties = [];
```
tdurnford, line 20: Did you move this somewhere else?

Composer/packages/types/src/telemetry.ts
```diff
@@ -294,3 +295,14 @@ export type TelemetryClient = {
...
+/**
+ * These events are ones which are always tracked regardless of whether tracking is opted
+ * into or not (i.e. those having to do with turning telemetry itself on or off). If
+ * the user opts out, we will strip all PII from the event before sending it. Every event
+ * whose name is in this array should have "enabled: boolean" in its fields.
+ */
+export const alwaysTrackEvents: TelemetryEventName[] = ['TelemetryOptInOut'];
+
+/** Names of the properties containing PII Data */
+export const piiProperties = ['userId'];
```
tdurnford, line 308: This is a better spot for the piiProperties array 👍🏻

Composer/packages/client/src/telemetry/AppInsightsClient.ts
beyackle: Fixed this in the latest version.

tdurnford (approved):

PR #8276, opened 2021-07-06
fix: correctly update time survey was taken

beyackle:
## Description

The HaTS survey wasn't saving the time it was taken to local storage, which was causing the "has it been long enough" check to always be true. Also, the useSurveyNotification hook wasn't recalculating eligibility, so if you were eligible once, you were eligible forever and the box would keep appearing.

## Task Item

#minor

### DIFF
--- a/Composer/packages/client/src/components/Notifications/useSurveyNotification.ts
+++ b/Composer/packages/client/src/components/Notifications/useSurveyNotification.ts
@@ -7,9 +7,9 @@ import formatMessage from 'format-message';
 import querystring from 'query-string';
 
 import { ClientStorage } from '../../utils/storage';
-import { surveyEligibilityState, dispatcherState, machineInfoState } from '../../recoilModel/atoms/appState';
+import { dispatcherState, machineInfoState } from '../../recoilModel/atoms/appState';
 import { MachineInfo } from '../../recoilModel/types';
-import { LAST_SURVEY_KEY, SURVEY_URL_BASE } from '../../constants';
+import { LAST_SURVEY_KEY, SURVEY_URL_BASE, SURVEY_PARAMETERS } from '../../constants';
 import TelemetryClient from '../../telemetry/TelemetryClient';
 
 const buildUrl = (info: MachineInfo) => {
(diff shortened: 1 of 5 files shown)


Composer/packages/client/src/components/Notifications/useSurveyNotification.ts
```diff
@@ -31,10 +31,12 @@ const buildUrl = (info: MachineInfo) => {
 };
 
 export const useSurveyNotification = () => {
-  const { addNotification, deleteNotification } = useRecoilValue(dispatcherState);
+  const { addNotification, deleteNotification, setSurveyEligibility } = useRecoilValue(dispatcherState);
   const surveyEligible = useRecoilValue(surveyEligibilityState);
   const machineInfo = useRecoilValue(machineInfoState);
 
+  setSurveyEligibility();
```
srinaath, line 38: This would keep calling the dispatcher quite a few times @beyackle . We should wrap it in a useEffect. Is there a reason you are placing it here?

srinaath (changes requested):

Composer/packages/client/src/components/Notifications/useSurveyNotification.ts
```diff
@@ -55,6 +57,7 @@ export const useSurveyNotification = () => {
               // This is safe; we control the URL that gets built
               // eslint-disable-next-line security/detect-non-literal-fs-filename
               window.open(url, '_blank');
+              surveyStorage.set('epochLastTaken', Date.now());
```
tdurnford, line 60: What is `epochLastTaken`?

tdurnford: You should also put this string in a variable in the surveyStorage file and import the variable to read/write to localstorage

beyackle: It's the timestamp for when the survey window was last opened (which is when we presume the user took the survey). It gets compared to Date.now() inside the hook so we can tell if it's been long enough to ask the user to take a survey again. Since this wasn't getting set, though, that logic assumed the user just never took the survey at all and so that check always passed (as if it were the first time the survey showed up).

tdurnford: Is there a more disruptive key we could use? lastSurveyTakenOnTimestamp

Composer/packages/client/src/components/Notifications/useSurveyNotification.ts
beyackle: This needs to run every time we could potentially display the notification. This dispatcher runs the logic to determine whether the user is eligible for a survey or not, including setting that to `false` once the survey has been taken. If we didn't do that, then the notification would just keep showing up because the eligibility would stay true.

Composer/packages/client/src/constants.tsx
```diff
@@ -533,4 +533,5 @@ export const SURVEY_PARAMETERS = {
   chanceToAppear: 0.3,
 };
 
+export const SURVEY_EPOCH_KEY = 'lastSurveyTakenOnTimestamp';
```
tdurnford, line 536: nit:
```suggestion
export const LAST_SURVEY_KEY = 'lastSurveyTakenOnTimestamp';
```

Composer/packages/client/src/components/Notifications/useSurveyNotification.ts
beyackle: Fixed - found a better place to run this logic.

srinaath (approved):

PR #8277, opened 2021-07-07
fix: LU editor would show an error if user tries to name an entity with a reserved name

### LINKED ISSUE #8055: Monaco editor for LU should show an error if user tries to name an entity with a reserved name
## Describe the bug

In the Monaco editor for LU, I can define an entity with the same name as a LUIS prebuilt. LUIS doesn't allow you to have custom entities of the same name as prebuilt entities, so it would be good to show a validation message.

## Version

Version:  2.0.0-nightly.246170.8fa93
Electron: 8.2.4
Chrome: 80.0.3987.165
NodeJS: 12.13.0
V8: 8.0.426.27-electron.0
## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [x] macOS
- [ ] Windows
- [ ] Ubuntu

## To Reproduce

1. Go to the trigger phrases and add a list entity using the same name as a LUIS prebuilt.
2. There is no indication that this name is reserved:

image

3. You only see an error when you try to start the bot:
image

## Expected behavior

Since we know that some names are reserved in LUIS, can we the LU language server show an error underneath the Monaco editor when an entity is defined with a reserved name? Similar to this:
image
but with message text like 

> L10:5 - L10:9 Invalid entity name. The name "age" is reserved for a prebuilt entity.

cosmicshuai:
closes #8055 
For prebuilt entities like `age, number, etc`. Now LU editor will show errors if user tries to name an entity with a reserved name.

For example:
image

Related PR in bf-lu:
https://github.com/microsoft/botframework-cli/pull/1250
https://github.com/microsoft/botframework-cli/pull/1260


hatpick:
@cosmicshuai can you attach the PR adding this to the bg-lu package?

cosmicshuai:
@hatpick 
I've attached the PRs related to this issue in the PR description filed.

hatpick (approved):

PR #8282, opened 2021-07-07
fix: categorizing LUIS and QNA resources correctly as required or optional in provision flow

### LINKED ISSUE #6775: QnA Maker and LUIS should be listed in required resources if needed in Publishing Profile

Today, we identify and help you provision LUIS and QnA Maker when you open your bot project, however when you have publishing profiles already created and / or you create a new one you should be told that these are required resources on the Add resources modal.
image

By intelligently adding the required resources based off of your project, this will make it where you don't get an error message when you go to publish.

Using the Next Steps menu to provide details that you should add the required resources to your publishing profiles to ensure that you give the user a heads up that all publishing profiles will now require this to be added and walk them through getting this done.  Maybe even disable publishing until these issues are addressed.

pavolum:
## Description

AzurePublish extension was always adding luis and qna as optional resources instead of required resources in the provision flow regardless of the underlying bot needing LUIS or QNA.

There is already a selector in client recoil state that does the parsing to determine whether LUIS/QNA are required in the current bot files (`allRequiredRecognizersSelector`)

This PR adds an accessor function for `allRequiredRecognizersSelector` that can be called through the Publish Plugin API such that the plug in can correctly categorize whether or not LUIS and QNA are optional.

## Task Item

fixes #6775 

## Screenshots

image

### DIFF
--- a/Composer/packages/extension-client/src/hooks/usePublishApi.ts
+++ b/Composer/packages/extension-client/src/hooks/usePublishApi.ts
@@ -58,6 +58,13 @@ export function usePublishApi() {
   function userShouldProvideTokens(): boolean {
     return window[ComposerGlobalName].userShouldProvideTokens();
   }
+  function getRequiredRecognizers(): {
+    projectId: string;
+    requiresLUIS: boolean;
+    requiresQNA: boolean;
+  }[] {
+    return window[ComposerGlobalName].getRequiredRecognizers();
+  }
   /** @deprecated use `userShouldProvideTokens` instead */
   function isGetTokenFromUser(): boolean {
     return window[ComposerGlobalName].userShouldProvideTokens();
@@ -78,5 +85,6 @@ export function usePublishApi() {
     userShouldProvideTokens,
     getTenantIdFromCache,
     setTenantId,
+    getRequiredRecognizers,
   };
 }
--- a/extensions/azurePublish/src/components/azureProvisionDialog.tsx
+++ b/extensions/azurePublish/src/components/azureProvisionDialog.tsx
@@ -14,13 +14,12 @@ import {
   getARMTokenForTenant,
   useLocalStorage,
   useTelemetryClient,
-  TelemetryClient,
   useApplicationApi,
 } from '@bfc/extension-client';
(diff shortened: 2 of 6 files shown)


Composer/packages/extension-client/src/hooks/usePublishApi.ts
```diff
@@ -58,6 +58,13 @@ export function usePublishApi() {
   function userShouldProvideTokens(): boolean {
     return window[ComposerGlobalName].userShouldProvideTokens();
   }
+  function getRequiredRecognizers(): {
+    projectId: string;
+    requiresLUIS: boolean;
+    requiresQNA: boolean;
+  }[] {
+    return window[ComposerGlobalName].getRequiredRecognizers();
```
benbrown, line 66: doesn't this need to pass the parameters through

benbrown: oh nevermind i see this is the return type.

extensions/azurePublish/src/components/azureProvisionDialog.tsx
```diff
@@ -601,7 +603,9 @@ export const AzureProvisionDialog: React.FC = () => {
 
   const getResources = async () => {
     try {
-      const resources = await getResourceList(currentProjectId, publishType);
+      const requireLUIS = requiredRecognizers.some((p) => p.requiresLUIS);
```
benbrown, line 606: why not move this into getResourceList instead of passing it as a parameter?

benbrown (approved):
As discussed, I would leave the optional resources in the list from the server, and just flip them to required on the front end.

Composer/packages/extension-client/src/hooks/usePublishApi.ts
pavolum: That function doesnt take params the bit of code 
```

extensions/azurePublish/src/components/azureProvisionDialog.tsx
pavolum: Gotcha

Issue #8283, opened 2021-07-07
App create failed when publishing from composer desktop 2.0 with new resources


stevengum:
## Describe the bug

The customer is publishing bot with LUIS from Composer desktop 2.0 with the option "Create new resources". The publish fails with the below error, which I describe in the repro steps.  

## Browser

- [x] Electron distribution

## To Reproduce

1. Customer signs into composer
2. Creates a new publishing profile
3. Chooses the tenant with multi factor auth enabled
4. Tries to publish and gets the below error

#### _This error has been anonymized and formatted for readability. See the linked incident for the original error._ 
```json
{
    "name": "CREATE_APP_REGISTRATION",
    "message": "App create failed! Please file an issue on Github.",
    "stack": "Error
        at new CustomizeError (C:\\Users\\John Doe\\AppData\\Local\\Programs\\Bot Framework Composer\\resources\\John.Doe.App\\extensions\\azurePublish\\dist\\extension.js:255547:25)
        at createCustomizeError2 (C:\\Users\\John Doe\\AppData\\Local\\Programs\\Bot Framework Composer\\resources\\John.Doe.App\\extensions\\azurePublish\\dist\\extension.js:255552:10)
        at BotProjectProvision.createApp (C:\\Users\\John Doe\\AppData\\Local\\Programs\\Bot Framework Composer\\resources\\John.Doe.App\\extensions\\azurePublish\\dist\\extension.js:256989:17)"
}
```

## Expected behavior

LUIS Resource should be published successfully.

## Additional context

For more information, please refer to this [incident](https://portal.microsofticm.com/imp/v3/incidents/details/246896943/home)  (**authentication required**).

stevengum:
Adding @Shameeritc and @victorlamana as FYI

@victorlamana I believe the issue you originally referenced is microsoft/botframework-composer#8249, correct?

victorlamana:
@stevengum hello Steven! Yeah, seems to be a similar issue

stevengum:
@victorlamana understood, thank you for the quick confirmation!

stevengum:
@natalgar, should @victorlamana & @shameeritc try the `v2.0.0-nightly.257298.f833d8e` nightly build of Composer to gather more info during the debugging process?

https://github.com/microsoft/BotFramework-Composer-Nightlies/releases/tag/v2.0.0-nightly.257298.f833d8e

natalgar:
@stevengum yes, the new build has additional logging

[...]

Issue #8249, opened 2021-07-01
Provision failure - App create failed!


victorlamana:
## Describe the bug

Hello Microsoft Team! Could you please help me on this issue?
Thanks a lot!

Provision failure
{"name":"CREATE_APP_REGISTRATION","message":"App create failed! Please file an issue on Github.","stack":"Error\n at new CustomizeError (C:\\Program Files\\Bot Framework Composer\\resources\\app.asar.unpacked\\extensions\\azurePublish\\dist\\extension.js:255547:25)\n at createCustomizeError2 (C:\\Program Files\\Bot Framework Composer\\resources\\app.asar.unpacked\\extensions\\azurePublish\\dist\\extension.js:255552:10)\n at BotProjectProvision.createApp (C:\\Program Files\\Bot Framework Composer\\resources\\app.asar.unpacked\\extensions\\azurePublish\\dist\\extension.js:256989:17)"}. See provision.Azure-Lab.log in your bot folder

## Version

2.0

## Browser

- [ ] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [X ] Edge

I don't think browsers make a difference here
## OS

- [ ] macOS
- [X ] Windows
- [ ] Ubuntu

## To Reproduce

1. Go to 'Publish'
2. Click on 'Publish Profile'
3.Create a new publish profile

## Expected behavior

Create the publish profile

luhan2017:
@victorlamana , it usually happens when the graph token is not valid, could you double check if you input a valid graph token? 

![image](https://user-images.githubusercontent.com/32191031/124212586-ccb8d100-db21-11eb-9a29-f5a18336c7d8.png)

victorlamana:
hello @luhan2017 i never saw this screen on Composer. I'm not sure if we're talking about the same thing:

![image](https://user-images.githubusercontent.com/52974336/124304379-fd0c6980-db39-11eb-8869-5b13150824b6.png)

My error happens when creating a publishing profile through this screen

![image](https://user-images.githubusercontent.com/52974336/124304457-1dd4bf00-db3a-11eb-903d-878134a73163.png)

victorlamana:
{
  "profileName": "Azure-LAB",
  "jobId": "2d6fe7e4-8891-404d-bdaa-8bba985238bb",
  "projectId": "88677.19435279138",
  "time": "2021-07-02T17:50:22.042Z",
  "log": [
    "Creating Azure resources...",
[...]

PR #8286, opened 2021-07-07
fix: log error message on final retry when "app create" or "add password" fails

natalgar:
## Description

log error message on final retry when "app create" or "add password" fails, assuming its an error from the Azure Graph API that returns error.message

## Task Item

refs #8249, #8283

### DIFF
--- a/extensions/azurePublish/src/node/provision.ts
+++ b/extensions/azurePublish/src/node/provision.ts
@@ -110,6 +110,14 @@ export class BotProjectProvision {
     // we need the `appId` and `id` fields - appId is part of our configuration, and the `id` is used to set the password.
     let appCreated;
     let retryCount = 3;
+
+    const logErrorMsg = (appCreateStep: string, errorMessage: string) => {
+      this.logger({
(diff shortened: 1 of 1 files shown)


extensions/azurePublish/src/node/provision.ts
```diff
@@ -110,6 +110,14 @@ export class BotProjectProvision {
     // we need the `appId` and `id` fields - appId is part of our configuration, and the `id` is used to set the password.
     let appCreated;
     let retryCount = 3;
+
+    const logErrorMsg = (appCreateStep: string, errorMessage: string) => {
```
GeoffCoxMSFT, line 114: Nit: Consider renaming to logFailedAfterRetries to indicate that it is specific to cases where the number of retries are exhausted.

GeoffCoxMSFT (approved):

stevengum:
Presumably available via the nightlies as of tag `v2.0.0-nightly.257298.f833d8e`

https://github.com/microsoft/BotFramework-Composer-Nightlies/releases/tag/v2.0.0-nightly.257298.f833d8e

PR #8288, opened 2021-07-07
fix: luisresource missing from import

### LINKED ISSUE #8244: Publishing Profile - Add New - Import missing JSON attribute for LUIS
## Describe the bug

Adding a new publishing profile to a project, select import, JSON schema generated is missing an attribute need to deploy SUCCESSFULLY with LUIS.

## Version
2.0

## Browser

- [ ] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ 93.0.926] Edge

## OS

- [ ] macOS
- [X ] Windows
- [ ] Ubuntu

## To Reproduce

1. select publish
2. select Publishing Profile tab
3. Click "Add new"
4. Select "Import"
5. Window pop with schema is missing  "luisResource": XXXXXXXXXX-luis" element at end of schema... will result in a failed deployment if you specify the other LUIS elements in the doc. 

## Expected behavior

 "luisResource": ""
-Should be last element in JSON
## Screenshots
BROKEN
image

CORRECT
image

natalgar:
## Description

add `luisResource` to the default schema for Azure Publishing

## Task Item

fixes #8244

### DIFF
--- a/extensions/azurePublish/src/node/schema.ts
+++ b/extensions/azurePublish/src/node/schema.ts
@@ -155,6 +155,7 @@ const schema: JSONSchema7 = {
     subscriptionId: '<id of your subscription>',
     region: '<region of your resource group>',
     appServiceOperatingSystem: 'windows',
+    luisResource: '<name of your luis resource>',
     settings: {
       applicationInsights: {
         InstrumentationKey: '<Instrumentation Key>',


extensions/azurePublish/src/node/schema.ts
```diff
@@ -155,6 +155,7 @@ const schema: JSONSchema7 = {
     subscriptionId: '<id of your subscription>',
     region: '<region of your resource group>',
     appServiceOperatingSystem: 'windows',
+    luisResource: '<name of your luis resource>',
```
natalgar, line 158: I put it up here with the other attributes of the same level. But if folks are used to seeing at at the bottom, below `MicrosoftAppId`, `MicrosoftAppPassword` I can move it there.

hatpick (approved):

extensions/azurePublish/src/node/schema.ts
benbrown: Yes having it at the top makes sense.

PR #8290, opened 2021-07-08
fix: some bugfix about skill

alanlong9278:
## Description

some bugfix about skill

### Can not delete skill folder when connecting to skill by zip

- Reason
Use the default folder structure of zip. But the zip structure may dont use skill name as folder name, so the skill name is not match to the folder name, we can not find the skill name folder when deleting a skill. Like the below image show, the folder name is manifest, but the skill name is empty.
image

- Solution
Replace the folder name with skill name.
skillZipDelete

### Can not select zip second time after fill a invalid url 
- Reason
 file input component can not trigger onChange function when the selected file is the same.
- Solution
Clean up input value after unzip selected zip file.
skillZipUrl

### replace real url with mock url for api test
```
    const url = 'https://xxx/manifests/Empty_45-2-1-manifest.json';
```

## Task Item

#minor


Composer/packages/server/src/models/bot/__tests__/botProject.test.ts
```diff
@@ -291,17 +291,31 @@ describe('skill operations', () => {
   xit('should create skill files', async () => {
     await proj.init();
 
-    const url = 'https://luhantest0625.azurewebsites.net/manifests/Empty_45-2-1-manifest.json';
+    const url = 'https://xxx/manifests/Empty_45-2-1-manifest.json';
```
hatpick, line 294: first, these tests are disabled, you should enable them and see if they pass.
second,. move this huge mocked zipContent outside each test and re-use it instead of re-defining it every time (assuming they are the same).

Composer/packages/server/src/models/utilities/util.ts
```diff
@@ -37,3 +37,7 @@ export const getRemoteFile = async (url): Promise<string> => {
   }
   return '';
 };
+
+export const convertFolderName = (path, folder) => {
```
hatpick, line 41: I dont understand what this util function is for, can you add types to the args, add better names to the args and function name, and add unit tests for it?

Composer/packages/client/src/components/AddRemoteSkillModal/BrowserModal.tsx
```diff
@@ -36,7 +37,14 @@ export const BrowserModal = (props) => {
         text={formatMessage('Browse')}
         onClick={onClickOpen}
       />
-      <input ref={inputFileRef} accept=".zip,.rar,.7zip" style={{ display: 'none' }} type="file" onChange={onChange} />
+      <input
+        ref={inputFileRef}
+        accept=".zip,.rar,.7zip"
+        max-file-size="1048576"
```
hatpick, line 43: how is this used? 
you should manually check the uploaded file size in the change callback and make sure it's within the limits of 1 GB use the following in your calculations instead of 1048576 ```const fileSizeLimit = 1024 * 1024```

hatpick (changes requested):

Composer/packages/client/src/components/AddRemoteSkillModal/BrowserModal.tsx
```diff
@@ -36,7 +37,14 @@ export const BrowserModal = (props) => {
         text={formatMessage('Browse')}
         onClick={onClickOpen}
       />
-      <input ref={inputFileRef} accept=".zip,.rar,.7zip" style={{ display: 'none' }} type="file" onChange={onChange} />
+      <input
+        ref={inputFileRef}
+        accept=".zip,.rar,.7zip"
```
hatpick, line 42: does JSZip support rar and 7zip?

srinaath: In addition, we do know that PVA is going to package it definitely as a zip. Is there a reason to support other formats?
@luhan2017

Composer/packages/server/src/models/utilities/util.ts
alanlong9278: Rename it & add test

Composer/packages/server/src/models/bot/__tests__/botProject.test.ts
alanlong9278: move the variable outside

Composer/packages/client/src/components/AddRemoteSkillModal/BrowserModal.tsx
alanlong9278: Have checked, JSzip just support .zip, so I remove the rar & 7zip

Composer/packages/server/src/models/bot/__tests__/botProject.test.ts
```diff
@@ -284,24 +284,30 @@ describe('qna operations', () => {
...
+  it('should create skill files', async () => {
     await proj.init();
 
-    const url = 'https://luhantest0625.azurewebsites.net/manifests/Empty_45-2-1-manifest.json';
-    const skillName = 'manifest';
-    const file = await proj.createSkillFiles(url, skillName, {});
+    const file = await proj.createSkillFiles(url, skillName, zipContent);
 
     expect(file).not.toBeUndefined();
   }, 10000);
```
srinaath, line 307: why do we need the wait time 10000? Could you assert for some property value inside the file instead of not undefined?

alanlong9278: Remove the wait time & add test 
```
    expect(file?.content).toContain(zipContent['manifests/Empty-2-1-manifest.json']);
```

hatpick (dismissed):

boydc2014 (approved):

boydc2014 (approved):

PR #8293, opened 2021-07-08
feat: add ability to resolve remote or relative schema $refs

### LINKED ISSUE #8271: [pva] add ability to resolve local and remote schemas
Support resolving local and remote schemas before sending to the client.

This enables schema authors (like PVA) to organize their schemas based on their own requirements.

a-b-r-o-w-n:
## Description

Adds ability to resolve $refs in external schemas (either remote or relative files).

## Task Item

closes #8271


Composer/packages/ui-plugins/schema-editor/src/Fields/ValueRefField.tsx
```diff
@@ -11,10 +11,12 @@ import { valueTypeDefinitions } from '../schema';
...
-    return Object.entries(valueTypeDefinitions || {}).map(([key, value]) => ({
-      key: `#/definitions/${key}`,
-      text: value?.title || startCase(key),
-    }));
+    return Object.entries(valueTypeDefinitions || {})
+      .map(([key, value]) => ({
+        key: `#/definitions/${key}`,
+        text: value?.title || startCase(key),
+      }))
+      .filter((o) => !o.key.includes('equalsExpression')); // a value must be a type, not just an expression
```
beyackle, line 19: I don't know if this is a big concern, but I think this would be slightly more efficient to do the filtering first (since you can do the same check for `equalsExpression` on the keys anyway).

beyackle (approved):
One small "fix it if you want to" thing, but this looks good otherwise, so I'll leave it to you if you want to do anything about it.

beyackle (approved):

PR #8301, opened 2021-07-08
feat: add cosmosdb to available resources

natalgar:
## Description

Cosmos DB - refactored in order to set up the resources config and dependencies in a central place, availableResourecs.

Work in progress for new azure provisioning service layer (this code is not live to customers yet). 

## Task Item

refs #7629

### DIFF
--- a/extensions/azurePublishNew/src/node/azureResources/cosmosDb.ts
+++ b/extensions/azurePublishNew/src/node/azureResources/cosmosDb.ts
@@ -5,8 +5,8 @@ import { TokenCredentials } from '@azure/ms-rest-js';
 import { CosmosDBManagementClient } from '@azure/arm-cosmosdb';
 
 import {
-  ProvisionServiceConfig,
   ProvisionMethod,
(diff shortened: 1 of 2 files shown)


extensions/azurePublishNew/src/node/azureResources/cosmosDb.ts
```diff
@@ -37,7 +39,7 @@ type CosmosDbProvisionResult = {
   collectionId: string;
 };
 
-type CosmosDbConfig = ResourceConfig & {
+export type CosmosDbConfig = ResourceConfig & {
```
VamsiModem, line 42: Consider moving this type to a common place, if used in other places.

natalgar: keeping it here because it most closely relates to the cosmosdb module

GeoffCoxMSFT (approved):

natalgar:
replacing this PR with #8321

PR #8306, opened 2021-07-09
fix: Add support of suggesting all entities in multiple import levels

### LINKED ISSUE #8295: LU intellisense only works for a single level of imports
## Describe the bug

When editing an LU file you get intellisense only if the definition is defined in that file or in an imported file.  You do not get intellisense if you have multiple import levels.

## Version

Version:  2.0.0-nightly.257298.f833d8e
Electron: 8.2.4
Chrome: 80.0.3987.165
NodeJS: 12.13.0
V8: 8.0.426.27-electron.0

## To Reproduce

1. Enable form generation in settings.
2. Create a form 'medical' 
3. Add a property 'name' of type 'personname'
4. Hit generate
5. Click on LU tab and select 'medical form dialog'
6. Click on 'show code'
7.  Add an utterance `- My {' and you should get a completion that includes 'Assign()'.

## Expected behavior

You should get completions for everything defined in the LU file and any imports.  

## Screenshots

No intellisense.
image

Intellisense if you go down one-level in imports.
image

cosmicshuai:
closes #8295 
For higher level imports, the LU editor now is able to suggest all entities defined. 
For example, in form dialog:
Screen Shot 2021-07-09 at 3 15 50 PM

Another example:
Empty.en-us.lu imports test.en-us.lu, test.en-us.lu imports compact2.en-us.lu and level2.en-us.lu, level2.en-us.lu imports level3.en-us.lu. Once user editing in Empty.en-us.lu, the editor can suggest all entities defined in all imported files.
image

### DIFF
--- a/Composer/packages/tools/language-servers/language-understanding/src/LUServer.ts
+++ b/Composer/packages/tools/language-servers/language-understanding/src/LUServer.ts
@@ -253,7 +253,7 @@ export class LUServer {
     luFeatures: any
   ): Promise<string[]> {
     let content = '';
-    const result: string[] = [];
+    let result: string[] = [];
     for (const importFile of imports) {
       try {
         content = (await importResolver('.', importFile.id)).content;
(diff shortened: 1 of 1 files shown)


chrimc62 (approved):
:shipit:

hatpick (approved):

PR #8307, opened 2021-07-09
fix: Adding fallback logic for readMe acquisition 

### LINKED ISSUE #8294: ReadMes not rendering for templates
## Describe the bug

In the template selection view, readMes are not rendering.

## Version

This is affecting all versions of Composer that use the new creation flow

## Browser

- [x] Electron distribution
- [x] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [x] macOS
- [x] Windows
- [ ] Ubuntu

## To Reproduce

Go to the template selection view and see that readMes do not populate

## Expected behavior

For the correct readMe to render

## Screenshots
 
 
image

## Additional context

I think its some obscure logic on how the npm registry handles readmes and versions. In the past the response was something like

```{versions:[{id:'1.0.0'},{id='rc15'},{id='rc14'}], readMe:'ReadMe Content for the package'}```

readme content was at the root of the response and was not tied to a particular version.Since updating we now see

```{versions:[{id:'r15'},{id: '1.0.0'},{id:'1.0.1-preview' readMe: 'Read Me COntent'}], readMe:''}```

So, since updating, the read me content is being stored within the version obj as opposed to the root like it was in the past and Composer is pulling readme from the root so it is pulling an empty string

I am unsure WHY this is happening but it IS the issue. The npm registry documentation is pretty lacking. Im wondering the best solution for this. At first glance we could just add more robust logic on the Composer side to check multiple areas for the readMe content if the expected field is blank. Alternatively we can go down the rabbit hole to see WHY these updates resulted in this change of registry schema but that might be quite an ambiguous feat to contend with.

### PR DESCRIPTION
**Pull request by pavolum**
## Description

We recently found out that pushing preview versions of packages changes where the readMe is stored in the npm registry API response. 

This PR enhances the readme parsing logic to account for this api behavior. 

If the readme is not at the root of the api response Composer will now look for it within the versions dictionary in the response. It searches the versions starting from the most recently published version and returning the first readMe it finds. 

If no readme is found then we render error text in the readMe section of template selection as opposed to no text. 

## Task Item
fixes #8294

## Screenshots

image

### CODE REVIEW

**Review by benbrown — APPROVED** [review]

File: Composer/packages/server/src/controllers/asset.ts [hunk 1]
```diff
@@ -118,7 +118,23 @@ export async function getTemplateReadMe(req: any, res: any) {
...
       const data = await response.json();
-      res.status(200).json(data?.readme || '');
+      // check for readme at root of response obj
+      let readMe = data?.readme;
+
+      // if no readme at root then pull readme from latest published version that has one
+      if (!readMe) {
+        const versionsDict = data?.versions;
+        if (versionsDict && Object.keys(versionsDict).length > 0) {
+          for (const versionKey of Object.keys(versionsDict).reverse()) {
```
cwhitten, line 128: this is written that the last key in the object returned from Object.keys is the latest. Object.keys() does not guarantee that. What other information is in this payload that can more deterministically find the latest version? Are dates associated with the version anywhere?

pavolum: There is a version string array that is ordered based on last published and that string value actually is the key to the versions dicitonary object so that should be a good way to get latest. Ill make that change now.

Composer/packages/server/src/controllers/asset.ts
```diff
@@ -124,10 +124,21 @@ export async function getTemplateReadMe(req: any, res: any) {
       // if no readme at root then pull readme from latest published version that has one
       if (!readMe) {
         const versionsDict = data?.versions;
-        if (versionsDict && Object.keys(versionsDict).length > 0) {
-          for (const versionKey of Object.keys(versionsDict).reverse()) {
-            if (versionsDict[versionKey] && versionsDict[versionKey]?.readme) {
-              readMe = versionsDict[versionKey].readme;
+        const versionTimesObj = data?.time;
+        if (versionsDict && versionTimesObj) {
+          const items = Object.keys(versionTimesObj).map((key) => {
```
hatpick, line 131: just use Object.entries

pavolum: Handled

benbrown (approved):

PR #8193, opened 2021-06-24
feature: offer users a HaTS survey on the design page

beyackle:
## Description

This is the completed feature branch for the HaTS survey feature. Please note that there is code in here that made it into the feature branch without being reviewed, as well as some late-entering bug fixes (I hadn't noticed that the card's own close box triggers the same Recoil dispatcher as I'd been calling manually, but also needed to insert a telemetry item there, so there are a few changes in that area), so this could still use a full review before it gets merged.

## Task Item

closes #8029

## Screenshots

image
(not shown in the screenshot is that the "Take the Survey" link routes to the correct URL with the query fields populated with OS, Composer version, machine ID, and Source)

PR #8308, opened 2021-07-09
chore: add unit tests to HaTS survey logic

beyackle:
## Description

This adds unit tests around the logic in `useSurveyNotification` that checks if a user is eligible to take the survey.

## Task Item

refs #8193


Composer/packages/client/src/components/Notifications/useSurveyNotification.ts
```diff
@@ -12,7 +12,7 @@ import { MachineInfo } from '../../recoilModel/types';
 import { LAST_SURVEY_KEY, SURVEY_URL_BASE, SURVEY_PARAMETERS } from '../../constants';
 import TelemetryClient from '../../telemetry/TelemetryClient';
 
-const buildUrl = (info: MachineInfo) => {
+export const buildUrl = (info: MachineInfo) => {
```
srinaath, line 15: We typically shouldn't export something just for testability. Was this exported just to write unit tests?

Composer/packages/client/src/components/Notifications/useSurveyNotification.ts
```diff
@@ -30,7 +30,7 @@ const buildUrl = (info: MachineInfo) => {
   return `${SURVEY_URL_BASE}?${querystring.stringify(parameters)}`;
 };
 
-const getSurveyEligibility = () => {
+export const getSurveyEligibility = () => {
```
srinaath, line 33: Same here. You should be able to test the logic it covers by testing the hook with different use cases

beyackle: Removed this.

Composer/packages/client/src/components/Notifications/useSurveyNotification.ts
beyackle: Removed this.

Composer/packages/client/src/components/Notifications/__tests__/useSurveyNotification.test.tsx
```diff
@@ -0,0 +1,152 @@
...
+const MOCK_VERSION = '2.3.4_jest';
+
+let surveyStorage;
+
+beforeAll(() => {
+  process.env.NODE_ENV = 'jest';
+  savedVersion = process.env.COMPOSER_VERSION;
+  process.env.COMPOSER_VERSION = MOCK_VERSION;
+
+  surveyStorage = new ClientStorage(window.localStorage, 'survey');
```
tdurnford, line 23: You should mock local storage

tdurnford (changes requested):

Composer/packages/client/src/components/Notifications/__tests__/useSurveyNotification.test.tsx
beyackle: For what I'm doing here with it, there's no need to; I don't need to check that stuff on localStorage is behaving right outside of the behaviors we see in the notification via the ClientStorage class, and `window.localStorage` is safe because JSDom provides it in a way that doesn't cause interference between different test files.

srinaath (approved):

tdurnford (approved):

PR #8372, opened 2021-07-13
fix: cypress/flake [Draft]

lei9444:
## Description
**root cause**
Some tests will create or update files and server may take some time to handle these files and if we change the triggers, the .lu.dialog file may also change. Next test will do clean first, but if the previous test is still handling the file, the remove task will fail. This is a race condition. And this will not block the previous test because we have a file persistence layer in client.

**fix**
add waiting time and rearrange the order of the tests.

## Task Item
#minor


lei9444:
reopen #8373

PR #8380, opened 2021-07-13
fix: validate endpoint url for connect to a skill

### LINKED ISSUE #8338: Importing the manifest should validate the endpoint url
## Describe the bug
The endpoint url is a required field for the Skill to function. 

Example:
[missingendpointurl.zip](https://github.com/microsoft/BotFramework-Composer/files/6804270/missingendpointurl.zip)

## Browser

- [x] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [x ] Windows
- [ ] Ubuntu

## To Reproduce

1. Create a zipped skill manifest which has an empty endpoint url
2. Import the manifest
3. an error should result, since the endpoint is missing

## Expected behavior
Validate required fields are present in the manifest during skill import.

## Screenshots

image

alanlong9278:
## Description

fix: validate endpoint url for connect to a skill
 image

## Task Item

closes #8338


Composer/packages/client/src/components/AddRemoteSkillModal/CreateSkillModal.tsx
```diff
@@ -46,6 +46,16 @@ export interface SkillFormDataErrors {
...
 const urlRegex = /^http[s]?:\/\/\w+/;
 const filePathRegex = /([^<>/\\:""]+\.\w+$)/;
+
+// All endpoints should have endpoint url
+const hasEndpointUrl = (content) => {
+  const endpoints = content.endpoints;
+  if (endpoints && endpoints.length > 0) {
+    return endpoints.every((endpoint) => !!endpoint.endpointUrl);
+  }
+  return true;
```
hatpick, line 56: if endpoints array is empty shouldn't you return false?

alanlong9278: Have checked, should return false if enpoints array is empty. Thanks

boydc2014 (approved):

PR #8381, opened 2021-07-13
fix: intent selected & description add scrollbar

alanlong9278:
### intent multiple select bug
 - Before

intentErr1
 
 - After
intentErr

### Add scroll bar for skill detail dom
image

## Task Item
#minor


tdurnford (approved):

hatpick:
Question: @luhan2017 Should the second screen disable the next button until at least one intent is selected?

luhan2017:
> Question: @luhan2017 Should the second screen disable the next button until at least one intent is selected?

In theory, this step is just to fetch the utterance of the selected intents as trigger phrases for skills. If user don't select any intent, they are also able to customize their own trigger phrases for skills.  @vivekkshankar, how do you think?

PR #8382, opened 2021-07-13
fix: Inserting text into the Skill host endpoint URL textbox error 

### LINKED ISSUE #8332: Inserting text into the Skill host endpoint URL textbox jumps to the end of the box
Version:  2.0.0-nightly.258070.287a4dc
Electron: 8.2.4
Chrome: 80.0.3987.165
NodeJS: 12.13.0
V8: 8.0.426.27-electron.0

1. Go to settings->skll configuration
2. With the mouse, click somewhere in the middle of the text box to set the insertion point there:
image
3. Try to type something and the cursor will jump to the end of the text box right after you type the first character
image

BTW, undo (CtrlZ) doesn't seem to work on this text box either.

alanlong9278:
### Fix Error about inserting text into the Skill host endpoint URL textbox jumps to the end of the box
inputErr

## Task Item

closes #8332


Composer/packages/client/src/pages/botProject/SkillHostEndPoint.tsx
```diff
@@ -87,15 +95,10 @@ export const SkillHostEndPoint: React.FC<SkillHostEndPointProps> = (props) => {
       <TextField
         ariaLabel={formatMessage('Skill host endpoint url')}
         data-testid={'SkillHostEndPointTextField'}
+        defaultValue={skillHostEndpoint}
         label={formatMessage('Skill host endpoint URL')}
         placeholder={formatMessage('Enter Skill host endpoint URL')}
-        value={skillHostEndpoint}
-        onChange={(e, value) => {
```
srinaath, line 93: Why does the jump in cursor position occur @alanlong9278?  Typically we debounce something only when a intensive operation is performed for the keystroke. It doesn't look to be that case here? If you dont want to update the recoil state immediately having a local state and then persisting to the recoil state onBlur seems better.

srinaath (changes requested):

Composer/packages/client/src/pages/botProject/SkillHostEndPoint.tsx
alanlong9278: 1. Cursor position error is because of recoil value change, so i set default value. 
2. Add debounce is for reducing api call. `setSettings` will trigger updating appSetting.json api call.

Composer/packages/client/package.json
```diff
@@ -77,6 +77,7 @@
     "react-timeago": "^4.4.0",
     "recoil": "^0.0.13",
     "sanitize-html": "2.3.3",
+    "sinon": "^11.1.1",
```
srinaath, line 80: We dont need sinon for fakeTimers. Use jest fake timers. 
https://jestjs.io/docs/timer-mocks

srinaath (changes requested):

Composer/packages/client/package.json
alanlong9278: Revert the debounce function

Composer/packages/client/src/pages/botProject/SkillHostEndPoint.tsx
alanlong9278: Use `state` & `onBlur` to call api to align with other setting property changes

srinaath (approved):

PR #8396, opened 2021-07-14
fix: breadcrumb resizing issue

### LINKED ISSUE #8361: [R14 bugbash] Resizing the Composer window gets ResizeObserver loop limit exceeded
Resizing Composer from full screen, I got a popup saying "ResizeObserver loop ..." ... "ResizeObserver loop limit exceeded".

What I did: 
Composer was on its "Home" page.
The window was maximized on my computer screen.
I moved it over to my external monitor. It maximized itself.
I clicked the resize button in top right corner. Error occurred.

zhixzhan:
## Description

set breadcrumb width to (85%, 15%), introduced in #7720

## Task Item
close #8361

### DIFF
--- a/Composer/packages/client/__tests__/pages/design/Design.test.tsx
+++ b/Composer/packages/client/__tests__/pages/design/Design.test.tsx
@@ -16,9 +16,6 @@ import { undoFunctionState } from '../../../src/recoilModel/undo/history';
 import mockProjectResponse from '../../../src/recoilModel/dispatchers/__tests__/mocks/mockProjectResponse.json';
 import DesignPage from '../../../src/pages/design/DesignPage';
 import { SAMPLE_DIALOG } from '../../mocks/sampleDialog';
-import ResizeObserver from '../../mocks/ResizeObserver';
-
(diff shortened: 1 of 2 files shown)


srinaath:
@zhixzhan were you able to repro the ResizeObserver Loop error and after the PR did the issue stop surfacing?

I think it still makes sense to merge the PR as we dont need to set the breadcrumb width through JS and we can take these changes in @sorgh @boydc2014

hatpick:
Do you have a solid repro for the issue? @zhixzhan 
if using percentage width (which was both me and Srinaath recommendation in the first place) solves the original issue, please go with it, it's better than manual calculations in JS.

beyackle:
I agree with the above - if this does solve the ResizeObserver issue, it should be approved and merged. I'm having trouble reproing the original bug, though, so it's hard for me to know whether this does or not.

zhixzhan:
@srinaath @hatpick @beyackle I can not repro this bug, after removing JS calculation, UI is still looks good, and should also improve page performance.

hatpick (approved):

PR #8406, opened 2021-07-14
fix: silently ignore ResizeObserver errors

### LINKED ISSUE #8361: [R14 bugbash] Resizing the Composer window gets ResizeObserver loop limit exceeded

beyackle:
## Description

This will cause all ResizeObserver errors to get swallowed by the window's custom error handler. These events are harmless and have been leading to issues.

## Task Item

close #8361

### DIFF
--- a/Composer/packages/client/src/components/ErrorBoundary.tsx
+++ b/Composer/packages/client/src/components/ErrorBoundary.tsx
@@ -35,6 +35,8 @@ interface ErrorBoundaryState {
   };
 }
 
+const patternsToIgnore = [/ResizeObserver/g];
+
 // only class component can be a error boundary
 export class ErrorBoundary extends Component<ErrorBoundaryProps, ErrorBoundaryState> {
   constructor(props: ErrorBoundaryProps) {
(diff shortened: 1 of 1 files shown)


Composer/packages/client/src/components/ErrorBoundary.tsx
```diff
@@ -57,6 +57,11 @@ export class ErrorBoundary extends Component<ErrorBoundaryProps, ErrorBoundarySt
   }
 
   onErrorHandler(message, source, lineno, colno, error) {
+    const patternsToIgnore = [/ResizeObserver/g];
```
tdurnford, line 60: Move `patternsToIgnore` outside of the function definition.

Composer/packages/client/src/components/ErrorBoundary.tsx
```diff
@@ -57,6 +57,11 @@ export class ErrorBoundary extends Component<ErrorBoundaryProps, ErrorBoundarySt
...
+    if (patternsToIgnore.some((regex) => regex.test(message))) {
+      // skip the console and application-level steps and just say we handled it
+      return true;
+    }
     console.error({ message, source, lineno, colno, error });
```
tdurnford, line 65: We should log the error to the console. Move the `console.log` before the if-statement. We don't want to hide errors from the developer.

tdurnford (changes requested):

tdurnford (approved):

PR #8408, opened 2021-07-15
fix: Support optional scm host domain for ASE

### LINKED ISSUE #8400: Publishing uses hard-coded SCM URL for zip deploy. This breaks deploy zip to ASE
## Describe the bug

Reported by  Behram Khan (Sr. Solutions Architect) on behalf of a customer trying to use Composer 2.0 to deploy into an App Service Environment (ASE).

Determined that ASE environments get special SCM URLS.  Instead of <appservicename>.scm.azurewebsites.net, they get <appservicename>scm.<asename>.p.azurewebsites.net.

Also found that the catch of 403 error reporting the token is invalid is misleading.

## Version

2.0 main (R14)

## Browser

- [ ] Electron distribution
- [X] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [X] Windows
- [ ] Ubuntu

## To Reproduce

1. Create ASE
2. Link App Service and App Service Plan to ASE
3. Use import in composer to identify existing assets
4. Publish

## Expected behavior

Publish succeeds.

## Actual behavior

Publish fails with 403 (forbidden) error and misreports that the token has expired.

GeoffCoxMSFT:
## Description

Added field to the publish profile for customers with ASE deployments to use to set the custom ASE SCM host domain.

## Task Item

fixes #8400 

## Screenshots

image

### DIFF
--- a/Composer/packages/lib/shared/src/settings.ts
+++ b/Composer/packages/lib/shared/src/settings.ts
@@ -70,5 +70,8 @@ export const applyPublishingProfileToSettings = (settings: DialogSetting, profil
   settings.MicrosoftAppId = profile.settings.MicrosoftAppId;
   settings.MicrosoftAppPassword = profile.settings.MicrosoftAppPassword;
 
+  // apply the SCM domain host
+  settings.scmHostDomain = profile.scmHostDomain;
+
   return settings;
 };
--- a/extensions/azurePublish/src/node/deploy.ts
+++ b/extensions/azurePublish/src/node/deploy.ts
@@ -157,7 +157,7 @@ export class BotProjectDeploy {
         status: BotProjectDeployLoggerType.DEPLOY_INFO,
         message: 'Publishing to Azure ...',
       });
-      await this.deployZip(this.accessToken, this.zipPath, name, environment, hostname);
+      await this.deployZip(this.accessToken, this.zipPath, name, environment, hostname, settings?.scmHostDomain);
       this.logger({
         status: BotProjectDeployLoggerType.DEPLOY_SUCCESS,
         message: 'Published successfully!',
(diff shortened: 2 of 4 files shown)


Composer/packages/lib/shared/src/settings.ts
```diff
@@ -70,5 +70,8 @@ export const applyPublishingProfileToSettings = (settings: DialogSetting, profil
   settings.MicrosoftAppId = profile.settings.MicrosoftAppId;
   settings.MicrosoftAppPassword = profile.settings.MicrosoftAppPassword;
 
+  // apply the SCM domain host
```
natalgar, line 73: do we need this comment? looks straightforward here

hatpick (approved):

extensions/azurePublish/src/node/deploy.ts
```diff
@@ -275,15 +275,22 @@ export class BotProjectDeploy {
...
+  ) {
     this.logger({
       status: BotProjectDeployLoggerType.DEPLOY_INFO,
       message: `Uploading zip file... to ${hostname ? hostname : name + (env ? '-' + env : '')}`,
     });
 
-    const publishEndpoint = `https://${
-      hostname ? hostname : name + (env ? '-' + env : '')
-    }.scm.azurewebsites.net/zipdeploy/?isAsync=true`;
+    const publishEndpoint = `https://${hostname ? hostname : name + (env ? '-' + env : '')}.${
```
natalgar, line 291: nit: can we put this logic in a method? Or evaluate each part as variables in lines above? Just a lot of logic to read in a template string!

natalgar (approved):

Composer/packages/lib/shared/src/settings.ts
GeoffCoxMSFT: I was following the comments in the existing function code, but good point.  I don't like obvious comments either:)

PR #8410, opened 2021-07-15
fix: set BeginSkill initial value allowInterruptions: 'false',

### LINKED ISSUE #8412: set default allowInterruptions=false in BeginSkill action
## Describe the bug

set default allowInterruptions=false in BeginSkill action 

## Browser

- [ ] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

zhixzhan:
## Task Item
fixes #8412

## Screenshots
image

### DIFF
--- a/Composer/packages/lib/shared/src/dialogFactory.ts
+++ b/Composer/packages/lib/shared/src/dialogFactory.ts
@@ -94,7 +94,7 @@ const initialDialogShape = () => ({
     botId: '=settings.MicrosoftAppId',
     skillHostEndpoint: '=settings.skillHostEndpoint',
     connectionName: '=settings.connectionName',
-    allowInterruptions: 'false',
+    allowInterruptions: false,
   },
   [SDKKinds.OnQnAMatch]: {
     $kind: SDKKinds.OnQnAMatch,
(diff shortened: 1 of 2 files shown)


luhan2017 (dismissed):

boydc2014 (approved):

Composer/packages/lib/shared/src/dialogFactory.ts
```diff
@@ -90,6 +90,12 @@ const initialDialogShape = () => ({
   [SDKKinds.SendActivity]: {
     activity: '',
   },
+  [SDKKinds.BeginSkill]: {
+    botId: '=settings.MicrosoftAppId',
+    skillHostEndpoint: '=settings.skillHostEndpoint',
+    connectionName: '=settings.connectionName',
+    allowInterruptions: 'false',
```
srinaath, line 97: This should be a boolean false and not stringified false. @zhixzhan  is there a reason you made it a string?

srinaath (changes requested):

Composer/packages/lib/shared/src/dialogFactory.ts
zhixzhan: @srinaath Thanks. it should be boolean.

boydc2014 (approved):

PR #8414, opened 2021-07-15
chore: remove version in manifest json file name

### LINKED ISSUE #8352: The version of the bot property should be used in the manifest path, not the skill manifest version
When I create a manifest for my bot, the **manifest version** is used in the path of the generated manifest URL,, we should probably use the bot version, not the manifest version for this.

The manifest version is implicit in the $schema attribute of the manifest file. 

If I try to generate a manifest based on these options:
image

The manifest URL will look like: 
https://composerbugbashrg.azurewebsites.net/manifests/PVAAsASkill001-2-1-manifest.json

It should have been something like (note the 1-0 in the URL
https://composerbugbashrg.azurewebsites.net/manifests/PVAAsASkill001-1-0-manifest.json

alanlong9278:
## Description

 use version to generate manifest json file name
version

## Task Item
closes #8352


luhan2017:
Let's hold on this change, this will generate one local manifest for each version, and we need to make sure version is a number and sorted correctly from oldest to newest. 

I have another proposal in the issue link. I think we can simplify this by

- disable the manfiest selection and always use schema 2.2 to generate manfiest file.
- just use "manifest.json" as the name, and always keep one manifest locally. (all the dialog assets doesn't have a version concept, it is not valid to only keep the manifest versions)

PR #8418, opened 2021-07-15
fix: fix tests by extending timeout

benbrown:
## Description

Fix some CI tests that seem to time out sometimes and cause errors.

#minor

### DIFF
--- a/Composer/packages/client/__tests__/components/ManageLuis/ManageLuis.test.tsx
+++ b/Composer/packages/client/__tests__/components/ManageLuis/ManageLuis.test.tsx
@@ -9,6 +9,7 @@ import { ManageLuis } from '../../../src/components/ManageLuis/ManageLuis';
 
 const serviceName = 'Language Understanding';
 const DOWN_ARROW = { keyCode: 40 };
+jest.setTimeout(10000);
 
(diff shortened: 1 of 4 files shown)


Composer/packages/client/__tests__/components/ManageLuis/ManageLuis.test.tsx
```diff
@@ -9,6 +9,7 @@ import { ManageLuis } from '../../../src/components/ManageLuis/ManageLuis';
 
 const serviceName = 'Language Understanding';
 const DOWN_ARROW = { keyCode: 40 };
+jest.setTimeout(10000);
```
srinaath, line 12: Do all tests in the suite need the big timeout value? If not we can use the individual timeout value
Screen Shot 2021-07-15 at 3 34 08 PM

taicchoumsft: Looks like we can be even more targeted by using the third param to set the timeout params (default is 1000) per [docs](https://testing-library.com/docs/dom-testing-library/api-async):
```
    const myGroup = await findByText('mockedGroup', undefined, {timeout: 10000});
```

PR #8420, opened 2021-07-16
fix: import fuzzy matched locale intent from PVA bot

### LINKED ISSUE #8411: Skill: support fuzzy(language code only) match for adding skill dispatch models
Currently when a root add a skill, root skill is doing an exact match for language codes.

For example, composer default and current authoring is "en-us", when adding a skill root bot will try to find “en-us” key in the dispatch models and show it in the trigger phrases window. 

Composer default language and most of the available languages are language-locale. But PVA only supports language only codes.

Composer need to support a fuzzy match for the languages. if composer's language is en-us, PVA is en, composer should be able to take and merge en utterances to composer en-us lu file.

liweitian:
## Description
import fuzzy matched locale lu intent from PVA bot
## Task Item
Closes #8411 
## Screenshots
test2

### DIFF
--- a/Composer/packages/client/src/components/AddRemoteSkillModal/SelectIntent.tsx
+++ b/Composer/packages/client/src/components/AddRemoteSkillModal/SelectIntent.tsx
@@ -172,7 +172,7 @@ export const SelectIntent: React.FC<SelectIntentProps> = (props) => {
   const [pageIndex, setPage] = useState(0);
   const [selectedIntents, setSelectedIntents] = useState<Array<string>>([]);
   // luFiles from manifest, language was included in root bot languages
-  const [lufilesOnLocale, setLufilesOnLocale] = useState<Array<{ locale: string; lufiles: LuFile }>>([]);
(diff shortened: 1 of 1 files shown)


Composer/packages/client/src/components/AddRemoteSkillModal/SelectIntent.tsx
```diff
@@ -80,27 +80,38 @@ const getRemoteLuFiles = async (
...
+  manifestDirPath: string,
+  locale: string
 ) => {
   const luFiles: Record<string, { id: string; content: string }[]> = {};
   try {
-    for (const [key, value] of Object.entries(skillLanguages)) {
-      if (composerLangeages.includes(key) && Array.isArray(value)) {
-        luFiles[key] = [];
-        for (const item of value) {
+    for (const cl of composerLangeages) {
```
luhan2017, line 88: Can we fix this typo? and add some comments here to explain what we are doing?

boydc2014 (approved):

PR #8553, opened 2021-08-13
feat: Add ability to configure template feed

### LINKED ISSUE #1241: Align Copy/Paste context menu to design
Current context menu in Visual Editor is not aligned to design. We can prioritize this work (design ready).

image
image

https://www.figma.com/file/04DymSq4xElhOBdAyutvvTbq/Composer-SH-06.19-?node-id=2124%3A1828

**Additional context**

### LINKED ISSUE #8477: No way to disable public repo access
Working with a large financial client that has a common industry security practice of filtering and blocking public repo access.  There is no method to turn off the default public repo access and Composer always attempts to access the repo locations at startup / going to the package manager.  For this client, this causes a security exception to be thrown on the developer's machine.

There needs to be a way to completely disable access or remove the public repo locations.

### LINKED ISSUE #8525: Allow multiple template feeds to drive template selection

There are Composer users who do not have access to public npm due to restrictions on their work machines. This results in the template selection view and template readmes to not render as they are sourced from public npm APIs as well as an error on app start when these api calls fail.

Related issues: #8478 #8477 #8188

- [x]  We need to add to the error handling when Composer calls to npm to fetch templates on start and provide a fallback experience for users who cannot hit the npm endpoints due to firewalls/proxy
  - proposed fallback: alert modal with info outlining that the users machine cannot access public npm and a deep link to the 'add alternate source for templates' feature
- [x] Enable users to provide alternate sources for templates
  - This will work for users who have private npm feeds or internal artifactory that have search apis similar to public npm such that they can provide a custom feed url that results in a similar structure to our [public npm registry api call] (https://registry.npmjs.org/-/v1/search?text=generator+keywords:bf-template+scope:microsoft). Then we simply need to add the ability to toggle between different, customer entered feeds
  - With this solution then the user would need to get approval to add the template packages to their private npm feed/artifactory, enter their private feed search URL (i.e. https://{Private npm feed root url}/-/v1/search?text=generator+keywords:bf-template+scope:microsoft) as a new source and they would have access to the public generators (as well as any custom ones they may choose to create and add to their private feed)
- [ ] Enable users to specify a local dir to treat as their 'npm feed' 
  - I suspect that there will be users who have internal artifact feeds that cannot be indexed in a similar way as our [public npm registry api call](https://registry.npmjs.org/-/v1/search?text=generator+keywords:bf-template+scope:microsoft). I am going to validate this with the multiple users who experienced this issue.
  - At first glance, one solution would be to enable users to add paths to local directories and implement the logic to treat a local path as a feed
    - That way users would go through the normal approval process to add public npm packages to their business artifactory, since that artifactory cannot be deterministically indexed by composer they would then have to download the packages locally and have Composer point to that local directory since Composer can deterministically parse through folder structure 

## Current work around 

There is a current work around that will resolve the issue  users assuming they can access public github repos.
1. Download the template generators from the [bot framework components repo's generators directory](https://github.com/microsoft/botframework-components/tree/main/generators)
2. Open a command line and navigate to where the Bot Framework Composer executable is (default on windows machine is `C:\Users\{USER NAME}\AppData\Local\Programs\Bot Framework Composer`)
3. run the following command: `set ADVANCED_TEMPLATE_OPTIONS=true` 
4. Run the following command: `"Bot Framework Composer.exe"`
5. This should open Composer and in the template selection view you should now see a 'create from local template' option. Select this option
6. in the local template path, place a path to the index.js file for the template you want to use which should now be on your machine from step 1. For example, if I wanted to create an empty bot then I would place the following path to my empty bot generators index.js file: `C:\Users\{USERNAME}\source\repos\botframework-components\generators\generator-bot-empty\generators\app\index.js`
7. Go through the creation modal and execute creation, you should have a resulting bot proj based on the template you chose in step 6.

## Additional context

- This issue is still being investigated and the above solutions are not the agreed upon best approach. We plan to have a agreed upon solution in the coming days as we gather more information from users experiencing the issue. The goal will be to add the solution to the patch release which will ship around end of August.
- This may be a larger issue if users in these enterprise networks cannot run `dotnet restore` or `npm install`. This would mean that that they would not only need to get approval for the generator npm pacakges to be added to their internal feeds but also all package dependencies of those generators.

pavolum:
## Description

**Ability to configure template feed**
- Adds a variable for a custom template feed url in user storage
- Adds a corresponding form in the app settings page to alter that template feed url variable
- Connect the stored template feed url with the call to fetch templates so the users custom source is used to populate template selection vs the default feed
- Fallback UI in template selection if custom feed returns no templates
- Deep link in creation flow to settings page where custom feed form is located

**Fallback experience for first party feed**
- If the first party feed url fails for any reason then a hardcoded back up set renders in the template selection screen
  - Hard coded template list has * for version number so it snaps to the latest stable version for that template

**Misc**
- Minor clean up throughout the template flow
- Unit tests for above new features
- No longer render just QNA template if feed url returns empty array

## Task Item
fixes #8477 
fixes #8525 
fixes #1241

## Screenshots

image

image

image


Composer/packages/client/src/components/CreationFlow/CreateBot.tsx
```diff
@@ -134,6 +144,10 @@ export function CreateBot(props: CreateBotProps) {
   const [localTemplatePathValid, setLocalTemplatePathValid] = useState<boolean>(false);
   const [displayedTemplates, setDisplayedTemplates] = useState<BotTemplate[]>([]);
   const [readMe] = useRecoilState(selectedTemplateReadMeState);
+  // const [isTemplateFeedModalVisible, setIsTemplateFeedModalVisible] = useState<boolean>(false);
```
benbrown, line 147: remove

Composer/packages/client/src/components/CreationFlow/CreateBot.tsx
```diff
@@ -264,22 +278,43 @@ export function CreateBot(props: CreateBotProps) {
 
   return (
     <Fragment>
+      {/* <TemplateFeedModal
```
benbrown, line 281: remove

Composer/packages/client/src/components/CreationFlow/CreateBot.tsx
```diff
@@ -264,22 +278,43 @@ export function CreateBot(props: CreateBotProps) {
...
-          }}
-        >
-          <PivotItem data-testid="dotnetFeed" headerText="C#" itemKey={csharpFeedKey}></PivotItem>
-          <PivotItem
-            data-testid="nodeFeed"
-            headerText={formatMessage('Node (Preview)')}
-            itemKey={nodeFeedKey}
-          ></PivotItem>
-        </Pivot>
+        <Stack horizontal styles={tabAndFeedBtnStyles}>
```
benbrown, line 290: why was this change needed to wrap everything in a stack? seems like this should be reverted since we no longer need the edit feeds button?

Composer/packages/client/src/components/CreationFlow/CreateBot.tsx
```diff
@@ -25,13 +25,21 @@ import { Link } from 'office-ui-fabric-react/lib/Link';
...
 
 import msftIcon from '../../images/msftIcon.svg';
 import { DialogCreationCopy } from '../../constants';
-import { creationFlowTypeState, fetchReadMePendingState, selectedTemplateReadMeState } from '../../recoilModel';
+import {
+  creationFlowTypeState,
+  dispatcherState,
+  fetchReadMePendingState,
+  selectedTemplateReadMeState,
+  templateFeedUrlState,
```
benbrown, line 37: not used?

Composer/packages/client/src/components/CreationFlow/CreateBot.tsx
```diff
@@ -25,13 +25,21 @@ import { Link } from 'office-ui-fabric-react/lib/Link';
...
+} from '../../recoilModel';
 import TelemetryClient from '../../telemetry/TelemetryClient';
 
 import { TemplateDetailView } from './TemplateDetailView';
+import { TemplateFeedModal } from './TemplateFeedModal';
```
benbrown, line 42: not used

Composer/packages/client/src/components/CreationFlow/TemplateFeedModal.tsx
```diff
@@ -0,0 +1,74 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
```
benbrown, line 2: this whole file is no longer needed right?

Composer/packages/client/src/recoilModel/dispatchers/project.ts
```diff
@@ -42,6 +42,7 @@ import {
   orchestratorForSkillsDialogState,
   selectedTemplateVersionState,
   watchedVariablesState,
+  templateFeedUrlState,
```
benbrown, line 45: why needed?

Composer/packages/client/src/recoilModel/dispatchers/storage.ts
```diff
@@ -153,9 +152,11 @@ export const storageDispatcher = () => {
         set(templateProjectsState, data);
       }
     } catch (err) {
+      TelemetryClient.track('TemplateNpmRegistryCallFailed', { error: err.message });
+
       set(applicationErrorState, {
         message: err.message,
-        summary: formatMessage('Error fetching runtime templates'),
+        summary: formatMessage('Error fetching not project templates'),
```
benbrown, line 159: "not project?" 

i think this should be "bot project"

Composer/packages/server/src/store/data.template.ts
```diff
@@ -10,6 +10,7 @@ import { BotProjectMetadata } from '../services/project';
 
 export default {
   version: 1,
+  templateFeedUrl: firstPartyTemplateFeed,
```
benbrown, line 13: this concerns me - as discussed, would prefer this to be absent unless customized to avoid zombie values if we change them later.  IS there a reason we can't do that? I believe all the other code is set up to use this default without it being present in the config

benbrown (changes requested):

Composer/packages/client/src/components/CreationFlow/CreateBot.tsx
pavolum: This is now a deep link that navigates you to the templae feed form in the settings page so the stack implementation is still needed

Composer/packages/server/src/store/data.template.ts
pavolum: Handled

pavolum:
TODO: 

- [ ] Unit tests
- [ ] Reorder setting to lower in the settings page and have deep link nav to that form within the page
- [ ] Add text validation and error message to the feed URL form to check if the feed url returns valid template format

Composer/packages/client/src/pages/setting/app-settings/TemplateFeedForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+export const TemplateFeedForm = () => {
+  const templateFeedUrl = useRecoilValue(templateFeedUrlState);
+  const templateProjects = useRecoilValue(templateProjectsState);
+  const { setTemplateFeedUrl, fetchTemplates } = useRecoilValue(dispatcherState);
+
+  const [urlValue, setUrlValue] = useState(templateFeedUrl ? templateFeedUrl : firstPartyTemplateFeed);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [fieldDisabled, setFieldDisabled] = useState(false);
+
+  const savePendingEdits = (isFirstPartyFeed?: boolean) => {
```
benbrown, line 40: is this additional flag necessary since you are already setting the value to the default?

Composer/packages/client/src/pages/setting/app-settings/TemplateFeedForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+      fetchTemplates([templateFeedUrl]);
+    } else {
+      fetchTemplates([firstPartyTemplateFeed]);
+    }
+  }, [templateFeedUrl]);
+
+  useEffect(() => {
+    setFieldDisabled(false);
+    if (templateProjects.length < 1) {
+      setErrorMessage('This feed did not return any templates, please confirm that this is a valid template feed URL.');
```
benbrown, line 71: formatMessage?

benbrown (approved):

Composer/packages/client/src/pages/setting/app-settings/TemplateFeedForm.tsx
pavolum: Yeah its still needed, the state change hasnt happened yet on the urlValue variable within the components lifecycle even though it is set in the code prior to calling this function so the conditional on line 41 would not be truthy.

benbrown: should this maybe be a useEffect on the feed value changing then?

benbrown: no I see.. this is fine!

benbrown (approved):

benbrown (approved):
re-approved

PR #8566, opened 2021-08-17
fix: the Value field of a Trace Activity is throwing warnings

### LINKED ISSUE #8472: with 2.1 the Value  field of a Trace Activity is throwing warnings where it did not previously
## Describe the bug

Pre 2.1, I could have the Value field of a trace activity, with a value type of string, without quotes and just call properties as needed. no warning was raised. 
`turn.activity.listOfScores = ${turn.activity.listOfScores}`
`turn.activity.value = ${turn.activity.value} and turn.activity = ${turn.activity}`

No single quotes were necessary as they were always implied. 
I now have hundreds of warnings for trace events properties where there wasnt a problem before.  

## Version

Version 2.1

## Browser

image

## OS

- [ ] Windows

## To Reproduce

Create a Trace Event  with a value type of string and enter 
a string with escaped properties. 
see screen shots for examples.
warning is received. 

## Expected behavior

the value fields with value type of string should not need quotes since its defined as a value type of string. 

## Screenshots

**Screenshot 1**
image

**Screenshot 2**
image

lei9444:
## Description
**Root cause**
In the current design, if the expression contain string type ,we don't parse the expression. And the the return type use binary number to store. But indexer only use the first element of the types and use "===" to check the type. we need to check all types in the array and use bit operation.

**Fix**
check all types and do bit operation

## Task Item
closes #8472

### DIFF
--- a/Composer/packages/lib/indexers/src/validations/expressionValidation/validation.ts
+++ b/Composer/packages/lib/indexers/src/validations/expressionValidation/validation.ts
@@ -55,7 +55,10 @@ export const checkExpression = (exp: any, required: boolean, types: number[]): n
       break;
     }
     case 'string': {
-      returnType = checkStringExpression(exp, types.length === 1 && types[0] === ReturnType.String);
+      returnType = checkStringExpression(
+        exp,
+        types.some((type) => type & ReturnType.String)
+      );
       break;
     }
     default:


srinaath (approved):

srinaath:
@lei9444  can we get this PR merged?

PR #8582, opened 2021-08-19
PVA V2 Demo - Project tree expand/collapse

pbuxton206:
## Description

PVA v2 Demo styling updates for a few components:
- Replaced "Add" and "Filter" buttons in project tree header with header title
- Added ability to expand / collapse the project tree using "X" button in project tree header and an expand button in the design Toolbar.
- Removed the <Split> component for the demo; note this means that the tree width cannot be resized.
- Replaced the breadcrumb header with a simple label indicating the dialog name.

## Reference
https://www.figma.com/file/frQviquVRVWnyr1JcAxrAz/Authoring-PVA-2.0?node-id=1240%3A7582

## Screenshots
pvav2testbot3


pbuxton206:
Adjusted a spacing issue on the Toolbar where button elements were pushed down slightly when expand button was visible.
image

PR #8844, opened 2021-12-09
fix: a11y keyboard interactions for configure and publish screens

OEvgeny:
## Description

Fixes the following keyboard navigation issues:
- On "Configure your bot" screen, "Development Resources" tab keyboard focus gets lost after closing modals. The issue happens because modal trigger buttons get disabled when "Set up Language Understanding" and "Set up QnA Maker" modals are opened. Removed modal trigger buttons disabling logic to fix the issue.
- On "Publish your bots" screen, "Publish" tab, "Publish target" drop-down causes navigation to the "Publishing profile" tab when it receives a focus event. The issue is caused by Dropdown component triggering `onChange` handler when keyboard focus received. Fixed by ignoring focus events for the particular "Manage profiles" drop-down item. Turned the drop-down to controlled one, so it always follows `item.publishTarget` selection rather internal state which allowed "Manage profiles" to be selected.

## Task Item

- [VSTS 70447](https://fuselabs.visualstudio.com/Composer/_workitems/edit/70447)
- [VSTS 70092](https://fuselabs.visualstudio.com/Composer/_workitems/edit/70092)

## Screenshots
preserve-focus-on-modals-kb webm
fix-publish-profile-kb webm

### DIFF
--- a/Composer/packages/client/src/pages/botProject/RootBotExternalService.tsx
+++ b/Composer/packages/client/src/pages/botProject/RootBotExternalService.tsx
@@ -410,7 +410,6 @@ export const RootBotExternalService: React.FC<RootBotExternalServiceProps> = (pr
           )}
         </div>
         <PrimaryButton
-          disabled={displayManageLuis || displayManageQNA}
           styles={{ root: { width: '250px', marginTop: '15px' } }}
           text={formatMessage('Set up Language Understanding')}
           onClick={() => {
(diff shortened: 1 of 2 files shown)


Composer/packages/client/src/pages/botProject/RootBotExternalService.tsx
```diff
@@ -479,7 +478,6 @@ export const RootBotExternalService: React.FC<RootBotExternalServiceProps> = (pr
           />
         </div>
         <PrimaryButton
-          disabled={displayManageLuis || displayManageQNA}
```
a-b-r-o-w-n, line 482: This seems a little risky to me. Is the problem that by the time the modal is dismissed, the buttons are still disabled so cannot get focused?

OEvgeny: Played with these buttons a bit: either currently focused button can't be remembered because it is disabled when modal gets opened (hence it cant be restored), or it can't be restored because the button is still disabled while modal is closing.
Tried different scenarios, and haven't found any regression with these buttons.

a-b-r-o-w-n (approved):

PR #8851, opened 2021-12-13
fix: change luis editor prebuilt entity error to warn

### LINKED ISSUE #8475: [MCS-DE] Syntax Validation LU preBuild Entities does not consider locale
With Composer v2.1 a preBuild entityname error for a reserved name is thrown. However the mentioned preBuild entity does not exist in the locale (de-DE) of the model, only in English.

Syntax checker should not throw an error in this case.

image

cosmicshuai:
fixes #8475 
Change luis authoring a prebuilt entity name reserved error to warning.
prebuilt_warn_editor
.
image


cwhitten (approved):

cwhitten:
Thank you @cosmicshuai

PR #8894, opened 2022-01-13
build: Switched production builds to yarn v3.1.1 (berry) for Cyber EO compliance

tonyanziano:
#minor https://msazure.visualstudio.com/CCI/_workitems/edit/13077543/

As part of the Cyber EO work, we need to pull our NPM dependencies from an internal Azure Artifacts feed when building in production.

The original plan was to dynamically change the URLs in the `yarn.lock` file at build time, like [we did in the Emulator.](https://github.com/microsoft/BotFramework-Emulator/pull/2363) However, the yarn v1 client can be unreliable when trying to pull packages from Azure Artifacts, as was told to us by the Azure Artifacts team. So the solution instead, was **to upgrade our build pipelines to build Composer using yarn berry (v2.x.x+)**.

This PR makes several changes that are needed for the pipelines to build and package Composer using yarn v3.1.1:

- Added `.yarnrc.yml` file to the root of the project (this will control the behavior of yarn 3)
- Added `yarn-berry.lock` files to the root and extensions (this is the yarn berry equivalent of a yarn file, and will not interfere with yarn 1 installations)
- Added a version-locked release of yarn 3.1.1 in the `.yarn/releases` folder that will be used by the build pipelines -- you have to opt-in to use 3.1.1 (v1 will still work locally)
- Locked several packages to their versions that were listed in the v1 `yarn.lock` files -- there were some bugs that surfaced with minor package bumps 
  - `tslib` needs to be added to all packages that consume the base `tsconfig.json` (almost every workspace) because it has the flag `importHelpers: true` enabled, which requires `tslib` as a dependency
  - `@reach/router` needed to be locked, because later versions were not automatically decoding URL parameters
  - `express` and `@types/express` needed to be locked due to some typing mismatches that were causing compilation errors
  - `cross-env` needed to be added to some packages that used it in their scripts
  - some `cross-env` scripts needed to be escaped with quotes
  - since `@bfc/server-workers` is unpacked from the app's standard asar, we need to stop hoisting dependencies so that it has all of its dependencies and can function independently outside of the asar


Composer/packages/tools/language-servers/language-generation/src/LGServer.ts
```diff
@@ -117,7 +117,7 @@ export class LGServer {
 
     this.connection.onRequest((method, params) => {
       if (initializeDocumentsMethodName === method) {
-        const { uri, lgOption }: { uri: string; lgOption?: LGOption } = params;
+        const { uri, lgOption }: { uri: string; lgOption?: LGOption } = params as { uri: string; lgOption?: LGOption };
```
tdurnford, line 120: Maybe extract `{ uri: string; lgOption?: LGOption }` out to a type since it's used twice. Or not.

Composer/packages/tools/language-servers/language-understanding/src/LUServer.ts
```diff
@@ -95,7 +95,7 @@ export class LUServer {
       if (method === LABELEXPERIENCEREQUEST) {
         this.labelingExperienceHandler(params);
       } else if (InitializeDocumentsMethodName === method) {
-        const { uri, luOption }: { uri: string; luOption?: LUOption } = params;
+        const { uri, luOption }: { uri: string; luOption?: LUOption } = params as { uri: string; luOption?: LUOption };
```
tdurnford, line 98: Consider extracting to type.

extensions/samples/package.json
```diff
@@ -9,7 +9,6 @@
     "build": "node ../scripts/build.js",
     "watch": "yarn build --watch"
   },
-  "dependencies": {},
```
tdurnford, line 12: Great cleanup!

tdurnford (dismissed):

PR #8906, opened 2022-01-14
fix: Make sure Orchestrator languages are not filtered when publishing

taicchoumsft:
This fixes an issue where the Orchestrator multi-language settings can disappear when the bot is using locales that are not supported by LUIS. 

The `dialogsWithLuProviderSelectorFamily` selector needs to be used to provide the `luProvider` property on the dialog. 

Without this property, `botindexer::filterLUISFilesToPublish` will always filter languages based on what LUIS supports.  When using Orchestrator, we don't want this filtering to occur, since Orchestrator supports all the languages that the Composer UI supports. 

## Task Item
Internal IcM. fixes #minor


srinaath (approved):

taicchoumsft:
> @taicchoumsft tests seem to break

Thanks @srinaath , I just updated branch and it unit tests appear to pass now.  Will run a few more times to be sure.

PR #8937, opened 2022-01-25
a11y: improve contextual menus and filtering

OEvgeny:
## Description

This improvements UX when navigating via keyboard through contextual menus and search/filter fields in the following ways:
- Search/filter is cleared on the first Escape key press, and gets closed on the second one
- When contextual menu is used with filter or search field, it respects arrows keys in addition to Escape key for toggling the menu.

These improvements resulted in the following changes:
- Swapped `onDismiss` for contextual menu items with `onMenuDismissed` since the later doesn't block the menu from being dismissed in contrast to the former.
- Added the logic to return focus back on filter blur action
- Refactored contextual menus hooks to avoid repeating code

## Task Item

- [VSTS 70101](https://fuselabs.visualstudio.com/Composer/_workitems/edit/70101)

## Screenshots

Desktop Recording 4 webm
Desktop Recording 6 webm

#minor

### DIFF
--- a/Composer/packages/client/src/components/ProjectTree/ProjectTreeHeader.tsx
+++ b/Composer/packages/client/src/components/ProjectTree/ProjectTreeHeader.tsx
@@ -6,10 +6,11 @@ import { jsx, css } from '@emotion/core';
 import { useEffect, useRef, useState } from 'react';
 import { FontSizes, NeutralColors } from '@uifabric/fluent-theme';
 import formatMessage from 'format-message';
-import { CommandButton } from 'office-ui-fabric-react/lib/Button';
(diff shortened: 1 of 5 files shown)


OEvgeny (commented):
Need to do some additional checks for:
```
Composer/packages/lib/code-editor/src/components/toolbar/ToolbarButtonMenu.tsx
Composer/packages/lib/code-editor/src/lu/DefineEntityButton.tsx
Composer/packages/lib/code-editor/src/lu/hooks/useLabelingMenuItems.tsx

```
Before this gets in

cwhitten (approved):

PR #8951, opened 2022-01-28
a11y: allow leaving flow editor and monaco editor using escape key

OEvgeny:
## Description

This allows keyboard `Escape` key to be used to leave editors (monaco editor and designer editor). This also fixes multiple issues with toolbar navigation. Now the focus order is as follows:
- Designer editor
- Designer toolbar
- Right pane

## Task Item

- [VSTS 70428](https://fuselabs.visualstudio.com/Composer/_workitems/edit/70428)

## Screenshots

image
image

Reelia-recording-2022-01-28 webm

#minor


Composer/packages/client/src/pages/design/VisualEditor.tsx
```diff
@@ -105,6 +105,11 @@ const VisualEditor: React.FC<VisualEditorProps> = (props) => {
         aria-label={formatMessage('Visual editor')}
         css={visualEditor(triggerButtonVisible || !selected)}
         data-testid="VisualEditor"
+        onKeyDownCapture={(ev) => {
+          if (ev.key === 'Escape') {
+            (document.activeElement as HTMLElement)?.blur();
```
OEvgeny, line 110: The `document.activeElement` is not always an `HTMLDocument` e.g. it might be an `SVGElement`. That's why this cast is needed.

cwhitten (approved):

PR #8976, opened 2022-02-08
a11y: make Select Dialog a listbox with options

OEvgeny:
## Description

The PR improves the button role by making the control to appear as a listbox with options which is closer to the control functionality.

## Task Item

- [VSTS 70436](https://fuselabs.visualstudio.com/Composer/_workitems/edit/70436)

## Screenshots

#minor


tonyanziano (changes requested):
The change looks good, but it looks like the `selectDialog.test.tsx` is failing:

```
Summary of all failing tests
FAIL packages/ui-plugins/select-dialog/src/__tests__/selectDialog.test.tsx (6.263 s)
  ● Select Dialog › should create a new dialog

    TestingLibraryElementError: Unable to find an accessible element with the role "menuitem"

    Here are the accessible roles:

      listbox:
```

OEvgeny:
Updated, thanks for looking into this!

cwhitten (approved):

PR #8998, opened 2022-02-16
a11y: handle keyboard navigation in array fields

OEvgeny:
## Description

This fixes the following issues found for ArrayField:
- The field disappears when moving focus back via keyboard (shift+tab)
- No chance to know current row while using screen reader when a field gets active
- Unclear focus behavior when pressing "Add new" button: focus stays on the button
- Multiple fields have the same id

The PR introduces the following changes to fix issues above:
- Remove field removal on blur because it works unstable. Instead disable "Add new" button until the previous field recieves the value, and remove any empty field in-between fields with values when adding a new empty field
- Add a `region` role to array fields with clear `aria-label` and make the region focusable
- Place focus on the region when creating a new field via keyboard (focus is not visible when clicking "Add new" via mouse)
- Add an index to array field ids, so every field has a different id

## Task Item

- [VSTS 70458](https://fuselabs.visualstudio.com/Composer/_workitems/edit/70458)

## Screenshots

array-field-kb webm

array-field-mouse webm

#minor

### DIFF
--- a/Composer/packages/adaptive-form/src/utils/arrayUtils.ts
+++ b/Composer/packages/adaptive-form/src/utils/arrayUtils.ts
@@ -33,6 +33,9 @@ const createArrayItem = <ItemType = unknown>(value: ItemType): ArrayItem<ItemTyp
   };
 };
 
+export const isItemValueEmpty = (value?: any) =>
+  !value || (typeof value === 'object' && !Object.values(value).some((val) => !!val));
+
 export const getArrayItemProps = <ItemType = unknown>(
   items: ArrayItem<ItemType>[],
   index: number,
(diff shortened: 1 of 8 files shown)


Composer/packages/adaptive-form/src/utils/arrayUtils.ts
```diff
@@ -95,7 +98,9 @@ export function useArrayItems<ItemType = unknown>(
   };
 
   const addItem = (newItem: ItemType) => {
-    handleChange(cache.concat(createArrayItem(newItem)));
+    // Allow only the last value in the cache to be empty
+    const newCache = isItemValueEmpty(newItem) ? cache.filter(({ value }) => !isItemValueEmpty(value)) : cache;
+    handleChange(newCache.concat(createArrayItem(newItem)));
```
tonyanziano, line 103: The `AllowedCallers.tsx` component also uses this function. Just checking to make sure that you are aware of this, and that this behavior is ok for all downstream components.

tonyanziano (approved):
I had only one comment, but other than that it looks great 👍

Composer/packages/adaptive-form/src/utils/arrayUtils.ts
OEvgeny: Let me double-check and circle back

OEvgeny: Works for me as before

tonyanziano: Great, thanks for checking

PR #9020, opened 2022-03-01
a11y: fix table-forms keyboard navigation

OEvgeny:
## Description

The PR aims to address the following issues found on pages with lists containing nested forms, including: "Bot Responses", "Knowledge (QnA)" and "Knowledge Base" pages.

Here is the list of issues addressed:
- Unable to navigate to fields nested in DetailsList cells using keyboard
- Some buttons are hidden while navigating via keyboard ("Clear field" button, "Add new" item button)
- Save via keyboard Enter key doesn't work under several circumstances
- Shift+Enter should persist caret position and the field in focus while adding a new line
- Visual focus indication for focused fields works unstable
- No error announcement
- No way to blur selected fields when the focus is on the DetailsList cell field

Highlights on changes made:
- Implement a CellFocusZone component to handle focus in DetailsLists' cells having form elements
- Move all related logic to the component, so it can: focus form elements, handle Escape and Enter key and handle focus loss cases
- Update EditField to play well with the new component
- Add a callback to be called on EditField when a new line gets added
- Handle Shift+Enter properly: from now it adds a new row to input and resets focus back to the input after it got switched to the textarea element
- Update CSS rules to handle focus for tables-view and EditableField components
- Refactor styles: move shared styles to the CellFocusZone
- Refactor table-views: introduce CellFocusZone and improve keyboard handling
- Knowledge Base: change the way alternative questions added (see the last gif: from now Shift+Enter equals clicking "Add new" button)
- Fix EditField styles to handle field height correctly
- Add announcements for EdtiField errors

## Task Item
- [VSTS 70470](https://fuselabs.visualstudio.com/Composer/_workitems/edit/70470)
- [VSTS 70527](https://fuselabs.visualstudio.com/Composer/_workitems/edit/70527)

## Screenshots
br-keyboard webm
ui-keyboard webm
kb-keyboard webm

#minor

### DIFF
--- a/Composer/packages/client/src/components/CellFocusZone.tsx
+++ b/Composer/packages/client/src/components/CellFocusZone.tsx
@@ -1,0 +1,100 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+/** @jsx jsx */
+import { jsx, css } from '@emotion/core';
+import React, { useCallback } from 'react';
+import { FocusZone, FocusZoneTabbableElements, IFocusZoneProps } from 'office-ui-fabric-react/lib/FocusZone';
+import { getFocusStyle, getTheme, mergeStyles } from 'office-ui-fabric-react/lib/Styling';
+
+export const formCell = css`
+  outline: none;
+  white-space: pre-wrap;
+  font-size: 14px;
+  line-height: 28px;
+`;
+
+export const formCellFocus = mergeStyles(
+  getFocusStyle(getTheme(), {
+    inset: -3,
+  })
+);
+
+/**
+ * CellFocusZone component props.
+ * Props are intently omitted as they're required for component to function correctly.
+ * Please manually test the component using keyboard in places it is used in case any changes made.
+ */
+type CellFocusZoneProps = Omit<
+  React.HTMLAttributes<unknown> & IFocusZoneProps,
+  | 'allowFocusRoot'
+  | 'data-is-focusable'
+  | 'isCircularNavigation'
+  | 'className'
+  | 'handleTabKey'
+  | 'shouldFocusInnerElementWhenReceivedFocus'
+  | 'tabIndex'
+  | 'onBlur'
+  | 'onKeyDown'
+>;
+
+const CellFocusZone: React.FC<CellFocusZoneProps> = (props) => {
+  const focusCurrentZoneAfterRender = useCallback((focusZoneEl: any) => {
+    const focusZoneId = focusZoneEl?.dataset?.focuszoneId;
+    if (!focusZoneId) {
+      return;
+    }
+    // wait for render to happen before placing focus back to the focus zone
+    setTimeout(() => {
+      const focusZone: HTMLElement | null = document.querySelector(`[data-focuszone-id=${focusZoneId}]`);
+      focusZone?.focus();
+    }, 100);
+  }, []);
+
+  const onCellKeyDown = useCallback((ev) => {
+    // ignore any of events coming from input fields
+    if (ev.target.localName === 'input' || ev.target.localName === 'textarea') {
+      ev.stopPropagation();
+      return;
+    }
+
+    // enter inside cell using Enter key
+    if (ev.target.dataset.focuszoneId && ev.key === 'Enter') {
+      const input: HTMLElement | null = (ev?.target as HTMLElement)?.querySelector('a, button, input, textarea');
+      input?.focus();
+      return;
+    }
+
+    // handle uncaught escape key events to allow returning back to navigation between cells
+    if (ev.key === 'Escape') {
+      ev.stopPropagation();
+      focusCurrentZoneAfterRender(ev.currentTarget);
+    }
+  }, []);
+
+  const onCellInnerBlur = useCallback((ev) => {
+    // this means we dont have an element to focus so we place focus back to the cell
+    if (!ev.relatedTarget) {
+      focusCurrentZoneAfterRender(ev.currentTarget);
+    }
+  }, []);
+
+  return (
+    <FocusZone
+      css={formCell}
+      {...props}
+      allowFocusRoot
+      data-is-focusable
+      isCircularNavigation
+      className={formCellFocus}
+      handleTabKey={FocusZoneTabbableElements.all}
+      shouldFocusInnerElementWhenReceivedFocus={false}
+      tabIndex={0}
+      onBlur={onCellInnerBlur}
+      onKeyDown={onCellKeyDown}
+    >
+      {props.children}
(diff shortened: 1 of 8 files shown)


tonyanziano:
Looks like the server package's typecheck is failing:

```
Error: [typecheck:client] src/pages/knowledge-base/table-view.tsx(546,47): error TS2345: Argument of type 'number | undefined' is not assignable to parameter of type 'SetStateAction<number>'.
[typecheck:client]   Type 'undefined' is not assignable to type 'SetStateAction<number>'.
```

OEvgeny:
@tonyanziano strange. It passes locally:
image

But `npm run typechec:client` does. Let me fix this

Composer/packages/client/src/components/CellFocusZone.tsx
```diff
@@ -0,0 +1,100 @@
...
+  | 'handleTabKey'
+  | 'shouldFocusInnerElementWhenReceivedFocus'
+  | 'tabIndex'
+  | 'onBlur'
+  | 'onKeyDown'
+>;
+
+const CellFocusZone: React.FC<CellFocusZoneProps> = (props) => {
+  const focusCurrentZoneAfterRender = useCallback((focusZoneEl: any) => {
+    const focusZoneId = focusZoneEl?.dataset?.focuszoneId;
```
tonyanziano, line 42: `focusZoneId` to follow the capitalization pattern

Composer/packages/client/src/components/CellFocusZone.tsx
```diff
@@ -0,0 +1,100 @@
...
+  | 'allowFocusRoot'
+  | 'data-is-focusable'
+  | 'isCircularNavigation'
+  | 'className'
+  | 'handleTabKey'
+  | 'shouldFocusInnerElementWhenReceivedFocus'
+  | 'tabIndex'
+  | 'onBlur'
+  | 'onKeyDown'
+>;
```
tonyanziano, line 38: Can you briefly explain the reason for this?

Composer/packages/client/src/components/EditableField.tsx
```diff
@@ -192,13 +197,29 @@ const EditableField: React.FC<EditableFieldProps> = (props) => {
...
       setMultiline(true);
+      updateField('value', e.target.value + '\n');
+      // wait for the textarea to be rendered and then restore focus and selection
+      setTimeout(() => {
+        const len = fieldRef.current?.value?.length;
+        fieldRef.current?.focus();
+        if (len) {
+          fieldRef.current?.setSelectionRange(len, len);
+        }
+      }, 100);
```
tonyanziano, line 215: is there a better way to do this then using `setTimeout()`? I noticed this same trick being used in `CellFocusZone`

tonyanziano (changes requested):

Composer/packages/client/src/components/CellFocusZone.tsx
OEvgeny: By limiting these props from being changed we ensure focus behavior is consistent for tables using the component. We may want to make the component even more non-configurable i.e. by not specifying `IFocusZoneProps` at all (in this case focus zone props can't be changed from the outside of the component).

Decided to go with a more relaxed version for now

Composer/packages/client/src/components/CellFocusZone.tsx
OEvgeny: Is this comment regarding `focusZoneEl?.dataset?.focuszoneId` part? If so, we can't affect this as it's how this is spelled in Fluent:

image

 the `dataset` element property does kebab case to camel case translation for us, so `data-focuszone-id` used by Fluent becomes `el.dataset.focuszoneId`,

Composer/packages/client/src/components/EditableField.tsx
OEvgeny: The `useLayoutEffect` hook would fit better here. There are two problems with it:
- It is not portable. By this I mean it can't be used in-place in the callback and requires adding more state to the component for a one-time thing
- The hook doesn't account for multiple re-renders happened in a row or with a small delay, so it won't cover the same case `setTimeout` covers here.

The current approach has an advantage in terms of that we have a control on a delay when it makes sense to account for re-renders. As an example 100ms should be a good value to return focus back if nothing is focused. We clearly can't do move focus back after one second because it might be confusing for end users as the focus can be moved anywhere else.

I was thinking on a custom hook implementation for both scenarios where we use `setTimeout` in the PR. The hook  may need to maintain the same 100ms delay though because of the reasons mentioned above.

Composer/packages/client/src/components/CellFocusZone.tsx
tonyanziano: Ok, that sounds good to me.

Composer/packages/client/src/components/CellFocusZone.tsx
tonyanziano: Ah, didn't realize this was controlled by Fluent. No problem then.

Composer/packages/client/src/components/EditableField.tsx
tonyanziano: > It is not portable. By this I mean it can't be used in-place in the callback and requires adding more state to the component for a one-time thing

I see what you mean here, you can't exactly call the logic inside of `useLayoutEffect` explicitly, so we would have to maintain some sort of state to trigger this which could get ugly.

> The hook doesn't account for multiple re-renders happened in a row or with a small delay, so it won't cover the same case setTimeout covers here.

We could potentially mitigate this by using a debounced function that eventually focuses the element.

After reading your response and thinking about it I think I'm fine with using this approach for now as it is only being used in these two places. If it starts to become something we need to use more commonly throughout the app then we can revisit the implementation.

Lastly, I like the idea of extracting this logic out into a hook that can be used by both components.

OEvgeny: Sounds good. I am going to investigate the hook approach, I don't see any other place we use it at the moment though

OEvgeny: Updated. So the hook does the following:
- place a callback to the macro tasks queue
- if re-render happen (react uses MessageChannel which guarantees it will be executed before setTimeout), pull out the task from queue and schedule another one
- since at this point it executes macro tasks, the task will be delayed which should give React scheduler time to schedule anything else if it wants to

With Fiber there may be different scheduler task priorities, so called out that this hook works only for immediate tasks

tonyanziano: Awesome!

tonyanziano (approved):

PR #9079, opened 2022-03-28
a11y: Add aria-required attribute to Form Editor required fields

OEvgeny:
## Description
These fields are visually indicated as required. so add `aria-required` attribute for screenreaders.

Alternative approach is might be in using a `required` property. It has more things to fix though:
- Need to remove double asterisk indication
- When placed in a native field, native tooltip pops-up which can't be customized

## Task Item

- [VSTS 70102](https://fuselabs.visualstudio.com/Composer/_workitems/edit/70102)

## Screenshots
image
image

#minor

### DIFF
--- a/Composer/packages/adaptive-form/src/components/fields/OpenObjectField/ObjectItem.tsx
+++ b/Composer/packages/adaptive-form/src/components/fields/OpenObjectField/ObjectItem.tsx
@@ -34,6 +34,7 @@ const ObjectItem: React.FC<ObjectItemProps> = ({
   onChange,
   onNameChange,
   onDelete,
+  required,
   ...rest
 }) => {
   const initialName = useMemo(() => originalName, []);
@@ -76,7 +77,7 @@ const ObjectItem: React.FC<ObjectItemProps> = ({
           label={formatMessage('Key')}
           name="key"
           placeholder={initialName || formatMessage('Add a new key')}
-          required={rest.required}
+          required={required}
           schema={{ type: 'string' }}
           uiOptions={{}}
           value={name}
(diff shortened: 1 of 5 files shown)


Composer/packages/adaptive-form/src/components/fields/OpenObjectField/ObjectItem.tsx
```diff
@@ -76,6 +76,7 @@ const ObjectItem: React.FC<ObjectItemProps> = ({
           label={formatMessage('Key')}
           name="key"
           placeholder={initialName || formatMessage('Add a new key')}
+          required={rest.required}
```
tonyanziano, line 79: nit: can you pull `required` off of `...rest` on line 36?

```ts
const ObjectItem: React.FC<ObjectItemProps> = ({
  definitions,
  name: originalName,
  formData,
  value,
  schema,
  onChange,
  onNameChange,
  onDelete,
  required, // <---
  ...rest
}) => {
```

Composer/packages/adaptive-form/src/components/fields/OpenObjectField/OpenObjectField.tsx
```diff
@@ -43,6 +43,7 @@ const OpenObjectField: React.FC<FieldProps<{
             formData={value}
             id={`${id}.value`}
             name={propertyName}
+            required={props.required}
```
tonyanziano, line 46: nit: pull `required` from props above on line 27

tonyanziano (approved):
Some minor comments but looks good.

tonyanziano (approved):

PR #9091, opened 2022-04-01
a11y: add support for intellisense

OEvgeny:
## Description

The PR adds accessibility properties for the intellisense component, wires them with input components and fixes some observations.

List of changes:
- Added `aria` attributes set into `Intellisense` component, so it to be passed down to any input rendered
- Updated fields. where it makes sense. to include a11y parameters passed by Intellisense
- Added scrollIntoView for CompletionElement to ensure the selected element is in the view
- Refactored CompletionList a11y properties. Now list has a listbox role when showing suggestions

## Task Item

- [VSTS 70456](https://fuselabs.visualstudio.com/Composer/_workitems/edit/70456)

## Screenshots

image

image

Suggestions-infocus webm

#minor

### DIFF
--- a/Composer/packages/adaptive-form/src/components/fields/IntellisenseFields.tsx
+++ b/Composer/packages/adaptive-form/src/components/fields/IntellisenseFields.tsx
@@ -47,9 +47,11 @@ export const IntellisenseTextField: React.FC<FieldProps<string>> = (props) => {
         onKeyDownTextField,
         onKeyUpTextField,
         onClickTextField,
+        aria,
       }) => (
         <StringField
           {...props}
+          aria={aria}
           cursorPosition={cursorPosition}
           focused={focused}
           id={id}
@@ -109,10 +111,12 @@ export const IntellisenseExpressionField: React.FC<FieldProps<string>> = (props)
         onKeyDownTextField,
         onKeyUpTextField,
         onClickTextField,
+        aria,
       }) => (
         <div ref={setContainerElm}>
           <StringField
             {...props}
+            aria={aria}
             cursorPosition={cursorPosition}
             focused={focused}
             id={id}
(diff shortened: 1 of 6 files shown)


Composer/packages/adaptive-form/src/components/fields/IntellisenseFields.tsx
```diff
@@ -161,9 +165,10 @@ export const IntellisenseNumberField: React.FC<FieldProps<string>> = (props) =>
...
       onChange={onChange}
     >
-      {({ textFieldValue, focused, onValueChanged, onKeyDownTextField, onKeyUpTextField, onClickTextField }) => (
+      {({ textFieldValue, focused, onValueChanged, onKeyDownTextField, onKeyUpTextField, onClickTextField, aria }) => (
         <NumberField
           {...props}
+          aria={aria}
           focused={focused}
           id={id}
           value={textFieldValue}
```
OEvgeny, line 174: There is also a `JsonField` wrapped into `Intellisense` down the lines. Verified that it doesn't make sense to add `aria` there:
image

- It renders Monaco editor which has it's own more complicated accessibility handling
- We don't provide any suggestions for the JSON field (AFIK) - only the placeholder message
- If we really want to provide suggestions for the JSON field, we should use Monaco built-in autocomplete feature as it is more accessible and user-friendly

Composer/packages/intellisense/src/components/CompletionElement.tsx
```diff
@@ -78,17 +78,31 @@ const renderDocumentation = (documentation: string | MarkupContent | undefined)
...
-  const { completionItem, isSelected, onClickCompletionItem } = props;
-
+  const { completionItem, isSelected, onClickCompletionItem, id } = props;
+  const rootRef = useRef<HTMLDivElement>(null);
   const additionalStyles = isSelected ? styles.selectedElement : {};
 
+  useLayoutEffect(() => {
+    if (isSelected && rootRef.current) {
+      rootRef.current.scrollIntoView?.();
+    }
```
tonyanziano, line 93: nit:

```ts
if (isSelected) {
  rootRef.current?.scrollIntoView?.();
}
```

tonyanziano (dismissed):
One minor comment, other than that it looks good 👍

Composer/packages/intellisense/src/components/CompletionElement.tsx
OEvgeny: Good point, fixed

tonyanziano (approved):

PR #9150, opened 2022-05-04
chore: upgrade webchat and emotion

OEvgeny:
## Description

* Upgrade WebChat packages to fix accessibility issues
* Upgrade emotion due to WebChat emotion bump
* Refactor codebase to meet `emotion@11` requirements
* Include required types for tsc to pass type checks

#minor


Composer/packages/adaptive-flow/src/adaptive-flow-renderer/components/OffsetContainer.tsx
```diff
@@ -2,10 +2,10 @@
 // Licensed under the MIT License.
 
 /** @jsx jsx */
-import { jsx } from '@emotion/core';
+import { jsx } from '@emotion/react';
 import React, { ReactNode } from 'react';
 
-interface OffsetContainerProps {
+interface OffsetContainerProps extends React.HTMLAttributes<HTMLElement> {
```
OEvgeny, line 8: Since v11 emotion removed `css` property from component props. See here: https://github.dev/emotion-js/emotion/blob/71ca9be3368ddb0a95ef5b2db7a97c67ab440593/packages/react/types/jsx-namespace.d.ts#L33

There are workarounds for turning it on for all components globally, but all of them assume either  react@17 or babel transform or a single `tsconfig.json`.

So, adding generic html element props directly to the props interface seems like a best fix for the issue at this point

PR #9193, opened 2022-05-18
fix: 34 security alerts at or above high severity

BruceHaley:
Fixes #minor

## Description

The [Component Detection task in the Azure DevOps pipeline](https://dev.azure.com/FuseLabs/Composer/_build/results?buildId=304886&view=logs&j=4b2e88cf-7f41-52d9-bc09-6abcb4f006a6&t=4df4959e-2e5c-519a-4a2a-a05a39029822) reports 34 security alerts at or above 'High' severity in the BotFramework-Composer repo. This fixes those vulnerabilities. They are:

ansi-regex 3.0.0
ansi-regex 4.1.0
ansi-regex 5.0.0
async 0.9.2
async 2.6.2
async 2.6.3
async 3.2.0
axios 0.21.1
ejs 3.1.6
eventsource 1.0.7
follow-redirects 1.13.0
follow-redirects 1.13.1
follow-redirects 1.13.3
follow-redirects 1.14.0
follow-redirects 1.14.1
glob-parent 3.1.0
minimist 0.0.10
minimist 1.2.5
moment 2.29.1
node-fetch 1.7.3
node-fetch 2.6.0
node-fetch 2.6.1
set-value 3.0.2
shelljs 0.8.4
simple-get 2.8.1
tar 4.4.8
tar 4.4.8
tar 4.4.8
tar 4.4.8
tar 4.4.8
tar 6.1.6
tar 6.1.6
tar 6.1.6
tmpl 1.0.4


tonyanziano:
Hey Bruce, just a heads up:

Back in January / February, we switched the Composer build tool over to Yarn v2+ (berry) to be compliant with Cyber EO requirements.

Yarn v2 operates on a completely separate set of lock files from yarn 1. You can see an example of such a lock file here: https://github.com/microsoft/BotFramework-Composer/blob/main/Composer/yarn-berry.lock

So the changes you have made in this PR will make dependabot and the security alert system happy, but it won't actually have any effect on what is packaged with the production Composer app until the v2 dependencies are updated.

@OEvgeny and I were considering completely getting rid of the v1 lock files and tooling so that we would no longer have to maintain two sets of lock files.

BruceHaley:
Abandoned: These, down to 20, will be revisited after #9298 is merged.

PR #9201, opened 2022-05-20
chore: use yarn 3 for development

OEvgeny:
* Cleanup scripts
* Fix linux specific issue with `@microsoft/orchestrator-core` not being built
* Split `.yarnrc.yml` files for dev and prod, so prod could use Azure registry
* Update yarn-berry lockfiles
* Remove old yarn lockfiles


Composer/packages/client/src/recoilModel/dispatchers/provisionQNA.ts
```diff
@@ -6,7 +6,7 @@ import { CallbackInterface, useRecoilCallback } from 'recoil';
 import { CognitiveServicesManagementClient } from '@azure/arm-cognitiveservices';
 import { TokenCredentials } from '@azure/ms-rest-js';
 import { SearchManagementClient } from '@azure/arm-search';
-import { SkuName } from '@azure/arm-search/lib/models';
+import { SkuName } from '@azure/arm-search/src/models';
```
tonyanziano, line 9: This doesn't seem right. What is the reasoning behind this change? Reminds of importing a Fluent component from `@fluent/src/Button` vs `@fluent/lib/Button` in which the latter is favorable.

scripts/switchYarnToProd.ps1
```diff
@@ -0,0 +1,7 @@
+<#
+  Swap .yarnrc.yml with .yarnrc.prod.yml to enforce auth and repo settings for production build.
+  We want the local dev environment to use npm registry for a sake of simplicity and for productin builds run in CI we use Azure packages registry.
+  So we run this script to swap the .yarnrc.yml file so that the CI can use Azure registry without affecting the local dev environment.
+#>
+
+Copy-Item -Path .yarnrc.prod.yml -Destination .yarnrc.yml
```
tonyanziano, line 7: just a note to myself to switch the build pipelines over to calling this script

Composer/packages/client/src/recoilModel/dispatchers/provisionQNA.ts
OEvgeny: Right.
Seems this package has newer version in yarn-berry which doesn't have lib folder. Changed to pull model type from esm folder instead of src.

scripts/switchYarnToProd.ps1
```diff
@@ -0,0 +1,7 @@
+<#
+  Swap .yarnrc.yml with .yarnrc.prod.yml to enforce auth and repo settings for production build.
+  We want the local dev environment to use npm registry for a sake of simplicity and for productin builds run in CI we use Azure packages registry.
```
tonyanziano, line 3: production**

Composer/packages/client/src/recoilModel/dispatchers/provisionQNA.ts
```diff
@@ -6,7 +6,7 @@ import { CallbackInterface, useRecoilCallback } from 'recoil';
 import { CognitiveServicesManagementClient } from '@azure/arm-cognitiveservices';
 import { TokenCredentials } from '@azure/ms-rest-js';
 import { SearchManagementClient } from '@azure/arm-search';
-import { SkuName } from '@azure/arm-search/lib/models';
+import { SkuName } from '@azure/arm-search/esm/models';
```
tonyanziano, line 9: just curious: why was this changed?

tonyanziano (dismissed):
2 minor comments, otherwise it looks good 👍

tonyanziano:
Looks like the most recent unit test run failed:

Summary of all failing tests
FAIL packages/lib/code-editor/src/components/toolbar/__tests__/ToolbarButtonMenu.test.tsx
  ● <ToolbarButtonMenu /> › property: Should render icon button with properties in menu when passed property payload
    TestingLibraryElementError: Unable to find an element with the text: /this/. This could be because the text is broken up by multiple elements. In this case, you can provide a function for your text matcher to make your matcher more flexible.

Composer/packages/client/src/recoilModel/dispatchers/provisionQNA.ts
OEvgeny: Yarn 3 resolved packages a bit differently and in the never package types moved to esm folder

scripts/switchYarnToProd.ps1
OEvgeny: Thanks, fixed

OEvgeny:
I believe worked died. Let's see if it passes

PR #9219, opened 2022-06-01
fix: get started content is not visible

### LINKED ISSUE #9202: Create Core Assistant bot shows blank "Get started" blade
Repro:

1. Open Composer.
2. Click Create new bot and use the Core Assistant Bot template.
3. Pick a target folder.

The message "Your new bot is almost ready!" appears in a blue box pointing to the empty "Get Started" blade:

image

Shouldn't the blade have some content? Possibly it should contain a hint about adding LUIS keys? Should an empty blade even be displayed? It does not look right.

Is this a bug?

Version = 2.1.2.

OEvgeny:
## Description

Render Get Started content in panel content area to avoid content truncation.

The initial approach was working on the previous fluent version due to relaxed styles for the panel header area. Now the overlapping content is truncated causing the panel to be empty.

It is not recommended to use the header area for rendering panel content.

## Task Item

fixes #9202

### DIFF
--- a/Composer/packages/client/src/components/GetStarted/GetStarted.tsx
+++ b/Composer/packages/client/src/components/GetStarted/GetStarted.tsx
@@ -3,7 +3,7 @@
 
 /** @jsx jsx */
 import { jsx } from '@emotion/react';
-import React from 'react';
+import React, { useState, useCallback } from 'react';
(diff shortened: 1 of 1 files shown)


tonyanziano (approved):
Looks good!

PR #9257, opened 2022-06-23
build: Fix Component Governance in Composer DevOps pipeline

BruceHaley:
Fixes #minor

## Description
Implementation of yarn v3 resulted in Component Governance analysis failing to find security vulnerabilities: CG does not handle the new v3 yarn-berry.lock files.

## Changes
Add a script to generate imitation yarn v1 yarn.lock files just before CG analysis runs.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -71,6 +71,8 @@ jobs:
       - powershell: |
           $find = 'npm:';
           $replace = '';
+          $find2 = 'resolution:';
+          $replace2 = 'resolved:';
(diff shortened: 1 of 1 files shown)


tonyanziano (approved):

tonyanziano:
Hmm, it looks like the new yarn broke the E2E tests. The E2E runner is somehow using the wrong version of Cypress which is breaking the tests.

I don't think this is related to your change, but the yarn 3 changes we merged earlier this week.

PR #9319, opened 2022-07-25
chore: update fluent to v 8.83

OEvgeny:
## Description
This fixes an issue with tables sorting not being announced. 

Fluent team did some changes in response to the issue we have, so this updates fluent version and adds required aria attributes to tables to address the issue.

## Task Item
[VSTS-69832](https://fuselabs.visualstudio.com/Composer/_workitems/edit/69832)

## Screenshots
image

#minor


tonyanziano (changes requested):
I think the fluent upgrade must have broken a few tests.

The `Manage___` tests are flaky and it's ok if they break, but there were 7 tests that failed. [The last PR to go in](https://github.com/microsoft/BotFramework-Composer/pull/9306) succeeded on all the checks so I think it's worth investigating.

tonyanziano (approved):
Awesome! Thanks :)

PR #9321, opened 2022-08-01
fix: [#9151] LG for 'choices' field in multi choice input results in Warning

### LINKED ISSUE #9151: LG for 'choices' field in multi choice input results in Warning
Repro bot attached.

I defined an option set as bot responses for YesNo() as shown in the image below

image

Then, I add a multi choice input which uses ${YesNo()} as array of choices

image

When doing this, I get a warning in composer

image

However, this usage should be correct and is even contemplated in the ChoiceInput code in the SDK:

```csharp
 if (Choices.ExpressionText != null && Choices.ExpressionText.TrimStart().StartsWith("${", StringComparison.InvariantCultureIgnoreCase))
            {
                // use ITemplate<ChocieSet> to bind (aka LG)
                return await new ChoiceSet(Choices.ExpressionText).BindAsync(dc, dc.State).ConfigureAwait(false);
            }
            else
            {
                // use Expression to bind
                return Choices.TryGetValue(dc.State).Value;
            }
```

The code excerpt is from here: https://github.com/microsoft/botbuilder-dotnet/blob/main/libraries/Microsoft.Bot.Builder.Dialogs.Adaptive/Input/ChoiceInput.cs#L236

Seems like the Choices field is ObjectExpression but should allow LG in the UI canvas.

ceciliaavila:
## Description

This PR fixes the validation of LG expressions to avoid showing a warning when LG expressions are used.
A `trimExpression` function was added to remove the `${}` characters of the expression allowing it to be correctly parsed.
It also adds a unit test to cover this new function.

## Task Item
Fixes #9151

## Screenshots
These images show the warning displayed before and the LG expression being properly validated after.

BEFORE
image
AFTER
image


OEvgeny (dismissed):

OEvgeny:
Should be good to go after related package lock hash updates

OEvgeny (approved):

Issue #9346, opened 2022-08-15
Bot frmaework Composer freezes very frequently - Out of memory exception. Both Web & Windows Version


selvaduraiece:
## Describe the bug
Bot frmaework Composer freezes very frequently - Out of memory exception. Both Web & Windows Version

![image](https://user-images.githubusercontent.com/40558980/184700294-f240f729-c227-4761-ac53-0a2137d51e14.png)

## Version
2.0

## Browser
All

- [ ] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS
Windows

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

## Screenshots
![image](https://user-images.githubusercontent.com/40558980/184700356-d1ae62fe-fad2-4783-9abe-7e577286966a.png)

selvaduraiece:
I did updated and used for 2 days and it still remanins same

selvaduraiece:
Extremely slow and taking more than 10min to add a simple SendResponse

dmvtech:
Linking to another issue with Composer and memory leak/Out of Memory which might be related or of help:
https://github.com/microsoft/BotFramework-Composer/issues/4728

stevkan:
Linking IcM [#335151352](https://portal.microsofticm.com/imp/v3/incidents/details/335151352/home), "High memory utilization by Bot framework composer", to issue. Cx is unable to move forward on project due to memory issue.

sw-joelmut:
Hi @tracyboehrer, we've been wondering if you can provide us with this information that could help us. Thanks.

> Linking IcM [#335151352](https://portal.microsofticm.com/imp/v3/incidents/details/335151352/home), "High memory utilization by Bot framework composer", to issue. Cx is unable to move forward on project due to memory issue.

stevkan:
@ceciliaavila, could you provide an update to on this?

ceciliaavila:
Hi @stevkan, we are working on this issue, trying to identify the reproduction steps. We'll let you know as soon as we have more information.

sw-joelmut:
We were able to reproduce the issue consistently by creating a sample with multiple actions, opening and closing the Composer canvas with other pages, and lowering the JS heap memory size limit (so it's easier to reproduce the error).
Moreover, we noticed each time the canvas page is closed the memory used is never released, causing to increment the memory usage over time.

[...]

PR #9418, opened 2022-11-08
fix: [#9346] Override global Window internal symbol property for the Monaco Editor

sw-joelmut:
Addresses #9346
#minor

## Description
This PR solves part of an issue where the Monaco Editor was generating a memory leak when the Authoring Canvas page is rendered multiple times.
> **Note:** This implementation will show an error message into the DevTools console saying that the "language Client services has been overridden", this error does not affect Composer functionality as it is an informative message.

## Specific Changes
- Remove use of global Window property object when installing Monaco Services.

## Testing
The following image shows the memory leak and the console error message.
imagen


sw-joelmut:
Closed as it's no longer needed

PR #9468, opened 2023-01-13
fix: [#9346] Fix memory leak issues in VisualPanel components

ceciliaavila:
#minor
## Description

This PR applies minor fixes in some of the **_VisualPanel_** components to avoid detached elements in memory.
Every time we open the design page in Composer, a detached element is added to the memory and it can't be removed by the garbage collector. If we switch between pages for a while, the memory usage increases and it may end up causing an out-of-memory exception.

## Task Item
Addresses #9346

## Screenshots
These images show the detached elements being removed by the garbage collector after the fixes.
image


tracyboehrer (approved):

tracyboehrer (approved):

Composer/packages/adaptive-flow/src/adaptive-flow-editor/AdaptiveFlowEditor.tsx
```diff
@@ -200,9 +192,9 @@ const VisualDesigner: React.FC<VisualDesignerProps> = ({ onFocus, onBlur, schema
                       }}
                     />
                   </div>
-                </FlowToolbar>
-              </MarqueeSelection>
-            </SelectionContext.Provider>
+                </MarqueeSelection>
```
OEvgeny, line 195: What is the reason for reordering the components?

ceciliaavila: Hi @OEvgeny, we found duplicated detached elements in memory and narrowed down the problem to these two components. We couldn't find the exact reason, but the issue was fixed after reordering the components.

OEvgeny (approved):

PR #9538, opened 2023-03-16
fix: [#9517][CSS] PRO | COMPOSER | In <dialogName>.triggers[0].actions[0]: Text: alwaysPrompt the return type does not match

### LINKED ISSUE #9517: [CSS] PRO | COMPOSER | In <dialogName>.triggers[0].actions[0]: Text: alwaysPrompt the return type does not match
ICM https://portal.microsofticm.com/imp/v3/incidents/details/367589004/home

[PROBLEM DESCRIPTION]
The customer developed their bot in Composer.

Version:  2.1.2
Electron: 8.2.4
Chrome: 80.0.3987.165
NodeJS: 12.13.0
V8: 8.0.426.27-electron.0

Customer receives this error: 
`In <dialogNmae>[0].actions[0]: Text: alwaysPrompt the return type does not match`
image

Attached is reproducible .zip composer bot:
[sampleRepro.zip](https://github.com/microsoft/BotFramework-Composer/files/10798022/VM-Credit-Card-VA_export.zip)

erquirogasw:
## Description

This PR fixes the issue by adding a new way of indexing the values in the cache. It will store the type along with the values in the cache after checking for errors to avoid overwriting the data and saving unuseful information.

## Task Item
Fixes #9517

## Screenshots

These images show the user sample and a Microsoft template working successfully with the new indexing.
image
image

### DIFF
--- a/Composer/packages/lib/indexers/src/validations/expressionValidation/index.ts
+++ b/Composer/packages/lib/indexers/src/validations/expressionValidation/index.ts
@@ -49,7 +49,7 @@ export const validateExpressions: ValidateFunc = (
     let errorMessage = '';
     let warningMessage = '';
     try {
-      const valueToValidate = cache?.[path] ? cache[path] : checkExpression(value, required, types)
+      const valueToValidate = cache?.[path] ? cache[path] : checkExpression(value, required, types);
       errorMessage = checkReturnType(valueToValidate, types);
       if (!errorMessage) {
         //First validate that the types of the value match and then store the type value in newCache using the path as key to avoid overwriting.
--- a/Composer/packages/lib/indexers/src/validations/expressionValidation/index.ts
+++ b/Composer/packages/lib/indexers/src/validations/expressionValidation/index.ts
@@ -49,10 +49,12 @@ export const validateExpressions: ValidateFunc = (
     let errorMessage = '';
     let warningMessage = '';
     try {
-      //the cacheKey adds the type of the value at the end, this allows a new indexing system to prevent the overwriting in cache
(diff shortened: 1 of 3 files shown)


Composer/packages/lib/indexers/src/validations/expressionValidation/index.ts
```diff
@@ -49,8 +49,10 @@ export const validateExpressions: ValidateFunc = (
     let errorMessage = '';
     let warningMessage = '';
     try {
-      newCache[value] = cache?.[value] ? cache[value] : checkExpression(value, required, types);
-      errorMessage = checkReturnType(newCache[value], types);
+      // The cacheKey adds the type of the value at the end, this allows a new indexing system to prevent the overwriting in cache.
+      const cacheKey = types.length > 1 ? value : `${value}.${types[0]}`;
```
OEvgeny, line 53: Can we store a type along with the cached result instead and check for types match before returning the value?

My only concern is cache pollution with values which probably won't be used anymore.

ceciliaavila: Hi @OEvgeny, we updated the solution to check for errors before saving the types in the cache.
Please let us know if this was not what you had in mind.
Thanks!

OEvgeny: Looks good. Please update the PR description with the details

ceciliaavila: Done. Thanks!

OEvgeny (approved):

PR #9549, opened 2023-04-03
fix: [#9528] Unable to add two PVA bot skills through bot framework composer with the option connect to a skill

### LINKED ISSUE #9528: Unable to add two PVA bot skills through bot framework composer with the option connect to a skill
## Describe the bug

We are able to add a PVA bot skill in bot composer successfully. When we try add another PVA skill, the below error is thrown in bot composer.
"This operation cannot be completed. The skill is already part of the Bot Project"
## Version

2.1.2
## Browser

Chrome

- [ ] Chrome

## OS

macOS
- [ ] macOS

## To Reproduce

1. Go to 'Bot Framework Composer'
2. Click on 'Connect to a skill'
3. Select PVA bot manifest zip file

## Screenshots

image

ceciliaavila:
## Description
This PR fixes the issue where Composer fails to connect to multiple PVA skills due to a validation error saying "The skill is already part of the Bot Project". This validation happened because a single manifest was saved in the `skills` folder, and adding another PVA bot would replace it.
Therefore, changing the folder structure to where the bot's manifest is saved solved the issue.
Other issues related to PVA skills were addressed:
- The Manifest URL didn't redirect to valid content, since the bot was loaded from a `.zip` file and no endpoint was available.
  - Instead of redirecting to the `manifest.json` file, it shows its content in a modal.
- Switching between PVA skill pages, the information didn't get refreshed.
  - Adding the `projectId` to the `useMemo` function solved the issue.
- When a PVA skill is added and the visual focus is on another bot that is not the root, it will throw an error as it didn't locate `.lg` and `.lu` files, failing to add the skill action trigger to the root bot.
  - Changing the current opened bot project id to use the root's id, solved the issue.

## Task Item
Fixes #9528

## Screenshots
The following images show the validation when adding an existing PVA skill and the folder structure, and the bots working as expected.
image
image

### DIFF
--- a/Composer/packages/client/src/pages/design/ManifestEditor.tsx
+++ b/Composer/packages/client/src/pages/design/ManifestEditor.tsx
@@ -18,7 +18,6 @@ import { Link } from '@fluentui/react/lib/Link';
 import get from 'lodash/get';
 
 import { SkillInfo, skillNameIdentifierByProjectIdSelector } from '../../recoilModel';
-import { isExternalLink } from '../../utils/httpUtil';
 
 import { formEditor } from './styles';
 import { PropertyEditorHeader } from './PropertyEditorHeader';
@@ -76,6 +75,10 @@ const styles = {
 
 const helpLink =
   'https://docs.microsoft.com/en-us/azure/bot-service/skills-write-manifest-2-1?view=azure-bot-service-4.0';
+
+const isExternalLink = (url: string): boolean => {
+  return /^http[s]?:\/\/\w+/.test(url);
+};
 
 export interface ManifestEditorProps {
   formData: SkillInfo;
--- a/Composer/packages/client/src/pages/design/ManifestEditor.tsx
+++ b/Composer/packages/client/src/pages/design/ManifestEditor.tsx
@@ -6,6 +6,8 @@ import { jsx, css } from '@emotion/react';
 import formatMessage from 'format-message';
 import ErrorBoundary from 'react-error-boundary';
 import React from 'react';
+import { useRecoilValue } from 'recoil';
+import { useShellApi } from '@bfc/extension-client';
 import { LoadingTimeout } from '@bfc/adaptive-form/lib/components/LoadingTimeout';
 import { FieldLabel } from '@bfc/adaptive-form/lib/components/FieldLabel';
 import ErrorInfo from '@bfc/adaptive-form/lib/components/ErrorInfo';
@@ -15,7 +17,8 @@ import { DetailsList, DetailsListLayoutMode, SelectionMode } from '@fluentui/rea
 import { Link } from '@fluentui/react/lib/Link';
 import get from 'lodash/get';
 
-import { SkillInfo } from '../../recoilModel';
+import { SkillInfo, skillNameIdentifierByProjectIdSelector } from '../../recoilModel';
+import { isExternalLink } from '../../utils/httpUtil';
 
 import { formEditor } from './styles';
 import { PropertyEditorHeader } from './PropertyEditorHeader';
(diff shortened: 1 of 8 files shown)


Composer/packages/server/src/models/utilities/util.ts
```diff
@@ -40,5 +40,11 @@ export const getRemoteFile = async (url): Promise<string> => {
 
 // convert zip folder name to skill name
 export const convertFolderNameToSkillName = (path, skillName) => {
-  return path.replace(/(\w+)\//, `${skillName}/`);
+  const hasFolder = path.match(/(\w+)\//);
```
OEvgeny, line 43: nit: this doesn't allow hyphens or dashes in folder name

Composer/packages/server/src/models/utilities/util.ts
```diff
@@ -40,5 +40,11 @@ export const getRemoteFile = async (url): Promise<string> => {
 
 // convert zip folder name to skill name
 export const convertFolderNameToSkillName = (path, skillName) => {
```
OEvgeny, line 42: suggestion use `path` module to operate with paths. Not tested but something like this should work:
```js
export const convertFolderNameToSkillName = (pathStr, skillName) => {
  const folderName = path.parse(pathStr).dir;
  const relativePath = path.relative(folderName, pathStr);
  
  return path.join(skillName, relativePath);
};
```

Composer/packages/client/src/recoilModel/dispatchers/botProjectFile.ts
```diff
@@ -14,7 +14,9 @@ import { rootBotProjectIdSelector } from '../selectors';
 import { setRootBotSettingState } from './setting';
 import { addSkillFiles, deleteSkillFiles } from './utils/skills';
 
-const urlRegex = /^http[s]?:\/\/\w+/;
+const isExternalLink = (url: string): boolean => {
```
OEvgeny, line 17: suggestion:
```js
const isExternalLink = (url: string): boolean => {
    const { origin } = location;
    const parsedUrl = new URL(url, origin);
    return parsedUrl.origin !== origin;
};
```
Note that in case the URL provided is invalid, this will throw.

My only concern is about protocols limiting to `http(s)`.

Also noted the implementation does not support relative protocol `//`. Though this is not supported in the code suggested as well, and I don't think it should be. Just raising the point in case there are known use-cases.

Composer/packages/client/src/pages/design/ManifestEditor.tsx
```diff
@@ -74,6 +76,10 @@ const styles = {
 const helpLink =
   'https://docs.microsoft.com/en-us/azure/bot-service/skills-write-manifest-2-1?view=azure-bot-service-4.0';
 
+const isExternalLink = (url: string): boolean => {
```
OEvgeny, line 79: suggestion: put the function to utilities and reuse here and in `Composer/packages/client/src/recoilModel/dispatchers/botProjectFile.ts`

ceciliaavila:
Hi @OEvgeny, all feedback applied. Thanks!

Composer/packages/server/src/models/utilities/__test__/util.test.ts
```diff
@@ -5,7 +5,7 @@ import { convertFolderNameToSkillName } from '../util';
 
 describe('Convert zip folder name to skill name', () => {
   it('should return path with skillName folder', () => {
-    const path = 'manifest/empty-manifest.json';
+    const path = 'manifest-folder/empty-manifest.json';
```
OEvgeny, line 8: :+1:

OEvgeny (approved):

PR #9554, opened 2023-04-07
feat: allow to disable oneauth

OEvgeny:
## Description

Allow to provide tokens directly in the electron app release by running with `COMPOSER_ENABLE_ONEAUTH=false` environment variable

#minor

### DIFF
--- a/Composer/packages/electron-server/src/preload.js
+++ b/Composer/packages/electron-server/src/preload.js
@@ -3,7 +3,7 @@
 
 const { ipcRenderer, contextBridge } = require('electron'); // eslint-disable-line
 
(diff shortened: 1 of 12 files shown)


Composer/packages/electron-server/src/preload.js
```diff
@@ -3,6 +3,8 @@
 
 const { ipcRenderer, contextBridge } = require('electron'); // eslint-disable-line
 
+const isOneAuthEanabled = !(process.env.COMPOSER_ENABLE_ONEAUTH === 'false' || process.platform === 'linux');
```
OEvgeny, line 6: Note we can't extract this into the module and share it with `Composer/packages/electron-server/src/auth/oneAuthService.ts` due to the preloader.js being used as is without bundling.

tdurnford (approved):

PR #9614, opened 2023-07-21
fix: Optimize memory usage when parsing LgFiles

sw-joelmut:
#minor

## Description
This PR reduces the memory usage when parsing the LgFiles ([lgParse.worker](https://github.com/microsoft/BotFramework-Composer/blob/main/Composer/packages/client/src/recoilModel/parsers/workers/lgParser.worker.ts)) by adding a `MapOptimizer` to the `LgFilesCache`.
This optimizer consists of, based on a capacity limit, the process will start removing LgFile parse results to free up space from memory. Additionally, it takes into account all the references and relationships between other LgFiles.

> **Note**: The `MapOptimizer` capacity limit is set to store a maximum of 10 parsed LgFiles, this amount can be modified through code.

## Screenshots
The following image shows loading 9 bot responses with the `MapOptimizer` capacity size of 3 (for testing purposes), seeing the improvement in memory usage.
image

### DIFF
--- a/Composer/packages/client/src/App.tsx
+++ b/Composer/packages/client/src/App.tsx
@@ -15,6 +15,7 @@ import { setOneAuthEnabled } from './utils/oneAuthUtil';
 import { LoadingSpinner } from './components/LoadingSpinner';
 import lgWorker from './recoilModel/parsers/lgWorker';
 import { LgEventType } from './recoilModel/parsers/types';
+import { useMount, useUnmount } from '@fluentui/react-hooks';
 
 setupIcons();
 
@@ -28,6 +29,7 @@ export const App: React.FC = () => {
   const { appLocale } = useRecoilValue(userSettingsState);
 
   const [isClosing, setIsClosing] = useState(false);
+  const [listener, setListener] = useState<{ destroy(): boolean }>({} as any);
 
   const {
     fetchExtensions,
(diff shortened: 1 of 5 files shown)


Composer/packages/client/src/App.tsx
```diff
@@ -34,11 +36,21 @@ export const App: React.FC = () => {
...
 
   useEffect(() => {
     loadLocale(appLocale);
   }, [appLocale]);
 
+  useEffect(() => {
+    lgWorker.listen(LgEventType.OnUpdateLgFile, (msg) => {
+      const { projectId, payload } = msg.data;
+      updateFile({ projectId, value: payload });
+    });
```
OEvgeny, line 51: This effect will re-run on every re-render. We should make it run only on mount instead, I believe

Also we should remove the listener when the component gets unmounted

Composer/packages/client/src/recoilModel/utils/mapOptimizer.ts
```diff
@@ -0,0 +1,103 @@
...
+interface MapOptimizerTree<Key> {
+  timestamp: number;
+  references: Key[];
+}
+
+interface OnUpdateMapOptimizerContext<Key> {
+  setReferences(references: Key[]): void;
+}
+
+export class MapOptimizer<Key, Value> {
```
OEvgeny, line 13: Could you add a doc comment to the class describing how it works?

OEvgeny:
Looks good, left a couple of comments to address.

Nice work!

sw-joelmut:
Hi @OEvgeny, the suggestions are applied, thanks!

OEvgeny (approved):

